PR #474, opened 2024-03-28
refactor(kas): remove kas oidc dependency

### LINKED ISSUE #580: remove `kas` oidc dependency
With the authn middleware `KAS` does not need to be aware of the oidc issuer and validate the incoming access token. This code should be removed. This will require us to forward the token along in the grpc metadata context so can kas can leverage it as a downstream service.

strantalis:
KAS Updates: 
- remove double access token verification
- always verify signed request token with dpop pubkey
- extract entity info
- remove bearer token from kas rewrap proto

SDK Updates:
- builds kas client when doing decrypt instead of on sdk initialization
- adds ability to pass in extra grpc dial options
- updates tdf tests to mock out kas grpc rewrap and public key endpoint

Resolves #580


mkleene (commented):
I don't love the kasClient refactoring for a few reasons

1. It makes it harder to use a single connection, which is a pretty reasonable option to keep open
2. We are taking a pretty straightforward service locator pattern and replacing it with creation of client classes deep in code so it's hard to see where dependencies are, let alone isolate them for testing if we need to
3. I don't think that it would be hard to specify the dial options in the `SDK` creation code

If we are going to do this we should also get rid of `auth_config` since it won't work for anyone. Also turning off `authn.Config.Enabled` won't work for rewrap so we might as well get rid of it, too.

service/kas/access/rewrap_test.go
```diff
@@ -394,93 +377,3 @@ func TestParseAndVerifyRequest(t *testing.T) {
...
-		URI:            *kasURI,
-		CryptoProvider: hsmSession,
-		OIDCVerifier:   nil,
-	}
-
-	body := `{"mock": "value"}`
-	_, err := kas.Rewrap(context.Background(), &kaspb.RewrapRequest{SignedRequestToken: body, Bearer: "invalidToken"})
-	status, ok := status.FromError(err)
-	assert.True(t, ok)
-	assert.Equal(t, codes.PermissionDenied, status.Code())
```
dmihalcik-virtru, line 485: I don't see any tests of returning PermissionDenied elsewhere? Can we move these tests to `authn_test.go`?

strantalis:
@mkleene 
> I don't love the kasClient refactoring for a few reasons
> 
> 1. It makes it harder to use a single connection, which is a pretty reasonable option to keep open
> 2. We are taking a pretty straightforward service locator pattern and replacing it with creation of client classes deep in code so it's hard to see where dependencies are, let alone isolate them for testing if we need to
> 3. I don't think that it would be hard to specify the dial options in the `SDK` creation code
> 
> If we are going to do this we should also get rid of `auth_config` since it won't work for anyone. Also turning off `authn.Config.Enabled` won't work for rewrap so we might as well get rid of it, too.

We can't use a single connection if we have multiple KAS's passed in. For example if we have `kas1.example.com` and `kas2.example.com` I don't think it's possible to re-use that connection. The same thing goes for decrypt with multiple key access objects. We could have `kas1.example.com` and `kas2.example.com`. Which all this might change when we get into ironing out multi-kas but this is currently how it works today. 

There is a new WithDialOptions that you can set in the sdk now. This was needed anyway so we could mock the KAS grpc service with the `buffconn` package.

service/kas/access/rewrap_test.go
strantalis: I think permission denied makes sense here but we probably want something to fail if the signed body request is signed with the wrong key. Let me add a few more tests.

sdk/tdf_test.go
```diff
@@ -493,23 +465,36 @@ func TestTDF(t *testing.T) {
...
+		testDecryptWithReader(s.T(), s.sdk, tdfFileName, decryptedTdfFileName, test)
 
 		// Remove the test files
 		_ = os.Remove(plaintTextFileName)
 		_ = os.Remove(tdfFileName)
 	}
 }
 
-func BenchmarkReader(b *testing.B) {
+/* func BenchmarkTDF(b *testing.B) {
```
strantalis, line 484: @sujankota Do we need this benchmark test? I was trying to update it to use testify suite but I don't think it will work. Also it didn't seem like we were doing the traditional golang benchmark test of 

```go
for i := 0; i < b.N; i++ {
  // Code to benchmark
}
```

pflynn-virtru (approved):

jrschumacher (approved):

dmihalcik-virtru (approved):

PR #489, opened 2024-04-02
fix(PLAT-3069):  opentdf/platform, gRPC: Namespace with existed attribute(s) can be deactivated w/o any prompts

mustyantsev:
Throw an error **unable to deactivate non-empty namespace. This namespace has %d attribute(s)** when namespace has attributes


jrschumacher:
@mustyantsev Thanks for this enhancement! Would you be up for updating the integration tests to validate that the attributes are deactivated when this function is called?

jakedoublev (approved):

jrschumacher (approved):

Issue #501, opened 2024-04-03
Test framework for OPA Policy Rego files


pflynn-virtru:
See https://www.openpolicyagent.org/docs/latest/policy-testing/

We use an embedded OPE engine which is hard to develop with and debug.

Goal:
- README with tips on how to develop
- Test framework which mirrors the OPA embedded engine approach
- Test framework which tests custom built-ins

PR #529, opened 2024-04-05
test(authorization): unit test rego

pflynn-virtru:
Adds a Go unit test to run the rego policies with various inputs

resolves https://github.com/opentdf/platform/issues/501

replaces https://github.com/opentdf/platform/pull/503 bad rebase


service/policies/entitlements/conditions-test.rego
dmihalcik-virtru, line 1: This maybe should be in a subfolder named `testdata`?

https://stackoverflow.com/a/76674532/23894185

Also naming it `-test.rego` sourncs like it contains unit tests, maybe a better name would be descriptive like `fixtures.rego` or `samples.rego` or something

dmihalcik-virtru (approved):

PR #536, opened 2024-04-08
chore(policy): cleanup allowed and normalized characters throughout attributes

jakedoublev:
1. Adds proto validation to namespace names, attribute names, and attribute values
2. Adds unit tests around those validation rules
3. Normalizes stored values to lower case in the DB
4. Integration tests that stored values were normalized in lower case correctly


service/policy/attributes/attributes_test.go
```diff
@@ -0,0 +1,284 @@
...
+func TestCreateAttribute_NameWithNonAlphanumeric_Fails(t *testing.T) {
+	names := []string{
+		"invalid@name",
+		"invalid#name",
+		"invalid$name",
+		"invalid%name",
+		"invalid^name",
+		"invalid:name",
+		"invalid/name",
+		"invalid&name",
```
jrschumacher, line 100: I think this is fine, but I'm not sure we need to test everything since everything outside the set is significantly larger than in the set.

jrschumacher (dismissed):

service/policy/attributes/attributes_test.go
jakedoublev: True, and that's worth clarifying in a comment. Updated commit: [04eef36](https://github.com/opentdf/platform/pull/536/commits/04eef36ef30f84e3abad08a54693ae8f06f8a39d)

jrschumacher (approved):

jrschumacher (approved):

PR #544, opened 2024-04-08
feat(sdk): don't require `client_id` in the auth token

mkleene:
This isn't a claim that is in any spec, except for token exchange drafts

Instead we should allow the roles specified in the token to dictate access to resources.


strantalis (approved):

jrschumacher:
Given that roles and groups are optional (at least with Keycloak they must be mapped) it would seem that we should utilize jq to allow the platform admin to validate any aspect of the claim. If they want to check for a client_id, role, group, etc then they *could*. 

I assume we can add this kind of enhancement latter. Should we create an issue?

strantalis:
> Given that roles and groups are optional (at least with Keycloak they must be mapped) it would seem that we should utilize jq to allow the platform admin to validate any aspect of the claim. If they want to check for a client_id, role, group, etc then they _could_.
> 
> I assume we can add this kind of enhancement latter. Should we create an issue?

I think we can add something to validate custom claims at a later time once someone maybe brings us a valid use case.

PR #546, opened 2024-04-09
feat(sdk): normalize token exchange

mkleene:
* move the keycloak provisioning code into a central location so it can be used in commands and sdk tests
* move the code that did token exchange into `oauth`
* add a functional option to specify token exchange parameters
* use pointers for client credentials option in config so it's obvious when they have been specified
* don't get rid of `authConfig` yet since this pull is already too big


sdk/options.go
```diff
@@ -65,3 +66,17 @@ func WithCustomAuthorizationConnection(conn *grpc.ClientConn) Option {
 		c.authorizationConn = conn
 	}
 }
+
+// WithTokenExchange specifies that the SDK should obtain its
+// access token by exchanging the given token for a new one. If
+// the subjectToken is for another client keycloak requires that
```
ttschampel, line 72: remove reference to keycloak

ttschampel (approved):

Dockerfile
```diff
@@ -5,6 +5,7 @@ WORKDIR /app
 COPY protocol/ protocol/
 COPY sdk/ sdk/
 COPY lib/ocrypto lib/ocrypto
+COPY lib/fixtures lib/fixtures
```
strantalis, line 8: Is our build dependent on fixtures package?

mkleene: yes, unit tests need it in order to generate test data

sdk/options.go
```diff
@@ -11,7 +11,8 @@ type Option func(*config)
 // Internal config struct for building SDK options.
 type config struct {
 	tls               grpc.DialOption
-	clientCredentials oauth.ClientCredentials
+	clientCredentials *oauth.ClientCredentials
+	tokenExchange     *oauth.TokenExchangeInfo
```
strantalis, line 15: Could we create a generic interface here for the different flows?

Dockerfile
strantalis: I see what you did now 😄

sdk/options.go
mkleene: I like the idea but I think that if we wanted to we'd need to do something like
```
`config` -> `flow interface` -> `use by sdk`
```
because of how the functional options work; they each set an individual property that influences
how things work together.

mkleene: I also think that the `TokenSource` might be this interface. The flow is then:
1. set options on config using function options
2. turn the options into an instance of `TokenSource`
3. use this `TokenSource` whenever we want an access token

sdk/idp_token_exchange_token_source.go
```diff
@@ -0,0 +1,49 @@
+package sdk
```
dmihalcik-virtru, line 1: Can this be placed in an `internal` package?

dmihalcik-virtru (commented):
Do you have a guide for how to test this?

mkleene:
> Do you have a guide for how to test this?

It has an integration test against keycloak atm which I think is a decent-ish doc. I can add some instructions or something in `examples` if you think that makes sense.

dmihalcik-virtru (approved):

strantalis (approved):

PR #553, opened 2024-04-10
chore(lint): gomnd lint fixes & nolint directives

dmihalcik-virtru:


service/internal/auth/authn.go
```diff
@@ -56,6 +56,8 @@ var (
 	}
 )
 
+const refreshInterval = 15 * time.Minute
```
jrschumacher, line 59: This seems like it should be configurable. Wouldn't the platform admin want to control this variable across PEPs and SDK alike?

dmihalcik-virtru: Maybe? This looks to be a parameter to see how long the IdP's cache parameter is stored by default. I don't think it has any effect other than avoiding overwhelming the IdP with requests for the current JWKS and improving response time for clients. This probably should be able to be set to 0 at least. I'll expose it

dmihalcik-virtru: I've the refresh interval duration as a new field in the existing `AuthN` config object. It looks like viper default values must be configured at the root, so I've added the value there but we could add a 'set defaults' function to the auth package if you think it is worthwhile

jrschumacher (approved):

PR #556, opened 2024-04-10
fix(kas): oidc issuer from config

pflynn-virtru:
For KAS, load the OIDC issuer from the `opentdf.yaml` config.

- Added `issuer` and `discovery` 
```yaml
  kas:
    enabled: true
    issuer: http://localhost:8888/auth/realms/opentdf
    discovery: http://127.0.0.1:4444
```

### DIFF
--- a/opentdf-example.yaml
+++ b/opentdf-example.yaml
@@ -11,6 +11,7 @@ logger:
 services:
   kas:
     enabled: true
+    issuer: http://localhost:8888/auth/realms/opentdf
   policy:
     enabled: true
   authorization:
--- a/opentdf-example.yaml
+++ b/opentdf-example.yaml
@@ -12,6 +12,7 @@ services:
   kas:
     enabled: true
     issuer: http://localhost:8888/auth/realms/opentdf
+    discovery: http://localhost:8888/auth/realms/opentdf
   policy:
     enabled: true
   authorization:
--- a/opentdf-with-hsm.yaml
+++ b/opentdf-with-hsm.yaml
@@ -11,6 +11,7 @@ logger:
 services:
   kas:
     enabled: true
+    issuer: http://localhost:8888/auth/realms/opentdf
   policy:
     enabled: true
   authorization:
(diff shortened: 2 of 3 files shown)


opentdf-example.yaml
```diff
@@ -11,6 +11,8 @@ logger:
 services:
   kas:
     enabled: true
+    issuer: http://localhost:8888/auth/realms/opentdf
+    discovery: http://localhost:8888/auth/realms/opentdf
```
pflynn-virtru, line 15: `discoveryBase` or anything camelcase doesn't work, so just one word, all lower

service/kas/kas.go
```diff
@@ -15,18 +15,29 @@ import (
...
 	ctx := context.Background()
+	if cfg.ExtraProps == nil || cfg.ExtraProps["issuer"] == nil {
+		panic(errors.New("services.kas.issuer is required"))
+	}
+	oidcIssuerURL, ok := cfg.ExtraProps["issuer"].(string)
+	if !ok {
+		panic(errors.New("services.kas.issuer must be a string"))
+	}
+	if cfg.ExtraProps == nil || cfg.ExtraProps["discovery"] == nil {
+		panic(errors.New("services.kas.discovery is required"))
```
dmihalcik-virtru, line 28: it can default to issuer if not found

opentdf-with-hsm.yaml
```diff
@@ -11,6 +11,8 @@ logger:
 services:
   kas:
     enabled: true
+    issuer: http://localhost:8888/auth/realms/opentdf
```
mkleene, line 14: `issuer` is listed in the OIDC discovery URL. is this `issuer` intended as an override? If it is maybe we should clarify.

pflynn-virtru: good point.  the code should change to reflect that this is an override.  The initial reason is private and public URLs can exist for the same IdP

strantalis:
#474 Resolves this.

PR #557, opened 2024-04-10
fix(policy): return `created_at` and `updated_at` timestamps in CREATE metadata

suchak1:


service/policy/db/attributes.go
```diff
@@ -468,7 +468,7 @@ func createAttributeSql(namespaceId string, name string, rule string, metadata [
 		Insert(t.Name()).
 		Columns("namespace_id", "name", "rule", "metadata").
 		Values(namespaceId, name, rule, metadata).
-		Suffix("RETURNING \"id\"").
+		Suffix("RETURNING \"id\", " + getMetadataField("", false)).
```
jakedoublev, line 471:
```suggestion
               //nolint:goconst // Dynamic SQL query substring should not be an abstracted constant
		Suffix("RETURNING \"id\", " + getMetadataField("", false)).
```

jakedoublev (dismissed):
LGTM, but I think we need clarity about committing the `go.work.sum` here and @opentdf/opentdf-architecture approval because of that changed file.

service/policy/db/attribute_values.go
```diff
@@ -145,7 +145,7 @@ func createAttributeValueSql(
 			value,
 			metadata,
 		).
-		Suffix("RETURNING id").
+		Suffix(createSuffix).
```
jrschumacher, line 148: Big brain @suchak1 🤌

service/policy/db/subject_mappings.go
```diff
@@ -324,9 +324,17 @@ func (c PolicyDBClient) CreateSubjectConditionSet(ctx context.Context, s *subjec
...
+	if err = r.Scan(&id, &metadataJSON); err != nil {
 		return nil, db.WrapIfKnownInvalidQueryErr(err)
 	}
+
+	if metadataJSON != nil {
+		if err = protojson.Unmarshal(metadataJSON, m); err != nil {
+			slog.Error("could not unmarshal metadata", slog.String("error", err.Error()))
+			return nil, err
+		}
+	}
```
jrschumacher, line 336: Along the lines of the other enhancements, would it make sense to abstract this as well?

```go
// the function could do the condition validation
if err := unmarshalMetadata(metadataJSON, m); err != nil
  // the function could slog the error on error
  return nil, err
}
```

suchak1: done

jakedoublev (approved):

PR #568, opened 2024-04-12
Subject mapping OPA builtin for condition evaluation and jq selection

elizabethhealy:


service/internal/subjectmappingbuiltin/jq_eval_test.go
pflynn-virtru, line 1: should `jq_eval_test.go` this be removed?

pflynn-virtru (approved):

service/internal/subjectmappingbuiltin/jq_eval_test.go
elizabethhealy: we still have that jq eval function its just not in a builtin anymore, just called from the subject mapping eval

PR #570, opened 2024-04-12
chore: updates provision for web clients

dmihalcik-virtru:
- Adds a password so `sampleuser` can log in directly over the web
- Adds redirect URIs to localhost so web apps can redirect to themselves


dmihalcik-virtru:
Do this in the client or e2e test code instead of hard coding sample data here

PR #573, opened 2024-04-12
feat(provisioning): Keycloak provisiong from custom config and policy data bootstrapping from config

elizabethhealy:


jrschumacher:
@elizabethhealy this is awesome work! Would this deprecate the other key cloak provisioner? If so, could we validate that it can do the same thing and remove the defunct code with this PR?

elizabethhealy:
> @elizabethhealy this is awesome work! Would this deprecate the other key cloak provisioner? If so, could we validate that it can do the same thing and remove the defunct code with this PR?

@jrschumacher not fully, this doesn't support the token exchange setup yet, i can work on adding that

strantalis:
@elizabethhealy Can you sync with @mkleene. I think he has some keycloak command changes in pr #546 as well.

elizabethhealy:
@jrschumacher i cleaned this up a little to remove some of the duplication with the other keycloak provisioning, the other kc provision command still exists but they now use the same functions

PR #574, opened 2024-04-12
fix(core): fix panic on nil pointer dereference by passing KAS the SDK instance on registration

jakedoublev:


service/kas/kas.go
```diff
@@ -58,7 +58,7 @@ func NewRegistration() serviceregistry.Registration {
 		ServiceDesc: &kaspb.AccessService_ServiceDesc,
 		RegisterFunc: func(srp serviceregistry.RegistrationParams) (any, serviceregistry.HandlerServer) {
 			// FIXME msg="mismatched key access url" keyAccessURL=http://localhost:9000 kasURL=https://:9000
```
jakedoublev, line 60: @pflynn-virtru was there an issue to track this?

service/kas/kas.go
```diff
@@ -58,7 +58,7 @@ func NewRegistration() serviceregistry.Registration {
...
-			kasURLString := "https://" + srp.OTDF.HTTPServer.Addr
+			kasURLString := "http://localhost" + srp.OTDF.HTTPServer.Addr
```
dmihalcik-virtru, line 61:
```suggestion
			hostWithPort := srp.OTDF.HTTPServer.Addr
			if strings.HasPrefix(hostWithPort, ":") {
				hostWithPort = "localhost" + hostWithPort
			}
			kasURLString := "http://" + hostWithPort
```

dmihalcik-virtru (changes requested):

service/kas/kas.go
dmihalcik-virtru: We probably should also somehow make the scheme (http or https) configurable

dmihalcik-virtru (approved):

service/kas/kas.go
dmihalcik-virtru: I've filed #578. Feel free to comment there with more details, @jakedoublev or @pflynn-virtru

PR #575, opened 2024-04-12
feat(sdk): add assertion to tdf3

### LINKED ISSUE #1242: Go SDK ZTDF Assertion Support
Update the Go sdk to support ZTDF Assertions
* creating assertions
* validating assertions 
* reading assertions metadata 

This support will also be added to the Java and Javascript SDKs.

sujankota:
Resolves #1242


sdk/assertion.go
ttschampel, line 1: Can we move the assertion and related structs to proto so we can use in other APIs

sdk/manifest.go
```diff
@@ -58,9 +58,30 @@ type EncryptionInformation struct {
 	IntegrityInformation `json:"integrityInformation"`
 }
 
+type Statement struct {
+	Format StatementFormat `json:"format,omitempty"`
+	Schema Schema          `json:"schema,omitempty"`
```
ttschampel, line 63: I think these two should be strings; both of theses could be custom per the deployment domain and the enumerated types defined are confusing b/c they both restrict us and don't actually contain enumerated types we know are needed in current day customer deployments

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+	var statementFormat string
+	err := json.Unmarshal(data, &statementFormat)
+	if err != nil {
+		return err
+	}
+	if *s, err = ParseStatementFormat(statementFormat); err != nil {
+		return err
+	}
+	return nil
+}
```
strantalis, line 268:
```suggestion
```

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+	var schema string
+	err := json.Unmarshal(data, &schema)
+	if err != nil {
+		return err
+	}
+	if *s, err = ParseSchema(schema); err != nil {
+		return err
+	}
+	return nil
+}
```
strantalis, line 205:
```suggestion
```

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+	"fmt"
+	"strings"
+)
+
+// AssertionType - Custom type to hold value for the assertion ranging from 0-1
+type AssertionType uint
+
+const (
+	handlingAssertion AssertionType = iota + 1
+	baseAssertion
```
strantalis, line 14: These need to be public so a pep can set this on the assertion.

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+	return nil
+}
+
+// AppliesToState - Custom type to hold value for the assertion ranging from 0-1
+type AppliesToState uint
+
+const (
+	encrypted AppliesToState = iota + 1
+	unencrypted
+)
```
strantalis, line 114: Need to be public

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+
+var (
+	AppliesToStateName = map[uint8]string{
+		1: "encrypted",
+		2: "unencrypted",
+	}
+	AppliesToStateValue = map[string]uint8{
+		"encrypted":   1,
+		"unencrypted": 2,
+	}
```
strantalis, line 124: I think these should be private

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+var (
+	AssertionTypeName = map[uint8]string{
+		1: "Handling",
+		2: "Base",
+	}
+	AssertionTypeValue = map[string]uint8{
+		"Handling": 1,
+		"Base":     2,
+	}
+)
```
strantalis, line 26: These can be private

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+const (
+	trustedDataObj Scope = iota + 1
+	paylaod
+	explicit
+)
+
+var (
+	ScopeName = map[uint8]string{
+		1: "TDO",
+		2: "PAYL",
```
strantalis, line 69:
```suggestion
		2: "PAYLOAD",
```

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+		3: "EXPLICIT",
```
strantalis, line 70: @ttschampel @biscoe916 We want to remove explicit right?
```suggestion
```

sdk/tdf_config.go
```diff
@@ -186,6 +186,16 @@ func WithSegmentSize(size int64) TDFOption {
...
 
+// WithAssertions  returns an Option that add assertions to TDF.
+func WithAssertions(assertionList ...Assertion) TDFOption {
+	return func(c *TDFConfig) error {
+		newAssertions := make([]Assertion, 0)
+		newAssertions = append(newAssertions, assertionList...)
+		c.assertions = newAssertions
+		return nil
+	}
+}
```
strantalis, line 197: Should we define an `AssertionConfig` struct so only the fields we want set are allowed to be. For example should a user be allowed to pass in their own binding?

sujankota: I'm not sure about binding. They can pass in RSA key for signing and verification. I agree we should have AssertionConfig so we can pass Keys.

sdk/assertion.go
strantalis: We shouldn't have anything in here specific to ictdf. Do we even need scope. Won't it always just be the TrustedDataObject when converting?

strantalis: If we need this then we should match what is defined in the acp which is `payload` and `tdo`

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+		return err
+	}
+	return nil
+}
+
+// BindingMethod -  Custom type to hold value for the binding method from 0
+type BindingMethod uint
+
+const (
+	JWT BindingMethod = iota + 1
```
strantalis, line 274:
```suggestion
	JWS BindingMethod = iota + 1
```

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+)
+
+var (
+	BindingMethodName = map[uint8]string{
+		1: "jwt",
```
strantalis, line 279:
```suggestion
		1: "jws",
```

sdk/assertion.go
```diff
@@ -1,4 +1,313 @@
...
+	}
+	BindingMethodValue = map[string]uint8{
+		"jwt": 1,
```
strantalis, line 282:
```suggestion
		"jws": 1,
```

lib/ocrypto/crypto_utils.go
```diff
@@ -74,3 +77,25 @@ func RandomBytes(size int) ([]byte, error) {
 
 	return data, nil
 }
+
+// PemToRSAPrivate Convert pem formatted key to  ASN.1 DER form
+func PemToRSAPrivate(privateKeyInPem string) (*rsa.PrivateKey, error) {
```
strantalis, line 82: @sujankota Is this being used somewhere?

sdk/manifest.go
```diff
@@ -51,9 +51,29 @@ type EncryptionInformation struct {
 	IntegrityInformation `json:"integrityInformation"`
 }
 
+type Statement struct {
```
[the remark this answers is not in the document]
strantalis: @dmihalcik-virtru I think [this](https://github.com/opentdf/spec/blob/feature/json-schema_steph/schema/manifest-json.md#assertions) branch in the spec repo or the acp has more information.

sdk/assertion.go
sujankota: image

I think we both are referring to different documents.

sdk/assertion.go
biscoe916: Yes, I don't think we need explicit.

sdk/assertion.go
biscoe916: > We shouldn't have anything in here specific to ictdf. Do we even need scope. Won't it always just be the TrustedDataObject when converting?

We do need scope. But when in ZTDF form, it should be "payload", and "tdo".

sdk/assertion.go
strantalis: I am wonder more and more if we should just do this for the whole manifest.

sdk/tdf_config.go
```diff
@@ -43,6 +44,101 @@ type KASInfo struct {
...
+	}
+	return AssertionSigningKeyType(value), nil
+}
+
+type TDFAssertionOption func(*AssertionConfig) error
+
+type AssertionConfig struct {
+	keyType AssertionSigningKeyType
+	payload []byte // If the type
+}
```
strantalis, line 89: I was thinking the assertion config would be a subset of fields we want the user to set. Because right now we have the WithAssertion option which takes an assertion allowing a user to set the binding themselves. Which I'm not sure we want.

@biscoe916 Any thoughts on this?

sdk/tdf_config.go
```diff
@@ -43,6 +44,101 @@ type KASInfo struct {
...
+
+// WithHS256UserDefinedKey returns an Option that add HS256 key for signing and verifying assertion.
+func WithHS256UserDefinedKey(key []byte) TDFAssertionOption {
+	return func(a *AssertionConfig) error {
+		a.keyType = AssertionSigningKeyHS256UserDefined
+		a.payload = key
+		return nil
+	}
+}
```
strantalis, line 140: I think having this many key options can get confusing. Can we do something similar like other pkgs and do WithAssertionKey(alg, key)

sdk/tdf_config.go
sujankota: The user should not set the binding. I'll make it private

strantalis: That is confusing from a developer experience perspective. That's why I'm suggesting changes.

biscoe916 (approved):

strantalis (approved):

PR #588, opened 2024-04-16
chore: rm go.work.sum

pflynn-virtru:
The Go toolchain behaves differently between Go versions.  The `go.work.sum` is one example.  Hence, we have removed and ignored, relying on IDEs and tools to compute it at develop/build time.


strantalis (dismissed):

pflynn-virtru:
on hold: seeing if https://github.com/opentdf/platform/pull/599 solves are issues

jrschumacher (approved):

PR #555, opened 2024-04-10
build: Updates make fix to run goimports

dmihalcik-virtru:

PR #592, opened 2024-04-17
feat(crypto): nanotdf crypto helper methods

sujankota:
Adds several EC functions to the ocrypto library

Needed for PLAT-3082


dmihalcik-virtru:
It looks like your IDE isn't using goimports for formatting. 

Feel free to approve https://github.com/opentdf/platform/pull/555 if you want something to do it outside of an IDE with an extension to `make fix`

lib/ocrypto/rsa_key_pair.go
```diff
@@ -72,3 +78,122 @@ func (keyPair RsaKeyPair) KeySize() (int, error) {
 	}
 	return keyPair.privateKey.N.BitLen(), nil
 }
+
```
patmantru, line 81: Shouldn't all this EC stuff be in ec_key_pair.go?

patmantru (changes requested):

pflynn-virtru (approved):

patmantru (approved):

patmantru (dismissed):

lib/ocrypto/ec_key_pair.go
```diff
@@ -97,3 +103,122 @@ func (keyPair ECKeyPair) KeySize() (int, error) {
...
+		return ConvertToECDHPublicKey(pub)
+	default:
+		break
+	}
+
+	return nil, fmt.Errorf("not an ec PEM formatted public key")
+}
+
+// ECPrivateKeyFromPem generate ec private from pem format
+func ECPrivateKeyFromPem(privateECKeyInPem string) (*ecdh.PrivateKey, error) {
```
strantalis, line 185:
```suggestion
func ECPrivateKeyFromPem(privateECKeyInPem []byte) (*ecdh.PrivateKey, error) {
```

lib/ocrypto/ec_key_pair.go
```diff
@@ -97,3 +103,122 @@ func (keyPair ECKeyPair) KeySize() (int, error) {
...
+	_, err := io.ReadFull(hkdfObj, derivedKey)
+	if err != nil {
+		return nil, fmt.Errorf("failed to derive hkdf key: %w", err)
+	}
+
+	return derivedKey, nil
+}
+
+// ECPubKeyFromPem generate ec public from pem format
+func ECPubKeyFromPem(pemECPubKey string) (*ecdh.PublicKey, error) {
```
strantalis, line 149:
```suggestion
func ECPubKeyFromPem(pemECPubKey []byte) (*ecdh.PublicKey, error) {
```

lib/ocrypto/ec_key_pair.go
```diff
@@ -97,3 +103,122 @@ func (keyPair ECKeyPair) KeySize() (int, error) {
...
+		return ConvertToECDHPrivateKey(privateKey)
+	default:
+		break
+	}
+
+	return nil, fmt.Errorf("not an ec PEM formatted private key")
+}
+
+// ComputeECDHKey calculate shared secret from public key from one party and the private key from another party.
+func ComputeECDHKey(privateKeyInPem string, publicKeyInPem string) ([]byte, error) {
```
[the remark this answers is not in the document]
strantalis: I have questions around this as well. I think we need to be careful wrapping functionality in to many helpers as this can sometimes make things more difficult to follow and understand.  

What is the use case for this function?

strantalis (commented):
@sujankota I think we need to avoid taking in pems as strings. Almost all the time if we read from a file it's going to be a `[]byte` and not a string. 

Mayb this [field](https://github.com/opentdf/platform/blob/main/service/kas/access/rewrap.go#L40) should be changed?

lib/ocrypto/ec_key_pair.go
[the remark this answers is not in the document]
sujankota: The use case: SDK public key(pem) from rewrap request with KAS private key to generate symmetric key. Both are in pem formats.

PR #600, opened 2024-04-17
fix(sdk): add example for token exchange

mkleene:
also
* pull internal sdk classes into internal module
* have a better (varargs) interface for audiences since there may be many, in general


dmihalcik-virtru (approved):

sdk/internal/idp_access_token_source.go
strantalis, line 1: @mkleene Did you mean to move these to the root of internal? They should live inside a oauth pkg if we are going ot move them.

PR #601, opened 2024-04-17
feat: ability to add public routes that bypass authn middleware

strantalis:


service/internal/auth/authn.go
```diff
@@ -452,3 +456,17 @@ func validateDPoP(accessToken jwt.Token, acessTokenRaw string, dpopInfo dpopInfo
 	}
 	return &dpopKey, nil
 }
+
+func (a Authentication) isPublicRoute(path string) func(string) bool {
+	return func(route string) bool {
+		pattern := fmt.Sprintf("^%s$", route)
```
dmihalcik-virtru, line 462: Use `filepath.Match` here instead of regex

https://pkg.go.dev/path/filepath#Match

service/internal/auth/config.go
```diff
@@ -4,8 +4,9 @@ import "fmt"
 
 // AuthConfig pulls AuthN and AuthZ together
 type Config struct {
-	Enabled     bool `yaml:"enabled" default:"true" `
-	AuthNConfig `mapstructure:",squash"`
+	Enabled      bool     `yaml:"enabled" default:"true" `
+	PublicRoutes []string `mapstructure:"-"`
```
dmihalcik-virtru, line 8: What does the `"-"` mean?

dmihalcik-virtru (changes requested):

service/internal/auth/config.go
strantalis: basically ignores the field when encoding/decoding the struct.

service/internal/auth/authn.go
strantalis: I am not sure if filepath match is flexible enough here.  If I have these allowed routes `"/static/*", "/static/*/*"` the below test will fail. 

`s.Require().True(slices.ContainsFunc(s.auth.publicRoutes, s.auth.isPublicRoute("/static/test/next/maybe")))`

dmihalcik-virtru: Do we need it to?

service/internal/auth/config.go
dmihalcik-virtru: Isn't the point that this can be extended? Or does this force it to be configured in code?

service/internal/auth/authn.go
dmihalcik-virtru: Also this is not a 'safe' way to convert to a regex. You should be splitting on the asterisks, then replacing each segment with a `regex.QuoteMeta` version of the segment

jrschumacher: Yea I like the glob approach better. I'm concerned that regex might introduce exposing a route that shouldn't be exposed unless we can bind it to the namespace.

service/internal/auth/authn.go
```diff
@@ -452,3 +456,17 @@ func validateDPoP(accessToken jwt.Token, acessTokenRaw string, dpopInfo dpopInfo
...
+		pattern = strings.ReplaceAll(pattern, "*", ".*")
```
jrschumacher, line 463: The risk here is that `.*` with turn into `..*` right?

service/internal/auth/authn.go
strantalis: Its updated to use `filepath.Match for now`. We can circle back if/when we need a regex solution.

service/internal/auth/authn.go
strantalis: This is removed in place of filepath.Match

service/internal/auth/config.go
strantalis: Forces it to be configured through code. I felt like it could potentially open up issues if we allowed it to be configured via the config file.

dmihalcik-virtru (approved):

jrschumacher (approved):

Issue #566, opened 2024-04-11
Add the ability to disable `DPoP`


mkleene:
Customers will have a variety of IdPs. Some of them do not implement DPoP. Allowing these customers to use these IdPs could be supported by adding a switch that disables DPoP checking.

This does weaken security somewhat so it's important to consider the implications.

Another idea would be to allow the IdP to control this. In this case we would only validate tokens that we detect are DPoP via the method here: https://github.com/opentdf/platform/issues/409

pflynn-virtru:
For maximum OIDC compatibility, It would be nice for the SDK and Service be able to adjust automatically to the capabilities of the IdP.
- SDK sends `dpop` header to token_endpoint, if it does not get a `cnf` field in the JWT, then set `Bearer` in `Authorization` header.
- Service  authentication does standard checks if `Bearer` is set, else DPoP check
- Add configuration to set `dpop.required=true` if the administrator of the OpenTDF service requires it.

pflynn-virtru:
relates to https://github.com/opentdf/platform/issues/409

mkleene:
> For maximum OIDC compatibility, It would be nice for the SDK and Service be able to adjust automatically to the capabilities of the IdP.
> 
>     * SDK sends `dpop` header to token_endpoint, if it does not get a `cnf` field in the JWT, then set `Bearer` in `Authorization` header.
> 
>     * Service  authentication does standard checks if `Bearer` is set, else DPoP check
> 
>     * Add configuration to set `server.auth.dpop.required=true` if the administrator of the OpenTDF service requires it.

I think that the second point isn't the best way to do things since it would enable spoofing. We should do things based on the structure of the token we are passed.

I like adding a config option that makes us enforce DPoP!

damorris25:
Agreed - we should allow the administrator to configure if DPoP is enabled or not. Trying to make the SDK's 'smart' enough to figure out if DPoP is enabled is likely a way to open a door for attack vs simply making it an administrator switch. That would match with how many other security settings are controlled in the overall software stack.

strantalis:
@dmihalcik-virtru We plan on moving forward with adding a enforce dpop setting in the platform so by default it will be more of a soft check. If dpop is presented in the request we will validate it. Otherwise we will just validate the bearer token. This means KAS will ignore checking the signed request body. 

Can you confirm we don't lose anything by not checking that signed body in a rewrap request?

mkleene:
[...]

PR #617, opened 2024-04-18
feat(sdk): make enforcement of DPoP optional

mkleene:
* add `auth.enforceDPoP`. enabling this setting tells the auth middleware to accept tokens that do not have a `cnf` claim as valid. since having DPoP or not is a property this config goes along with each issuer
* change the return of `checkToken` to return a context that has the right stuff in it so that we don't have to validate that something that came back without an error is non-nil
* make it possible to disable auth on rewrap. currently we can't because we require there be a dpop JWK

addresses https://github.com/opentdf/platform/issues/566


service/internal/auth/config.go
```diff
@@ -4,12 +4,12 @@ import "fmt"
 
 // AuthConfig pulls AuthN and AuthZ together
 type Config struct {
-	Enabled     bool `yaml:"enabled" default:"true" `
 	AuthNConfig `mapstructure:",squash"`
 }
 
 // AuthNConfig is the configuration need for the platform to validate tokens
 type AuthNConfig struct {
+	AllowNoDPoP       bool   `yaml:"allowNoDPoP" json:"allowNoDPoP" default:"true"`
```
dmihalcik-virtru, line 12: Does this change the behavior?

```suggestion
	AllowNoDPoP       bool   `yaml:"allowNoDPoP" json:"allowNoDPoP"`
```

service/internal/auth/config.go
```diff
@@ -4,12 +4,12 @@ import "fmt"
 
 // AuthConfig pulls AuthN and AuthZ together
 type Config struct {
-	Enabled     bool `yaml:"enabled" default:"true" `
```
dmihalcik-virtru, line 7: Suggestion: Rename field to `DeprecatedEnabled` and print a warning or error out if it is set to `false`. This will help people transition to the new config format

dmihalcik-virtru (approved):
Optional improvement to encourage upgrading to new config

service/internal/auth/config.go
mkleene: I like it! I'll do that. thanks!

dmihalcik (approved):

dmihalcik-virtru (approved):

dmihalcik-virtru (dismissed):

docs/configuration.md
```diff
@@ -41,6 +41,7 @@ The server configuration is used to define how the application runs its server.
 | `tls.key` | The path to the tls key. | |
 | `auth.audience` | The audience for the IDP. | |
 | `auth.issuer` | The issuer for the IDP. | |
+| `auth.allowNoDPoP` | If true, we allow access tokens that do not have DPoP bindings. | `false` |
```
strantalis, line 44: The only concern I have here is we need a general platform setting that will drive the sdk as well. Unless we can get that information from the IDP openid-configuration which didn't sound reliable from @mkleene. 

@dmihalcik-virtru @biscoe916  Any thoughts on this?

strantalis: Ok I can get on board with that. But can we change the name of the config option. 

- skipDPoPVerification
- bypassDPoPCheck
- disableDPoPValidation
- dpopDisabled

mkleene: Those are all OK but they seem inaccurate to me. What we are doing is allowing access tokens that are not DPoP bound. All of those suggest (to me) that we are going to allow DPoP that is wrong through.

jrschumacher: OpenSSH seems to utilizes the word `strict` when it comes to this kind of behavior (i.e. `StrictModes`, `StrictHostKeyChecking`, `GSSAPIStrictAcceptorCheck`) both on client and in `/etc/sshd/sshd_config`.

What I would encourage is to focus firstly on a standard config syntax to communicate various behaviors. In this way, customer understanding today and the future is derived by manuals and consistency.

For this specific problem, it seems that Keycloak uses a positive language which resolves to `yes|no`. [DPOP_BOUND_ACCESS_TOKENS](https://github.com/keycloak/keycloak/blob/main/server-spi-private/src/main/java/org/keycloak/protocol/oidc/OIDCConfigAttributes.java#L49)

The language in their documentation seems to align with our implementation, no?

> If the switch OAuth 2.0 DPoP Bound Access Tokens Enabled is off, the **client can still send** DPoP proof in the token request. In that case, {project_name} will verify DPoP proof and will add the thumbprint to the token. But if the switch is off, DPoP binding is not enforced by the {project_name} server for this client. It is recommended to have this switch on if you want to make sure that particular client always uses DPoP binding.
>
> [Ref](https://github.com/keycloak/keycloak/blob/071032a108bd9e9fce9e66d00c36d56bd4b334df/docs/documentation/server_admin/topics/clients/oidc/con-advanced-settings.adoc#L100:~:text=OAuth%202.0%20Demonstrating%20Proof%2Dof%2DPossession%20at%20the%20Application%20Layer)

service/internal/auth/config.go
```diff
@@ -4,12 +4,13 @@ import "fmt"
...
-	AuthNConfig `mapstructure:",squash"`
+	DeprecatedDisabled bool `yaml:"deprecatedDisabled" default:"false"`
```
strantalis, line 7: @mkleene 
We can't rename the config value. This would become a breaking change. We should leave the option as is. Add comments and docs  that this is deprecated. Then remove it in the future.

mkleene: removed it

docs/configuration.md
mkleene: I think that any of these are fine. My only reservation (at the code level) is that `false` is the zero value for booleans so if this ever gets created without reading the docs carefully then we won't be verifying DPoP.

dmihalcik-virtru: I'm fine with this for now. Some suggestions for a false-first name for this would be:

- `dpopOptional`
- `allowUnboundTokens

dmihalcik-virtru (approved):

Issue #505, opened 2024-04-04
Policy API: Extend error messaging with wrong parameter info to improve clarity and not mislead user.


sievdokymov-virtru:
**Overview:**
In some cases response error message looks misleading as it doesn’t contain specific parameter names which may cause wrong users perception. So it would be good to extend error messages with more detailed info e.g. wrong params info. 

E.g. case of AttributeService.CreateAttributeValue gRPC request complaining on wrong _**value**_ parameter while actually 'members' param value causes an error: 
------

**Steps to Reproduce:**
1. Run the backend server for the opened/platform (as described in the [Readme](https://github.com/opentdf/platform/blob/main/README.md)).
2. Create a Namespace and an Attribute.
3. Perform AttributeService.CreateAttributeValue gRPC request with filling next parameters: valid _id_ of existed attribute,  valid _value_ of String format, INVALID 'members' value of string format (while it should be uuid as per docs) .
4. Invoke the request and observe response.

**Actual Result:**
Request fails with “3 INVALID_ARGUMENT” code in response complaining on non-uuid format of the **_value_** parameter:
<img width="1278" alt="ExtendErrorMessage" src="https://github.com/opentdf/platform/assets/100794336/4c27884b-c187-4f3f-855f-63dfc695be79">

So it looks like _value_ param is wrong while actually _members_ value is wrong.

**Expected Result:**
Extend the error message with wrong param info so that user can clearly understand what is wrong in the request.

jakedoublev:
I think this might be an error message issue @sievdokymov-virtru and actually stemming from invalid `members` in your example. Could you please let me know if you're able to reproduce using valid members? 

The following succeeds for me, where `e1536f25-d287-43ed-9ad9-2cf4a7698e5f` is one of the sample provisioned attribute ids and already exists in the DB. 
```bash
grpcurl -plaintext -d '{"attribute_id":"e1536f25-d287-43ed-9ad9-2cf4a7698e5f","metadata":{"labels":[{"value":"id","key":"quis esse do tempor"}]},"value":"VALUE123"}' localhost:8080 policy.attributes.AttributesService/CreateAttributeValue
{
  "value": {
    "id": "decfde39-fd1a-4959-aebe-c3d21b46d12b",
    "attribute": {
      "id": "e1536f25-d287-43ed-9ad9-2cf4a7698e5f"
    },
    "value": "value123",
    "active": true,
    "metadata": {
      "labels": {
        "quis esse do tempor": "id"
      }
    }
[...]

PR #633, opened 2024-04-22
fix(db): invalid uuid error message

suchak1:


jakedoublev:
@sievdokymov-virtru this meets your needs, correct?

jakedoublev (changes requested):
This is nice cleanup, but it looks like the same error is returned from this wrapper function, `db.ErrUUIDInvalid`, which reads `value not a valid UUID`? In that case, it seems like it will still not be clear to the requester that the _member_ id was invalid and not the _value_ id as the original opener of the issue found.

suchak1:
Server responses are again with the `StatusifyError` fx which replaces `ErrUUIDInvalid` with `ErrTextEnumValueInvalid`. I believe external users and @sievdokymov-virtru use the server / AttributeService to place the request in https://github.com/opentdf/platform/issues/505. If we believe internal developers will be confused by `ErrUUIDInvalid`, I can replace the text for that as well.

jakedoublev (approved):

strantalis (approved):

PR #635, opened 2024-04-22
fix(entity-resolution): Support keycloak entity resolution for jwt entities

elizabethhealy:
Allow for entity resolution for jwts, get the client/user instead of pulling the subject external value field from the jwt directly

Question to think about: for a clientID, should we return both the client and its service account user (if exists)?


service/internal/idpplugin/keycloak_plugin.go
```diff
@@ -58,14 +58,47 @@ func EntityResolution(ctx context.Context,
...
+		case *authorization.Entity_Jwt:
+			entityJwt := ident.GetJwt()
+			// decode the jwt
+			_, mapClaims, err := connector.client.DecodeAccessToken(ctx, entityJwt, kcConfig.Realm)
+			if err != nil {
+				return &authorization.IdpPluginResponse{},
+					status.Error(codes.Internal, db.ErrTextCreationFailed)
+			}
+			// if it has client ID use that
+			clientID, ok := (*mapClaims)["client_id"].(string)
```
pflynn-virtru, line 83: @mkleene you found an oddity with `client-id`, I believe.  Is this correct logic for Keycloak?

mkleene: according to this: https://datatracker.ietf.org/doc/html/rfc9068, it's required, but there are lots of cases where it's not included

mkleene: and that's a proposed standard so it makes sense

pflynn-virtru (dismissed):

service/internal/idpplugin/keycloak_plugin.go
```diff
@@ -94,95 +127,90 @@ func EntityResolution(ctx context.Context,
...
-					if exErr != nil {
+		} else if getUserParams != (gocloak.GetUsersParams{}) {
+			// if getUserParams not empty
+			users, err := connector.client.GetUsers(ctx, connector.token.AccessToken, kcConfig.Realm, getUserParams)
+			if err != nil {
+				slog.Error(err.Error())
+				return &authorization.IdpPluginResponse{},
+					status.Error(codes.Internal, db.ErrTextGetRetrievalFailed)
+			}
+			if len(users) == 1 {
```
strantalis, line 138: Is there ever a chance we will somehow get more than 1 user?

service/internal/idpplugin/keycloak_plugin.go
```diff
@@ -94,95 +127,90 @@ func EntityResolution(ctx context.Context,
...
-						entityNotFoundErr = authorization.EntityNotFoundError{Code: int32(codes.NotFound), Message: db.ErrTextGetRetrievalFailed, Entity: ident.GetEmailAddress()}
-					case *authorization.Entity_UserName:
-						entityNotFoundErr = authorization.EntityNotFoundError{Code: int32(codes.NotFound), Message: db.ErrTextGetRetrievalFailed, Entity: ident.GetUserName()}
-					// case "":
-					// 	return &authorization.IdpPluginResponse{},
-					// 		status.Error(codes.InvalidArgument, db.ErrTextNotFound)
-					default:
-						slog.Error("Unsupported/unknown type for", "entity", ident.String())
-						entityNotFoundErr = authorization.EntityNotFoundError{Code: int32(codes.NotFound), Message: db.ErrTextGetRetrievalFailed, Entity: ident.String()}
+					if len(groups) == 1 {
```
strantalis, line 159: Same here is it possible to get more than one group?

pflynn-virtru: it is possible, let's modify the `slog.Error` to "Unexpected group count found for "entity", ident.String(), "count",  len(groups)"

PR #647, opened 2024-04-23
feat(core): add service scoped database clients

jrschumacher:
This PR introduces a significant change to how we perceive database clients within the platform. The original implementation has a concept of a single database client which services could use. As work was being executed, it became clear that the policy service was one (if not only) service which required a database client. 

Recently, we learned that downstream projects that wrap OpenTDF required access to a database client, but this posed a couple concerns:

1. Downstream clients need the ability to add migrations and have those bound to the migration CLI tool as well as the auto-migration feature
2. We need to ensure a bad actor cannot poison a PEP and thus have easy access to mutate the policy outside an audited and authenticated flow
3. We need to ensure PEPs are interacting with services via contracts and not by direct database access, which could slow down development

## Supported features

- create a new database client for each service / PEP
	- this sets us up to support limiting scope in the future by giving theoretical access to managing multiple DB user identities per instance of the platform
- scope the database client to a schema scoped to the service namespace
	- this will give us theoretical control over [limiting the access](https://stackoverflow.com/a/17355059/79201) of the database user
	- this might need to change to a database per service depending on the access controls needed
- add support for service based migrations
	- having all services use the same migrations directory is brittle since clients are scoped to schemas we can support multiple migration directories
	- additionally, when deploying a policy service apart from another PEP we don't want to automatically migrate a schema that won't be used
- migration commands now support service and/or `--all`
	- this enables developers to migrate up and down as they are exploring the process of supporting a new feature
	- this gives us the opportunity to enhance the tool to support migrations for services that are currently "enabled"

## Unsupported features

- robust support for service auto-migration based on server registration
- rigorous performance testing
- DX review


service/kas/access/accessPdp.go
```diff
@@ -47,7 +47,7 @@ func checkDissems(dissems []string, ent *authorization.Entity) (bool, error) {
 
 func checkAttributes(ctx context.Context, dataAttrs []Attribute, ent *authorization.Entity, sdk *otdf.SDK) (bool, error) {
 	ec := authorization.EntityChain{Entities: make([]*authorization.Entity, 0)}
-	ec.Entities = append(ec.Entities, ent)
+	ec.Entities = append(ec.GetEntities(), ent)
```
jrschumacher, line 50: @dmihalcik-virtru this was a lint issue

service/pkg/server/services.go
```diff
@@ -27,3 +31,56 @@ func registerServices() error {
...
+	// Conditionally set the db client if the service requires it
+	// Currently, we are dynamically registering namespaces and don't offer the ability to apply
+	// config at the NS layer. This poses a problem where services under a NS want to share a
+	// database connection.
+	// TODO: this should be reassessed with how we handle registering a single namespace
+
+	slog.Info("creating database client", slog.String("namespace", reg.Namespace))
+	// Make sure we only create a single db client per namespace
+	var err error
+	d, err = db.New(cfg,
```
strantalis, line 56: @jrschumacher Maybe I'm missing something but are you creating a new pool for each service registered? 

If that's the case I think we need to share the connection pool. Especially if each service is using the shared db config.  

We could easily overwhelm any connection limits because it would be max connections * the number of services registered. Or we need to capture service specific db config potentially.

jrschumacher: Yea I expected there would be performance issues as we move to a production environment. 

I think this could get complex. If we have separate database servers per service we'd need to d to make sure they have seperate pools right?

We should probably talk about what we're going to support day one vs where we'd like to be in a few months.

strantalis: > I think this could get complex. If we have separate database servers per service we'd need to d to make sure they have seperate pools right?

Yes that's how it would work. But would require db specific config for each service. 

The way I have been thinking about this is if you need that separation just run that one service separate so it can have a different db config and isolation. There are probably couple ways to tackle this so we should probably sync on this sometime today.

strantalis (approved):

PR #651, opened 2024-04-23
chore(auth): DPoP and public fixes

dmihalcik-virtru:
- expose legacy public key endpoint as public
- DPoP `htu` should include origin part of url
- clarify error messages for dpop

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -56,8 +56,6 @@ server:
     key: /path/to/key
   auth:
     enabled: true
-    allowedHosts:
-      - "https://example.com"
     audience: https://example.com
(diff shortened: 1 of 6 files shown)


mkleene (commented):
Do we gain anything by only allowing some hosts? This seems to make sure that the token is for us. I'd say that the `aud` parameter is what we want here and this is just another thing that we need to configure.

There might be something that I'm missing about what this buys us, though.

dmihalcik-virtru:
Now it checks the `htu` field of the DPoP token matches the request Origin + Path

mkleene (approved):
This looks great!

Issue #658, opened 2024-04-24
Enable SonarCloud analysis and collect code coverage info


sievdokymov-virtru:
To add SonarCloud static code analysis for the opentdf/platform repo to identify actual and potential issues:
- enable sonar integration for the repo in terraform + disable automatic analysis for the repo (to be able to run CI-based analysis) 
- add Sonar analysis step/workflow to CI pipelines

Correspondent JIRA issue: https://virtru.atlassian.net/browse/PLAT-3122

sievdokymov-virtru:
Solution is merged in https://github.com/opentdf/platform/pull/659, workflow is green. Closing as Done.

PR #659, opened 2024-04-24
chore(main): Enable SonarCloud analysis with collecting code coverage values

sievdokymov-virtru:
In scope of #658 / PLAT-3122

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,8 +1,7 @@
 name: SonarCloud
 on:
-  push:
-    branches:
-      - feat/sonarcloud-analysis
+  schedule:
+    - cron: "0 5 * * 2,4,6"
 #  pull_request:
 #    types:
 #      - opened
@@ -43,7 +42,7 @@ jobs:
         with:
           name: code-coverage-report
           path: coverage.*
-          retention-days: 1
+          retention-days: 3
 
   sonarcloud:
     name: 'SonarCloud'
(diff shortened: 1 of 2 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,69 @@
+name: SonarCloud
+on:
+  schedule:
+    - cron: "0 5 * * 2,4,6"
+#  pull_request:
```
sievdokymov-virtru, line 5: Not sure whether we want to use it as a quality gate on PRs since it may introduce excessive confusion w/o proper announcement and preparation (At first, I'd agreed on following code coverage, code duplication rules internally)

dmihalcik-virtru: Additionally, PRs from forks will not have access to the secret. I'd suggest adding and `if` block to the `sonarcloud` job instead.

dmihalcik-virtru (approved):

.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: 'Download Code Coverage Report'
+        uses: actions/download-artifact@v3
+        with:
+          name: code-coverage-report
+
+      - name: 'SonarCloud Scan'
+        uses: SonarSource/sonarcloud-github-action@master
```
strantalis, line 67: This should reference a release commit sha. Not a branch. 

https://github.com/SonarSource/sonarcloud-github-action/releases

sievdokymov-virtru: Updated.

dmihalcik-virtru (approved):

Issue #567, opened 2024-04-11
Entitlements Endpoint: Update to align with ADR 472


ttschampel:
Update Entitlements Endpoint to Align with ADR #472  

The rough task list:

- Define entity resolution service interface (ERS)
- Move Keycloak IDP Plugin to a service implementing the ERS contract
- Move the JQ Custom Built-In Function to a Subject Mapping Built-In Function `resolve` encapsulating both the JQ Selector and condition evaluation currently residing in Rego.  
- Update the platform to use a single Core Rego file that consists of fetching the entity representation for a configured ERS endpoint and then passing to the new Subject Mapping resolve function
- Remove Keycloak specific inputs in OPA engine call.

jrschumacher:
@jakedoublev anything we should consider for policy with regards to acceptance criteria?

elizabethhealy:
i'm getting started on the third bullet `Move the JQ Custom Built-In Function to a Subject Mapping Built-In Function resolve encapsulating both the JQ Selector and condition evaluation currently residing in Rego.`

PR #660, opened 2024-04-24
feat(ers): Create entity resolution service, replace idp plugin

elizabethhealy:
[#567
](https://github.com/opentdf/platform/issues/567)
Define entity resolution service interface (ERS)
Move Keycloak IDP Plugin to a service implementing the ERS contract
Update the rego to fetch the entity representation for a configured ERS endpoint
Remove Keycloak specific inputs in OPA engine call.


ttschampel (commented):
Can we remove entitlements.rego and rename entitlements-keycloak.rego to entitlements.rego since it is now generic?

service/policies/entitlements/entitlements-keycloak.rego
```diff
@@ -24,10 +16,18 @@ else := {"entities": [{
...
+
 attributes := [attribute |
 	# external entity
-    response := keycloak.resolve.entities(idp_request, idp_config)
-    entity_representations := response.entityRepresentations
+    response := http.send({
+        "method" : "POST",
+        "url": input.ers_url,
+        "body": ers_request
+    })
```
strantalis, line 26: @elizabethhealy 
How is an auth header being added to this request? Or are we assuming this will be a public endpoint?

elizabethhealy: ooh this is a good question, i usually have auth disabled and forget about it.
we could assign a client to the authorization service and have that client retrieve a token under getEntitlements thats then passed into opa
other option is guess is to pass the authorization serivice assigned client and secret into opa and retrieve it from there but seems less safe

elizabethhealy: @strantalis is calling the token endpoint in authroization service then passing the token to rego any better? that way i guess we wouldnt have to pass the client credentials to rego to get the token

opentdf-dev.yaml
```diff
@@ -15,11 +15,17 @@ services:
     enabled: true
   authorization:
     enabled: true
-    url: http://localhost:8888
-    client: "tdf-entity-resolution"
-    secret: "secret"
+    ersUrl: http://localhost:8080/entityresolution/resolve
+    clientid: tdf-authorization-svc
+    clientsecret: secret
+    tokenEndpoint: http://localhost:8888/auth/realms/opentdf/protocol/openid-connect/token
```
strantalis, line 21: @elizabethhealy Can we resolve this from the issuer that is set in the auth config some how?

service/entityresolution/entity_resolution.proto
```diff
@@ -26,12 +23,12 @@ message IdpConfig {
...
 
 */
-message IdpPluginRequest {
+message ResolveEntitiesRequest {
     repeated authorization.Entity entities = 1;
 }
 
 
-message IdpEntityRepresentation {
+message ERSEntityRepresentation {
```
strantalis, line 31: Do we need to prefix this with ERS?

service/entityresolution/keycloak/keycloak_entity_resolution.go
```diff
@@ -33,28 +34,17 @@ type KeyCloakConnector struct {
 }
 
 func EntityResolution(ctx context.Context,
-	req *authorization.IdpPluginRequest, config *authorization.IdpConfig) (*authorization.IdpPluginResponse, error) {
+	req *entityresolution.ResolveEntitiesRequest, kcConfig KeycloakConfig) (entityresolution.ResolveEntitiesResponse, error) {
 	// note this only logs when run in test not when running in the OPE engine.
-	slog.DebugContext(ctx, "EntityResolution", "req", fmt.Sprintf("%+v", req), "config", fmt.Sprintf("%+v", config))
-	jsonString, err := json.Marshal(config.GetConfig().AsMap())
+	slog.DebugContext(ctx, "EntityResolution", "req", fmt.Sprintf("%+v", req), "config", fmt.Sprintf("%+v", kcConfig))
```
jakedoublev, line 39: Is it a problem if we log this config when it contains the client id and secret?

service/internal/auth/casbin.go
```diff
@@ -96,10 +96,12 @@ p,	role:readonly,		/kas.AccessService/LegacyPublicKey,			read,			allow
 ## gRPC routes
 p,	role:unknown,			kas.AccessService/LegacyPublicKey,			other,	allow
 p,	role:unknown,			kas.AccessService/PublicKey,						other,	allow
+p,	role:unknown,			entityresolution.EntityResolutionService.ResolveEntities,										write,		allow
```
jakedoublev, line 99: I could be missing something, but are these public routes, or service to service routes only?

service/policies/entitlements/entitlements.rego
```diff
@@ -2,32 +2,65 @@ package opentdf.entitlements
...
+	condition_group_evaluate(entity_representation.additionalProps, condition_group.boolean_operator, condition_group.conditions)
+}
 
 # condition_group
 condition_group_evaluate(payload, boolean_operator, conditions) if {
 	# AND
 	boolean_operator == 1
 	some condition in conditions
-	condition_evaluate(jq.evaluate(payload, condition.subject_external_selector_value), condition.operator, condition.subject_external_values)
+	# TODO: additional_props is a list of entity representations
```
jakedoublev, line 48: Does an issue already exist to track this question?

service/entityresolution/keycloak/keycloak_entity_resolution.go
elizabethhealy: with the debug context it should only log when running tests, not when running in the OPA engine

lib/fixtures/keycloak.go
```diff
@@ -290,6 +291,20 @@ func SetupKeycloak(ctx context.Context, kcConnectParams KeycloakConnectParams) e
...
 	}
 
+	// Create TDF Authorization Svc Client
+	_, err = createClient(ctx, client, token, &kcConnectParams, gocloak.Client{
+		ClientID:                gocloak.StringP(opentdfAuthorizationClientID),
+		Enabled:                 gocloak.BoolP(true),
+		Name:                    gocloak.StringP(opentdfAuthorizationClientID),
+		ServiceAccountsEnabled:  gocloak.BoolP(true),
+		ClientAuthenticatorType: gocloak.StringP("client-secret"),
+		Secret:                  gocloak.StringP("secret"),
```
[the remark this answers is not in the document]
elizabethhealy: this is the regular provision keycloak command, the other keycloak-from-config reads whats in the config

service/internal/auth/casbin.go
elizabethhealy: the casbin policy yes but auth requires a jwt, right now we dont care about what role the person has, just that a valid jwt is provided

elizabethhealy: its being hit from the opa engine (outside the service)

service/policies/entitlements/entitlements.rego
elizabethhealy: yes i have an ADR for this that i hope to discuss in OH today around multi-entity representations https://github.com/opentdf/platform/issues/735

service/entityresolution/entity_resolution.proto
elizabethhealy: removed, there used to be another EntityRepresentaion object and i wanted to make sure they were distinguishable

opentdf-dev.yaml
elizabethhealy: possibly, but the current issuer doesnt include the realm, also each service only seems to have access to `srp.Config` its own values in the config

strantalis: ok. we can circle back on this in the future.

service/entityresolution/keycloak/keycloak_entity_resolution.go
strantalis: Even when log level is set to `debug` it won't log?

elizabethhealy: ah no you're right, i forgot this was when it was an opa plugin, i moved the config logging to a separate log and commented it out

service/internal/auth/casbin.go
elizabethhealy: added

strantalis (approved):

PR #664, opened 2024-04-24
docs: how to run platform locally

strantalis:
Updates readme with instructions on how to run the platform locally without having to worry about setting up a go development environment.

### DIFF
--- a/.github/scripts/init-temp-keys.sh
+++ b/.github/scripts/init-temp-keys.sh
@@ -2,7 +2,7 @@
 # init-temporary-keys.sh
 # Initialize temporary keys for use with a KAS
 
-USAGE="Usage:  ${CMD:=${0##*/}} [(-v|--verbose)] [-H|--hsm]"
+USAGE="Usage:  ${CMD:=${0##*/}} [(-v|--verbose)] [-H|--hsm] [-o|--output <path>]"
 
 # helper functions
 exit2() {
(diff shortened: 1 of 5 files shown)


khvirtru (approved):
LGTM!

services started correctly by following the README.

opentdf-example.yaml
```diff
@@ -15,7 +15,7 @@ services:
     enabled: true
   authorization:
     enabled: true
-    url: http://localhost:8888
+    url: http://keycloak:8888
```
pflynn-virtru, line 18: does this work?  where is the setup for this?

strantalis: I added a `opentdf-dev.yaml` for local development. The readme should be updated. But yeah this will only work in docker compose setup like I documented.

opentdf-example.yaml
```diff
@@ -69,15 +69,15 @@ server:
     standard:
       rsa:
         123:
-          privateKeyPath: kas-private.pem
-          publicKeyPath: kas-cert.pem
+          privateKeyPath: /keys/kas-private.pem
```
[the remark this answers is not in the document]
strantalis: I think we just need to update this https://github.com/strantalis/platform/blob/documentation-runnning-locally/.gitignore#L33 ?

pflynn-virtru (approved):

opentdf-example.yaml
pflynn-virtru: got it now.  You run a container that connects to the already running docker-compose network

PR #667, opened 2024-04-25
feature(sdk): Adds TLS Certificate Exchange Flow 

imdominicreed:


.gitignore
```diff
@@ -38,3 +38,4 @@ tmp-gen/
 
 *.zip
 sensitive.txt.tdf
+certs/
```
pflynn-virtru, line 41: FYI another PR is adding `keys/`

pflynn-virtru:
some conflicts with https://github.com/opentdf/platform/pull/664

sdk/internal/oauth/cert-test/ca.jks
pflynn-virtru, line 1: use standard folder name `testdata`

sdk/options.go
```diff
@@ -40,6 +43,12 @@ func WithClientCredentials(clientID, clientSecret string, scopes []string) Optio
 	}
 }
 
+func WithTLSCredentials(tls *tls.Config, audience []string) Option {
```
strantalis, line 46: @imdominicreed I just noticed this is how we do it with token exchange but the more I look at this its not clear that this also requires `WihClientCredentials` option to work right?

imdominicreed: The ``WithClientCredential`` should be required, I believe. The X509 Certificate Authentication Flow is binded to a Keycloak Client.

pflynn-virtru (approved):

sdk/options.go
strantalis: How do we make sure it's clear from an end user perspective that you need both options. I am thinking it might be easier to understand for these to take the client creds as well. 

```go 
sdk.New("platform",
     sdk.WithClientCredentials("opentdf", "secret", nil),
     sdk.WithTLSExchange(subjectToken, "opentdf-sdk")
)
```


#### Proposal

```go
sdk.New("platform",
     sdk.WithTLSExchange( tls.Config, oauth2.ClientCredentials, scopes )
)
```

imdominicreed: I agree with this, and when I was first was writing this,  I had it like that. However, Client Credentials is always required if you want to authentication [https://github.com/opentdf/platform/blob/main/sdk/sdk.go#L119](url). Also, oauth.ClientCredentials is an internal library, so end users will not have access to the struct.

imdominicreed: Also in that example code you would need ```sdk.WithEndpoint("https://localhost:8443/auth/realm/opentdf").```

imdominicreed: Personally. I was thinking we had an interface for ExchangeInfo.

```go
type ExchangeInfo struct {
    ExchangeType() string // or maybe even doExchange, but that would take a bit of refactoring
}
```
and then we can have something closer to what you proposed
```
sdk.New("platform",
     sdk.WithExchangeInfo( oauth2.ExchangeInfoTokenExchange{SubjectToken: token}, oauth2.ClientCredentials, scopes )
)

```

and then for tls it would look like

```go 
sdk.New("platform",
     sdk.WithExchangeInfo( oauth2.ExchangeInfoTLSAuth{TLSConfig: config}, oauth2.ClientCredentials, scopes )
)

```

imdominicreed: If we do, ``DoExchange() Token`` this would also allow end users to create their own exchanges that the SDK doesn't have implemented yet.

strantalis: I think the oauth2 pkg in the stdlib has come good stuff we could leverage or at least try to align with as well. For example [TokenSource](https://pkg.go.dev/golang.org/x/oauth2#TokenSource) interface. But also we can and should just move to leveraging the [clientcredentials.Config](https://pkg.go.dev/golang.org/x/oauth2@v0.19.0/clientcredentials#Config)

strantalis: I think because we require client creds right now we are okay.

strantalis (approved):

pflynn-virtru (approved):

PR #668, opened 2024-04-26
fix(policy): normalize FQN lookup to lower case

### LINKED ISSUE #669: Policy API: FQN lookup should be normalized to lower case to match create/update of policy objects
### Background

Some PDPs and PEPs may store FQNs out of sync with lowercase FQNs expected by the platform and stored in the DB. Since Create/Update of policy objects are normalized to lowercase, lookup by FQN should be normalized as well.

### Acceptance Criteria

1. FQN lookup is normalized to lowercase
2. Proto comments document the behavior
3. Integration test to prove this behavior is satisfied

jakedoublev:
Closes #669


jakedoublev:
There will be conflicts with https://github.com/opentdf/platform/pull/647 which should be merged in first due to priority/impact.

Issue #631, opened 2024-04-20
add `WithGrpcPlainText` sdk option


strantalis:
We are currently using the `WithInsecure` option in the sdk to signal in my opinion the following tls config option. 

```
// InsecureSkipVerify controls whether a client verifies the server's
// certificate chain and host name. If InsecureSkipVerify is true, crypto/tls
// accepts any certificate presented by the server and any host name in that
// certificate. In this mode, TLS is susceptible to machine-in-the-middle
// attacks unless custom verification is used. This should be used only for
// testing or in combination with VerifyConnection or VerifyPeerCertificate.
InsecureSkipVerify [bool](https://pkg.go.dev/builtin#bool)
```

This is not the case though . We should add a new option called `WithGrpcPlainText` maybe to support non-tls connections.

pflynn-virtru:
jOr require all endpoints be Secure in development.  Some libraries like MSAL will not work at all in clear (even in dev mode)

strantalis:
@pflynn-virtru I think that will align with #451 so we are always testing against a secure connection. Maybe we shouldn't even add the ability to do a WithPlainText like grpccurl.

pflynn-virtru:
Feedback from the larger group, gave use a requirement to support clear.
Documentation about how to use it and where not to use it.  Provide warning about its usage

pflynn-virtru:
Should this propagate to all clients perfuming a connection?  Http and Grpc?

PR #670, opened 2024-04-26
feat(sdk): insecure plaintext and skip verify conn

pflynn-virtru:
resolves https://github.com/opentdf/platform/issues/631

_breaking change in the SDK_

```
// WithInsecurePlaintextConn returns an Option that sets up HTTP connection sent in the clear.
func WithInsecurePlaintextConn() Option {

// WithInsecureSkipVerifyConn returns an Option that sets up HTTPS connection without verification.
func WithInsecureSkipVerifyConn() Option {
```

- Some log statement changes


mkleene (approved):
looks good! good call on propagating the TLS settings to the IDP connections.

jrschumacher:
In the future, use the [breaking change convention](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-both--and-breaking-change-footer)

Subject `feat(sdk)!: insecure plaintext and skip verify conn`

Body/description:

```
BREAKING CHANGE:
.```
// WithInsecurePlaintextConn returns an Option that sets up HTTP connection sent in the clear.
func WithInsecurePlaintextConn() Option {

// WithInsecureSkipVerifyConn returns an Option that sets up HTTPS connection without verification.
func WithInsecureSkipVerifyConn() Option {
.```

- Some log statement changes"
```

Issue #704, opened 2024-04-30
Encrypt Failure: Unable to retrieve public key from KAS


cakeholeDC:
As of APR 29 2024 `otdfctl encrypt` fails with the following error message:

```
otdfctl encrypt plaintext.txt -o encrypted
{"time":"2024-04-29T11:05:12.640013-07:00","level":"DEBUG","msg":"LoadConfig: file and key not provided, using default file","config file":""}
{"time":"2024-04-29T11:05:12.936563-07:00","level":"DEBUG","msg":"getting new access token"}
{"time":"2024-04-29T11:05:12.936615-07:00","level":"DEBUG","msg":"Building DPoP Proof"}
 ERROR    Failed to encrypt: unable to retrieve public key from KAS at [http://localhost:8080]: error making request to KAS: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp [::1]:8080: connect: connection refused
```

Path to recreate:
Spin up local opentdf/platform following the instructions on the `main` branch (`0e59213e127e8b6a0b071a04f3ce380907fe494e`):
1. Kas Keys
2. Postgres & Keycloak
3. Bootstrap Keycloak
4. Start the platform

```
otdfctl auth client-credentials --client-id opentdf --client-secret secret

echo "Hello World!" > plaintext.txt

otdfctl encrypt plaintext.txt -o encrypted 
```

The `otdfctl encrypt` function worked as expected with a previous iteration of the opentdf platform. And if we roll back to a commit from last week and re-deploy the platform, the encrypt function begins to work again as expected

```
# opentdf/platform repo
git checkout f9e9d7288c1f63fdc1ffb0916fdb9ae4c390cee8
```
Follow README steps to run platform locally. 
_**‼️ Note, the process is different on this commit**_
1. docker-compose up
2. OpenTDF Config
3. Provision Keycloak
4. Kas Keys
5. Start the Platform

elizabethhealy:
If you are using the docker run command from the "Running the Platform Locally" section of the readme, you may need to add a flag to publish the port to the host ex:
```
docker run --network opentdf_platform -p "127.0.0.1:8080:8080"\
      -v "$(pwd)/kas-keys/:/keys/" \
      -v "$(pwd)/opentdf.yaml:/home/nonroot/.opentdf/opentdf.yaml" \
      -it registry.opentdf.io/platform:nightly start
```

b-long:
[...]

PR #709, opened 2024-05-01
fix(documentation): Add flag to run command in instructions for "Running the Platform Locally"

elizabethhealy:
Addresses https://github.com/opentdf/platform/issues/704
Adds a flag to publish the port to the host so the platform is available to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,10 @@ Bootstrap Keycloak
    ```
 4. Start the platform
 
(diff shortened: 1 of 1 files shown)


b-long:
FYI, I've tested this and it does look like this solves https://github.com/opentdf/platform/issues/704 👍 

Thank you!

jakedoublev (approved):

Issue #711, opened 2024-05-01
Distinguish project Consumers and Contributors


b-long:
Amazing project and kudos to the whole team! 🎉 🌮 🍰 

This issue is sort of a bug and a feature request.  If there's a bug, it's that a lot of folks have trouble following directions 😆 🤦.  

Recently, on another project, some teammates and I found it useful to distinguish between those who are _"**Consuming**"_ the project and those that are _"**Contributing**"_  to the project.

This had a few benefits:

1. It cleaned up our primary (top-level) README.md 
2. It's simple (K.I.S.S.) and user-friendly 
3. It means we can move faster, thinking about these two use-cases.

All it takes is a bit of text like this: 

> 
> ### 1. Consuming
> Consumers of `$PROJECT` should begin their journey [here](https://github.com/opentdf/platform/blob/main/Consuming.md).
> 
> ### 2. Contributing
> To contribute to `$PROJECT`, you'll need bit more set setup and should start [here](https://github.com/opentdf/platform/blob/main/Contributing.md).

I think this could go a long way to help both personas be more effective.  Plus, it could improve Developer Experience, [which we know is valuable](https://github.blog/2023-06-08-developer-experience-what-is-it-and-why-should-you-care/).

If you all are open to it, I'll open a PR to suggest some updates.

Issue #565, opened 2024-04-11
Undefined symbols when running the opentdf service according to the documentation/instructions


rsundaar1:
After downloading the latest opentdf app, tried running "go run github.com/opentdf/platform/service provision keycloak". I see thefollowing errors:

# github.com/opentdf/platform/service/internal/security
service\internal\security\hsm.go:51:13: undefined: pkcs11.Ctx
service\internal\security\hsm.go:52:13: undefined: pkcs11.SessionHandle
service\internal\security\hsm.go:71:27: undefined: pkcs11.ObjectHandle
service\internal\security\hsm.go:73:26: undefined: pkcs11.ObjectHandle
service\internal\security\hsm.go:147:57: undefined: pkcs11.Ctx
service\internal\security\hsm.go:149:16: undefined: pkcs11.New
service\internal\security\hsm.go:160:39: undefined: pkcs11.Ctx
service\internal\security\hsm.go:168:33: undefined: pkcs11.Ctx
service\internal\security\hsm.go:208:38: undefined: pkcs11.Ctx
service\internal\security\hsm.go:346:64: undefined: pkcs11.ObjectHandle
service\internal\security\hsm.go:149:16: too many errors

jakedoublev:
Thanks for opening this issue @rsundaar1! Are you still experiencing it if you pull the latest? This was likely fixed by some of the pulls around HSM and the crypto utilized.

dmihalcik-virtru:
This is likely still an issue. We might need to add build tags to prevent 

We can do this by refactoring `hsm.go` so it uses a build tag, adding 

```go
//go:build hsm
```

to the top of the file, then adding another bit of code that lets it compile when the `hsm` tag is not set during `go build`.

dmihalcik-virtru:
@rsundaar1 

To address the issue above you need to install a PKCS 1.1 compatible library, called in the code an HSM library. These libraries tend to be platform specific so for ease of development we generally use the [softHSM](https://www.opendnssec.org/softhsm/) library, which provides that API on top of openssl.

dmihalcik-virtru:
People internally are often running into this, especially when trying to use platform/service as a library.

dmihalcik-virtru:
This is merged as of https://github.com/opentdf/platform/pull/830 and should be available in 0.4.3 when we tag that

PR #713, opened 2024-05-01
feat(docs): Distinguish developers & users

b-long:
This PR builds on https://github.com/opentdf/platform/pull/709 .

Resolves https://github.com/opentdf/platform/issues/711 .


Consuming.md
```diff
@@ -0,0 +1,28 @@
...
+4. Start the platform
+
+   Exposes the server at localhost:8080
+   ```sh
+   docker run --network opentdf_platform \
+      -p "127.0.0.1:8080:8080" \
+      -v "$(pwd)/kas-keys/:/keys/" \
+      -v "$(pwd)/opentdf.yaml:/home/nonroot/.opentdf/opentdf.yaml" \
+      -it registry.opentdf.io/platform:nightly start
+   ```
```
b-long, line 28: There's an opportunity here (at the end of this file) to boost engagement / adoption, and help users.

```suggestion
   ```

## 🎉 Your platform is ready to use!

See: http://localhost:8080/ ... 

## Next steps

* Connect with the team
* Join our slack channel ([click here](https://join.slack.com/t/opentdf/shared_invite/zt-1e3jhnedw-wjviK~qRH_T1zG4dfaa~3A))
* Try out our CLI (`otdfctl`): https://github.com/opentdf/otdfctl
```

dmihalcik-virtru:
The stuff in `Consuming.md` should be moved into the READMEs - e.g. configuring the services should be in `service/README.md` and using the SDK should be in `sdk/README.md`

Consuming.md
```diff
@@ -0,0 +1,28 @@
+# Consumers 
+
+The information below is meant for users of the OpenTDF platform.
+
+To contribute/develop, see [here](./Contributing.md).
+
+## Running the Platform Locally
+
+1. Initialize KAS Keys ```.github/scripts/init-temp-keys.sh -o kas-keys```
```
dmihalcik-virtru, line 9: We need to make this accessible to people using the project via `docker run` or `go run`. This would could mean making the script available directly, or including it as an application subcommand.

b-long: I think that's a great idea, absolutely 👍 That sort of change could also help address some parts of https://github.com/opentdf/platform/issues/758

Consuming.md
```diff
@@ -0,0 +1,28 @@
+# Consumers
```
b-long, line 1: Quoting @dmihalcik-virtru , so we can have a threaded convo 😅 

> The stuff in Consuming.md should be moved into the READMEs - e.g. configuring the services should be in service/README.md and using the SDK should be in sdk/README.md

I see the value in that idea 👍 I like it as a contributor to the project.

But, I'm not sure that clarifies users versus developers.  You need to know different things to use `opentdf/platform` versus contribute to `opentdf/platform`.

Perhaps we can do both?  It looks like we already have two use-cases / personas:
* HSM: https://github.com/opentdf/platform/blob/main/README.md
* Non-HSM: https://github.com/opentdf/platform/blob/main/README-hsm.md

Couldn't we just expand this thinking to give an end-user a nice welcoming / happy path?

dmihalcik-virtru (approved):
Okay I filed some follow-up tickets:

1. #768 Update platform module READMEs to looks good in https://pkg.go.dev/
2. #769 Move init-temp-keys.sh to platform/fixtures tool so people can easily run it without checking out the repo

Consuming.md
dmihalcik-virtru: It looks like the 'default' behavior will be non-HSM, so maybe separating out 'how to set up hsm' makes sense, after we fix building without an hsm library (see https://github.com/opentdf/platform/issues/565)

dmihalcik-virtru (approved):

Consuming.md
b-long: I think this is resolved via https://github.com/opentdf/otdfctl/issues/127 👍

PR #738, opened 2024-05-02
fix(core): Fix several misspellings 

b-long:
This PR fixes several misspellings.

Since doing this work manually seems like a waste of time, you might consider introducing automation to prevent spelling errors (I make them constantly).
 
One automatic spell-checker is https://github.com/codespell-project/codespell .  It's very easy to run both locally and in CI, using pre-commit, e.g. the following will lint Go, Markdown, and YAML:

```yaml
# To learn about "pre-commit", see:
#   https://github.com/pre-commit/pre-commit

repos:
-   repo: https://github.com/codespell-project/codespell
    rev: v2.2.6
    hooks:
    - id: codespell
      files: ^.*\.(go|md|yaml)$
      args: ["--write-changes", "--ignore-words-list", "MYCOOLWORD,ANOTHAONE"]
```

### DIFF
--- a/README-hsm.md
+++ b/README-hsm.md
@@ -3,7 +3,7 @@
 The `opentdf` services can use hardware security modules (HSMs)
 to protect access to high value private key data,
 notably KAS private keys used for long lived identity of the server.
-The servcies can use PKCS #11 bindings to communicate with a system or network HSM.
+The services can use PKCS #11 bindings to communicate with a system or network HSM.
 To configure a development environment,
 we use [softHSM](https://github.com/opendnssec/SoftHSMv2).
 
--- a/examples/cmd/encrypt.go
+++ b/examples/cmd/encrypt.go
@@ -49,7 +49,7 @@ func encrypt(cmd *cobra.Command, args []string) error {
 		//sdk.WithDataAttributes("https://example.com/attr/attr1/value/value1"),
 		sdk.WithKasInformation(
 			sdk.KASInfo{
-				// examples assume unsecure http
+				// examples assume insecure http
 				URL:       fmt.Sprintf("http://%s", cmd.Flag("platformEndpoint").Value.String()),
 				PublicKey: "",
 			}))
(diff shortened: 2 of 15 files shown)


jrschumacher:
This is a great idea! We already use golangci-lint which has the ability to add [misspell](https://golangci-lint.run/usage/linters/#misspell) to check for spelling errors. I'd prefer to minimize the number of external tools if we can. Would you give this a try to see how it compares?

Also I'm not sure we want to introduce precommits as this will need to be a more detailed conversation. Precommits are great for small teams but large opensource projects they can become burdensome and many projects lean on CI linters.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,9 @@
+# To learn about "pre-commit", see:
+#   https://github.com/pre-commit/pre-commit
+repos:
+-   repo: https://github.com/codespell-project/codespell
+    rev: v2.2.6
+    hooks:
+    - id: codespell
+      files: ^.*\.(go|md|yml|yaml|sh)$
+      args: ["--write-changes", "--ignore-words-list", "MYCOOLWORD,ANOTHAONE"]
```
dmihalcik-virtru, line 9: I'm guessing
```suggestion
      args: ["--write-changes", "--ignore-words-list", "virtru,opentdf"]
```

.github/workflows/lint-all.yaml
```diff
@@ -44,3 +44,11 @@ jobs:
       - uses: bufbuild/buf-lint-action@044d13acb1f155179c606aaa2e53aea304d22058
         with:
           input: service
+  
+  pre-commit:
```
dmihalcik-virtru, line 48: Can we move this to checks.yaml and delete lint-all?

jrschumacher (approved):

.github/workflows/lint-all.yaml
b-long: Resolved by deleting this bit of code 👍

.pre-commit-config.yaml
b-long: Resolved by deleting this bit of code 👍

PR #740, opened 2024-05-03
feat(core): Service Level Child Loggers

### LINKED ISSUE #710: child logger for services
In our log messages it isn't clear which service the logs are coming from and makes it difficult to debug. With our service registration pattern we should introduce the concept of a child logger. This way as new services are added we can add more context around that service automatically to the logger.

https://betterstack.com/community/guides/logging/logging-in-go/#creating-and-using-child-loggers

pbacon-blaber:
- Adds child loggers to services
- Closes #710


service/authorization/authorization.go
```diff
@@ -67,6 +69,7 @@ var retrieveEntitlements = func(ctx context.Context, req *authorization.GetEntit
 }
 
 func (as AuthorizationService) GetDecisions(ctx context.Context, req *authorization.GetDecisionsRequest) (*authorization.GetDecisionsResponse, error) {
+	as.logger.Debug("testing 123")
```
strantalis, line 72:
```suggestion
```

strantalis (approved):

PR #747, opened 2024-05-06
chore: rewrap fixes for client-web

dmihalcik-virtru:
- Fixes several issues surfaced while fixing up client web, notably:
- work with empty attribute lists (for both subjects and objects). evaluates to 'permit' when both are empty
- remove duplicate dpop analysis code
- adds more stuff to dpop/auth context to make up for the missing re-parsing code
- implement SRT parsing correctly using dpop jwk if present and requested

### DIFF
--- a/service/kas/access/rewrap.go
+++ b/service/kas/access/rewrap.go
@@ -117,12 +117,11 @@ func verifySignedRequestToken(ctx context.Context, in *kaspb.RewrapRequest) (*Re
 	// First load legacy method for verifying SRT
 	md, exists := metadata.FromIncomingContext(ctx)
 	if !exists {
-		slog.WarnContext(ctx, "missing metadata")
+		slog.WarnContext(ctx, "missing metadata for srt validation")
 		return nil, errors.New("missing metadata")
 	}
 	if vpk, ok := md["X-Virtrupubkey"]; ok && len(vpk) == 1 {
 		slog.InfoContext(ctx, "Legacy Client: Processing X-Virtrupubkey")
-		
 	}
 
 	// get dpop public key from context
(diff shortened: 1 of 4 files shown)


service/kas/access/rewrap.go
```diff
@@ -75,49 +86,64 @@ func generateHMACDigest(ctx context.Context, msg, key []byte) ([]byte, error) {
 	return mac.Sum(nil), nil
 }
 
-func verifySignedRequestToken(ctx context.Context, in *kaspb.RewrapRequest) (*RequestBody, error) {
+func noverify(ctx context.Context, srt string) (string, error) {
+	token, err := jwt.Parse([]byte(srt), jwt.WithValidate(false))
```
strantalis, line 90: I haven't really thought about this but should we still validate the srt?

service/kas/access/rewrap.go
```diff
@@ -141,16 +167,55 @@ func verifySignedRequestToken(ctx context.Context, in *kaspb.RewrapRequest) (*Re
...
-		return requestBody, nil
+		return &requestBody, nil
 	default:
 		slog.WarnContext(ctx, fmt.Sprintf("clientPublicKey not a supported key, was [%T]", clientPublicKey))
 		return nil, err400("clientPublicKey unsupported type")
 	}
 }
 
+func verifySRT(ctx context.Context, srt string, dpopJWK jwk.Key) (string, error) {
+	v, err := jws.NewVerifier(jwa.RS256)
```
strantalis, line 181: @dmihalcik-virtru  I might be missing something but why did you move away from `jwt.Parse` ? When passing in `jwt.WithKey` it should verify the signature.

mkleene, line 181: Do we want to fix the algorithm here? `RS256` is pretty popular but it seems like it might be good to allow others.

service/kas/access/rewrap.go
```diff
@@ -141,16 +167,55 @@ func verifySignedRequestToken(ctx context.Context, in *kaspb.RewrapRequest) (*Re
...
+		slog.ErrorContext(ctx, "failed to decode payload", "srt", srt, "err", err)
+		return "", err400("signed content failure")
+	}
+	var srtDecoded SignedRequestBody
+	err = json.Unmarshal(payload, &srtDecoded)
+	if err != nil {
+		slog.WarnContext(ctx, "unable to unmarshall srt", "err", err, "srt", srt, "jwk", dpopJWK)
+		return "", err400("signed content failure")
+	}
+	return srtDecoded.RequestBody, nil
```
mkleene, line 216: The manual parsing and verification is surprising to me. Could we have a comment to understand why we can't verify the JWT using a higher level API?

service/kas/access/rewrap.go
dmihalcik: We don't really need to however we are signing it and I'd like to continue supporting this functionality until we decide to remove it explicitly with an ADR

service/kas/access/rewrap.go
dmihalcik-virtru: This is a JWS not a JWT.

mkleene: In our case they are pretty much the same. Our payload is JSON and since we are verifying using a signature algorithm our JWS will always be a JWT.

mkleene: ah, interesting, I thought the hash size was part of the name of the algorithm. but yeah, we control this so there isn't really a point to supporting more things.

dmihalcik-virtru: yeah I guess we actually are doing JWT stuff - notably, we support setting the expiry. I've added it back

service/kas/access/rewrap.go
dmihalcik-virtru: reverted to using the higher level API now that i have the correct options

mkleene (approved):

PR #750, opened 2024-05-06
chore(ci): enforce scoping of the commit messages

jrschumacher:


.github/workflows/checks.yaml
```diff
@@ -24,6 +24,33 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          # Types include:
+          #   - fix: fixes
+          #   - feat: features and enhancements
+          #   - chore: non-feature or enhancement (i.e. docs, ci, linting, automated, etc)
+          types: |
+            fix
+            feat
+            chore
```
strantalis, line 35:
```suggestion
            chore
            refactor
            revert
```

strantalis: @jrschumacher Should we also keep the `test` type. Referencing the default types here. 

https://github.com/commitizen/conventional-commit-types/blob/master/index.json

.github/workflows/checks.yaml
```diff
@@ -24,6 +24,35 @@ jobs:
...
+          #   - authorization: related to the authorization service changes (i.e. /service/authorization)
+          scopes: |
+            main
+            core
+            ci
+            docs
+            sdk
+            policy
+            key-access
+            authorization
```
dmihalcik-virtru, line 55: Characters are at a premium here
```suggestion
            kas
            authz
```

jrschumacher: Yea good point with the commit signature. We'll want to expand those in the changelog I assume.

.github/workflows/checks.yaml
```diff
@@ -24,6 +24,35 @@ jobs:
...
+            revert
+          # Scopes include:
+          #   - main: used for automated releases
+          #   - core: related to any core need such as the core service or monorepo
+          #   - ci: anything related to ci
+          #   - docs: anything related solely to documentation
+          #   - sdk: related to sdk changes in the /sdk directory
+          #   - policy: related to policy service changes (i.e. /service/policy)
+          #   - key-access: related to key access service changes (i.e. /service/kas)
+          #   - authorization: related to the authorization service changes (i.e. /service/authorization)
```
jrschumacher, line 46:
```suggestion
          #   - kas: related to key access service changes (i.e. /service/kas)
          #   - authz: related to the authorization service changes (i.e. /service/authorization)
```

strantalis (approved):

PR #753, opened 2024-05-06
feat(kas): Allow setting DialOptions for KAS / multi KAS.

ntrevino-virtru:
This allows setting DialOptions for KAS. This is especially important for decrypt with external servers.


sdk/tdf_config.go
```diff
@@ -30,6 +32,8 @@ const kHTTPOk = 200
 
 // KASInfo contains Key Access Server information.
 type KASInfo struct {
+	// Can be empty. If empty, it will inherit the global SDK DialOptions.
```
dmihalcik-virtru, line 35: Do we want to have separate dialoptions for KAS or perhaps every service?

dmihalcik-virtru (approved):

sdk/tdf_config.go
ntrevino-virtru: I could see that making sense. KAS definitely.

jrschumacher:
The original intention of the SDK abstraction of the services was for convenience when you have a monolith running. Since we haven't yet gotten to a place where performance concerns require the splitting of services from and OpenTDF perspective could we take a step back for a moment and consider our options. 

At minimum I feel that one option (which requires no development) would be spinning up multiple clients and pointing each one at a different platform. Since we don't have the concept of a central platform that negotiates the connections to other instances. 

This causes confusion since each instance of the SDK will invoke a separate abstraction to every service even when not needed. So that would beg the question if you should specify the service abstractions you want to define when invoking the SDK. 

Alternatively if you require multiple KAS or a have a deployment model that is not monolithic maybe the utility feature isn't for you and you are required to dial every service? (Solved through documentation)

Lastly, if we had the concept of a core platform and further extended the well known then the platform could communicate all the dialoptions required or proxy requests to the required service. (Service development)

Issue #727, opened 2024-05-01
register current platform services to healthz


strantalis:
Once #726 is completed we need to update the current services to register their own readiness status that the health check service can invoke. 

This will require updating 

- Policy Services
- KAS
- Authorization Service

Issue #726, opened 2024-05-01
ability for services to register readiness check


strantalis:
Add registration option that allows a service to register its own health check to the platform health check service.  

Not sure if this was the best implementation but we can mimic how the WellknownService was designed. 

Private package variable: https://github.com/opentdf/platform/blob/main/service/wellknownconfiguration/wellknown_configuration.go#L22

Register function: https://github.com/opentdf/platform/blob/main/service/wellknownconfiguration/wellknown_configuration.go#L26-L34

Service Registration Param: https://github.com/opentdf/platform/blob/main/service/pkg/serviceregistry/serviceregistry.go#L34

The health check service should now call out to each of these registered functions to get the status of a service and respond with a 200 if all services are healthy.

Not sure how customizable the proto response is but it would be nice to show which service is down or up as well. This might require us to define our own health check response proto.

Issue #700, opened 2024-04-30
ADR: service health check roll up


jrschumacher:
## Problem

As part of the work to enable scoped database connections we introduced a defect within the healthcheck #699. This issue revealed that services need to own their healthcheck and our main platform needs to roll these up as part of a high level healthcheck. 

## Solution

Implement an optional parameter as part of the registration called `healthz`. This parameter will enable services to append their own healthcheck process as part of a roll up within the platform healthcheck. 

When services require database connections it will be their burden to implement a healthz endpoint that pings their database connection to make sure it's healthy.

strantalis:
At a high level what we want is for [this](https://github.com/opentdf/platform/blob/main/service/health/health.go) health service to call out to those registered services healthz function that were registered. 

If they don't register a healthz function we should return a status of `unknown`. 

We probably can mimic what I had originally [setup](https://github.com/opentdf/platform/blob/main/service/pkg/serviceregistry/serviceregistry.go#L34) for WellknownConfig. Allowing the service to register something like `HealthzRegistration(namespace string, healthz func()) error`

jakedoublev:
> At a high level what we want is for the [health service](https://github.com/opentdf/platform/blob/main/service/health/health.go?rgh-link-date=2024-05-01T16%3A49%3A44Z) to call out to a registered services healthz function that is registered.
> 
> If they don't register a healthz function we should return a status of `unknown`.

What if `healthz` is broken out into separate liveness/readiness concerns? From the perspective of the rolled up healthz service, if a service is registered and the platform is running, doesn't that mean it's inherently `live` and not `unknown`? Maybe just the readiness probe is what needs to be able to be registered? `HealthzReadinessRegistration(namespace string, readiness func()) error`.

I wonder what the value is in requiring a liveness check defined separately from registration when it seems like successful registration to the running platform instance without liveness couldn't happen?

strantalis:
That makes sense to me. I don't think a service needs a liveness check. 

Also if a service returns an error I guess we need to mark the overall readiness of the platform down? This might get complicated the more I think about it because policy can operate fine without kas or authorization service being operational. But maybe this is where we push for services to be split out in the future.

jakedoublev:
> Also if a service returns an error I guess we need to mark the overall readiness of the platform down? This might get complicated the more I think about it because policy can operate fine without kas or authorization service being operational. But maybe this is where we push for services to be split out in the future.

[...]

PR #755, opened 2024-05-07
fix(core): rollup readiness checks to central health service

strantalis:
This pr introduces the ability for services to register with a centralized platform health check. 

```mermaid
block-beta
columns 1
HC["Platform Health Check"]
space
block:ID
 K["KAS"]
 A["Authorization Service"]
 P["Policy Service"]
end
K --> HC
A --> HC
P --> HC
```

Resolves 
- #700 
- #727 
- #726

### DIFF
--- a/service/pkg/serviceregistry/serviceregistry.go
+++ b/service/pkg/serviceregistry/serviceregistry.go
@@ -27,13 +27,14 @@ type RemoteServiceConfig struct {
 }
 
 type RegistrationParams struct {
-	Config          ServiceConfig
-	OTDF            *server.OpenTDFServer
-	DBClient        *db.Client
-	Engine          *opa.Engine
-	SDK             *sdk.SDK
-	WellKnownConfig func(namespace string, config any) error
-	Logger          *logger.Logger
+	Config                 ServiceConfig
+	OTDF                   *server.OpenTDFServer
+	DBClient               *db.Client
+	Engine                 *opa.Engine
+	SDK                    *sdk.SDK
+	WellKnownConfig        func(namespace string, config any) error
+	RegisterReadinessCheck func(namespace string, check func(context.Context) error) error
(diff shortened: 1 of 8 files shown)


service/pkg/serviceregistry/serviceregistry.go
```diff
@@ -26,12 +26,13 @@ type RemoteServiceConfig struct {
...
+	Config                 ServiceConfig
+	OTDF                   *server.OpenTDFServer
+	DBClient               *db.Client
+	Engine                 *opa.Engine
+	SDK                    *sdk.SDK
+	WellKnownConfig        func(namespace string, config any) error
+	RegisterReadinessCheck func(namespace string, check func(context.Context) error) error
 }
 type HandlerServer func(ctx context.Context, mux *runtime.ServeMux, server any) error
 type RegisterFunc func(RegistrationParams) (Impl any, HandlerServer HandlerServer)
```
dmihalcik-virtru, line 38: Alternative design idea - Create a HealthCheckingService interface with an `IsReady` function and let the respondants implement that if they want to implement health checks. Then when the service is activated, check the `Impl` object for `.type(HealthCheckingService)`. This would scale to more aspects instead of having to keep adding to the `RegistrationParams` object. That said, this seems fine so I'll approve

dmihalcik-virtru (approved):

pflynn-virtru (approved):

service/pkg/serviceregistry/serviceregistry.go
strantalis: I agree with this needing to change. Created a placeholder ticket so we can track. 

https://github.com/opentdf/platform/issues/757

pflynn-virtru (approved):

pflynn-virtru (approved):

ttschampel (approved):

Issue #749, opened 2024-05-06
Add GetDecisionsByToken function to address #735


elizabethhealy:
Resolves ADR https://github.com/opentdf/platform/issues/735
GetDecisionsByToken (public)
- disassemble jwt into entity chain

Config to support jwt decomposition
- we need to know what fields corresponds to entities
- get example tokens for each flow client_creds, password, token, authorization code for unit testing

Work TODO:
- proto updates for new function
- proto updates to remove jwt entity type and add uuid entity type
- jwt research, do the relevant idps have fields that we can use for pe & npe
- change authorization call to new endpoint in kas

PR #759, opened 2024-05-07
feat(authz): Handle jwts as entity chains in decision requests

elizabethhealy:
Resolves https://github.com/opentdf/platform/issues/749
- add endpoint GetDecisionsByToken
- remove jwt entity type
- add support for uuid entity type
- add configuration for different jwt entity selectors


service/kas/access/accessPdp.go
```diff
@@ -16,16 +16,18 @@ const (
...
-			return false, err
-		}
-		return dissemAccess, nil
+		// dissemAccess, err := checkDissems(policy.Body.Dissem, token)
+		// if err != nil {
+		// 	return false, err
+		// }
+
+		// throw an error, dont support dissems?
+		return false, errors.New("don't support dissems with v2 platform kas")
```
elizabethhealy, line 27: can we remove dissems checks since were just using the token?

strantalis: @biscoe916 Dissems is deprecated in the spec last time I looked. Can we move forward with removing it? 

If we still need to support Dissems this probably isn't the right place for the check anymore. 

cc: @jrschumacher @ttschampel

examples/cmd/encrypt.go
```diff
@@ -46,7 +46,7 @@ func encrypt(cmd *cobra.Command, args []string) error {
 	defer tdfFile.Close()
 
 	tdf, err := client.CreateTDF(tdfFile, strReader,
-		//sdk.WithDataAttributes("https://example.com/attr/attr1/value/value1"),
+		// sdk.WithDataAttributes("https://example.com/attr/attr1/value/value1"),
```
elizabethhealy, line 49: fixture provisioning is broken so this doesn't work in ci https://github.com/opentdf/platform/actions/runs/9057319882/job/24881138541?pr=759

service/authorization/authorization.proto
```diff
@@ -7,6 +7,12 @@ import "google/protobuf/any.proto";
 
 import "policy/objects.proto";
 
+
+message Token {
+  string id = 1; // ephemeral id for tracking between request and response
+  string jwt = 2; // the token
+}
```
pflynn-virtru, line 14: Can we add `TokenCustom`?   like other Custom, `EntityCustom`

pflynn-virtru (dismissed):

service/authorization/authorization.proto
elizabethhealy: sure! how would that differ from the string jwt? also what would that look like in terms of selector extraction? we would have to assume the token represented by TokenCustom is a map[string]interface{}

service/kas/access/accessPdp.go
damorris25: we should not drop dissems as eventually we will need it, or some equivalent - however, it could be moved somewhere else

elizabethhealy:
https://github.com/opentdf/platform/pull/790 -- follow up with detailed e2e testing

service/authorization/authorization.go
```diff
@@ -76,20 +99,33 @@ func NewRegistration() serviceregistry.Registration {
 					panic("Error casting clientSecert to string")
 				}
 			}
```
imdominicreed, line 101: NIT: clientSecert has a typo

also i think you want clientSecret and clientId to be lowercase since the config is forced lowercase.

service/authorization/authorization.go
```diff
@@ -2,13 +2,15 @@ package authorization
...
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
 	"strconv"
 	"strings"
 	"time"
 
+	"github.com/golang-jwt/jwt"
```
strantalis, line 13: Can you switch any usage of `golang-jwt` to the `lestrrat-go`.

opentdf-dev.yaml
```diff
@@ -19,6 +19,26 @@ services:
...
+      conditionalselectors:
+        - ifselector: client_id
+          present: false
+          selector: preferred_username
+          entitytype: userName
+        # - ifselector: some_claim
+        #   present: true
+        #   equalto: something
+        #   selector: other_selector
+        #   entitytype: email
```
strantalis, line 41: @elizabethhealy 
Is this saying if `client_id` is missing then try to get `preferred_username`? And is the second example saying if `some_claim` is present and equal to `something` then try to get the selector?

service/authorization/authorization.proto
```diff
@@ -174,6 +188,7 @@ message EntityEntitlements {
 
 //A logical bucket of attributes belonging to a "Resource"
 message ResourceAttribute {
+  string resource_attributes_id = 1;
```
jakedoublev, line 191: Is this id ephemeral and not long-lasting? Is it possible to rely on an index instead of an on-the-fly id if its purpose is to track decision responses by the `repeated ResourceAttribute` within the token decision request?

service/authorization/authorization.proto
```diff
@@ -158,7 +166,13 @@ message GetDecisionsResponse {
          "id": "e2",
          "emailAddress": "alice@example.org"
        }
-     ]
+     ],
```
jakedoublev, line 169: I've been wondering if we need scopes on GetEntitlements. Is a scoped GetEntitlements call different than a GetDecision request with a known set of attributes and a set of entities?

jakedoublev: This comment should be captured in an issue instead of PR comments. Closing out.

service/kas/access/accessPdp.go
strantalis: @elizabethhealy We discussed this today and you can remove the dissems check from kas. 

I created a place holder issue to track re-implementing it with the GetDecision call on the future.

https://github.com/opentdf/platform/issues/812

service/authorization/authorization.go
elizabethhealy: thanks good catch! i changed the case in the config but forgot the change it here

service/authorization/authorization.proto
elizabethhealy: probably, i can break this off into a separate issue since we do the same thing for entities when getting multiple of them
also this was just a quick fix, the field was supposed to be there and is referenced in the docs but must have gotten removed

elizabethhealy: https://github.com/opentdf/platform/issues/814

service/authorization/authorization.go
elizabethhealy: switched

service/entityresolution/entity_resolution.proto
```diff
@@ -62,11 +62,61 @@ message EntityNotFoundError {
...
+
+
 service EntityResolutionService {
   rpc ResolveEntities(ResolveEntitiesRequest) returns (ResolveEntitiesResponse) {
     option (google.api.http) = {
         post: "/entityresolution/resolve"
         body: "*";
     };
   }
+  rpc GetEntityChainFromJwt(GetEntityChainFromJwtRequest) returns (GetEntityChainFromJwtResponse) {
```
jakedoublev, line 116: This is probably a Nit, but it would likely keep the Casbin route authz policy cleaner and easier to read if we call the RPCs `CreateEntityChainFromJwt` or `BuildEntityChainFromJwt` instead of `Get...` since this is a POST http verb and not a GET?

jakedoublev: I don't think this is a blocker regardless because this route is internal to the platform (for now) so it can be changed later.

jakedoublev (dismissed):
I'm not a Codeowner so it both needs and should get a further review, but this LGTM per the ADR outcome.

service/entityresolution/entity_resolution.proto
elizabethhealy: i agree i like CreateEntityChainFromJwt better, ill update it

strantalis (approved):

elizabethhealy:
TODO: followup with additional integration testing for token exchange flows

service/authorization/authorization.go
```diff
@@ -48,35 +49,33 @@ func NewRegistration() serviceregistry.Registration {
...
-			val, ok = srp.Config.ExtraProps["clientId"]
+			val, ok = srp.Config.ExtraProps["clientid"]
 			if ok {
 				clientID, ok = val.(string)
 				if !ok {
 					panic("Error casting clientID to string")
 				}
 			}
-			val, ok = srp.Config.ExtraProps["clientSecert"]
+			val, ok = srp.Config.ExtraProps["clientsecert"]
```
dmihalcik-virtru, line 71: Should this be `clientSecret` or `clientSecert`?

PR #762, opened 2024-05-07
chore(ci): Validates internal deps are up-to-date

dmihalcik-virtru:
Most of our tests currently use the go workspaces feature, which allows us to test with internal 'head' (unreleased) versions of the code in sync. However. our current build infra generates per-module tags and releases, which means we need to make sure that downstream users get a coherent set of modules.

To validate a release, this removes the 'upstream' deps from the go.work before running any test. This only happens on release branches


jrschumacher (approved):
Approving so we can maintain the workspace files.

jrschumacher (approved):

PR #763, opened 2024-05-07
chore(core): Dep updates

dmihalcik-virtru:


elizabethhealy (approved):

dmihalcik-virtru:
I'm pretty sure this is still broken so I'm putting this on ice until we have better CI around internal deps

dmihalcik-virtru:
merged separately

PR #767, opened 2024-05-08
chore(ci): Enforce go mod tidy and go fmt

dmihalcik-virtru:


jrschumacher:
Should this use #772?

dmihalcik-virtru:
> Should this use #772?

There are two differences between this code and the makefile code.

1. the `Makefile` uses [goimports](https://pkg.go.dev/golang.org/x/tools/cmd/goimports#pkg-overview) as its formatter; I didn't want to have to install another tool here but we could do it
2. This is done per module using a matrix, unlike the `Makefile` which works on all the modules. I can use `addprefix` to generate a list of targets so we can do things like `make tidy:service` and so on, but it the makfile would be a little harder to read

pflynn-virtru (approved):

strantalis (approved):

Issue #719, opened 2024-05-01
Epic: Audit 


strantalis:
The original implementation of the platform had introduced audit logs into specific locations. This is a higher level issue to track the work that needs to be done to introduce this functionality back in. 

- #730
- #884 
- #885

cassandrabailey293:
initial implementation has been completed - see children tickets.

PR #774, opened 2024-05-09
feat(core): audit logging

### LINKED ISSUE #730: audit package
In order to generate audit events within the platform we need an audit package that implements `Leveler`.

https://betterstack.com/community/guides/logging/logging-in-go/#creating-custom-log-levels

This package should be public and be something services or the sdk could easily call into.  Maybe this fails under the logger package but the idea is to keep the audit centralized so in the future we could potentially look into things like tamper proof audit logs generated from the platform. 

Example of how it was originally done in python. 

https://github.com/opentdf/backend/blob/dbfcea7594fa72dd49d02a19aa01b36f3a32b7e6/containers/kas/kas_app/tdf3_kas_app/plugins/audit_hooks.py#L31-L77

pbacon-blaber:
- Adds audit logging level to internal logger
- Closes #730 

The main changes for the PR are in `logger.go` and `audit.go` where we add the new AUDIT log level and then add a specific method to the logger to log rewrap requests as an example. More detailed changes for other audit calls will be done as part of #719 

Examples of rewrap success log:
```json
{
  "time": "2024-05-21T11:03:18.910724-04:00",
  "level": "AUDIT",
  "msg": "{\"id\":\"4e027444-3496-4a6d-9e95-f44be19e341c\",\"object\":{\"type\":\"data_object\",\"id\":\"a96b1e6a-177d-11ef-a5c4-06f27729b9a5\",\"attributes\":{\"attrs\":[\"https://example.com/attr/attr1/value/value1\"],\"dissem\":[],\"permissions\":[]}},\"action\":{\"type\":\"read\",\"result\":\"success\"},\"actor\":{\"id\":\"68f9d36d-e103-4518-8258-70fe7cc87c90\",\"attributes\":{\"attrs\":[],\"dissem\":[],\"permissions\":null}},\"eventMetaData\":{},\"clientInfo\":{\"userAgent\":\"\",\"platform\":\"kas\",\"requestIp\":\"\"},\"diff\":{},\"timestamp\":\"2024-05-21T11:03:18-04:00\"}",
  "namespace": "kas"
}
```

Parsed JSON string of audit log message:
```json
{
  "id": "4e027444-3496-4a6d-9e95-f44be19e341c",
  "object": {
    "type": "data_object",
    "id": "a96b1e6a-177d-11ef-a5c4-06f27729b9a5",
    "attributes": {
      "attrs": [
        "https://example.com/attr/attr1/value/value1"
      ],
      "dissem": [],
      "permissions": []
    }
  },
  "action": {
    "type": "read",
    "result": "success"
  },
  "actor": {
    "id": "68f9d36d-e103-4518-8258-70fe7cc87c90",
    "attributes": {
      "attrs": [],
      "dissem": [],
      "permissions": null
    }
  },
  "eventMetaData": {},
  "clientInfo": {
    "userAgent": "",
    "platform": "kas",
    "requestIp": ""
  },
  "diff": {},
  "timestamp": "2024-05-21T11:03:18-04:00"
}
```


service/internal/logger/logger.go
```diff
@@ -17,10 +18,34 @@ type Config struct {
...
 
-func (l *Logger) With(key string, value string) *Logger {
-	return &Logger{
-		Logger: l.Logger.With(key, value),
+// From the Slog docs (https://betterstack.com/community/guides/logging/logging-in-go/#customizing-slog-levels):
+// The log/slog package provides four log levels by default, with each one
+// associated with an integer value: DEBUG (-4), INFO (0), WARN (4), and ERROR (8).
+const (
+	// Currently setting AUDIT level to 6, between WARN and ERROR
+	LevelAudit = slog.Level(6)
```
pbacon-blaber, line 26: @strantalis, wondering what would be more appropriate here. I'm thinking maybe between INFO and WARN might be better (2)?

strantalis: Can we make it higher than error and not allow them to turn it off?

jrschumacher:
Can we abstract this to define enumerated events and the service? I'm concerned that without predefined events it's going to be hard to document and if we don't automatically scope to a service then there could be naming conflicts with the events.

jrschumacher:
@strantalis informed me that there is already a plan in place that will address my concerns!

service/internal/logger/logger.go
pbacon-blaber: Yeah I made it level 10, and the way the logger is initialized the `getLevel` function only accepts the options "debug", "info" and "error" and will throw an error if anything else is provided as the level, so in that way it would never be possible to turn off

service/internal/logger/audit.go
```diff
@@ -0,0 +1,102 @@
...
+	Body PolicyBody `json:"body"`
+}
+
+type PolicyBody struct {
+	DataAttributes []SimpleAttribute `json:"dataAttributes"`
+	Dissem         []string          `json:"dissem"`
+}
+
+type SimpleAttribute struct {
+	URI string `json:"attribute"`
```
jakedoublev, line 20: Nit: FQN (fully qualified name) seems like it would be better here for consistency

pbacon-blaber: I was mostly basing this off of [how it was implemented in Python](https://github.com/opentdf/backend/blob/dbfcea7594fa72dd49d02a19aa01b36f3a32b7e6/containers/kas/kas_app/tdf3_kas_app/plugins/audit_hooks.py#L31-L77) as well as [the current policy struct](https://github.com/opentdf/platform/blob/7a3379b6878562e4958e61516335e912716588b7/service/kas/access/attribute.go#L10)  (as this is just a simplification of the existing policy struct for logging purposes)

I can change it no problem, but wondering if it's a wider change we'd want to make including the attribute struct as well in KAS?

pbacon-blaber: Awesome, I'm going to resolve this conversation then.

service/kas/access/rewrap.go
```diff
@@ -318,27 +318,33 @@ func (p *Provider) tdf3Rewrap(ctx context.Context, body *RequestBody, entity *en
 
 	access, err := canAccess(ctx, ent, *policy, p.SDK)
 
+	auditPolicy := transformAuditPolicy(policy)
+
 	if err != nil {
-		slog.WarnContext(ctx, "Could not perform access decision!", "err", err)
+		p.Logger.WarnContext(ctx, "Could not perform access decision!", "err", err)
+		p.Logger.AuditRewrap(ctx, *auditPolicy, false)
```
jrschumacher, line 325: What kind of errors are returned? I do get that technically the key was not released but it does feel like audit on errors could be confusing 🤷

pbacon-blaber: This is not as much of an error as an error log. See this part of the Python implementation: https://github.com/opentdf/backend/blob/dbfcea7594fa72dd49d02a19aa01b36f3a32b7e6/containers/kas/kas_app/tdf3_kas_app/plugins/audit_hooks.py#L85

service/internal/logger/audit.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"context"
+	"encoding/json"
+	"log/slog"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+type PolicyLog struct {
```
pbacon-blaber, line 12: is there a better way to do this in Go? I feel like this object type definition is disgustingly taking over half the file here. Should I move the entire `PolicyLog` definition into a new file?

pbacon-blaber: I moved them into auditPolicy.go

service/internal/logger/audit.go
```diff
@@ -0,0 +1,89 @@
...
+		auditLog.Object.Attributes.Attrs = append(auditLog.Object.Attributes.Attrs, value.URI)
+	}
+	auditLog.Object.Attributes.Dissem = policy.Body.Dissem
+
+	// Actor
+	auditLog.Actor.ID = policy.Actor.ID
+
+	auditLogJSONBytes, err := json.Marshal(auditLog)
+	auditLogJSONString := string(auditLogJSONBytes)
+	if err != nil {
```
elizabethhealy, line 83: nit: check the err before string conversion to avoid nil pointer issues (not 100% sure if that would apply with this string func)

service/internal/logger/logger.go
```diff
@@ -39,22 +64,42 @@ func NewLogger(config Config) (*Logger, error) {
...
+		Level:       LevelAudit,
+		ReplaceAttr: customReplaceAttributes,
+	})
+
+	auditLoggerBase := slog.New(auditLoggerHandler)
+	auditLogger := CreateAuditLogger(*auditLoggerBase)
+
 	return &Logger{
 		Logger: logger,
+		Audit:  auditLogger,
```
elizabethhealy, line 92: is there a way to disable audit logging? (or do we want that?)

elizabethhealy (approved):

service/internal/logger/audit.go
pbacon-blaber: Good catch, I was originally doing string conversion in the log method itself and then swapped to this and put it on the wrong line

service/internal/logger/logger.go
pbacon-blaber: So for this, I originally had the level set to in between error and warning, and then @strantalis I believe suggested it be higher than error and not something we can turn off

The way it's designed now we could probably pass a config option to turn off audit logging, and if that is false all the `logger.Audit` sub logger could be turned to level "off" instead of whatever it is set at

@strantalis what do you think, should we do that?

elizabethhealy (approved):

PR #783, opened 2024-05-13
chore(ci): fixes for partial workspace init

dmihalcik-virtru:
- Remove set -e, which is in conflict with how I've been writing these scripts
- Instead, use && to leave return value set to error state if any phase fails


dmihalcik-virtru:
This is blocking releases currently, see https://github.com/opentdf/platform/actions/runs/9040705249/job/24845206095#step:4:5

elizabethhealy (approved):

Issue #789, opened 2024-05-13
Add full round trip e2e testing for all attribute rule types


elizabethhealy:
We need to test the full flow of the platform. Add integration tests that encrypt and decrypt with different types of attributes. Ensure that certain clients/users are able to decrypt and others are not.

Possible setup:
Attributes:
- AllOf -- Language: French, English, Spanish
- AnyOf -- Colors: Red, Blue, Green
- Hierarchy -- Cards: King, Queen, Jack

// maybe test with a user as well
Mappings for '.clientId' == opentdf-sdk:
- English, Red, Queen

Runs:
Success:
- encrypt with English
- encrypt with Red
- encrypt with Red and Green
- encrypt with Jack
- encrypt with Queen
- encrypt wtih Engligh, Red, Green, Jack

Failures:
- encrypt with English and French
- encrypt with Blue
- encrypt with King
- encrypt with English, French, Red, Green, Queen
- encrypt with English, Blue, Green, Queen
- encrypt with English, Red, Green, King

elizabethhealy:
Blocked by https://github.com/opentdf/platform/pull/759 and possibly https://github.com/opentdf/platform/issues/786

PR #790, opened 2024-05-13
feat(ci): Add e2e roundtrip tests for different attribute combinations

elizabethhealy:
Resolves https://github.com/opentdf/platform/issues/789
tests expected successes and failures of each attribute rule and combinations of attributes

Ensure the kas -> authorization flow is working as expected


elizabethhealy:
follow up would be to add all the different token types - i.e. run with jwt of client+user w/ password flow, token exchange, etc, to make sure they all work

elizabethhealy:
depends on https://github.com/opentdf/platform/pull/759 to get the flow working

e2e/cmd/roundtrip_tests.go
```diff
@@ -0,0 +1,178 @@
...
+		testConfig := *(cmd.Context().Value(RootConfigKey).(*TestConfig))
+		return runRoundtripTests(&testConfig)
+	},
+}
+
+func init() {
+	E2ECmd.AddCommand(RTTestCmd)
+}
+
+// TODO: uncomment when authorization bug fixed preventing attribute lookup
```
elizabethhealy, line 28: https://github.com/opentdf/platform/issues/792

jrschumacher:
@elizabethhealy Could we talk about moving these to opentdf/otdfctl? This way we can make sure that the CLI tool will continue to be maintained and have the separation of the platform in a way where we will truly have e2e running?

Or maybe we could make e2e a module and have it pull in the CLI? Without incorporating the CLI we could end up making changes thinking we tested everything and forget that we broke the CLI.

elizabethhealy:
@jrschumacher ya i can break off the actual code to an e2e module, tbh it doesnt really even need to be cli since its just tests, would it be preferable to just remove the cli aspect and run it as a script?

elizabethhealy:
TODO: add some more flows with user + client in the jwt, one having access and one not, etc

elizabethhealy:
@jrschumacher I made the updates we discussed, lmk if this new structure is ok

service/rttests/rt_test.go
```diff
@@ -0,0 +1,390 @@
...
+			}},
+				{Value: &policy.Action_Standard{
+					Standard: policy.Action_STANDARD_ACTION_TRANSMIT,
+				}},
+			},
+			NewSubjectConditionSet: &subjectmapping.SubjectConditionSetCreate{
+				SubjectSets: []*policy.SubjectSet{
+					{ConditionGroups: []*policy.ConditionGroup{
+						{Conditions: []*policy.Condition{{
+							SubjectExternalSelectorValue: ".clientId",
```
imdominicreed, line 268: is this a working subjectmapping. I have something like this, but i had to do "\".clientId\"". seems like it needs to be quoted due to here https://github.com/opentdf/platform/blame/41ca82d692209d120bfa52800fa0988bf373b0b5/service/internal/jqbuiltin/jq_builtin.go#L90

elizabethhealy: it works without the additional quotes, i believe that unescape query function is intended to remove extra sets of quotes if present

imdominicreed: ahh, you know what i figured out what happened. my mistake fixed two things at once, and had the wrong conclusion.

Issue #786, opened 2024-05-13
Policy fixture provisioning does not fill example data


elizabethhealy:
Running `go run ./service provision fixtures` does not have any errors on the client side but calling ListAttributes or ListSubjectMappings after the provisioning does not return anything. The issue exists when running in CI as well. https://github.com/opentdf/platform/actions/runs/9057319882/job/24881138541?pr=759#step:42:156

Locally in the postgres logs I see
```
opentdf-v2-poc-opentdfdb-1   | 2024-05-13 04:29:17.189 UTC [137] ERROR:  relation "goose_db_version" does not exist at character 32
opentdf-v2-poc-opentdfdb-1   | 2024-05-13 04:29:17.189 UTC [137] STATEMENT:  SELECT tstamp, is_applied FROM goose_db_version WHERE version_id=$1 ORDER BY tstamp DESC LIMIT 1
opentdf-v2-poc-opentdfdb-1   | 2024-05-13 04:30:34.495 UTC [253] ERROR:  relation "goose_db_version" does not exist at character 32
opentdf-v2-poc-opentdfdb-1   | 2024-05-13 04:30:34.495 UTC [253] STATEMENT:  SELECT tstamp, is_applied FROM goose_db_version WHERE version_id=$1 ORDER BY tstamp DESC LIMIT 1

```

jrschumacher:
@elizabethhealy looks like `runMigrations` is false https://github.com/opentdf/platform/actions/runs/9057319882/job/24881138541?pr=759#step:42:35

elizabethhealy:
> @elizabethhealy looks like `runMigrations` is false https://github.com/opentdf/platform/actions/runs/9057319882/job/24881138541?pr=759#step:42:35

@jrschumacher thats just for the authorization namespace cause it doesnt have a db client, the attributes service registration runs them https://github.com/opentdf/platform/actions/runs/9057319882/job/24881138541?pr=759#step:42:52, as well as the subjectmapping service, namespace service etc

jrschumacher:
That's strange. `runMigrations` is a global config option maybe there's another defect there on the logs.

strantalis:
@jrschumacher @elizabethhealy 

The fixtures look to be getting provisioned against the wrong schema.

```
2024/05/13 14:48:28 INFO 📦 running migrations in schema schema=opentdf
2024/05/13 14:48:28 INFO running migration up schema=opentdf database=opentdf
```

[...]

PR #794, opened 2024-05-13
fix(ci): Use the correct schema with the provision fixture command

elizabethhealy:
addresses https://github.com/opentdf/platform/issues/786


jakedoublev (approved):
Thanks for this.

PR #795, opened 2024-05-13
fix(sdk): Reduces sdk go requirement to 1.21

dmihalcik-virtru:
- Updates to latest internal deps, which have had their version requirement reduced as well
- tidy elsewhere as well


examples/go.mod
```diff
@@ -1,12 +1,10 @@
 module github.com/opentdf/platform/examples
 
-go 1.22
-
-toolchain go1.22.3
+go 1.22.3
```
jrschumacher, line 3: @dmihalcik-virtru looks like it undid your other work.

dmihalcik-virtru: yeah the problem is the `go mod tidy` analysis uses the released versions of the modules, not the workspace, as there is no way to guess what the future release number will be. So we'll have to do `sdk`, then `service`, and finally `examples` each separately, with releases in between

jrschumacher (approved):

PR #796, opened 2024-05-13
fix(core): Updates logs statements to log errors

dmihalcik-virtru:
Fixes unbalanced/unknown key issues in slog

### DIFF
--- a/lib/fixtures/keycloak.go
+++ b/lib/fixtures/keycloak.go
@@ -845,7 +845,7 @@ func createTokenExchange(ctx context.Context, connectParams *KeycloakConnectPara
 	}
 	if err := client.UpdatePermissionScope(ctx, token.AccessToken, connectParams.Realm,
 		*realmManagementClientID, tokenExchangePolicyScopePermissionID, permissionScopePolicyRepresentation); err != nil {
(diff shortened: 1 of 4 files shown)


jrschumacher (approved):

lib/fixtures/keycloak.go
```diff
@@ -845,7 +845,7 @@ func createTokenExchange(ctx context.Context, connectParams *KeycloakConnectPara
 	}
 	if err := client.UpdatePermissionScope(ctx, token.AccessToken, connectParams.Realm,
 		*realmManagementClientID, tokenExchangePolicyScopePermissionID, permissionScopePolicyRepresentation); err != nil {
-		slog.Error("Error creating permission scope : %s", err)
+		slog.Error("Error creating permission scope", "error", err)
```
strantalis, line 848: @dmihalcik-virtru Can we stick with key/value. 

```suggestion
		slog.Error("Error creating permission scope", slog.Any("error", err))
```
or `slog.String("error",err.Error())

We can use this linter to enforce the style we want going forward. https://github.com/go-simpler/sloglint

dmihalcik-virtru: Why?

PR #797, opened 2024-05-13
feat(sdk): Allow setting TDF mime type

ntrevino-virtru:


sdk/tdf.go
```diff
@@ -204,7 +204,11 @@ func (s SDK) CreateTDF(writer io.Writer, reader io.ReadSeeker, opts ...TDFOption
 	tdfObject.manifest.EncryptionInformation.Method.IsStreamable = true
 
 	// add payload info
-	tdfObject.manifest.Payload.MimeType = defaultMimeType
+	mimeType := tdfConfig.mimeType
+	if mimeType == "" {
+	  mimeType = defaultMimeType
```
jrschumacher, line 209:
```suggestion
		mimeType = defaultMimeType
```

ntrevino-virtru:
@strantalis I'd personally vote no on that because that leaves room for unpredictability which I'm not a fan of. I'm just me though, so happy to if the general consensus is in favor.

dmihalcik-virtru (approved):

PR #824, opened 2024-05-16
fix(policy): make resource-mappings update patch instead of put in RESTful gateway

### LINKED ISSUE #313: REST HTTP paths have collisions; some paths are not currently usable
Our current use of gRPC Gateway with gorilla.mux results in several name collisions. Notably, the `attributes/namespaces` collides with the `attributes/_` path.

jakedoublev:
Closes #313 

All CRUD routes for Get, List, Update, Deactivate/Delete, Create were tested over RESTful gateway for the following policy:
1. attribute namespaces
2. attribute definitions
3. attribute values
4. subject condition sets
5. subject mappings
6. resource mappings

All were found to be accessible and working as expected, with the one exception being Update of resource mappings as POST instead of PATCH which this PR fixes.


dmihalcik-virtru (approved):
I'll file an issue for integration testing with the REST api, which we are currently lacking. Once we are generating client-web source code we can use it for the tests here.

jrschumacher (approved):

jrschumacher:
> I'll file an issue for integration testing with the REST api, which we are currently lacking. Once we are generating client-web source code we can use it for the tests here.

Curious if each service maintainer should own these tests so they have to manage them with proto changes?

PR #830, opened 2024-05-16
feat(core): Adds opentdf.hsm build constraint

dmihalcik-virtru:
- Adds a new build tag, `opentdf.hsm`
- when flag is set, HSM crypto provider is available; absent if not
- this should prevent compile time errors in downstream projects that do not require an HSM
- Adds flag to Makefile by default. We may want to add an easy way to disable it


service/internal/security/standard_crypto.go
```diff
@@ -139,11 +139,11 @@ func (s StandardCrypto) GenerateNanoTDFSymmetricKey([]byte) ([]byte, error) {
 	return nil, errNotImplemented
 }
 
-func (s StandardCrypto) GenerateEphemeralKasKeys() (PrivateKeyEC, []byte, error) {
+func (s StandardCrypto) GenerateEphemeralKasKeys() (any, []byte, error) {
```
pflynn-virtru, line 142: why `any` here?

pflynn-virtru: reason - this was tied to HSM

pflynn-virtru (approved):

strantalis (approved):

Issue #630, opened 2024-04-19
ADR: Open Policy Agent (OPA) removal


pflynn-virtru:
# Open Policy Agent removal

## Context and Problem Statement
Currently, we use [OPA](https://www.openpolicyagent.org) for our entitlement PDP.

<img width="998" alt="Screenshot 2024-04-22 at 2 25 22 PM" src="https://github.com/opentdf/platform/assets/43211074/3f8c904d-34e9-40fa-b169-c66eb905c913">

This is the next iteration, soon-to-be-merged.  Note most logic is removed from .rego, basically it is just glue at this point.

<img width="636" alt="Screenshot 2024-04-24 at 11 21 18 AM" src="https://github.com/opentdf/platform/assets/43211074/1ea3bc18-686d-4ec3-b3b4-e9f9d9e6448f">

### Relevant ADR
Entitlement Policy Deployment Model https://github.com/opentdf/platform/issues/472
Configuration Service https://github.com/virtru-corp/adr/blob/main/docs/decisions/ar-8-config-service-v2.md

## Decision Outcome
- 📜 No change
  - Keep OPA
  - Re-evaluate after upcoming major milestone

### Decision Notes
- 🟩 Infinite flexibility in entitlement PDP
- 🟨 More data points needed - this will come after the next major milestone
- 🟨 Support should be like other sub-system integrations, ex. database, IdP
- 🟥 Developers' velocities will be impacted when working around OPA

## Decision Drivers
- OPA development challenges
  - type conversion
  - built-in development
  - testing
  - small rego codebase < 50 loc
- OPA customer challenges
  - non-support if changed
- OPA capabilities
  - Federation of policies
  - Policy signing
- OpenTDF subject mappings solution
- Possible performance bottleneck (not enough data)

## Considered Options
- Remove OPA library
- Optional OPA library
- Optional OPA service integration
- Support and Keep OPA

###  ⚖️ Remove OPA library

Create a pure Go implementation of the Entitlement PDP.  Mimic what is done with [Access PDP](https://github.com/opentdf/platform/blob/main/service/internal/access/pdp.go)

* 🟩 Developers happy
* 🟨 Moves integration point to Entity Resolution Service (ERS)
* 🟥 One less integration point

### ⚖️ Optional OPA library

Create a pure Go implementation of the Entitlement PDP.
Leave OPA-Entitlement PDP.
Create a configuration to enable.
```
authorization:
  entitlement:
    # native Golang
    pdp: default
[...]

PR #833, opened 2024-05-17
feat(authz): Allow un-scoped GetEntitlements calls

### LINKED ISSUE #823: Authorization: GetEntitlements should allow unscoped calls per proto definition
### Background

The Authorization Service `GetEntitlements` RPC proto indicates that the [scopes are optional](https://github.com/opentdf/platform/blob/41ca82d692209d120bfa52800fa0988bf373b0b5/service/authorization/authorization.proto#L165-L168), but the RPC handler currently [errors out if scopes are not provided](https://github.com/opentdf/platform/blob/41ca82d692209d120bfa52800fa0988bf373b0b5/service/authorization/authorization.go#L243-L246). This presents a challenge for downstream use cases like: https://github.com/opentdf/platform/issues/705.

### Acceptance Criteria

1. `GetEntitlements` can accept unscoped calls, relying on the policy `MatchSubjectMappings` rpc
2. The `MatchSubjectMappings` policy API is called with a flattened Entity Representation's selector fields/values (ERS is called first)
3. Tests

elizabethhealy:
Resolves #823 
Per the proto definitions scope should be optional for getEntitlements
However there will be performance implications on un-scoped calls since we will have to evaluate all subject mappings

Temporary solution: retrieve all the attribute fqns then get their subject mappings


jakedoublev (approved):
Approving as known tech debt associated with OPA https://github.com/opentdf/platform/issues/630

The alternatives are to:
1. move the ERS call/resolution outside OPA so the result is available when calling policy MatchSubjectMappings before OPA resolves them
2. call policy MatchSubjectMappings from inside OPA over REST after the OPA call to ERS

Of this PR's implementation and those two options, moving the ERS call outside OPA (and/or removing OPA entirely) are the preferred options to resolve https://github.com/opentdf/platform/issues/823, but they are larger lifts and should not block this need for the time being.

PR #861, opened 2024-05-23
fix(core): Update hkdf salt

dmihalcik-virtru:
> salt - This is a non-random value tied to the magic number and version SHA256(MAGIC_NUMBER + VERSION). So for this version of the nanotdf the value of the salt is 3de3ca1e50cf62d8b6aba603a96fca6761387a7ac86c3d3afe85ae2d1812edfc in hex.

source: https://github.com/opentdf/spec/tree/main/schema/nanotdf#4-ecc-encryption-key-derivation

### DIFF
--- a/sdk/nanotdf.go
+++ b/sdk/nanotdf.go
@@ -3,6 +3,7 @@ package sdk
 import (
 	"bytes"
 	"context"
+	"crypto/sha256"
 	"encoding/binary"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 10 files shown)


pflynn-virtru (approved):

sdk/nanotdf.go
```diff
@@ -774,3 +777,9 @@ type keyAccess struct {
 	URL           string `json:"url"`
 	Protocol      string `json:"protocol"`
 }
+
+func versionSalt() []byte {
```
sujankota, line 781: Do we need this because go HKDF does sha256?

sujankota: nvm, since JS is failing you changes looks corredct.

sujankota (approved):

PR #863, opened 2024-05-23
feat(tests): Add roundtrip tests for token exchange flow

elizabethhealy:
Adds roundtrip testing for token exchange flow
Verifies functionality of multi entity decisions via jwt

### DIFF
--- a/service/rttests/rt_test.go
+++ b/service/rttests/rt_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/opentdf/platform/protocol/go/policy/subjectmapping"
 	"github.com/opentdf/platform/sdk"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/oauth2"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/encoding/protojson"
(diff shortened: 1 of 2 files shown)


lib/fixtures/keycloak.go
```diff
@@ -120,6 +121,9 @@ func SetupKeycloak(ctx context.Context, kcConnectParams KeycloakConnectParams) e
 
 	opentdfClientID := "opentdf"
 	opentdfSdkClientID := "opentdf-sdk"
+	opentdfSdkCertExchangeClientID := "opentdf-sdk-cert-exchange"
```
elizabethhealy, line 124: create separate client for cert exchange, i was having issues with opentdf-sdk since you cant do both token exchange and cert exchange

service/rttests/rt_test.go
```diff
@@ -71,12 +82,77 @@ var failureAttributeSets = [][]string{
...
+		"https://example.com/attr/color/value/red",
+		"https://example.com/attr/color/value/blue",
+		"https://example.com/attr/color/value/green",
+		"https://example.com/attr/cards/value/jack"},
+}
+
+var failureTokenExchangeSets = [][]string{
+	{"https://example.com/attr/language/value/french"},
+	{"https://example.com/attr/language/value/english", "https://example.com/attr/language/value/french"},
+	{"https://example.com/attr/color/value/red"},
```
elizabethhealy, line 106: possible bug: what is the expected behavior for anyOf with multiple entities in the chain?

pflynn-virtru: @ttschampel  all entities need anyOf?

pflynn-virtru (approved):

service/rttests/rt_test.go
elizabethhealy: ^ ok that is how its functioning now, i vaguely remember with v1 we had it be permissive so if any entity fufilled anyOf then it passed

PR #872, opened 2024-05-24
fix(sdk): if we encounter an error getting an access token then don't make the request

mkleene:
This has been confusing for other teams that have experienced errors with authentication


elizabethhealy:
LGTM but i dont have approval permissions

dmihalcik-virtru (approved):

sdk/auth/token_adding_interceptor.go
```diff
@@ -52,15 +54,18 @@ func (i TokenAddingInterceptor) AddCredentials(
 	if err == nil {
 		newMetadata = append(newMetadata, "Authorization", fmt.Sprintf("DPoP %s", accessToken))
 	} else {
-		slog.ErrorContext(ctx, "error getting access token. request will be unauthenticated", "error", err)
-		return invoker(ctx, method, req, reply, cc, opts...)
+		slog.ErrorContext(ctx, "error getting access token", "error", err)
+		return status.Error(codes.Unauthenticated, err.Error())
```
dmihalcik-virtru, line 58: This feels like CORS to me where a client decides to report a server error type. This is probably fine, but maybe we should consider a local error type that wraps a grpc/status instead? They should still work with errors.As (grpc.Status), but may provide some hint to the application that they are misconfigured

mkleene: In this case we are constrained to being 'compatible' with errors returned by the `status` package due to the fact that this code lives in the GRPC pipeline. My read was that we needed to have a GRPC code included.

Issue #881, opened 2024-05-28
support additional elliptic curves


strantalis:
We need to support additional elliptic curves in kas and the sdk.  Specifically P384 but should be designed in a way to support  others as well.

https://pkg.go.dev/crypto/elliptic#Curve

pflynn-virtru:
>  No Elliptic Curves less than 256-bits are supported

https://github.com/opentdf/spec/tree/main/schema/nanotdf#31-assumptions

pflynn-virtru:
Is only one key active at a time? meaning `P-384` is the only active key that is served from `kas_public_key` endpoint?
Or are all curves active, and the SDK client can use any for its use case?

PR #883, opened 2024-05-29
feat(kas): Partial implementation EC384 and EC521 support in KAS

pflynn-virtru:
Enabled EC384 and EC521 support in KAS and added relevant unit tests. The config file now supports the new EC384 and EC521 key types. Also, test keys for both types have been generated and added.

This will enable creation of a nanotdf with different curves.  This will NOT enable decryption of nanotdf since we have no place to put the key identifier, see: https://github.com/opentdf/platform/issues/900

partially resolves: https://github.com/opentdf/platform/issues/881


dmihalcik-virtru (changes requested):
This is not backward compatible and will break hsm.go. Can you test with hsm enabled? 

Also, we may need instructions about how to upgrade - e.g. what do people need to change in their opentdf.yaml files and/or rename their existing EC key files? Will this break existing nanotdf documents?

pflynn-virtru:
@biscoe916 can we get a definitive decision on HSM?

.github/scripts/init-temp-keys.sh
```diff
@@ -65,15 +65,22 @@ if [ "$opt_hsm" = true ]; then
 fi
 
 openssl req -x509 -nodes -newkey RSA:2048 -subj "/CN=kas" -keyout "$opt_output/kas-private.pem" -out "$opt_output/kas-cert.pem" -days 365
-openssl ecparam -name prime256v1 >ecparams.tmp
-openssl req -x509 -nodes -newkey ec:ecparams.tmp -subj "/CN=kas" -keyout "$opt_output/kas-ec-private.pem" -out "$opt_output/kas-ec-cert.pem" -days 365
+# P-256 aka secp256r1 , prime256v1
```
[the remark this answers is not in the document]
pflynn-virtru: we do not need to change them.  openssl handles both aliases

sdk/nanotdf.go
```diff
@@ -841,9 +857,20 @@ func (s SDK) ReadNanoTDF(writer io.Writer, reader io.ReadSeeker) (uint32, error)
...
+	switch ecMode {
+	case ocrypto.ECCModeSecp256r1:
+		req.Algorithm = "ec:secp256r1" // security.AlgorithmECP256R1
+	case ocrypto.ECCModeSecp384r1:
+		req.Algorithm = "ec:secp384r1" // security.AlgorithmECP384R1
+	case ocrypto.ECCModeSecp521r1:
+		req.Algorithm = "ec:secp521r1" // security.AlgorithmECP521R1
+	case ocrypto.ECCModeSecp256k1:
+		req.Algorithm = "ec:secp256k1" // not supported
+	}
```
patmantru, line 871: Needs a `default` case that throws an error

patmantru (changes requested):

sdk/nanotdf.go
pflynn-virtru: updated

docs/openapi/authorization/authorization.swagger.json
elizabethhealy, line 1: why remove all the swagger docs?

opentdf-dev.yaml
```diff
@@ -17,6 +17,16 @@ services:
       - kid: e1
         alg: ec:secp256r1
         legacy: true
+      - kid: e2
+        alg: ec:secp384r1
+      - kid: e2
+        alg: ec:secp384r1
+        legacy: true
```
strantalis, line 24: Do we need to legacy on everything?

dmihalcik-virtru: We probably should add a 'default' or 'current' boolean as well so we don't have to mark the same key twice in the keyring. At present, you have to list the same key twice to get it to show up in `PublicKey` and to be tested for kidless headers ('legacy' mode)

So we could replace:

```yaml
    keyring:
      - alg: ec:secp3841r1
        kid: a
      - alg: ec:secp3841r1
        kid: a
        legacy: true
```
with:

```yaml
    keyring:
      - alg: ec:secp3841r1
        kid: a
        current: true
        legacy: true
```

dmihalcik-virtru: We could also have legacy default to true, or current default to true, or maybe have current default to !legacy or something more confusing (or rename current to something else or have an enum for these options?)

docs/openapi/authorization/authorization.swagger.json
dmihalcik-virtru: I think the current plan is to only use gateway for TDF/nanoTDF clients 

All clients that use these endpoints so far use gRPC directly - even client-web: https://github.com/opentdf/client-web/pull/292

examples/cmd/decrypt.go
```diff
@@ -40,42 +39,38 @@ func decrypt(cmd *cobra.Command, args []string) error {
...
-	case n < 3:
-		return errors.New("file too small; no magic number found")
-	case bytes.HasPrefix(magic[:], []byte("L1L")):
-		isNano = true
-	default:
-		isNano = false
-	}
-	_, err = file.Seek(0, 0)
+	var isTDF = true
+	tdfreader, err := client.LoadTDF(file)
```
dmihalcik-virtru, line 47: aw, I miss my magic header check. I was kinda hoping to move this into sdk, so maybe we should move this to an issue saying we dshould have a unified API where nano/whatever is an option instead of having separate methods?

patmantru (approved):

examples/cmd/decrypt.go
pflynn-virtru: SDK function to detect isTDF() and to detect the format https://github.com/opentdf/platform/issues/1000

opentdf-dev.yaml
pflynn-virtru: is there an action to be done in this PR?

test/tdf-roundtrips.bats
```diff
@@ -134,8 +134,8 @@ server:
           private_key_path: kas-private.pem
           public_key_path: kas-cert.pem
         r2:
-          private_key_path: kas-r2-private.pem
-          public_key_path: kas-r2-cert.pem
+          private_key_path: kas-private.pem
+          public_key_path: kas-cert.pem
```
dmihalcik-virtru, line 138:
```suggestion
          private_key_path: kas-r2-private.pem
          public_key_path: kas-r2-cert.pem
```

opentdf-dev.yaml
dmihalcik-virtru: No we can update it later

opentdf-dev.yaml
```diff
@@ -17,6 +17,16 @@ services:
...
+      - kid: e3
+        alg: ec:secp521r1
```
dmihalcik-virtru, line 26: Can you add a variable to the `encrypt` that is like `--algorithm` or `--curve` so we can use these in the bats test?

opentdf-dev.yaml
dmihalcik-virtru: No we can update it later

dmihalcik-virtru (approved):
If you want to submit this as is we can do the e2e test in the bats as a follow up

patmantru (approved):

dmihalcik-virtru (approved):

strantalis (dismissed):

dmihalcik-virtru (approved):

dmihalcik-virtru (approved):

pflynn-virtru:
replaced with kid work

PR #888, opened 2024-05-30
fix(core): Return 404 if public key not found

dmihalcik-virtru:
Updates kas_public_key endpoints to return a 404 instead of a 500 if the public key lookup fails. Potential ways a public key is not found:

1. unsupported algorithm
2. unrecognized key identifier (either user supplied TK or default)

This will still 500 on errors in config format or key format

TK should also return 400 on invalid fmt

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -191,7 +191,7 @@ jobs:
       - run: curl --show-error --fail-with-body --insecure localhost:8080/kas/v2/kas_public_key
       - run: curl --show-error --fail-with-body --insecure localhost:8080/kas/v2/kas_public_key?algorithm=ec:secp256r1
       - run: |-
-          curl_status=$(curl -o -I -L -s -w "%{http_code}" localhost:8080/kas/v2/kas_public_key?algorithm=invalid)
+          curl_status=$(curl -o /dev/null -s -w "%{http_code}" localhost:8080/kas/v2/kas_public_key?algorithm=invalid)
           [ $curl_status = 404 ]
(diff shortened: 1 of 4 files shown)


pflynn-virtru (dismissed):

jakedoublev:
Will this also fix https://github.com/opentdf/platform/issues/902?

.github/workflows/checks.yaml
```diff
@@ -188,7 +188,11 @@ jobs:
       - run: grpcurl -plaintext localhost:8080 list
       - run: grpcurl -plaintext localhost:8080 grpc.health.v1.Health.Check
       - run: grpcurl -plaintext localhost:8080 kas.AccessService/PublicKey
-      - run: curl --show-error --fail --insecure localhost:8080/kas/v2/kas_public_key
+      - run: curl --show-error --fail-with-body --insecure localhost:8080/kas/v2/kas_public_key
+      - run: curl --show-error --fail-with-body --insecure localhost:8080/kas/v2/kas_public_key?algorithm=ec:secp256r1
+      - run: |-
+          curl_status=$(curl -o /dev/null -s -w "%{http_code}" localhost:8080/kas/v2/kas_public_key?algorithm=invalid)
+          [ $curl_status = 404 ]
```
strantalis, line 195: Should we validate the grpc output as well?

dmihalcik-virtru: ok

dmihalcik-virtru:
> Will this also fix #902?

I've added a fix for this as well

pflynn-virtru (approved):

strantalis (approved):

Issue #804, opened 2024-05-14
go logging: Have some standards for logging


dmihalcik-virtru:
Alternatives:

1. prefer `LogAttrs` for efficiency
2. Use custom logger type (platform.logger) to enable ease of `Audit` events. Prefer using Context and Attr structs.
3. Use default logger with custom Level and no requirements.

---
## LogAttrs

- https://pkg.go.dev/log/slog#hdr-Attrs_and_Values
- https://github.com/uber-go/zap?tab=readme-ov-file#performance

How it looks:

```go
import (
	"log/slog"
	"time"
	"github.com/platform/service/audit"
)

...

func x() {
	slog.LogAttrs(
		ctx,
		audit.Level,
		"x function started",
		slog.Time("now", time.Now()),
	)
}
```

## Custom Logger

> Move internal/logger/logger.go to service/logger/logger.go
> Update linter to look for use of default logger or other brand of logger and block them.

## No Rules

> Generate a custom handler that logs certain levels as audit or when various attr/context matchers are hit.
> Register custom handler at service start

jrschumacher:
Are you suggesting we use this syntax `logger.LogAttrs(ctx, slog.LevelInfo, "hello", slog.Int("count", 3))` rather than `slog.InfoContext(ctx, "hello", slog.Int("count", 3))`?

What kind of efficiency gains do we get? Seems that this could prove challenging to enforce.

dmihalcik-virtru:
It looks like the efficiency gains are minimal. [zap maintains a performance comparison](https://github.com/uber-go/zap?tab=readme-ov-file#performance) and it looks like we only save 2 allocs (out of about 4x the number of attributes we allocate in a line).

Issue #328, opened 2024-03-07
platform logging improvements


pflynn-virtru:
- Formalize logging message format - lowercase, only period punctionation.
- Decide on and record Error codes
- Create logging integration test to enforce the above

- #710
- #712
- #715
- #716
- #720

pflynn-virtru:
notes:
- Don't lose error, have an error stack trace, keep chain intact
- No fat. or log. statements.
- slog with errors.join
  - everytime you create an error, slog it, with some golang line or code block identifier 
  - ideally everything in one line log
  - slog.XXXWithContext
  - add service to Context so log messages show which service
- Include user error message, not just technical  -- error id, user message 
- Follow Java 

Users:
library users - error messaging through API and SDK
end users messaging

pflynn-virtru:
possible option to log line number and package

pflynn-virtru:
reference: https://betterstack.com/community/guides/logging/logging-in-go/

biscoe916:
@strantalis @pflynn-virtru We're going to do this, we just need to flesh this ticket out a bit more.

jrschumacher:
Could we also get something added to Github Discussions or the README about our logging practice to better onboard teams?

Issue #115, opened 2024-02-01
Policy API: Enable admins to perform unsafe mutations on data in their platform


jrschumacher:
Implement the unsafe mutation RPCs so that admins can have an audit trail when they need to make unsafe changes to their platform

## Depends on

- #115 

## Acceptance Criteria

- Create an UnsafeService which captures the unsafe update RPC to empower platform admins
	- e.g. `UnsafeService.UnsafeUpdateAttributeName()`
	- e.g. `UnsafeService.UnsafeDeleteAttribute()`
- Every unsafe property should require passing a value that would provoke the user to question what they are doing or why the platform requires it

jakedoublev:
Moved back to `todo` and unassigned myself as other priorities consistently took precedence this week.

jakedoublev:
Namespaces
- [ ] reactivate (no cascade)
- [ ] update name (and fqn)
Definitions
- [ ] reactivate (no cascade)
- [ ] update name (and fqn)
- [ ] update rule
Values 
- [ ] reactivate
- [ ] update value (and fqn)

jakedoublev:
Out of scope: audit events for unsafe policy events, which will be handled in a separate issue to follow.

PR #889, opened 2024-05-30
feat(core): Rewrap and Policy CRUD Audit Events

### LINKED ISSUE #884: Audit Events: Rewrap and Policy CRUD
Parent: #719 

Get audit events firing for rewrap requests and policy CRUD.

Living document for the audit events and their structure: 
https://whimsical.com/audit-HkQXHSMxu8xaWZUznhFCPW

### Rewrap

Rewrap requests are mostly done and already being emitted. The data being passed in just needs to be tweaked to match the schema defined in the whimsical document above

```json
{
  "object": {
    "type": "key_object",
    "id": "uuid", // policy uuid 
    "name": "string - optional", // not needed
    "attributes": {
      "assertions": [], // new field
      "attrs": [], // data attributes from policy body
      "permissions": [] // - keep empty
    }
  },
  "action": {
    "type": "rewrap",
    "result": "success | failure"
  },
  "owner": {
    "id": "ad484b11-c3ee-44d1-b552-7ee9bdcf33b0",
    "orgId": "ad484b11-c3ee-44d1-b552-7ee9bdcf33b0"
  },
  "actor": {
    "id": "string", // sub claim in token
    "attributes": null
  },
  "eventMetaData": {
    "keyID": "string",
    "policyBinding": "string",
    "tdfFormat": "nanotd ,ztdf, ictdf",
    "algorithm": "..."
  },
  "clientInfo": {
    "userAgent": "string",
    "platform": "string - secure reader, bp, etc..",
    "requestIp": "string",
  },
  "diff": "show's what changed in the object - only needed for update actions",
  "requestId": "uuid", // New Field
  "timestamp": "date"
}
```

### Policy CRUD

This hasn't been implemented yet so will need to be started and will be the bulk of the work for this issue. The format found in the whimsical is here:

```json
{
  "object": {
    "type": "string - [subject_mapping, resource_mapping, attribute_definition, attribute_value, namespace, condition_set, kas_registry,kas_attribute_assignment, ]",
    "id": "uuid",
    "name": "string - optional",
    "attributes": {
      "attrs": [],
      "permissions": [] // - only for user_objects
    }
  },
  "action": {
    "type": "string - [create,  update, delete ]",
    "result": "string - [success, error]"
  },
  "owner": {
    "id": "ad484b11-c3ee-44d1-b552-7ee9bdcf33b0",
    "orgId": "ad484b11-c3ee-44d1-b552-7ee9bdcf33b0"
  },
  "actor": {
    "id": "string", // sub out of access token
    "attributes": {
      "permissions": [],
      "attrs": []
    }
  },
  "eventMetaData": {
    // <meta data>
  },
  "clientInfo": {
    "userAgent": "string",
    "platform": "string - secure reader, bp, etc..",
    "requestIp": "string",
  },
  "diff": "show's what changed in the object - only needed for update actions", // Only on update policy objects
  "timestamp": "date"
}
```

pbacon-blaber:
- cleans up structure of audit logging
- adds fields as discussed to the rewrap audit event
- adds policy CRUD audit events
- implements diff generation using https://github.com/wI2L/jsondiff
- closes #884

### Running Questions

Some questions I had as I worked on this, to be discussed in this PR or on Slack. I will update the questions with their answers below when they are answered

#### Deactivated Events

> Should deactivated actions be considered as `delete` or `update` action types in the audit event?

After discussion with Pat, I'm leaning towards making these `update` action types as the resource still technically exists and could be "activated" with a small code change (although currently there are no endpoints to support activating anything that has been deactivated)

#### Update `diff` vs Performance

> When we perform an update, we're doing a SQL update call, in order to generate a "diff" for the audit event we'd have to load the resource which adds another DB call. Is this something we want to do?

I confirmed with @jakedoublev that it's correct we don't have access to the existing data at the time of update. Also confirmed with @strantalis that we should load the existing item in order to calculate the diff before the update as well.

#### Rewrap

- [x] rewrap

#### Policy CRUD

- [x] attribute definition
  - [x] created
  - [x] updated
  - [x] updated (deactivated)
- [x] attribute value
  - [x] created
  - [x] updated
  - [x] updated (deactivated)
- [x] kas attribute definition assignment
  - [x] created
  - [x] deleted
- [x] kas attribute value assignment
  - [x] created
  - [x] deleted
- [x] subject mapping
  - [x] created
  - [x] updated
  - [x] deleted
- [x] condition sets
  - [x] created
  - [x] updated
  - [x] deleted
- [x] resource mappings
  - [x] created
  - [x] updated
  - [x] deleted
- [x] namespaces
  - [x] created
  - [x] updated
  - [x] updated (deactivated)
- [x] kas registry
  - [x] created
  - [x] updated
  - [x] deleted


pflynn-virtru:
I can toggle those checkboxes for some reason

pbacon-blaber:
> I can toggle those checkboxes for some reason

I think that's fine, I was mainly using that as just tracking my own work to make sure I didn't miss anything.

service/internal/logger/audit/policy.go
```diff
@@ -0,0 +1,84 @@
...
+	ObjectTypeSubjectMapping                   = "subject_mapping"
+	ObjectTypeResourceMapping                  = "resource_mapping"
+	ObjectTypeAttributeDefinition              = "attribute_definition"
+	ObjectTypeAttributeValue                   = "attribute_value"
+	ObjectTypeNamespace                        = "namespace"
+	ObjectTypeConditionSet                     = "condition_set"
+	ObjectTypeKasRegistry                      = "kas_registry"
+	ObjectTypeKasAttributeDefinitionAssignment = "kas_attribute_definition_assignment"
+	ObjectTypeKasAttributeValueAssignment      = "kas_attribute_value_assignment"
+)
```
strantalis, line 20: golang doesn't have the concept of enums but you can still sort of represent them if you think it might be beneficial here and also for action types.

Here are a couple of examples
- https://www.sohamkamani.com/golang/enums/
- https://builtin.com/software-engineering-perspectives/golang-enum

pbacon-blaber: @strantalis made those updates now and put them inside `audit/constants.go`

service/internal/server/server.go
```diff
@@ -263,6 +269,14 @@ func newGrpcServer(c Config, a *auth.Authentication) (*grpc.Server, error) {
 		i...,
 	))
 
+	// Chain relaip interceptor
+	trustedPeers := []netip.Prefix{} // TODO: add this as a config option?
```
strantalis, line 273: @pbacon-blaber Can we track this enhancement in an issue.

pbacon-blaber: https://github.com/opentdf/platform/issues/957

service/internal/logger/audit/serviceInterceptor.go
```diff
@@ -0,0 +1,33 @@
...
+// to the context of incoming requests. This metadata is used to log audit events
+func UnaryServerInterceptor(ctx context.Context, req any, _ *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (any, error) {
+	// Generate and set a request ID on incoming requests
+	requestUUID := uuid.New()
+	ctx = context.WithValue(ctx, RequestIDContextKey, requestUUID)
+
+	// Get metadata from the context
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil, status.Error(codes.Unauthenticated, "missing metadata")
```
jakedoublev, line 23: Are there any public unauthenticated routes this interceptor expects to hit? If so, is there a more accurate code than unauthenticated? https://grpc.io/docs/guides/status-codes/#the-full-list-of-status-codes

service/policy/attributes/attributes.go
```diff
@@ -39,13 +40,21 @@ func (s AttributesService) CreateAttribute(ctx context.Context,
...
+		ObjectType: audit.ObjectTypeAttributeDefinition,
+		ActionType: audit.ActionTypeCreate,
+	}
+
 	item, err := s.dbClient.CreateAttribute(ctx, req)
 	if err != nil {
+		s.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
 		return nil, db.StatusifyError(err, db.ErrTextCreationFailed, slog.String("attribute", req.String()))
 	}
-	rsp.Attribute = item
```
jakedoublev, line 46: Was this intentionally removed?

jakedoublev:
> After discussion with Pat, I'm leaning towards making these update action types as the resource still technically exists and could be "activated" with a small code change (although currently there are no endpoints to support activating anything that has been deactivated)

I concur with this opinion. There is a need to support unsafe actions (renaming namespaces, deleting policy, reactivating deactivated policy) and those unsafe deletes will be true deletes. https://github.com/opentdf/platform/issues/115

service/policy/attributes/attributes.go
```diff
@@ -99,11 +108,30 @@ func (s *AttributesService) UpdateAttribute(ctx context.Context,
...
+
+	updated, err := s.dbClient.UpdateAttribute(ctx, req.GetId(), req)
+	if err != nil {
+		s.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
 		return nil, db.StatusifyError(err, db.ErrTextUpdateFailed, slog.String("id", req.GetId()), slog.String("attribute", req.String()))
 	}
-	rsp.Attribute = a
+
+	auditParams.Original = original
+	auditParams.Updated = updated
```
jakedoublev, line 131: Have you validated you have the expected `updated` policy objects available for the audit diff? I believe most of the update SQL statements only return the ID from the updated row so your expected diff may require an additional roundtrip. If so, that should be fine performance-wise as reads are expected to be thousands to millions more common than writes to policy objects.

service/policy/attributes/attributes.go
pbacon-blaber: nope, good catch

service/policy/attributes/attributes.go
pbacon-blaber: I'll dig down and double check these here

service/internal/logger/audit/serviceInterceptor.go
pbacon-blaber: Yeah I guess I just defaulted to this because I originally had this in the authn.go file before I decided it didn't fit there that well. I would imagine the best code from that list might be `INVALID_ARGUMENT`?

jakedoublev: I think that makes sense. It's the equivalent of a 400 in gRPC.

pbacon-blaber: Updated to INVLAID_ARGUMENT

service/policy/attributes/attributes.go
pbacon-blaber: @jakedoublev made those changes to fetch the updated version from the DB again, I noticed that the `deactivate` methods seem to actually return the entire object instead of just the ID so we don't need to fetch again for those methods

https://github.com/opentdf/platform/blob/da4d78873cf4a7baa125647930dbf7963626fd68/service/policy/db/namespaces.go#L306

jakedoublev (approved):
Is there a configuration story being tracked for audit? Some things that come to mind are:
1. enabling/disabling logging for various services based on config enablement state
2. writing audit logs to a different output location than stdout, including a remote store, and/or a different output than other generic logs

Also, given that we're now doing two more database roundtrips per policy `update` call to complete the audit diff with a proper pre- and post-update state, it could get expensive and we could consider making it a little more async. Can we open an issue for some audit profiling and link it to the existing performance testing ticket (https://github.com/virtru-corp/data-security-platform/issues/63)? I still think updates (safe or unsafe) will be relatively rare, but it's worth testing at some point to inform any assumptions we make.

There are existing issues around standardizing logging and linting of log calls we may want to follow up on after this work is merged: https://github.com/opentdf/platform/issues/328 & https://github.com/opentdf/platform/issues/804

strantalis (approved):

strantalis:
> 1. enabling/disabling logging for various services based on config enablement state

If a service is disabled there won't be any logs. Are you asking to change the log levels per service?

> 2. writing audit logs to a different output location than stdout, including a remote store, and/or a different output than other generic logs

This is not being asked for at this time. So it's not being tracked as future work.

jakedoublev:
@strantalis 

> 1. enabling/disabling logging for various services based on config enablement state
If a service is disabled there won't be any logs. Are you asking to change the log levels per service?

> @jakedoublev
> 
> > 1. enabling/disabling logging for various services based on config enablement state
> 
> If a service is disabled there won't be any logs. Are you asking to change the log levels per service?

I was thinking it could be valuable to disable audit itself within a running service, but that would only really come into play if audit were a standalone service itself responsible for sending audit logs to a remote store. If that's not on the roadmap, it's not relevant.

pbacon-blaber:
> @strantalis
> 
> > 1. enabling/disabling logging for various services based on config enablement state
> >    If a service is disabled there won't be any logs. Are you asking to change the log levels per service?
> 
> > @jakedoublev
> > > 1. enabling/disabling logging for various services based on config enablement state
> > 
> > 
> > If a service is disabled there won't be any logs. Are you asking to change the log levels per service?
> 
> I was thinking it could be valuable to disable audit itself within a running service, but that would only really come into play if audit were a standalone service itself responsible for sending audit logs to a remote store. If that's not on the roadmap, it's not relevant.

I had this question when I started developing audit and I think at the moment we decided that it will not be possible to configure audit to toggle it off. It will always be on (as of now). 

I think at the time we had that discussion we didn't realize we had to make two extra round trips for update events, so maybe that adds more in favor of allowing audit (and these round trips) to be toggled off. It could warrant further discussion

PR #901, opened 2024-05-31
chore(ci): Updates to latest golangci-lint

dmihalcik-virtru:


.golangci.yaml
```diff
@@ -31,7 +33,7 @@ linters-settings:
         # Default: true
         skipRecvDeref: false
 
-  gomnd:
+  mnd:
```
pflynn-virtru, line 36: why the rename?

patmantru (approved):

.golangci.yaml
dmihalcik-virtru: I don't know why, but golangci replaced it. It still supports gomnd as an alias but reports warnings, so I went ahead and renamed it.

https://golangci-lint.run/product/changelog/#v1580

PR #906, opened 2024-05-31
feat(core)!: Single module

dmihalcik-virtru:
Instead of keeping sdk, service, in separate modules and working to improve automation around them, let's simplify our release procedure by making a single module and versioning that.


b-long:
Do we have any measurements / guesses on impact to build times & artifact size (e.g. for someone that depends on just the SDK)?  That said, I will say I'm not terribly concerned about build times/artifact size, but I do hope this doesn't break the process described below 🙊 


We do have a use case that is **only** depending on the SDK.  Since https://github.com/opentdf/client-python was made private, we starting producing our own Python wheel by compiling the OpenTDF Go SKD using [`gopy`](https://github.com/go-python/gopy).  Tangentially, we'd be happy to make that project public and support it if there's any interest.

Previously, in that project our `go.mod` file used `require github.com/opentdf/platform/sdk v0.2.0` and produced a 9.1MB Python wheel for Linux in 1m10s.

Now, our `go.mod` file uses `require github.com/opentdf/platform/sdk v0.2.5` and produces a 9.1MB Python wheel for Linux in 1m9s.

Very excited about the future of OpenTDF 🎉 🙂  Just wanted to mention that there is a project that exists today and only consuming the SDK.

damorris25:
If you want to use Python, why not generate the SDK from the protobufs?

jrschumacher:
> Tangentially, we'd be happy to make that project public and support it if there's any interest.

@b-long since Python isn't on the roadmap for official SDK support, why don't you create it under your personal GitHub account. I'll bring up the topic of adding a section on the [README](https://github.com/opentdf/.github/blob/main/profile/README.md) for Community Supported SDKs.

jrschumacher:
@dmihalcik-virtru my only concern with this is that we will get a number of irrelevant indirect dependencies pulled into otdfctl and downstream Go PEPs. For instance, there are a number of DB dependencies in platform that SDK consumers don't need. 

Could we not revert #495 so the root go.mod would be the platform which would simplify the platform release while keeping SDK deps clean?

b-long:
> If you want to use Python, why not generate the SDK from the protobufs?

Will do! 🙂  In part, I just wanted to try `gopy` wheels (we've used gRPC successfully before) 👍 I'll add this to our roadmap, and plan to open source it under my personal account.

Issue #900, opened 2024-05-31
ADR: NanoTDF KAS resource locator path and key identifier


pflynn-virtru:
# NanoTDF KAS resource locator path and key identifier

### Problem 
1. KAS resource locator usage varies
2. No identifier for the KAS key in a NanoTDF

The NanoTDF specification requires enhancements to support key identifier and multiple ways to access KAS.

> This section contains a Resource Locator type that allows describing access to a resource. In the case of the KAS, the Resource Locator defines how to access a KAS.

> The Resource Locator is a way for the nanotdf to represent references to external resources in as succinct a format as possible.

See https://github.com/opentdf/spec/tree/main/schema/nanotdf#3312-kas

Example body with protocol values:
```
https://secure.virtru.com/api/kas
http://kas.example.com:9000
\\securehost\kas
```

### How to access a KAS

1. Parse KAS resource locator field from NanoTDF header
3. Create a URL using protocol enum and body.  Body usually contains domain with a partial path, if needed, to the KAS service
4. Append `/v2/rewrap` or `/kas/v2/rewrap` or `/rewrap` to URL from step 2 (_varies by SDK_)
5. Perform a RESTful or gRPC call using the [TDF protocol](https://github.com/opentdf/spec/tree/main/protocol) (_varies by SDK_)

### Which KAS key

As we introduce multiple KAS keys and perform key rotations, we need a key identifier `kid` used in creating the NanoTDF so a rewrap operation can use the same key.

### Policy Key Access

This section allows for an ephemeral key other than the Payload key to encrypt the policy.

See https://github.com/opentdf/spec/tree/main/schema/nanotdf#342323-optional-policy-key-access

## Goal

1. Clarify specification in regards to "How to access a KAS":
- Define how to build the URL to perform a rewrap request including `kid`
- Define what parts of the URL are defined where
- Define where to find what RPC methods are available
- Define a strategy on how to find KAS if relocated in the future
2. Update specification in regards to "Which KAS key":
- Choose `kid` or public key
- Clarify keys (policy and payload key?)
- Define where to locate `kid` on decryption
- Define how to determine `kid` on encryption
- Define what format it takes
- Define how to use it when accessing KAS

## Related

- https://github.com/opentdf/spec/pull/21
- https://github.com/opentdf/platform/issues/917
- https://github.com/opentdf/platform/issues/873

Included here because a possible version change to NanoTDF specification could influence this decision.

### ⚖️ Add or Use Key Identifier section
[...]

PR #939, opened 2024-06-06
feat(core): New cryptoProvider config

dmihalcik-virtru:
- define new `cryptoProvider` config structs to support rotation with different keys
- this means the algorithm must be present
- I'm using some heuristics to maintain backward compatibility of standard crypto configs, but hsm configs must be updated
- Adds `kid` in response to kas_public_key
- Adds `kid` in key access objects produced by SDK
- Still no support in nanotdf, as we have not agreed on how/where to store the `kid` value in the header. See https://github.com/opentdf/platform/issues/900

New Config:

```yaml
server:
  cryptoProvider:
    type: standard
    standard:
      keys:
        - kid: r1
          alg: rsa:2048
          private: kas-private.pem
          cert: kas-cert.pem
        - kid: r0
          alg: rsa:2048
          private: kas-private-old.pem
          cert: kas-cert-old.pem
        - kid: e1
          alg: ec:secp256r1
          private: kas-ec-private.pem
          cert: kas-ec-cert.pem
services:
  kas:
    enabled: true
    keyring:
      - alg: rsa:2048
        kid: r1
      - alg: rsa:2048
        kid: r0
        legacy: true
      - alg: ec:secp256r1
        kid: e1
```

some notes:

- `kid` values should be unique, preferably for the lifetime of the kas host domain name
- `kid` values should short strings (I'd suggest maxing out at 44 characters)
- `private` and `cert` indicate the location of private key and a certificate, if available
- For `hsm` keys, these should be label values
- For `standard` keys, these should be paths to PEM files relative to the current working directory
- I've deprecated the `eccertid` for a new `keyring` parameter which describes how KAS will interpret the key. So we have two sections: `server.cryptoProvider` describes what keys are available, while `service.kas.keyring` describes how KAS uses those keys.
- We don't have a 'rotate' script yet. To do this manually, update the init-temp-keys script to use a new name/label and rerun and add the new keys to the list, updating the `certid` fields to point to the new values

To come:

1. nanoTDF support


strantalis:
@dmihalcik-virtru I opened up this issue earlier today. I think @ttschampel ran into similar issues. But it's not clear what eccertid is needed for or what it does. 

Also what are your thoughts on moving the crypto provider. It has felt kind of odd living under the server block when kas is the only service that leverages it. 

https://github.com/opentdf/platform/issues/954

@biscoe916 @jrschumacher Would we use this same crypto for signing attributes or other resources in the future?

pflynn-virtru (approved):

dmihalcik-virtru:
@strantalis We can probably move all of the config into the `keyring` section with little disruption to KAS. The code is written at the moment to allow all services to use a shared crypto provider, which is really an abstraction layer that allows us to switch between a pkcs#11 backend and a golang crypto backend - which could be something other services desire but in practice so far we haven't needed it elsewhere

patmantru (approved):

PR #941, opened 2024-06-06
fix(sdk): convert platform endpoint to grpc dial format

jrschumacher:


sdk/sdk.go
```diff
@@ -143,6 +152,34 @@ func New(platformEndpoint string, opts ...Option) (*SDK, error) {
...
+
+	if !regexp.MustCompile(`^(https?:\/\/)?(([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(:\d+)?|(localhost)(:\d+)?)(\/.*)?$`).MatchString(e) {
+		return "", errors.New("platform endpoint is not a valid URL")
+	}
+
+	u, err := url.ParseRequestURI(e)
+	if err != nil {
+		return "", errors.Join(fmt.Errorf("cannot parse platform endpoint(%s)", e), err)
+	}
+	fmt.Printf("u: %v\n", u)
```
jrschumacher, line 169:
```suggestion
```

sdk/sdk_test.go
```diff
@@ -125,3 +125,84 @@ func Test_ShouldCreateNewSDKWithBadEndpoint(t *testing.T) {
 	require.NoError(t, err)
 	assert.NotNil(t, s)
 }
+
+func Test_ShouldSantizePlatformEndpoint(t *testing.T) {
```
dmihalcik-virtru, line 129: 🎅🏻 🎅🏻 🎅🏻 🎅🏻

sdk/sdk.go
dmihalcik-virtru, line 171:
```suggestion
```

sdk/sdk.go
```diff
@@ -143,6 +152,34 @@ func New(platformEndpoint string, opts ...Option) (*SDK, error) {
 	}, nil
 }
 
+func SanitizePlatformEndpoint(e string) (string, error) {
+	// check if there's a scheme, if not, add https
+	if !regexp.MustCompile(`^https?://`).MatchString(e) {
+		e = "https://" + e
+	}
+
+	if !regexp.MustCompile(`^(https?:\/\/)?(([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(:\d+)?|(localhost)(:\d+)?)(\/.*)?$`).MatchString(e) {
```
dmihalcik-virtru, line 163:
```suggestion
	if !regexp.MustCompile(`^(https?:\/\/)?(([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(:\d+)?|(localhost)(:\d+)?)\/?$`).MatchString(e) {
```
Do we want to allow paths on the platform endpoints?

sdk/sdk.go
```diff
@@ -143,6 +152,34 @@ func New(platformEndpoint string, opts ...Option) (*SDK, error) {
...
+		return "", errors.New("platform endpoint is not a valid URL")
```
dmihalcik-virtru, line 164:
```suggestion
		return "", errors.New("platform endpoint is not valid")
```

sdk/options.go
```diff
@@ -151,3 +152,11 @@ func WithPlatformConfiguration(platformConfiguration PlatformConfiguration) Opti
 		c.platformConfiguration = platformConfiguration
 	}
 }
+
+// WithIPC returns an Option that indicates the SDK should use IPC for communication
+// this will allow the platform endpoint to be an empty string
+func WithIPC() Option {
```
dmihalcik-virtru, line 158: Suggestion - why not use a custom scheme, since this implies a different transport?

sdk/sdk.go
jrschumacher: Probably not. Rather tell people they need to run this on the root of the domain.

sdk/options.go
jrschumacher: The issue with this is that there is no scheme at all. When we bind the SDK into the platform, it's an empty string `sdk.New("")` I'm open to a new name... maybe `WithNoEndpoint()` or something like that?

dmihalcik-virtru: I'm fine with this since we intend to support it indefinitely

dmihalcik-virtru (dismissed):
It looks like you need to run `make fix` to get some transitive updates.

strantalis (approved):

PR #961, opened 2024-06-10
fix: deprecate the consumer section

jrschumacher:
The work to simplify consumption of the platform didn't work as expected. Removing that until we have time to work on a docker compose.


opentdf-example.yaml
```diff
@@ -15,23 +15,23 @@ services:
     enabled: true
   entityresolution:
     enabled: true
-    url: http://keycloak:8888/auth
+    url: http://localhost:8888/auth
     clientid: "tdf-entity-resolution"
     clientsecret: "secret"
     realm: "opentdf"
-    legacykeycloak: true
+    legacylocalhost: true
```
elizabethhealy, line 22: this should still be `legacykeycloak`

jrschumacher, line 22:
```suggestion
    legacykeycloak: true
```

Issue #950, opened 2024-06-07
authz: Return DENY decision instead of throwing error when encountering attribute that doesn't exist


elizabethhealy:
Right now in getDecisions, if we encounter an attribute that does not exist, we just return the error that we get from GetAttributeVluesByFqns.
Instead we should catch that specific error and return a deny decision.

pflynn-virtru:
KAS is enforcing the attribute rules from a trusted attribute authority (not to mention entitlement authority).
The attribute authority has a set of namespaces.  

If an attribute namespaces is in the system and the attribute name or value is not then DENY. 
If an attribute namespaces is not in the system then ALLOW

elizabethhealy:
^ might need some more informant errors from the attributes service if we do this. rn we call getAttrValueByFqn and i believe it just returns "not found" if that fqn doesnt exist. we might need some more detail around whether its the namespace, attribute, or value that doesnt exist

jakedoublev:
> ^ might need some more informant errors from the attributes service if we do this. rn we call getAttrValueByFqn and i believe it just returns "not found" if that fqn doesnt exist. we might need some more detail around whether its the namespace, attribute, or value that doesnt exist

I also encountered an unexpected behavior today where we return an error from `GetAttributesByValueFqns` if [a single fqn is not found](https://github.com/opentdf/platform/blob/main/service/policy/db/attribute_fqn.go#L188-L191). I'm thinking this should be updated to log and ignore that single nonexistent fqn in the response without erroring in the entire request. @elizabethhealy @pflynn-virtru does that align with your thinking?

PR #962, opened 2024-06-11
fix(authz): Return deny on GetDecision if resource attribute lookup returns not found

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/950


service/authorization/authorization.go
```diff
@@ -175,7 +175,28 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
 			// get attribute definition/value combinations
 			dataAttrDefsAndVals, err := retrieveAttributeDefinitions(ctx, ra, as.sdk)
 			if err != nil {
-				// TODO: should all decisions in a request fail if one FQN lookup fails?
+				// if attribute an FQN does not exist
+				// return deny for all entity chains aginst this RA set and continue to next
+				if strings.Contains(err.Error(), db.ErrTextNotFound) {
```
dmihalcik-virtru, line 180: Can we somehow use `errors.Is` here instead of seeking? I suspect this is about as performant, so I'm okay if this is the recommended solution

dmihalcik-virtru (approved):

service/authorization/authorization.go
strantalis: I don't think those db errors actually implement the error interface. We should probably fix that if we plan to leverage these in other places besides policy. 

cc: @jakedoublev

jakedoublev: The [`db.ErrTextNotFound` is an error message](https://github.com/opentdf/platform/blob/1d76303af08bc1f479f7163a32141a5506ebe6ea/service/pkg/db/errors.go#L101) passed to [`status.Error()` here along with the not found code](https://github.com/opentdf/platform/blob/1d76303af08bc1f479f7163a32141a5506ebe6ea/service/pkg/db/errors.go#L116-L118), and that function signature [expects a code and a message](https://github.com/grpc/grpc-go/blob/v1.64.0/status/status.go#L58).

It's the safe error message, not a Go error. Perhaps a rename to `db.ErrMsgTextNotFound` as the safe external RPC error message would be the right call? And possibly checking for the code of the response as `codes.NotFound` instead of the error message string?

patmantru (approved):

PR #963, opened 2024-06-11
fix(core): Add dependencies to service configs

elizabethhealy:
Some services require others and make calls to others. Define dependencies for each service that are also required to be enabled for that service to function.
This can be expanded upon to do a waterfall startup if a service ends up requiring another on start.

### DIFF
--- a/service/pkg/server/start.go
+++ b/service/pkg/server/start.go
@@ -72,6 +72,19 @@ func Start(f ...StartOptions) error {
 		return fmt.Errorf("issue registering services: %w", err)
 	}
 
+	// verify dependencies
+	slog.Info("verifying dependencies")
+	for name, service := range conf.Services {
+		if service.Enabled && (service.Dependencies != nil) {
+			for _, dep := range service.Dependencies {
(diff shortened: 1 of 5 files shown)


elizabethhealy:
maybe this should be hardcoded instead of in the config?

service/pkg/server/start.go
```diff
@@ -72,6 +72,19 @@ func Start(f ...StartOptions) error {
 		return fmt.Errorf("issue registering services: %w", err)
 	}
 
+	// verify dependencies
+	slog.Info("verifying dependencies")
+	for name, service := range conf.Services {
+		if service.Enabled && (service.Dependencies != nil) {
```
strantalis, line 78: @elizabethhealy I think we want to check if remote is defined here because in the near future for example kas will run a a separate deployment.

elizabethhealy: if Remote is defined is should enabled be true? or does `enabled` pertain to whether its enabled here, not just anywhere

strantalis: enabled is only for within the current running service.

PR #969, opened 2024-06-11
chore(ci): Adds ci for key rotation example

dmihalcik-virtru:


test/service-start.bats
```diff
@@ -71,15 +71,3 @@
...
-  go run ./examples encrypt "Hello Zero Trust"
-  go run ./examples decrypt sensitive.txt.tdf
-  go run ./examples decrypt sensitive.txt.tdf | grep "Hello Zero Trust"
-}
-
-@test "examples: roundtrip nanoTDF" {
-  go run ./examples encrypt -o sensitive.txt.ntdf --nano "Hello NanoTDF"
-  go run ./examples decrypt sensitive.txt.ntdf
-  go run ./examples decrypt sensitive.txt.ntdf | grep "Hello NanoTDF"
-}
```
pflynn-virtru, line 85: Some people are using this as a reference.  Can we keep it?  or leave some comment

test/tdf-roundtrips.bats
```diff
@@ -0,0 +1,220 @@
+#!/usr/bin/env bats
+
```
pflynn-virtru, line 2: Add description of process below.  Not long

pflynn-virtru (approved):

test/service-start.bats
dmihalcik-virtru: These are moved over to tdf-roundtrips.bats. I've refactored that file to be easier to read (moving helper methods to the end instead of the start) and added a note at the start of this file that it is just for tests that make sure the system if running and exposing the expected featuers.

test/tdf-roundtrips.bats
```diff
@@ -0,0 +1,163 @@
...
+@test "examples: legacy key support Z-TDF" {
+  echo "[INFO] validating default key is r1"
+  [ $(grpcurl -plaintext "localhost:8080" "kas.AccessService/PublicKey" | jq -e -r .kid) = r1 ]
+
+  echo "[INFO] encrypting samples"
+  go run ./examples encrypt -o sensitive-with-no-kid.txt.tdf --no-kid-in-kao "Hello Legacy"
+  go run ./examples encrypt -o sensitive-with-kid.txt.tdf "Hello with Key Identifier"
+
+  echo "[INFO] decrypting..."
+  go run ./examples decrypt sensitive-with-no-kid.txt.tdf | grep "Hello Legacy"
```
strantalis, line 41: should we use the bats-assert helper here?

test/tdf-roundtrips.bats
dmihalcik-virtru: Better?

test/tdf-roundtrips.bats
dmihalcik-virtru: Pro: Simpler test assertions and clearer errors make easier debugging.
Con: Another dep and thing to learn. Hard to install on macos

Moved to another PR for discussion:

https://github.com/opentdf/platform/pull/981/commits/82ce1f5a35b6640e994806daf450eb5465fd8fd8

strantalis (approved):

PR #977, opened 2024-06-12
chore(ci): Disallows breaking proto changes

dmihalcik-virtru:


pflynn-virtru (approved):

.github/workflows/checks.yaml
```diff
@@ -214,8 +214,6 @@ jobs:
         with:
           input: service
       - uses: bufbuild/buf-breaking-action@a074e988ee34efcd4927079e79c611f428354c01
-        # TODO(#212) Block on breaking changes after protos are frozen
-        continue-on-error: true
```
pflynn-virtru, line 218: I assume `continue-on-error` is default of `false`

pflynn-virtru (approved):

PR #981, opened 2024-06-13
chore(ci): Adds bats-assert as test dep 

dmihalcik-virtru:
Should we do this?

Pros:
- tests are easier to read
- test output is significantly more actionable in the event of an error

Cons:
- Installing libraries requires more work for developers
- harder to write - developers need to think about the BATs model and know names of the helper functions and helper variables.


pflynn-virtru:
bats are added

Issue #970, opened 2024-06-12
introduce `TRACE` level logs


strantalis:
Some logs are currently verbose in debug mode and should be moved to a `TRACE` level type log. A good example is in this issue #966.

PR #968, opened 2024-06-11
fix(core): reduce noise in casbin debug logging

jakedoublev:
Closes #966

PR #989, opened 2024-06-13
feat: adds new trace log level

strantalis:
This adds a new log level of `TRACE` that we can take advantage of for logs that we might find noisy on debug but might need to capture during a triaging scenerio. 

Resolves:
- #968 
- #970


jakedoublev (dismissed):
Looks like the linter caught something legitimate, but this lgtm otherwise

jakedoublev (approved):

Issue #644, opened 2024-04-23
Platform library should expose Entity Representation flattening-related functions


jakedoublev:
### Background
As a result of the [accepted ADR to employ flattening of an Entity Representation into known fields and values](https://github.com/opentdf/platform/issues/610), the platform needs a small library of functions to do flattening and comparison.

### Acceptance Criteria

This set of functions, consumable downstream in PEPs or tools like `otdfctl`:
1. Take arbitrary JSON and flatten it into an array of fields and values (see [existing proto](https://github.com/opentdf/platform/blob/main/service/policy/objects.proto#L217), and [ADR example](https://github.com/opentdf/platform/issues/610#:~:text=...will%20be%20flattened%20to%3A)). Every array value should appear twice in the output, once with the index wildcard as `*` and once with the index specified numerically. This gives admins the ability to entitle based on order in an array or presence without the platform driving that opinion.
2. Take an Entity Representation and a flattened set of fields and values and compare to find the matched field/value pairs.
3. Take an Entity Representation and a set of selected fields and provide the values for whichever are present, even if they differ

Unit tests for all of the above.

Issue #645, opened 2024-04-23
Auth Service should consume Entity Representation flattening instead of jq selector syntax when making access decisions


jakedoublev:
### Background
As a result of the [accepted ADR to employ flattening of an Entity Representation into known fields and values](https://github.com/opentdf/platform/issues/610), the platform needs to consume the [flattening functions](https://github.com/opentdf/platform/issues/644) instead of jq syntax when making an access decision.

### Acceptance Criteria

1. any jq evaluation is removed from Access PDP, OPA, and Auth Service
2. flattening functions are employed in their place to provide context to the entitlement of a subject

jakedoublev:
Blocked by https://github.com/opentdf/platform/issues/646 & https://github.com/opentdf/platform/issues/644

biscoe916:
@jakedoublev @strantalis @biscoe916 to discuss in arch sync tomorrow

PR #990, opened 2024-06-14
feat(authz): Add lib with entity representation flattening

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/644
breaking https://github.com/opentdf/platform/issues/645 into separate PR


service/internal/subjectmappingbuiltin/subject_mapping_builtin.go
```diff
@@ -187,7 +188,11 @@ ConditionEval:
 }
 
 func EvaluateCondition(condition *policy.Condition, entity map[string]any) (bool, error) {
-	mappedValues, err := ExecuteQuery(entity, condition.GetSubjectExternalSelectorValue())
+	flattenedEntity, err := flattening.Flatten(entity)
+	if err != nil {
+		return false, err
```
jakedoublev, line 193: Can we wrap these errors? 
```suggestion
		return false, fmt.Errorf("failure to EvaluateCondition in subject mapping builtin: %w", err)
```

jakedoublev (dismissed):

service/internal/subjectmappingbuiltin/subject_mapping_builtin.go
elizabethhealy: will do in second PR

lib/flattening/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/opentdf/platform/lib/flattening
 
-go 1.22.4
+go 1.21
```
jakedoublev, line 3: Was this downgrade intentional?

jakedoublev: Hmm, my diff view got out of sync for some reason. Disregard.

jakedoublev (approved):
Not sure just by looking at this if the pipeline is perfectly updated but these changes look good to me.

strantalis (approved):

PR #1001, opened 2024-06-18
chore: add friendly reminders if go mod tidy needs to be run

jrschumacher:


strantalis (dismissed):

dmihalcik-virtru (dismissed):
Suggestion: Prefer `go work edit` to shell scripts, as it is more portable and easier to read

dmihalcik-virtru (approved):

PR #1003, opened 2024-06-18
feat(policy): add unsafe service protos and unsafe service proto Go gencode

jakedoublev:
First PR related to #115


jrschumacher:
@jakedoublev I'm wondering if we want the client to return some value to verify they was to perform this destructive behavior. For instance if you want to delete you need to provide the id and the name.

jrschumacher:
Then we can just pass the user's input from CLI or web directly to the server. No need for user clients to add that layer.

jakedoublev:
> @jakedoublev I'm wondering if we want the client to return some value to verify they was to perform this destructive behavior. For instance if you want to delete you need to provide the id and the name.

That's a good idea @jrschumacher. I think the behavior that would guard against is a mistaken UUID being utilized to delete the policy object. Some other options would be:
1. requiring passage of parent/children (i.e. provide definition ids under the namespace when deleting a namespace, and values under a definition when deleting a definition)
2. adding another param like `acknowledgeRisk: boolean`

Of those, I think 1 is too burdensome and 2 is awkward and clunky. Your suggestion is the best to prevent a mistake and for consumer DX/UX downstream.

jakedoublev:
I think `reactivation` is okay with just an `id` because it is not cascading (like `delete`), which is most dangerous, and unsafe `update`s are likely also fine with just the `id` and the mutations.

The protos ensure gencode in each language and the HTTP path behind `/unsafe` will indicate the danger of these mutations.

jrschumacher (approved):

PR #1018, opened 2024-06-20
feat(policy): policy unsafe namespace RPCs wired up to database

jakedoublev:


service/policy/db/namespaces.go
```diff
@@ -267,11 +267,44 @@ func (c PolicyDBClient) UpdateNamespace(ctx context.Context, id string, r *names
...
 
-func deactivateNamespaceSQL(id string) (string, []interface{}, error) {
+func unsafeUpdateNamespaceSQL(id string, name string) (string, []interface{}, error) {
 	t := Tables.Namespaces
 	return db.NewStatementBuilder().
 		Update(t.Name()).
-		Set("active", false).
+		Set("name", name).
+		Where(sq.Eq{t.Field("id"): id}).
+		Suffix("RETURNING \"id\"").
```
jrschumacher, line 276: I think we should return the entire structure. That way someone can undo if it was accidental.

jakedoublev: I like that. This will be slightly different behavior than safe updates, then, but I think that's fine because it's a separate service altogether. We'll return the updated. See [newly pushed commit](https://github.com/opentdf/platform/pull/1018/commits/3c384d943d98c4672abd3cfb140f979217c80a39) to do so.

jrschumacher (approved):

PR #1038, opened 2024-06-26
feat(sdk): Support custom key splits

dmihalcik-virtru:
Implementation of https://github.com/opentdf/spec/pull/32

This is a proposal to allow customizing how a client shares key data across multiple KASes. With a split, you can copy the same share to multiple providers, allowing for robustness if a given KAS is unavailable - or if a decrypting user or application does not have authorization with that KAS.

### DIFF
--- a/sdk/kas_client.go
+++ b/sdk/kas_client.go
@@ -67,7 +67,7 @@ func newKASClient(dialOptions []grpc.DialOption, accessTokenSource auth.AccessTo
 }
 
 // there is no connection caching as of now
-func (k *KASClient) makeRewrapRequest(keyAccess KeyAccess, policy string) (*kas.RewrapResponse, error) {
+func (k *KASClient) makeRewrapRequest(ctx context.Context, keyAccess KeyAccess, policy string) (*kas.RewrapResponse, error) {
 	rewrapRequest, err := k.getRewrapRequest(keyAccess, policy)
 	if err != nil {
 		return nil, err
@@ -83,7 +83,6 @@ func (k *KASClient) makeRewrapRequest(keyAccess KeyAccess, policy string) (*kas.
 	}
 	defer conn.Close()
 
-	ctx := context.Background()
 	serviceClient := kas.NewAccessServiceClient(conn)
 
 	response, err := serviceClient.Rewrap(ctx, rewrapRequest)
@@ -94,8 +93,8 @@ func (k *KASClient) makeRewrapRequest(keyAccess KeyAccess, policy string) (*kas.
 	return response, nil
 }
 
-func (k *KASClient) unwrap(keyAccess KeyAccess, policy string) ([]byte, error) {
-	response, err := k.makeRewrapRequest(keyAccess, policy)
+func (k *KASClient) unwrap(ctx context.Context, keyAccess KeyAccess, policy string) ([]byte, error) {
+	response, err := k.makeRewrapRequest(ctx, keyAccess, policy)
 	if err != nil {
 		return nil, fmt.Errorf("error making request to kas: %w", err)
 	}
@@ -156,7 +155,7 @@ func (k *KASClient) getNanoTDFRewrapRequest(header string, kasURL string, pubKey
 	return &rewrapRequest, nil
 }
 
-func (k *KASClient) makeNanoTDFRewrapRequest(header string, kasURL string, pubKey string) (*kas.RewrapResponse, error) {
+func (k *KASClient) makeNanoTDFRewrapRequest(ctx context.Context, header string, kasURL string, pubKey string) (*kas.RewrapResponse, error) {
 	rewrapRequest, err := k.getNanoTDFRewrapRequest(header, kasURL, pubKey)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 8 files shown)


sdk/tdf.go
```diff
@@ -392,12 +442,21 @@ func (s SDK) LoadTDF(reader io.ReadSeeker) (*Reader, error) {
...
+	if r.payloadKey != nil {
+		return nil
+	}
+	return r.doPayloadKeyUnwrap(ctx)
+}
+
 // Read reads up to len(p) bytes into p. It returns the number of bytes
 // read (0 <= n <= len(p)) and any error encountered. It returns an
 // io.EOF error when the stream ends.
 func (r *Reader) Read(p []byte) (int, error) {
```
pflynn-virtru, line 457: I wonder if we should break the contract and pass in context

pflynn-virtru (approved):

sdk/tdf.go
dmihalcik-virtru: yeah, I kinda wish there was a 'ReadWithContext' interface. It is very common pattern for read to use network resources, either explicitly (e.g. using an HTTP request with a Range header, like we do for remote decrypt), or implicitly using something like NFS. I'd rather leave this as is for now and see how people use it to decide to change it

sdk/kas_client.go
```diff
@@ -287,26 +285,67 @@ func (k *KASClient) getRewrapRequest(keyAccess KeyAccess, policy string) (*kas.R
...
+	url, algorithm string
+}
+
+type timeStampedKASInfo struct {
+	KASInfo
+	time.Time
+}
+
+// Caches the most recent key info for a given KAS URL and algorithm
+type kasKeyCache struct {
```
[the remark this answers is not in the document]
dmihalcik-virtru: We probably do want the application to have *some* control over it, although yes it may make sense to restrict access in some fashion. Notably, in the test code I have to call 'clear' after each test case that involves rotation. So I might want to change this.

sdk/kas_client.go
```diff
@@ -287,26 +285,67 @@ func (k *KASClient) getRewrapRequest(keyAccess KeyAccess, policy string) (*kas.R
...
+}
+
+func (c *kasKeyCache) get(url, algorithm string) *KASInfo {
+	cacheKey := kasKeyRequest{url, algorithm}
+	now := time.Now()
+	cv, ok := c.c[cacheKey]
+	if !ok {
+		return nil
+	}
+	ago := now.Add(-1 * time.Hour)
```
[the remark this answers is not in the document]
dmihalcik-virtru: My thinking was we'd want to use the same key for each split that is writing to the same KAS. This method isn't sufficient to guarantee that, either. For that I'd need some kind of mark/release feature.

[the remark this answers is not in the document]
dmihalcik-virtru: I've filed a separate issue for this; https://github.com/opentdf/platform/issues/1107

sdk/kas_client.go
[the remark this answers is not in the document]
dmihalcik-virtru: See https://github.com/opentdf/platform/issues/1107

patmantru (approved):

sdk/nanotdf.go
```diff
@@ -749,6 +749,11 @@ func (s SDK) CreateNanoTDF(writer io.Writer, reader io.Reader, config NanoTDFCon
 
 // ReadNanoTDF - read the nano tdf and return the decrypted data from it
 func (s SDK) ReadNanoTDF(writer io.Writer, reader io.ReadSeeker) (uint32, error) {
+	return s.ReadNanoTDFContext(context.Background(), writer, reader)
+}
+
+// ReadNanoTDFContext - allows cancelling the reader
+func (s SDK) ReadNanoTDFContext(ctx context.Context, writer io.Writer, reader io.ReadSeeker) (uint32, error) {
```
strantalis, line 756: @dmihalcik-virtru  Are we planning on deprecating `ReadNanoTDF` or should we just break the method signature and add the context as the first argument.

sdk/tdf_config.go
```diff
@@ -118,6 +123,14 @@ func WithKasInformation(kasInfoList ...KASInfo) TDFOption {
 	}
 }
 
+func withSplitPlan(p ...splitStep) TDFOption {
```
strantalis, line 126: @dmihalcik-virtru  Is there a reason this is private?

sdk/nanotdf.go
dmihalcik-virtru: I was modeling this after the `slog.[Level]Context` [methods](https://pkg.go.dev/golang.org/x/exp/slog#hdr-Contexts)

dmihalcik-virtru: So no I was not planning to deprecate. We should encourage use of the Context variant wherever possible (e.g. within a context-having thread of execution)

sdk/tdf_config.go
dmihalcik-virtru: For now, I'd rather keep it private until we need to use it externally. This will be used by the public `WithAutoconfigure` option that is being added in https://github.com/opentdf/platform/pull/1051

strantalis (approved):

PR #1039, opened 2024-06-26
chore: Audit Logging Tests

pbacon-blaber:
- add tests to auidt events

### DIFF
--- a/service/internal/logger/audit/logger.go
+++ b/service/internal/logger/audit/logger.go
@@ -25,7 +25,7 @@ type Logger struct {
 
 // Used to support custom log levels showing up with custom labels as well
 // see https://betterstack.com/community/guides/logging/logging-in-go/#creating-custom-log-levels
-func AuditLevelReplaceAttr(_ []string, a slog.Attr) slog.Attr {
+func ReplaceAttrAuditLevel(_ []string, a slog.Attr) slog.Attr {
 	if a.Key != slog.LevelKey {
 		return a
 	}
--- a/service/internal/logger/audit/logger.go
+++ b/service/internal/logger/audit/logger.go
@@ -26,18 +26,19 @@ type Logger struct {
 // Used to support custom log levels showing up with custom labels as well
 // see https://betterstack.com/community/guides/logging/logging-in-go/#creating-custom-log-levels
 func CustomReplaceAttr(_ []string, a slog.Attr) slog.Attr {
-	if a.Key == slog.LevelKey {
-		level, ok := a.Value.Any().(slog.Level)
-		if !ok {
-			return a
-		}
-		levelLabel, exists := logLevelNames[level]
-		if !exists {
-			levelLabel = level.String()
-		}
-		a.Value = slog.StringValue(levelLabel)
+	if a.Key != slog.LevelKey {
+		return a
+	}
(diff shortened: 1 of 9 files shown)


service/internal/logger/audit/logger.go
```diff
@@ -15,10 +15,32 @@ const (
...
+			return a
+		}
+		levelLabel, exists := logLevelNames[level]
+		if !exists {
+			levelLabel = level.String()
+		}
+		a.Value = slog.StringValue(levelLabel)
+	}
+
+	return a
```
dmihalcik-virtru, line 41:
```suggestion
	if a.Key != slog.LevelKey {
		return a
	}
	level, ok := a.Value.Any().(slog.Level)
	if !ok {
		return a
	}

	levelLabel, exists := logLevelNames[level]
	if !exists {
		levelLabel = level.String()
	}
	a.Value = slog.StringValue(levelLabel)
	return a
```

service/internal/logger/audit/logger.go
```diff
@@ -15,10 +15,32 @@ const (
...
+	LevelAudit: LevelAuditStr,
+}
+
 type Logger struct {
 	logger *slog.Logger
 }
 
+// Used to support custom log levels showing up with custom labels as well
+// see https://betterstack.com/community/guides/logging/logging-in-go/#creating-custom-log-levels
+func CustomReplaceAttr(_ []string, a slog.Attr) slog.Attr {
```
dmihalcik-virtru, line 28: When is this function used? Should it be given a name that reflects how or why it will be used? This name feels like a placeholder

service/internal/logger/audit/utils.go
```diff
@@ -78,8 +78,14 @@ type ContextData struct {
 // Gets relevant audit data from the context object.
 func GetAuditDataFromContext(ctx context.Context) ContextData {
 	// Extract the request ID from context
-	requestID, requestIDOk := ctx.Value(sdkAudit.RequestIDContextKey).(uuid.UUID)
-	if !requestIDOk {
+
+	requestIDString, ok := ctx.Value(sdkAudit.RequestIDContextKey).(string)
+	if !ok {
+		requestIDString = ""
+	}
```
dmihalcik-virtru, line 85:
```suggestion
	requestIDString, _ := ctx.Value(sdkAudit.RequestIDContextKey).(string)
```

dmihalcik-virtru (approved):
Some minor nits. Also, you may want to use [testify](https://github.com/stretchr/testify) for future tests, which improves test readability and error output, while in my experience also making it a little easier to write more robust assertions.

service/internal/logger/audit/logger.go
pbacon-blaber: It's used as the `ReplaceAttr` function passed into the slog logger (see link in comment). I can update the name to say that it's like the audit level creator or something along those lines

pbacon-blaber:
> Some minor nits. Also, you may want to use [testify](https://github.com/stretchr/testify) for future tests, which improves test readability and error output, while in my experience also making it a little easier to write more robust assertions.

Nice I will use this in the future it seems much better, the comparison then `t.Error` seemed super verbose to me but I was just doing what was in other files I was looking at at the time.

dmihalcik-virtru (approved):

PR #1041, opened 2024-06-27
fix(core): update casbin policy to allow authorization service

strantalis:
This adds authorization service rpc to `org-admin` and `admin` role. It also adds kas `rewrap` to the unknown role.

### DIFF
--- a/service/internal/auth/casbin.go
+++ b/service/internal/auth/casbin.go
@@ -61,7 +61,7 @@ p,	role:org-admin,		/unsafe*,										            *,			allow
 p,	role:admin,		    policy.*,																*,			allow
 p,	role:admin,		    kasregistry.*,													*,			allow
 p,	role:admin,		    kas.AccessService/Rewrap, 			        *,			allow
-p,  role:org-admin,   authorization.*,                        *,      allow
+p,  role:admin,   authorization.*,                        *,      allow
 ## HTTP routes
 p,	role:admin,		/attributes*,																*,			allow
 p,	role:admin,		/namespaces*,																*,			allow
--- a/service/internal/auth/casbin.go
+++ b/service/internal/auth/casbin.go
@@ -45,35 +45,35 @@ var defaultPolicy = `
 ## gRPC routes
 p,	role:org-admin,		policy.*,																*,			allow
 p,	role:org-admin,		kasregistry.*,													*,			allow
-p,	role:org-admin,		kas.AccessService/Rewrap, 			            *,			allow
+p,	role:org-admin,		kas.AccessService/Rewrap, 			        *,			allow
+p,  role:org-admin,   authorization.*,                        *,      allow
 ## HTTP routes
 p,	role:org-admin,		/attributes*,														*,			allow
 p,	role:org-admin,		/namespaces*,														*,			allow
 p,	role:org-admin,		/subject-mappings*,											*,			allow
 p,	role:org-admin,		/resource-mappings*,										*,			allow
 p,	role:org-admin,		/key-access-servers*,										*,			allow
-p,	role:org-admin, 	/kas/v2/rewrap,						  		*,      allow
-p,	role:org-admin,		/unsafe*,										*,			allow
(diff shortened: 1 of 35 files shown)


elizabethhealy (approved):
does it need an entry for http authorization endpoints? idk if anyones using them

service/internal/auth/casbin.go
```diff
@@ -45,35 +45,35 @@ var defaultPolicy = `
...
 # Role: Admin
 ## gRPC routes
-p,	role:admin,		policy.*,																		*,			allow
-p,	role:admin,		kasregistry.*,															*,			allow
-p,	role:admin,		kas.AccessService/Info,					            *,			allow
-p,	role:admin,		kas.AccessService/Rewrap, 			            *,			allow
+p,	role:admin,		    policy.*,																*,			allow
+p,	role:admin,		    kasregistry.*,													*,			allow
+p,	role:admin,		    kas.AccessService/Rewrap, 			        *,			allow
+p,  role:org-admin,   authorization.*,                        *,      allow
```
jakedoublev, line 64:
```suggestion
p,  role:admin,   authorization.*,                        *,      allow
```

service/internal/auth/casbin.go
```diff
@@ -45,35 +45,35 @@ var defaultPolicy = `
...
 ## HTTP routes
 p,	role:admin,		/attributes*,																*,			allow
 p,	role:admin,		/namespaces*,																*,			allow
 p,	role:admin,		/subject-mappings*,													*,			allow
 p,	role:admin,		/resource-mappings*,												*,			allow
 p,	role:admin,		/key-access-servers*,												*,			allow
-p,	role:admin,		/kas/v2/rewrap,						  				*,      allow
+p,	role:admin,		/kas/v2/rewrap,						  				        *,      allow
```
jakedoublev, line 71: should we just make rewrap a public endpoint instead?

service/internal/auth/casbin.go
```diff
@@ -87,10 +87,13 @@ p,	role:standard,		/entityresolution/resolve,							write,  	allow
 # Public routes
 ## gRPC routes
 ## for ERS, right now we don't care about requester role, just that a valid jwt is provided when the OPA engine calls (enforced in the ERS itself, not casbin)
-p,	role:unknown,			entityresolution.EntityResolutionService.ResolveEntities,										write,		allow
+p,	role:unknown,			entityresolution.EntityResolutionService.ResolveEntities,					write,		allow
+p,	role:unknown,     kas.AccessService/Rewrap, 			                                  write,	  allow
 ## HTTP routes
 ## for ERS, right now we don't care about requester role, just that a valid jwt is provided when the OPA engine calls (enforced in the ERS itself, not casbin)
-p,	role:unknown,			/entityresolution/resolve,										write,		allow
+p,	role:unknown,			/entityresolution/resolve,							  write,		allow
```
jakedoublev, line 94: Not introduced by your PR, but is giving any unknown role access to `/entityresolution/resolve` actually a problem because any authenticated user could reach out to this endpoint to get the idP's entity attributes for any user? Maybe I'm reading it wrong but it seems like the malicious user would just need user names or client ids in their request? @elizabethhealy

elizabethhealy: ya i think that was a "for now" fix when setting it up, and because consumers will have to implement their own ERS we expect them to set up their own policy
we can change it to org admin or something, it should definitely be something more restricted since keycloak restricts who can query clients and users, so maybe a new role that specific to that functionality

service/internal/auth/casbin.go
strantalis: We still need the access token validated and some information extracted for KAS.

service/internal/auth/casbin.go
jakedoublev: We can track this within a separate issue. https://github.com/opentdf/platform/issues/1046

jakedoublev (dismissed):

jakedoublev (approved):

jakedoublev (approved):

PR #1045, opened 2024-06-27
chore(ci): Update rt-test to use a suite

dmihalcik-virtru:
This improves error output and makes it easier to use tools to run just the failing test


elizabethhealy (dismissed):
lgtm, you can prolly remove the roundtrip() func since it looks like its not being used anymore

elizabethhealy (approved):

Issue #1033, opened 2024-06-25
new capability to set a platform `mode`


strantalis:
Currently, when running the platform, we need to explicitly set whether a service is enabled. This works, but as we start to look at running a service like `KAS` independently, we need to consider how to manage this configuration more efficiently. It will become cumbersome if every service within the platform needs to be individually enabled or disabled.

Additionally, not every service requires the same tooling. For example, `KAS` does not need `opa`, yet `opa` is created regardless of the platform configuration. 

https://github.com/opentdf/platform/blob/c3828d088a3483b78079cd257b4237291cf7b6f0/service/pkg/server/start.go#L64-L69

First, we need to define what constitutes the core platform.

**Core Platform:**
  - Well-known (Platform Discovery) Endpoint
  - Policy

The two other services that aren't part of the core are:

**Authorization Service**  
**Key Access Service (KAS)**

If we agree that these are the main services, we can define the following modes:

- **all** - Runs all services as a monolith
- **core**
- **authorization**
- **kas**

As users potentially need to extend or build on top of OpenTDF they could also potentially need to define their own modes as well. This gives us the ability to provide that. So as they add new services under `services` that will essentially become a mode.

Current configuration looks something like this where each service is explicitly defined under the services block. 

https://github.com/opentdf/platform/blob/c3828d088a3483b78079cd257b4237291cf7b6f0/opentdf-dev.yaml#L11-L39

In the new approach, the `enabled` field would be removed as everything would be driven by the `mode`.

https://github.com/opentdf/platform/blob/main/service/pkg/serviceregistry/serviceregistry.go#L22

We wouldn't be required to set defaults for those services like we do today.

https://github.com/opentdf/platform/blob/main/service/internal/config/config.go#L27

#### Example All Mode

```yaml
mode: all
credentials:
  client_id: xxxxxx
  client_secret: xxxxx
db:
  <db config>
services:
  kas:
   <any kas config>
  core:
    <any core config>
  authorization:
    <an authorization config>
```

#### Example Core Mode

```yaml
mode: core # Not sure if we want to call this mode something else?
credentials:
  client_id: xxxxxx
  client_secret: xxxxx
db:
  <db config>
services:
  core:
    <any core config >
```

#### Example KAS Mode

```yaml
mode: kas
credentials:
  client_id: xxxxxx
  client_secret: xxxxx
remote_services:
  core: https://core.opentdf.io:9000
services:
  kas:
    <any core config>
```

[...]

Issue #751, opened 2024-05-06
config loaded debug statement logs secrets


pflynn-virtru:
Expected: no secrets from `opentdf.yaml` should be logged

Actual:
```
time=2024-05-06T14:22:09.179-04:00 level=DEBUG msg="config loaded" config="&{DB:{Host:localhost Port:5432 Database:opentdf User:postgres Password:changeme RunMigrations:true SSLMode:prefer Schema:opentdf VerifyConnection:true MigrationsFS:<nil>} OPA:{Path:./opentdf-opa.yaml Embedded:true Logger:<nil>} Server:{Auth:{Enabled:true PublicRoutes:[] AuthNConfig:{EnforceDPoP:false Issuer:http://localhost:8888/auth/realms/master Audience:http://localhost:8080 OIDCConfiguration:{Issuer: AuthorizationEndpoint: TokenEndpoint: JwksURI: ResponseTypesSupported:[] SubjectTypesSupported:[] IDTokenSigningAlgValuesSupported:[] RequireRequestURIRegistration:false} Policy:{Default: RoleClaim: RoleMap:map[] Csv: Model:} CacheRefresh:15m}} GRPC:{ReflectionEnabled:true} CryptoProvider:{Type:standard HSMConfig:{Enabled:false ModulePath: PIN: SlotID:0 SlotLabel: Keys:map[]} StandardConfig:{RSAKeys:map[123:{PrivateKeyPath:../kas-private.pem PublicKeyPath:../kas-cert.pem} 456:{PrivateKeyPath:../kas-private.pem PublicKeyPath:../kas-cert.pem}] ECKeys:map[123:{PrivateKeyPath:../kas-ec-private.pem PublicKeyPath:../kas-ec-cert.pem}]}} TLS:{Enabled:false Cert: Key:} CORS:{Enabled:true AllowedOrigins: AllowedMethods: AllowedHeaders: ExposedHeaders: AllowCredentials:false MaxAge:0} WellKnownConfigRegister:<nil> Port:8080 Host:} Logger:{Level:debug Output:stdout Type:text} Services:map[authorization:{Enabled:true Remote:{Endpoint:} ExtraProps:map[client:tdf-entity-resolution legacy:true realm:opentdf secret:secret url:http://localhost:8888]} kas:{Enabled:true Remote:{Endpoint:} ExtraProps:map[]} policy:{Enabled:true Remote:{Endpoint:} ExtraProps:map[]}]}"
```

[...]

PR #1048, opened 2024-06-28
fix(authz): move opa out of startup call

strantalis:
This pull request helps address https://github.com/opentdf/platform/issues/1033 and introducing the concept of a mode. 

Currently we are starting `opa` for every service. When in practice it is only consumed by the authorization service that is dependent on executing the `rego` policy.

In this update, I have removed OPA and instead directly eval the Rego policy, which can be either embedded or passed in via configuration. Until we have clear use cases for the additional management functionality that OPA provides, we can focus on just executing our entitlements policy.


service/authorization/authorization.go
```diff
@@ -57,28 +65,59 @@ func NewRegistration() serviceregistry.Registration {
...
+
 			if err := mapstructure.Decode(srp.Config.ExtraProps, &authZCfg); err != nil {
 				panic(fmt.Errorf("invalid auth svc cfg [%v] %w", srp.Config.ExtraProps, err))
 			}
 
-			config := clientcredentials.Config{ClientID: authZCfg.ClientID, ClientSecret: authZCfg.ClientSecret, TokenURL: authZCfg.TokenEndpoint}
-			slog.Debug("authorization service client config", slog.Any("config", config))
-			newTokenSource := oauth2.ReuseTokenSourceWithExpiry(nil, config.TokenSource(context.Background()), tokenExpiryDelay)
-			slog.Debug("authorization service token source created", slog.Any("token_source", newTokenSource))
+			slog.Debug("authorization service config", slog.Any("config", authZCfg))
```
jakedoublev, line 88: Did we end up deploying a global logging solution for masking config secrets in logs?

service/authorization/authorization.go
```diff
@@ -337,15 +373,8 @@ func (as *AuthorizationService) GetEntitlements(ctx context.Context, req *author
 		}
 		var attributeFqns []string
 		for _, attr := range listAttributeResp.GetAttributes() {
-			ns := attr.GetNamespace().GetName()
-			an := attr.GetName()
 			for _, val := range attr.GetValues() {
```
jakedoublev, line 376: can avoid this extra looping once https://github.com/opentdf/platform/pull/1050 is merged

service/authorization/authorization.go
strantalis: No it was only implemented for the root config object and I don't think its working last time I looked.

strantalis: https://github.com/opentdf/platform/issues/751

pflynn-virtru (dismissed):

service/authorization/authorization.go
```diff
@@ -23,31 +24,41 @@ import (
...
+	config      Config
 	logger      *logger.Logger
 	tokenSource *oauth2.TokenSource
+	eval        rego.PreparedEvalQuery
 }
 
 type Config struct {
 	// Entity Resolution Service URL
-	ERSURL string `mapstructure:"ersurl"`
+	ERSURL string `mapstructure:"ersurl" default:"http://localhost:8080/entityresolution/resolve"`
```
jrschumacher, line 46: I'm not sure we need to define the default here if the plan is to remove it.

jakedoublev (approved):

PR #1068, opened 2024-07-02
fix(ci): temporarily allow breaking proto changes

jakedoublev:
Unblocks #1067

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -1,4 +1,4 @@
-name: "Checks"
+name: 'Checks'
 
 env:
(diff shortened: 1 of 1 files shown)


elizabethhealy (approved):

.github/workflows/checks.yaml
```diff
@@ -251,7 +251,7 @@ jobs:
 
   ci:
     needs:
-      - buflint
+      # - buflint
```
dmihalcik-virtru, line 254: Instead of disabling the whole job, do something smaller in scale like ignoring unsafe.proto

jakedoublev:
Closing in favor of suggested change

PR #1075, opened 2024-07-02
feat(core): enable extending default policy

jrschumacher:
Resolves https://github.com/opentdf/platform/issues/1104


jakedoublev:
The most recent commit handles a case where a PEP tries to extend a default policy but the platform is running with custom policy. It should not panic like the below:

```shell
r.res,\tp.res) && keyMatch(r.act,\tp.act)\n}}" isDefaultModel=true isDefaultPolicy=false isDefaultRoleMap=false isDefaultRoleClaim=false
time=2024-07-03T11:29:55.761-07:00 level=DEBUG msg="authentication interceptor enabled"
time=2024-07-03T11:29:55.766-07:00 level=WARN msg="policies not added because they are not the default"
time=2024-07-03T11:29:55.766-07:00 level=ERROR msg="issue adding authz policies" error="could not add policies"
time=2024-07-03T11:29:55.766-07:00 level=INFO msg="shutting down http server"
time=2024-07-03T11:29:55.766-07:00 level=INFO msg="shutting down in process grpc server"
time=2024-07-03T11:29:55.766-07:00 level=INFO msg="shutdown complete"
time=2024-07-03T11:29:55.766-07:00 level=INFO msg="Stopping bundle loader." name=test plugin=bundle
time=2024-07-03T11:29:55.766-07:00 level=INFO msg="Stopping decision logger." plugin=decision_logs
panic: issue adding authz policies: could not add policies
```

The new behavior is below (log a warning with the policies that did not get added), but don't panic and just trust the custom policy configured in yaml:

```shell
time=2024-07-03T11:33:54.935-07:00 level=DEBUG msg="authentication interceptor enabled"
time=2024-07-03T11:33:54.939-07:00 level=INFO msg="creating crypto provider" type=standard
time=2024-07-03T11:33:54.940-07:00 level=WARN msg="default authz policy could not be not extended because policies are not the default" unextended_policies="[[admin admin write]]"
time=2024-07-03T11:33:54.940-07:00 level=INFO msg="registering services"
```

jakedoublev:
Closing in favor of #1105

Issue #1069, opened 2024-07-02
Need ability to join attributes on key access grants when needed


jrschumacher:
https://github.com/opentdf/platform/blob/4a3042625d0d08951bc36053ba053dc44fcffe99/service/policy/db/attributes.go#L415-L450

Issue #864, opened 2024-05-23
Explore using sqlc instead of squirrel for policy


jrschumacher:
Look at using ResourceMappings or SubjectMappings since they are smaller and less complex.

https://sqlc.dev/

jrschumacher:
@strantalis we've been using this in COP and it's been a great experience. I'll review with you next week.

jakedoublev:
Relevant blogpost: https://brandur.org/sqlc

PR #1077, opened 2024-07-02
feat(policy): add support for key access grants returned

jrschumacher:
Closes https://github.com/opentdf/platform/issues/1069

Returns KAS grants on lookup by FQN for `values` and `definitions` of attributes.

We have [an open issue](https://github.com/opentdf/platform/issues/864) to use `sqlc` for our query generation instead of squirrel, and this attributes select function is a prime candidate for a refactor splitting into separate SQL queries.


jakedoublev (approved):
I don't expect this to join KAS Grants for attribute definitions, and I don't think we have integration tests that it works for attribute values, but approving this to unblock and keep working it. 

https://github.com/opentdf/platform/blob/468aa26e357c6f091be84563e5a4369ccb86d823/service/policy/db/attributes.go#L105-L123

dmihalcik-virtru:
image

pflynn-virtru (approved):

ttschampel (approved):

PR #1091, opened 2024-07-03
fix(core): OPA consoleLogger only DEBUG

pflynn-virtru:
The consoleLogger is now initialized based on the logger level. If the logger level is equal to or less than opalog.Debug, the consoleLogger is set to nil, otherwise it is set to logger. This change allows for more flexible handling of logging within the application.


strantalis:
@pflynn-virtru I have this pr which moves us to execute rego directly versus leveraging opa. 

https://github.com/opentdf/platform/issues/1048

PR #1100, opened 2024-07-05
fix: fixed masking all sensitive keywords

mustyantsev:


service/internal/config/config.go
```diff
@@ -19,12 +20,13 @@ import (
...
-	OPA      opa.Config                               `yaml:"opa"`
-	Server   server.Config                            `yaml:"server"`
-	Logger   logger.Config                            `yaml:"logger"`
-	Services map[string]serviceregistry.ServiceConfig `yaml:"services" default:"{\"policy\": {\"enabled\": true}, \"health\": {\"enabled\": true}, \"authorization\": {\"enabled\": true}, \"wellknown\": {\"enabled\": true}, \"kas\": {\"enabled\": true}, \"entityresolution\": {\"enabled\": true}}"`
+	DevMode       bool                                     `mapstructure:"dev_mode"`
+	DB            db.Config                                `yaml:"db"`
+	OPA           opa.Config                               `yaml:"opa"`
+	Server        server.Config                            `yaml:"server"`
+	Logger        logger.Config                            `yaml:"logger"`
+	SensitiveKeys map[string]bool                          `yaml:"sensitiveKeys"`
```
dmihalcik-virtru, line 28: Can you make this a list instead of a dictionary?

```suggestion
	SensitiveKeys []string                          `yaml:"sensitiveKeys"`
```

mustyantsev: done

service/internal/config/config.go
```diff
@@ -89,7 +91,15 @@ func LoadConfig(key string, file string) (*Config, error) {
 }
 
 func (c *Config) LogValue() slog.Value {
-	redactedConfig := util.RedactSensitiveData(c)
+	var sensitiveKeys = c.SensitiveKeys
+
+	redactedInterface := util.RedactSensitiveData(c, sensitiveKeys)
+	redactedConfig, ok := redactedInterface.(*Config) // Direct type assertion
+	if !ok {
+		log.Println("Error asserting redacted configuration to *Config type")
```
dmihalcik-virtru, line 99: We prefer`slog` to `log`, but given you are already processing a logline at this point maybe you can just add the error to the output? maybe 

```
  values = append(values, slog.String("logError", "unable to redact config"))
```
or similar

dmihalcik-virtru (approved):
one optional request

jrschumacher:
Adding this to the config file feels like it's going to be a footgun. I can understand adding extra options but known values should either be struct tags or constants even if the hardcoded is hacky at first.

strantalis:
@mustyantsev Did we ever try https://github.com/m-mizutani/masq.

mustyantsev:
> @mustyantsev Did we ever try https://github.com/m-mizutani/masq.

yes, tried but did not succeed

mustyantsev:
> https://github.com/m-mizutani/masq

hello @jrschumacher , so how do you see it should be implemented?

Issue #1080, opened 2024-07-02
Add contains subject operator type


ttschampel:
Add support for a **contains** subject mapping operator type to support string contains functionality on the subject mapping external selected value.  

Reference - currently SubjectMappingOperatorEnum types

```
enum SubjectMappingOperatorEnum {
  SUBJECT_MAPPING_OPERATOR_ENUM_UNSPECIFIED = 0;
  SUBJECT_MAPPING_OPERATOR_ENUM_IN = 1;
  SUBJECT_MAPPING_OPERATOR_ENUM_NOT_IN = 2;
}

```

### Contains Example : 
The below would be a positive match for the condition & user

**The Condition:**
```
 {
     "subject_external_selector_value": "'.email'",
     "operator": "SUBJECT_MAPPING_OPERATOR_ENUM_CONTAINS",
     "subject_external_values" : ["@opentdf.io"]
   }
```
 **user**:
```
 {"email":"sample@opentdf.io"}
```

jrschumacher:
@ttschampel could we add

- contains
- begins with
- ends with

ttschampel:
I'm good with contains, starts with, ends with;  might as well add these all at once

cassandrabailey293:
we can split out "starts with" and "ends with" into separate issues so we can make sure we address this this week.

PR #1109, opened 2024-07-08
feat(core): CONTAINS SubjectMapping Operator

pflynn-virtru:
This change introduces an additional enum value in the SubjectMappingOperatorEnum in the Policy Objects protos. The new option "CONTAINS" extends the operator's ways to condition subject mappings. Corresponding updates are applied to documentation and swagger files to reflect this addition.

resolves https://github.com/opentdf/platform/issues/1080


service/policy/objects.proto
```diff
@@ -105,6 +105,7 @@ enum SubjectMappingOperatorEnum {
   SUBJECT_MAPPING_OPERATOR_ENUM_UNSPECIFIED = 0;
   SUBJECT_MAPPING_OPERATOR_ENUM_IN = 1;
   SUBJECT_MAPPING_OPERATOR_ENUM_NOT_IN = 2;
+  SUBJECT_MAPPING_OPERATOR_ENUM_CONTAINS = 3;
```
jakedoublev, line 108: Could we please add comments to the proto/gencode about how (if I read correctly), contains evaluates a string match against every value in the selected values? It's a little hard to decipher how it's different from "in" from just looking at the proto, but I think it's OR/ANY-evaluation with substring matches.

service/internal/subjectmappingbuiltin/subject_mapping_builtin_test.go
```diff
@@ -90,6 +90,31 @@ func Test_EvaluateConditionNOTINFalse(t *testing.T) {
 	assert.False(t, res)
 }
 
+// evaluate condition CONTAINS
+var containsCondition1 = policy.Condition{
+	SubjectExternalSelectorValue: ".attributes.testing[]",
+	Operator:                     policy.SubjectMappingOperatorEnum_SUBJECT_MAPPING_OPERATOR_ENUM_CONTAINS,
+	SubjectExternalValues:        []string{"option"},
```
jakedoublev, line 97: can we please add test cases for how multiple values in a list where only one matches are expected to be evaluated?

service/policy/objects.proto
jakedoublev: Alternatively we could just look at the first value and document that the contains (and future startsWith/endsWith) only look at one value within the selected values list. Someone could always move it up out of a single Condition into multiple Conditions in an OR ConditionGroup if they need more complex evaluation.

jakedoublev: And if we do want to only consider one value in a list for `contains/startsWith/endsWith`, we could validate the length of the `external_selector_values` on the create request in policy to enforce that API behavior.

jakedoublev (approved):

Issue #1063, opened 2024-07-01
Entity Resolution Service unknown user support


ttschampel:
There are valid user flows where an entitlement request can be made where the user has not been seen by the ERS backed IDP (in case of IDP mapping like LDAP+Sync) or is a member of a third party organization.   

A concrete use case would be an entity by email address.

The logic involving how unmatched entities are handled should be a function with in the ERS; as opposed to the core authorization service.

### Support for this in the sample Keycloak ERS:
Add configuration to Keycloak ERS to support something to describe unknown entity support yes/no and what entity types:
e.g.
```
unknownEntities:
  enabled: true
  entityTypes: [email]
```

Given this configuration Keycloak ERS would echo back the entity representation to the caller in the case of no match (for entity type = email).  This enables downstream subject mappings to still inform entitlements for these unknown users.

pflynn-virtru:
If the email is unknown to a keycloak instance, shouldn't we do a lookup to another identity service that does know?  
Unknown entities should not be giving equivalent entitlements as known entities.

ttschampel:
Agreed @pflynn-virtru , but this is up to the customer's ERS implementation. 

In most cases all entities should be validated as known and the ERS implementation + IDP will have federation support and/or consumers can customize entitlement rego.

this ticket is just for the sample Keycloak ERS to support this feature in a deployment that does not have this rich federation/IDP mapping in place.  Also, this feature can be disabled if not wanted.

cassandrabailey293:
here is a bug caused by this issue: https://github.com/opentdf/platform/issues/1036

cassandrabailey293:
Fixed by #1113

Issue #1055, opened 2024-06-28
`GetEntitlements` or `EntityResolution` unexpected entitlements 


strantalis:
I am seeing weird behavior when calling `GetEntitlements` . See the example code below. 

```go
package main

import (
	"context"
	"log"

	"github.com/opentdf/platform/protocol/go/authorization"
	"github.com/opentdf/platform/sdk"
)

func main() {

	platformEndpoint := "http://localhost:9002"

	// Create a new client
	client, err := sdk.New(
		platformEndpoint,
		sdk.WithInsecurePlaintextConn(), // This is for example
		sdk.WithClientCredentials("opentdf", "secret", nil),
	)

	if err != nil {
		log.Fatal(err)
	}

	// Get Entitlements

	entitlementReq := &authorization.GetEntitlementsRequest{
		Entities: []*authorization.Entity{
			{
				Id: "entity-1",
				// EntityType: &authorization.Entity_ClientId{
				// 	ClientId: "opentdf",
				// },
			},
		},
	}

	entitlements, err := client.Authorization.GetEntitlements(context.Background(), entitlementReq)
	if err != nil {
		log.Fatal(err)
	}

	log.Printf("Entitlements: %v", entitlements.GetEntitlements())

}
```

I am getting the output below which I would have expected it to be an empty set entitlement set.
```
2024/06/28 15:44:06 Entitlements: [entity_id:"entity-1"  attribute_value_fqns:"https://example.net/attr/attr1/value/value1"  attribute_value_fqns:"https://example.com/attr/attr1/value/value1"]
```

## Logs

```
time=2024-06-28T15:54:55.796-04:00 level=DEBUG msg="root claim found" claim=realm_access.roles claims="map[roles:[default-roles-opentdf offline_access uma_authorization]]"
time=2024-06-28T15:54:55.796-04:00 level=DEBUG msg="checking role" role=default-roles-opentdf map=standard
time=2024-06-28T15:54:55.796-04:00 level=DEBUG msg="checking role" role=default-roles-opentdf map=admin
time=2024-06-28T15:54:55.796-04:00 level=DEBUG msg="checking role" role=default-roles-opentdf map=org-admin
time=2024-06-28T15:54:55.796-04:00 level=DEBUG msg="checking role" role=offline_access map=standard
time=2024-06-28T15:54:55.796-04:00 level=DEBUG msg="checking role" role=offline_access map=admin
[...]

Issue #1036, opened 2024-06-25
Unexpected emails in access decisions returns a 500 error


BenGirone:
To reproduce, put an email in an access decision like `bob@unexpected.com`

I can provide some logs and specifics if it isn't immediately reproducible.

jrschumacher:
@BenGirone will you provide sample input and the output you get? Also, what is the behavior you are expecting?

BenGirone:
> @BenGirone will you provide sample input and the output you get? Also, what is the behavior you are expecting?

Cody was able to reproduce it again. simply put an email that the system won't expect in an entity and you should get an output like this

`ime=2024-07-08T11:19:35.395-05:00 level=ERROR msg="error from entity resolution: {\"body\": {\"code\": 2, \"details\": [], \"message\": \"code:5 message:\\\"resource retrieval failed\\\" entity:\\\"unexpected@example.com\\\"\"}, \"headers\": {\"content-length\": [\"110\"], \"content-type\": [\"application/json\"], \"date\": [\"Mon, 08 Jul 2024 16:19:35 GMT\"], \"vary\": [\"Origin\"]}, \"raw_body\": \"{\\\"code\\\":2,\\\"message\\\":\\\"code:5 message:\\\\\\\"resource retrieval failed\\\\\\\" entity:\\\\\\\"unexpected@example.com\\\\\\\"\\\",\\\"details\\\":[]}\", \"status\": \"500 Internal Server Error\", \"status_code\": 500}" namespace=sharepoint error="error from entity resolution: {\"body\": {\"code\": 2, \"details\": [], \"message\": \"code:5 message:\\\"resource retrieval failed\\\" entity:\\\"unexpected@example.com\\\"\"}, \"headers\": {\"content-length\": [\"110\"], \"content-type\": [\"application/json\"], \"date\": [\"Mon, 08 Jul 2024 16:19:35 GMT\"], \"vary\": [\"Origin\"]}, \"raw_body\": \"{\\\"code\\\":2,\\\"message\\\":\\\"code:5 message:\\\\\\\"resource retrieval failed\\\\\\\" entity:\\\\\\\"unexpected@example.com\\\\\\\"\\\",\\\"details\\\":[]}\", \"status\": \"500 Internal Server Error\", \"status_code\": 500}" extra="getEntitlements request failed"
[...]

PR #1113, opened 2024-07-09
feat(authz): Keycloak ERS ability to handle clients, users, and emails that dont exist

elizabethhealy:
should resolve all
https://github.com/opentdf/platform/issues/1063
https://github.com/opentdf/platform/issues/1036
https://github.com/opentdf/platform/issues/1055

### DIFF
--- a/opentdf-dev.yaml
+++ b/opentdf-dev.yaml
@@ -37,11 +37,10 @@ services:
     clientsecret: 'secret'
     realm: 'opentdf'
     legacykeycloak: true
-    unknown:
-      entities:
-        types:
-            email: true
-            username: true
+    inferid:
+      from:
+        email: true
+        username: true
 server:
   auth:
     enabled: true
--- a/opentdf-dev.yaml
+++ b/opentdf-dev.yaml
@@ -37,9 +37,11 @@ services:
     clientsecret: 'secret'
     realm: 'opentdf'
     legacykeycloak: true
-    unknownentities:
-      enabled: true
-      entitytypes: [emailAddress, userName, clientId]
+    unknown:
+      entities:
+        types:
+            email: true
+            username: true
 server:
   auth:
     enabled: true
--- a/opentdf-dev.yaml
+++ b/opentdf-dev.yaml
@@ -37,6 +37,9 @@ services:
     clientsecret: 'secret'
     realm: 'opentdf'
     legacykeycloak: true
+    unknownentities:
+      enabled: true
+      entitytypes: [emailAddress, userName, clientId]
 server:
   auth:
     enabled: true
--- a/service/entityresolution/keycloak/keycloak_entity_resolution.go
+++ b/service/entityresolution/keycloak/keycloak_entity_resolution.go
@@ -44,6 +44,7 @@ type InferredIdentityConfig struct {
 }
 
 type EntityImpliedFrom struct {
+	ClientID bool `json:"clientid,omitempty"`
 	Email    bool `json:"email,omitempty"`
 	Username bool `json:"username,omitempty"`
 }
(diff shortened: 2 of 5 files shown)


opentdf-dev.yaml
```diff
@@ -37,6 +37,9 @@ services:
     clientsecret: 'secret'
     realm: 'opentdf'
     legacykeycloak: true
+    unknownentities:
+      enabled: true
+      entitytypes: [emailAddress, userName, clientId]
```
dmihalcik-virtru, line 42: You probably don't need `enabled`, right? Also, I'd suggest layering it a bit so you get better legibility and an extension point in case we need to add more detail somewhere.

```suggestion
    unknowns:
      entities:
        types: [emailAddress, userName, clientId]
```

service/entityresolution/keycloak/keycloak_entity_resolution.go
```diff
@@ -28,12 +30,18 @@ const UsernameConditionalSelector = "client_id"
...
+	Realm               string              `json:"realm"`
+	ClientID            string              `json:"clientid"`
+	ClientSecret        string              `json:"clientsecret"`
+	LegacyKeycloak      bool                `json:"legacykeycloak" default:"false"`
+	SubGroups           bool                `json:"subgroups" default:"false"`
+	UnknownEntityConfig UnknownEntityConfig `json:"unknownentities" default:"false"`
+}
+
+type UnknownEntityConfig struct {
+	Enabled     bool     `json:"enabled"`
```
dmihalcik-virtru, line 43: Instead of a type here, you can have a method that checks if EntityTypes is empty

service/entityresolution/keycloak/keycloak_entity_resolution.go
```diff
@@ -28,12 +30,18 @@ const UsernameConditionalSelector = "client_id"
...
+	EntityTypes []string `json:"entitytypes"`
```
dmihalcik-virtru, line 44: I'd rather not have bad configs cause failures during runtime; instead maybe do a custom type here so we can validate this during parse?

https://sagikazarmark.hu/blog/decoding-custom-formats-with-viper/

dmihalcik-virtru:
Were you able to reproduce the panic mentioned in https://github.com/opentdf/platform/issues/1036?

strantalis:
@dmihalcik-virtru Our code doesn't panic. GetDecision doesn't handle something when an invalid email is sent. So the client gets a 500 error instead of a decision denied.

elizabethhealy:
> @dmihalcik-virtru Our code doesn't panic. GetDecision doesn't handle something when an invalid email is sent. So the client gets a 500 error instead of a decision denied. 

@dmihalcik-virtru @strantalis i changed it to a 404 not found

strantalis:
@elizabethhealy Do we need to update GetDecisions or GetEntitlements to handle these new responses?

service/entityresolution/keycloak/keycloak_entity_resolution.go
jrschumacher: In another repo we use https://github.com/go-playground/validator to validate the config at runtime and indicate where the error occurs. Not sure if that would be useful here.

elizabethhealy: can i break this into a separate issue? cause we have the same problem with the authorization config parsing

elizabethhealy: actually i see sean did something with the validator for the authz config in his opa pr, i'll just dupe that here

dmihalcik-virtru: I simplified the type so it no longer contains string comparisons at runtime

opentdf-dev.yaml
dmihalcik-virtru: I've changed this a bit.

dmihalcik-virtru: New format:

```yaml
    inferid:
      from:
        clientid: false
        email: true
        username: true
```

ttschampel (approved):
may want to add test cases to cover this new feature

PR #1121, opened 2024-07-09
feat(core): GetEntitlements with_comprehensive_hierarchy

### LINKED ISSUE #1054: Authorization GetEntitlements should consider hierarchical attribute values in entitlement values response
### Background

As the behavior of the platform is that any TDF can be encrypted with any attribute FQN without precondition or validation that the attributes exist on TDF creation, it is a PEP concern to drive UX around TDF creation.

If I am entitled to an `anyOf` attribute value, an `allOf` attribute value, and a specific `hierarchy` attribute value that is third of five values, the current Authorization Service behavior is to return the precise values I am entitled to, which would be `anyOf:exampleValue`, `allOf:exampleValue`, `hierarchy:third` even though I can access `hierarchy:fourth` and `hierarchy:fifth` due to the attribute definition rule.

Due to this, a PEP built on top of the platform must do their own work to fetch a user's entitlements, seek out any hierarchical attributes, and check a user's entitlement level against the potential values to drive a UX where they are able to TDF data at lower values than their entitlement level. This is intuitive: a GetDecisions request would allow the user to access `hierarchy:fourth` and `hierarchy:fifth` given a user's entitlement to `hierarchy:third`, so the PEP should drive UX where those are "accessible" values.

### Acceptance Criteria

1. `GetEntitlements` should support a new query parameter to return all attributes an entity would be entitled to
2. If that new parameter is provided, the response should consider both the attribute rule _and_ the resolved subject mappings instead of _only_ the subject mappings (i.e. `with_comprehensive` or some other selector parameter) so the hierarchy scenario is handled elegantly
3. The current API behavior without that parameter should remain the same
4. Tests

pflynn-virtru:
If a hierarchy attribute is entitled, lower entitlements are added as well when `with_comprehensive_hierarchy` is set in GetEntitlements

Hierarchy order is determined by creation order
Creation order is guaranteed unless unsafe operations used
```
  // WARNING!!
  // Unsafe reordering requires the full list of values in the new order they should be stored. Updating the order of values in a HIERARCHY-rule Attribute Definition
  // will retroactively alter access to existing TDFs containing those values. Replacing values on an attribute in place is not supported; values can be unsafely deleted
  // deleted, created, and unsafely re-ordered as necessary.
  ```

If the above is not followed, then you will have a very bad day

resolves #1054

### CODE REVIEW

File: service/authorization/authorization.go [hunk 1]
```diff
@@ -419,19 +419,56 @@ func (as *AuthorizationService) GetEntitlements(ctx context.Context, req *author
...
+						if isFollowing {
+							saa = append(saa, followingAttrVal.Fqn)
+						} else {
+							// if fqn match, then rest are added
+							// order is determined by creation order
+							// creation order is guaranteed unless unsafe operations used
+							isFollowing = followingAttrVal.GetFqn() == fqnStr
+						}
+					}
+					break
```
elizabethhealy, line 462: why break here?

service/authorization/authorization.go
```diff
@@ -419,19 +419,56 @@ func (as *AuthorizationService) GetEntitlements(ctx context.Context, req *author
...
+				if attrDef.GetRule() == policy.AttributeRuleTypeEnum_ATTRIBUTE_RULE_TYPE_ENUM_HIERARCHY {
+					// add the following fqn in the hierarchy
+					isFollowing := false
+					for _, followingAttrVal := range attrDef.GetValues() {
+						if isFollowing {
+							saa = append(saa, followingAttrVal.Fqn)
+						} else {
+							// if fqn match, then rest are added
+							// order is determined by creation order
+							// creation order is guaranteed unless unsafe operations used
```
elizabethhealy, line 458: ty for the comment w/ explanation here, i had these questions lol

elizabethhealy (approved):

strantalis (approved):

PR #1122, opened 2024-07-10
fix(sdk): Let WithDataAttributes take values

dmihalcik-virtru:


dmihalcik-virtru:
This is not backwards compatbile (breaks existing calls to WithDataAttributes that use string arrays)

Issue #983, opened 2024-06-13
remove tokenendpoint from authorization service config


strantalis:
We should be able to get the token endpoint from either the wellknown config or derive it from the platform issuer.

https://github.com/opentdf/platform/blob/d6c753ed6047afcb461655a6a344c6bd546b23b1/service/authorization/authorization.go#L78-L84

strantalis:
This is resolved.

Issue #1046, opened 2024-06-27
EntityResolution service should be more secured by RBAC casbin privileges


jakedoublev:
### Background 

The current RBAC Casbin `defaultPolicy` gives ERS route access to the `standard` role. This route should be of higher privilege to avoid an authenticated user querying another entity's idP value entity resolution.

https://github.com/opentdf/platform/blob/db4f06fdb9314747d9a95a5a09f974d86a1f0f29/service/internal/auth/casbin.go#L85

### Acceptance Criteria

1. ERS is protected by a higher privilege role
2. (optional) roles are enhanced to include a new intra-platform service or PEP role that is intended to be distinct from administrators or workflows

jrschumacher:
@elizabethhealy @jakedoublev is this done with the reduced complexity of our default authz policy?

elizabethhealy:
@jrschumacher yes it is restricted to admin with the current casbin policy https://github.com/opentdf/platform/blob/main/service/internal/auth/casbin_policy.csv

PR #1129, opened 2024-07-11
fix(authz): remove tokenendpoint configuration

strantalis:
This pr removes the need for a token endpoint in the authz configuration. It also gives the `org-admin` permission to call the entity resolution service. 

Resolves: #983

### DIFF
--- a/service/authorization/authorization.go
+++ b/service/authorization/authorization.go
@@ -22,6 +22,7 @@ import (
 	attr "github.com/opentdf/platform/protocol/go/policy/attributes"
 	otdf "github.com/opentdf/platform/sdk"
 	"github.com/opentdf/platform/service/internal/access"
+	"github.com/opentdf/platform/service/internal/auth"
 	"github.com/opentdf/platform/service/internal/entitlements"
 	"github.com/opentdf/platform/service/internal/logger"
 	"github.com/opentdf/platform/service/internal/logger/audit"
@@ -50,7 +51,7 @@ type Config struct {
 	// OAuth Client secret
 	ClientSecret string `mapstructure:"clientsecret" validate:"required"`
 	// OAuth token endpoint
-	TokenEndpoint string `mapstructure:"tokenendpoint" validate:"required,http_url"`
+	tokenEndpoint string
 	// Custom Rego Policy To Load
 	Rego CustomRego `mapstructure:"rego"`
 }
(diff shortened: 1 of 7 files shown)


service/authorization/authorization.go
```diff
@@ -109,6 +110,15 @@ func NewRegistration() serviceregistry.Registration {
 				}
 			}
 
+			// Get Token Endpoint
+			oidcCfg, err := auth.DiscoverOIDCConfiguration(context.Background(), srp.OTDF.AuthN.GetIssuer(), srp.Logger)
+			if err != nil {
+				logger.Error("failed to discover OIDC configuration", slog.String("error", err.Error()))
+				panic(fmt.Errorf("failed to discover OIDC configuration: %w", err))
```
dmihalcik-virtru, line 117: I'd rather do this during a readiness/startup check. The smell here is using context.Background. The problem is this will cause panics during startup for transient oidc connectivity issues, e.g. if you start keycloak and platform at the same time, this will almost certainly panic at first. Instead, it should just report as not started/ not ready.

service/authorization/authorization.go
```diff
@@ -50,7 +51,7 @@ type Config struct {
 	// OAuth Client secret
 	ClientSecret string `mapstructure:"clientsecret" validate:"required"`
 	// OAuth token endpoint
-	TokenEndpoint string `mapstructure:"tokenendpoint" validate:"required,http_url"`
+	tokenEndpoint string
```
dmihalcik-virtru, line 54: I'd rather this be left in to allow someone to use an OIDC endpoint that does not support discovery, which AFAIK is optional.

If you do want to require discovery, you still may want to leave this exposed and mark as `//Deprecated` and leave the previous behavior in place (i.e. only use service discovery if not set already)

service/authorization/authorization.go
dmihalcik-virtru: To do this move all the logic into another helper method that is then passed into the `srp.RegisterReadinessFunc()`

service/authorization/authorization.go
strantalis: If we need to support an idp that doesn't expose discovery then that functionality should be built into the authn configuration or how we initialize the sdk. 

This configuration was always temporary until we make an ers call outside of rego. Which we will need to do when we transition to the `MatchSubjectMappings` rpc.

dmihalcik-virtru (approved):

service/authorization/authorization.go
dmihalcik-virtru: I'll file a follow up about moving logic to readiness checks and improving error conditions of service handlers

service/authorization/authorization.go
dmihalcik-virtru: I guess? As someone who has struggled with keycloak's OIDC discovery configuration a lot, I do think you are making it difficult to work around the very common scenario where discovery is just wrong

jakedoublev:
Should we resolve this https://github.com/opentdf/platform/issues/1046 while we're updating ERS Casbin permissions?

service/authorization/authorization.go
dmihalcik-virtru: Filed as https://github.com/opentdf/platform/issues/1165

PR #1140, opened 2024-07-12
fix: CreateNanoTDF not using functional options

mustyantsev:


sdk/nanotdf_config.go
```diff
@@ -30,7 +30,7 @@ type NanoTDFConfig struct {
 type NanoTDFOption func(*NanoTDFConfig) error
 
 // NewNanoTDFConfig - Create a new instance of a nanoTDF config
-func (s SDK) NewNanoTDFConfig() (*NanoTDFConfig, error) {
```
dmihalcik-virtru, line 33: This is already used in DSP so you'll have to mark it as deprecated and not rename it

dmihalcik-virtru (changes requested):

sdk/nanotdf.go
```diff
@@ -657,7 +657,7 @@ func NewNanoTDFHeaderFromReader(reader io.Reader) (NanoTDFHeader, uint32, error)
 // ============================================================================================================
 
 // CreateNanoTDF - reads plain text from the given reader and saves it to the writer, subject to the given options
-func (s SDK) CreateNanoTDF(writer io.Writer, reader io.Reader, config NanoTDFConfig) (uint32, error) {
+func (s SDK) CreateNanoTDF(writer io.Writer, reader io.Reader, opts ...NanoTDFOption) (uint32, error) {
```
dmihalcik-virtru, line 660: Again, this will break existing code. Unfortunately:

> There is, in fact, no backward-compatible change you can make to a function’s signature.
   - [go/module compatibility](https://go.dev/blog/module-compatibility)
```suggestion
func (s SDK) CreateNanoTDFOpts(writer io.Writer, reader io.Reader, opts ...NanoTDFOption) (uint32, error) {
	config, err := s.newNanoTDFConfig(opts...)
	if err != nil {
		return 0, fmt.Errorf("NewTDFConfig failed: %w", err)
	}
	return s.CreateNanoTDF(writer, reader, config)
}

// Deprecated
func (s SDK) CreateNanoTDF(writer io.Writer, reader io.Reader, config NanoTDFConfig) (uint32, error) {
```

sdk/nanotdf_config.go
```diff
@@ -56,6 +54,29 @@ func (s SDK) NewNanoTDFConfig() (*NanoTDFConfig, error) {
 	return c, nil
 }
 
+// NewNanoTDFConfig - Create a new instance of a nanoTDF config
+// Deprecated: Use newNanoTDFConfig instead
```
dmihalcik-virtru, line 58:
```suggestion
// Deprecated: Use NanoTDFOptions with CreateNanoTDFOpts
```

dmihalcik-virtru (dismissed):

patmantru (approved):

PR #1143, opened 2024-07-12
fix(examples): let attributes example work

dmihalcik-virtru:
- adds tokenEndpoint and creds params to root example command
- Use the admin creds in a test instead of the 'standard' creds
- namespaces must be valid dns names

### DIFF
--- a/test/service-start.bats
+++ b/test/service-start.bats
@@ -73,3 +73,10 @@
   echo "$output"
   [[ $output = *NotFound* ]]
 }
+
+@test "gRPC: attributes example" {
+  run go run ./examples --creds opentdf:secret attributes
+  echo "$output"
+  [[ $output = *listing namespaces* ]]
+  [ $status = 0 ]
+}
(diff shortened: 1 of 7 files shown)


test/service-start.bats
```diff
@@ -73,3 +73,10 @@
   echo "$output"
   [[ $output = *NotFound* ]]
 }
+
+@test "gRPC: attributes example" {
+  run go run ./examples --creds opentdf:secret attributes
+  echo "$output"
+  [ $output = *listing namespaces* ]
+  [ $status = 0 ]
+}
```
elizabethhealy, line 82: nit dont have to address: maybe put it in a new bat file, since this one just seems to be testing whether the services are up, and the other is round trip specific, maybe a new one for attribute/policy creation, or it could go in the other one and we just rename the bat idc

elizabethhealy (dismissed):

pflynn-virtru (approved):

elizabethhealy (approved):

PR #1150, opened 2024-07-14
fix(policy): move policy sql logs to trace level to reduce noise

jakedoublev:


service/pkg/server/services.go
```diff
@@ -98,7 +98,7 @@ func startService(
 		var err error
 
 		logger.Info("creating database client", slog.String("namespace", s.Namespace))
-		d, err = db.New(ctx, cfg.DB,
+		d, err = db.New(ctx, cfg.DB, cfg.Logger,
```
strantalis, line 101: Should we provide additional context with this logger? Maybe we need to move this up https://github.com/opentdf/platform/pull/1150/files#diff-6fb82367e0e4c4e917e2e0fad446315e7125e98a9e4bffea82c039c51214225fR155

jakedoublev: This logger is for a db client instance, which means it'll be associated with a database schema we could use for log context (see [68863af](https://github.com/opentdf/platform/pull/1150/commits/68863af8912fe8e1c0ebcc862876a582ca5aa7e3)), but some consumers of db clients are not under a specific service namespace like migration CLI commands, migrations on startup when configured, and provisioners (integration test policy fixtures and dev fixtures). I'd rather not have to create a logger just to pass into a db client instantiation for those cases. What do you think about associating with the schema?

strantalis: That works for me and we could always evolve it in the future.

strantalis (approved):

PR #1160, opened 2024-07-15
fix(sdk): Remove case sensitivity of attr values

dmihalcik-virtru:
- Lets autoconfigure group attributes by case insenstiive name
- This still may have bugs in case sensitive values
- And also it is case sensitive by KAS url, resulting in unnecessary splits to the same server

### DIFF
--- a/sdk/internal/autoconfigure/granter.go
+++ b/sdk/internal/autoconfigure/granter.go
@@ -33,33 +33,40 @@ type SplitStep struct {
 }
 
 // Utility type to represent an FQN for an attribute.
-type AttributeNameFQN string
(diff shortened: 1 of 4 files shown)


jrschumacher:
Where is `NewAttributeNameFQN` being used? When I search the repo I only find the internal package, but no usage of it.

dmihalcik-virtru:
I'm exploring different methods of validating attribute FQNs and trying to hoist that type information (validated FQN) into the code instead of just passing around strings. These may be useful downstream for people who don't want proto-generated objects that might not be hydrated and are pretty heavy.

jrschumacher (approved):

PR #1163, opened 2024-07-16
feat(examples): adds attribute `ls,add,rm` cmds

dmihalcik-virtru:
- `example attribute ...`
  - `ls [-N http://name.space]`: List all attributes and values, or all in the requested namespace
   - `add -a https://name.space/attr/attrname -v "v1 v2 v3" [-r anyof]` Defines and attribute
   - `rm -a https://a.b/attr/n [-v "just the values to remove"]`

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -200,7 +200,7 @@ jobs:
       - run: test/policy-service.bats
       - name: create roundtrip test data and run tests
         run: go test ./service/rttests -v
-      - name: enable staic entitlements rego policy
+      - name: enable static entitlements rego policy
         run: yq eval '.services.authorization.rego.path = "./test/rego/static-entitlements.rego"' -i opentdf.yaml
       - run: sleep 30
       - name: validate static rego policy
--- a/examples/README.md
+++ b/examples/README.md
@@ -12,7 +12,7 @@ go build
 ```
 
 ```shell
-./examples attributes
+./examples attributes ls
 ```
 
 Examples:
--- a/examples/cmd/attributes.go
+++ b/examples/cmd/attributes.go
@@ -22,6 +22,7 @@ var (
 	ns         string
 	attr       string
 	kas        string
+	longformat bool
 	rule       string
 	values     string
 	unsafeBool bool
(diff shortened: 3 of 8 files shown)


jrschumacher:
@dmihalcik-virtru this seems like its a duplication of opentdf/otdfctl which is powering the opentdf CLI administration as well as downstream implementations. Is there a reason why we are rebuilding the wheel?

dmihalcik-virtru:
> @dmihalcik-virtru this seems like its a duplication of https://github.com/opentdf/otdfctl which is powering the opentdf CLI administration as well as downstream implementations. Is there a reason why we are rebuilding the wheel?
> 
> Concern is that we'll have to maintain this example if people start using it to manage their platforms rather than pointing to the CLI.

This is intended as example code and for use with integration tests. We do not publish releases of `examples` and do not document it as such. `otdfctl` is intended to be production quality, and likely not incorporate such features as are present here that are experimental or dangerous

dmihalcik-virtru:
As to specific motivations for doing this now, we need integration tests here that bring up multiple copies of the core services to validate splitting and sharing key data across different servers, validating failure modes when attribute defintiions disagree, and more. otdfctl is difficult to use with unreleased features, as it is hosted in a different repo. Additionally, it isn't clear that we want to expose *all* features through the CLI before vetting them as much as possible.

pflynn-virtru (approved):

strantalis:
@dmihalcik-virtru Is otdfctl missing features that a consumer would take advantage of?

jrschumacher:
> Additionally, it isn't clear that we want to expose all features through the CLI before vetting them as much as possible.

One nice feature of cobra is that we can use the `hidden` flag to prevent showing it in the help menu. This would allow us to experiment without supporting it. 

Also, I'm fine using build tags so we can build experimental versions of the CLI and stable versions. This is especially true since we tend to put the CLI cobra instructions in separate files.

dmihalcik-virtru:
> One nice feature of cobra is that we can use the `hidden` flag to prevent showing it in the help menu. This would allow us to experiment without supporting it. 

We may want to do this soon for jwk set management, or other emerging features. However, without moving `otdfctl` repo into this one, it will be difficult to do it on a branch; it would still either require coordinating work there with a branch from a PR here, or something more complex where we check out a branch from there and add it to the workspace after checkout (or look into a submodule)

PR #1171, opened 2024-07-17
chore: Auto Approve Version Bumps

pbacon-blaber:
Auto approves version bump PRs that satisfy the following conditions

- author is `opentdf-automation`
- PR title starts with `fix(core): Autobump`

I believe this will have the PR be approved by the GH actions bot. WE may need to add that bot as a codeowner to the mod files or maybe generate a token for a dedicated bot separate from opentdf-automation?

### DIFF
--- a/.github/workflows/auto-approve-version-bumps.yaml
+++ b/.github/workflows/auto-approve-version-bumps.yaml
@@ -1,0 +1,14 @@
+name: Auto Approve Version Bumps
+
+on: pull_request_target
+
+jobs:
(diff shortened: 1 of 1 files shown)


dmihalcik-virtru (approved):
Should/could we add some more assertions to make sure this a go.mod/go.sum only change?

pbacon-blaber:
> Should/could we add some more assertions to make sure this a go.mod/go.sum only change?

@dmihalcik-virtru: https://github.com/opentdf/platform/pull/1174

PR #1175, opened 2024-07-18
feat(core): Adds authn time skew config

dmihalcik-virtru:
- Adds acceptable skew control to oidc configuration
  - for idp provided tokens, use `skew` (default 1 minute)
  - for client provided tokens (dpop tokens), use `dpopskew` (default 1 hour)
- This should reduce flakiness on some tests

### DIFF
--- a/service/internal/auth/authn.go
+++ b/service/internal/auth/authn.go
@@ -222,7 +222,6 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 			m: r.Method,
 		}, r.Header["Dpop"])
 		if err != nil {
-			a.logger.WarnContext(r.Context(), "failed to validate token", slog.String("error", err.Error()))
 			http.Error(w, "unauthenticated", http.StatusUnauthorized)
 			return
 		}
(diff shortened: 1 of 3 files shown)


strantalis (approved):

service/internal/auth/authn.go
```diff
@@ -415,7 +417,7 @@ func GetRawAccessTokenFromContext(ctx context.Context, l *logger.Logger) string
 	return ""
 }
 
-func validateDPoP(accessToken jwt.Token, acessTokenRaw string, dpopInfo receiverInfo, headers []string, logger *logger.Logger) (jwk.Key, error) {
+func (a Authentication) validateDPoP(accessToken jwt.Token, acessTokenRaw string, dpopInfo receiverInfo, headers []string) (jwk.Key, error) {
```
elizabethhealy, line 420: nit: sp -- "acessTokenRaw" -> "accessTokenRaw" but looks like this is separate from your changes so feel free to ignore for now

elizabethhealy (approved):

service/internal/auth/authn.go
dmihalcik-virtru: I think there has been a move to add a spellchecker to our linter, but we haven't turned it on

PR #1219, opened 2024-07-26
feat(core): MIC-934 Moves logger out of internal folder

danielRicaud:


service/logger/logger.go
```diff
@@ -7,7 +7,7 @@ import (
 	"log/slog"
 	"os"
 
-	"github.com/opentdf/platform/service/internal/logger/audit"
+	"github.com/opentdf/platform/service/logger/audit"
```
dmihalcik-virtru, line 10: This doesn't just make the logger public - it also makes the audit package public, and exposes the internal structure of the `logger.Logger` type. And the audit package seems so tightly coupled with this that maybe they should be a single package with one object? Not sure, maybe we need the ability to extend or have mulitple audit objects?

Options I see:

1. Consolidate on just the audit name and move everything into that
2. replace the `Audit` field in logger with a (lowercase) `audit` and add an `Audit()` getter method or audit proxy methods to hide the internals of this type
3. Don't move everything out of internal. Instead create this as a new package and have `internal/logger/Logger` implement an interface defined here.

Who designed/is using the Audit stuff? I'll see if I can get them to review.

pbacon-blaber: I don't necessarily think that the internals of the Audit stuff is sensitive as I believe the audit events we're publishing are public?

That being said I think any of these would be fine. I had originally made all the audit stuff just part of the logger package but made it a "sub package" as it got more and more complex. I think it being separate also makes it easier to understand / add and edit existing audit events as well.

service/pkg/server/start.go
```diff
@@ -11,18 +11,18 @@ import (
...
 
 const devModeMessage = `
 ██████╗ ███████╗██╗   ██╗███████╗██╗      ██████╗ ██████╗ ███╗   ███╗███████╗███╗   ██╗████████╗    ███╗   ███╗ ██████╗ ██████╗ ███████╗
 ██╔══██╗██╔════╝██║   ██║██╔════╝██║     ██╔═══██╗██╔══██╗████╗ ████║██╔════╝████╗  ██║╚══██╔══╝    ████╗ ████║██╔═══██╗██╔══██╗██╔════╝
-██║  ██║█████╗  ██║   ██║█████╗  ██║     ██║   ██║██████╔╝██╔████╔██║█████╗  ██╔██╗ ██║   ██║       ██╔████╔██║██║   ██║██║  ██║█████╗  
-██║  ██║██╔══╝  ╚██╗ ██╔╝██╔══╝  ██║     ██║   ██║██╔═══╝ ██║╚██╔╝██║██╔══╝  ██║╚██╗██║   ██║       ██║╚██╔╝██║██║   ██║██║  ██║██╔══╝  
+██║  ██║█████╗  ██║   ██║█████╗  ██║     ██║   ██║██████╔╝██╔████╔██║█████╗  ██╔██╗ ██║   ██║       ██╔████╔██║██║   ██║██║  ██║█████╗
+██║  ██║██╔══╝  ╚██╗ ██╔╝██╔══╝  ██║     ██║   ██║██╔═══╝ ██║╚██╔╝██║██╔══╝  ██║╚██╗██║   ██║       ██║╚██╔╝██║██║   ██║██║  ██║██╔══╝
 ██████╔╝███████╗ ╚████╔╝ ███████╗███████╗╚██████╔╝██║     ██║ ╚═╝ ██║███████╗██║ ╚████║   ██║       ██║ ╚═╝ ██║╚██████╔╝██████╔╝███████╗
-╚═════╝ ╚══════╝  ╚═══╝  ╚══════╝╚══════╝ ╚═════╝ ╚═╝     ╚═╝     ╚═╝╚══════╝╚═╝  ╚═══╝   ╚═╝       ╚═╝     ╚═╝ ╚═════╝ ╚═════╝ ╚══════╝
```
pbacon-blaber, line 25: also unusure if this whitespace was meant to be left inhere to format this block message or what

danielRicaud: Thanks @pbacon-blaber I'll restore the whitespace just to be safe!

service/logger/logger.go
jrschumacher: There will be a need to extend audit in DSP if we want a uniform audit solution. How this is done is open to discussion.

strantalis: I think moving audit out is fine. What we need is a generic audit method that peps can call and pass the following struct.

https://github.com/opentdf/platform/blob/9c1053f154bb2b8b9d2170b677f84729f70ab64c/service/internal/logger/audit/utils.go#L18-L29

danielRicaud: Thank you all, for now should we leave audit exposed and follow back on adding that generic audit method in a follow-up task?

dmihalcik-virtru (approved):

jrschumacher (approved):

PR #1230, opened 2024-07-31
feat(sdk): Allow for payload key retrieval.

### LINKED ISSUE #1310: Add UnsafePayloadKey function
- Unsafe Payload key function will allow developers to retrieve the payload key after doing an unwrap. The idea is that this will be useful for retrieving the key when Virtru is operating as a EKM, where the encryption and decryption is done by a 3rd party. 
- The protection is that an NPE must first have access to retrieve the key. This is no different than a user directly calling KAS /rewrap and performing the unwrap themselves.

c-r33d:
resolves #1310 

- Add the ability to retrieve the DEK using the SDK. 

>[!NOTE]
> If the NPE already has access to do the unwrap, then add a method to allow NPEs to retrieve the key.
> This is helpful for use cases where the encryption/decryption is not done by Virtru, but by a 3rd party.
> In such as case, Virtru is acting as an external key manager, and the key needs to be provided to the
> party doing the enc/decr.


sdk/tdf.go
```diff
@@ -718,6 +718,26 @@ func (r *Reader) DataAttributes() ([]string, error) {
...
+/*
+*
+Retrieve the payload key, either from performing an unwrap or from a previous unwrap,
+and write it to a user buffer.
+
+OUTPUTS:
+  - []byte - Byte array containing the DEK.
+  - error - If an error occurred while processing
+*/
+func (r *Reader) PayloadKey() ([]byte, error) {
```
jrschumacher, line 730:
```suggestion
/*
*WARNING:* Using this function is unsafe since KAS will no longer be able to prevent access to the key.

Retrieve the payload key, either from performing an unwrap or from a previous unwrap,
and write it to a user buffer.

OUTPUTS:
  - []byte - Byte array containing the DEK.
  - error - If an error occurred while processing
*/
func (r *Reader) UnsafePayloadKeyRetrieval() ([]byte, error) {
```

c-r33d: Why would it be unsafe? KAS has to allow access to the key before this function can be used.

c-r33d: Agreed, this should be properly marked. I have committed your suggestions.

strantalis (approved):

PR #1245, opened 2024-08-01
feat(core): ability to run a set of isolated services

strantalis:
This pr resolves https://github.com/opentdf/platform/issues/1033

It introduces two new config values `mode` and `sdk_config` at the root level. It also removes the `enabled` field on a service. 

Mode now takes a list which could be `all`, `core` or a specific namespaced service for example `kas`. You can also combine modes like `core,kas`.

When running in something other than `core` or `all` you are required to specify a `sdk_config` object. This is because the core services will be remote now and not over the internal IPC server.


service/pkg/server/services.go
```diff
@@ -19,162 +22,138 @@ import (
...
-				// Only report on started services
-				if !r.Started {
-					continue
-				}
-				logger.Info("stopping service", slog.String("namespace", ns), slog.String("service", r.ServiceDesc.ServiceName))
-				if r.Close != nil {
-					r.Close()
-				}
-			}
+func registerCoreServices(reg serviceregistry.Registry, mode []string) ([]string, error) {
```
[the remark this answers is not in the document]
strantalis: We could probably filter out the mode if it exists.  The idea is mode could be defined as `mode: core,pep1,pep2,...`

service/internal/config/config.go
```diff
@@ -18,11 +18,21 @@ import (
...
+	DB      db.Config     `mapstructure:"db"`
+	Server  server.Config `mapstructure:"server"`
+	Logger  logger.Config `mapstructure:"logger"`
+	// Defines which services to run
+	Mode      []string                                 `mapsctructure:"mode"`
+	SDKConfig SDKConfig                                `mapstructure:"sdk_config"`
+	Services  map[string]serviceregistry.ServiceConfig `mapstructure:"services"`
+}
+
+type SDKConfig struct {
```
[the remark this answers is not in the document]
strantalis: My thought is services running together in the monolith will share the same client. If they truly need that level of separate isolation they could then split off and run the service on its own with the new `mode` pattern.

service/pkg/db/db_migration.go
```diff
@@ -86,7 +86,7 @@ func (c *Client) RunMigrations(ctx context.Context, migrations *embed.FS) (int,
 			applied++
 		}
 	}
-
+	c.ranMigrations = true
```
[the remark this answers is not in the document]
strantalis: Migrations are tracked in the database and also run within a transaction according the goose docs. So if the transaction never was committed it should rollback. 

https://github.com/pressly/goose?tab=readme-ov-file#sql-migrations

> By default, all migrations are run within a transaction. Some statements like CREATE DATABASE, however, cannot be run within a transaction. You may optionally add -- +goose NO TRANSACTION to the top of your migration file in order to skip transactions within that specific migration file. Both Up and Down migrations within this file will be run without transactions.

service/pkg/server/options.go
```diff
@@ -18,37 +24,69 @@ func WithConfigName(name string) StartOptions {
...
 		return c
 	}
 }
 
+// WithAuthZDefaultPolicyExtension option allows for extending the default casbin poliy
+// Example:
+//
+//	opentdf.WithAuthZDefaultPolicyExtension([][]string{
+//				buildEndpointAuthzPolicy("org-admin", "pep*", "*", true),
+//				buildEndpointAuthzPolicy("standard", "shared.*", "read", true),
```
jakedoublev, line 66:
```suggestion
//				{"p","role:org-admin", "pep*", "*", true),
```

pflynn-virtru:
suite

service/pkg/server/services.go
```diff
@@ -19,162 +22,185 @@ import (
 	wellknown "github.com/opentdf/platform/service/wellknownconfiguration"
 )
```
jrschumacher, line 24:
```suggestion

const MODE_ALL = "all"
const MODE_ESSENTIAL = "esssential"
const MODE_CORE = "core"
const SERVICE_KAS = "kas"
const SERVICE_POLICY = "policy"
const SERVICE_AUTHORIZATION = "authorization"
const SERVICE_WELLKNOWN = "wellknown"

```

jrschumacher: We should use consts rather than strings to ensure no typos.

service/pkg/server/services.go
```diff
@@ -19,162 +22,185 @@ import (
...
+	migrationsDisabled := (required && !client.MigrationsEnabled())
+
+	reason := "undetermined"
+	switch {
+	case requiredAlreadyRan:
+		reason = "required migrations already ran"
+	case noDBRequired:
+		reason = "service does not require a database"
+	case migrationsDisabled:
+		reason = "migrations are disabled"
```
jrschumacher, line 203: Should we capture a failed migration state?

service/pkg/serviceregistry/serviceregistry.go
```diff
@@ -88,35 +79,129 @@ type DBRegister struct {
...
+
+// RegisterHTTPServer registers an HTTP server with the service.
+// It takes a context, a ServeMux, and an implementation function as parameters.
+// If the service did not register a handler, it returns an error.
+func (s *Service) RegisterHTTPServer(ctx context.Context, mux *runtime.ServeMux) error {
+	if s.handleFunc == nil {
+		return fmt.Errorf("service did not register a handler")
+	}
+	return s.handleFunc(ctx, mux, s.impl)
+}
```
jrschumacher, line 132: I think we should deprecate this. If we need we should require using gRPC and api gateway or connectrpc

jrschumacher: If we need it I'd rather make it difficult to use so that we encourage protos since they are selfdocumenting.

service/pkg/server/services.go
strantalis: This is what we had existing. I just moved stuff around.

jrschumacher:
I think we should describe what we expect to happen.... and on this note maybe we need to add config versions so we can detect changes in the future? This won't work for config maps.

1. Without Mode the platform should fail to start and make note of the recent changes
2. If Services still have Enabled we should ignore unless validation exists

There are libs that validate yaml files and refuse to function if they don't meet the schema. For instance, WebPack and Rollup in JS land. Would this be critical to add?

jrschumacher:
> Is the migration stuff part of this PR or could it be moved to its own?

@dmihalcik-virtru this refactor is really focused on the serviceregistry since the modes impacts the way that was implemented. 

Sean brought me in during his refactor (pair programming) to check with me on the original implementation. I agree with this refactor since it's more clear. Also with respect to OpenTDF Policy is the only service that uses a DBClient, and we have extensive integration tests that are passing. 

If I missed a specific section you're concerned about, would you point that out to me?

strantalis:
@jrschumacher @dmihalcik-virtru The default `mode` is `all`. In opentdf I don't think we ever disable any services that come out of the box. So for now its an upgrade documentation exercise for our internal product. I haven't even begun to think about versioning the config and what complexities that brings versus just have really detailed upgrade notes.

jrschumacher:
I agree. Default should be monolith, and we can state that this is not a recommendation for a secure production environment. Additionally, I don't think OpenTDF as a group should be offering production advice, since that typically involves liability and insurance.

PR #1248, opened 2024-08-02
fix(main): adds viper decoder config option to respect yaml tags

0xArch3r:


jrschumacher:
@0xArch3r thanks for the PR. We'll look at this in the next couple of days.

dmihalcik-virtru:
It looks like this is not recommended by the viper developers

https://github.com/spf13/viper/issues/617#issuecomment-451579262

Issue #1220, opened 2024-07-29
feat: typed entities


strantalis:
Based on the outcome of the following adr https://github.com/opentdf/platform/issues/1181 this issue is to track the implementation of `typed` Entities. 

Update Entity Proto to add the ENUM `type` field that only supports `SUBJECT` or `ENVIRONMENT`

https://github.com/opentdf/platform/blob/5aef245720cb417a51622ffd1276f57163319c6d/service/authorization/authorization.proto#L17-L29

Next the example keycloak ers needs to be updated to classify entities as either `SUBJECT` or `ENVIRONMENT`. The distinction is we are classifying  the client in a client credentials flow as `SUBJECT` now and the client as `ENVIRONMENT` in a user based oauth2 flow. 

> [!NOTE]  
> Refer to ADR for other oauth2 access token examples

Finally in the access pkg update the logic to ignore `ENVIRONMENT` typed entities in the decisioning process.
> [!IMPORTANT]
> This is only until obligations are implemented based on the following ADR https://github.com/opentdf/platform/issues/874

Acceptance Criteria:
- Documentation updates around entity `type`
- Test Updates to cover changes in access decision pkg and ers changes to make sure we get back correct entity `type`
- Create new issues to track issues or future work as needed

PR #1249, opened 2024-08-02
feat(authz): Typed Entities

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/1220


service/authorization/authorization.go
```diff
@@ -282,12 +287,23 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
+						}
 						auditECEntitlements = append(auditECEntitlements, audit.EntityChainEntitlement{
-							EntityID:                 e.GetEntityId(),
+							EntityID:                 entityID,
 							AttributeValueReferences: e.GetAttributeValueFqns(),
 						})
-						entityAttrValues[e.GetEntityId()] = e.GetAttributeValueFqns()
+						entityCategory := entities[idx].GetCategory()
+
+						// If entity type unspecified, include in access decision to err on the side of caution
```
elizabethhealy, line 301: lmk if this is the right behavior

strantalis: I think its better to be safer here and check unspecified.

ttschampel (commented):
Does the ERS take care of the confidential client = [environment][subject] entities (copies of each other)?

pflynn-virtru (approved):

pflynn-virtru (approved):

PR #1251, opened 2024-08-05
chore(ci): check for associated issue on pull request

strantalis:
resolves https://github.com/opentdf/platform/issues/1252

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -67,6 +67,7 @@ jobs:
           with:
             github-token: ${{ secrets.GITHUB_TOKEN }}
             script: |
+              core.debug(JSON.stringify(context));
               const query = `query ($owner: String!, $name: String!, $number: Int!) {
                 repository(owner: $owner, name: $name) {
                   pullRequest(number: $number) {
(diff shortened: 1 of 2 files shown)


strantalis:
test comment

strantalis:
Example of the workflow working. 

https://github.com/strantalis/platform/pull/1

.github/workflows/checks.yaml
```diff
@@ -56,6 +56,96 @@ jobs:
             authz
             examples
 
+  associated-issue:
+    name: Issue Accociated with Pull Request
+    if: contains(fromJSON('["pull_request", "pull_request_target"]'), github.event_name)
```
dmihalcik-virtru, line 61:
```suggestion
    if: >-
      contains(fromJSON('["pull_request", "pull_request_target"]'), github.event_name)
      && !contains(fromJSON('["dependabot", "opentdf-automation"]'), github.actor)
```

dmihalcik-virtru (changes requested):
I haven't tested my change so please double check it

.github/workflows/checks.yaml
strantalis: I updated the if condition with what I think will work. Not an easy way to test this until we merge in I think.

strantalis: I updated the if condition with what I think will work. Not an easy way to test this until we merge in I think.

dmihalcik-virtru (approved):

Issue #984, opened 2024-06-13
ADR: Remove Attribute Value Members from Policy and instead house Value groupings in configuration of PDPs


jakedoublev:
### Background

Members were a legacy artifact of the OpenTDF Platform v1 (this is v2) and were kind of a virtualization of attribute values. Some of the following were revealed with the refactor of policy:

1. Value members seemed to only apply to `anyOf` rule attributes, where values were groupings of other values

     - `https://demo.com/attr/association/value/group_c` which has as members the values `group_a` and `group_b`
     - `hierarchy` and `allOf` rules don't really make sense for this grouping

2. It was additional burden placed on platform admins to understand entitlement policy around members, with some significant questions

	- are members always `anyOf` within themselves?
	- how is entitlement resolved if an admin has Subject Mappings pointing directly to a value with members? Is this behavior that should be explicitly prevented? 
	- how is entitlement resolved if an attribute value with members adds members of another attribute definition?

3. Tagging PDP and other PDPs can separately hold configured groupings of values as it makes sense contextually to roll up values appropriately outside of policy decisioning logic

Therefore, the decision was reached to `Remove Attribute Value Members from Policy and instead house Value groupings in configuration of PDPs`

The benefits are:

1. simplified administrative burden when there is only one way to entitle (subject mappings)
2. reduction of policy complexity
3. PDP-specific configuration flexibility without solving every PDP's contextual needs with a one-size-fits-all solution
4. greater unity across definition types (since members only seemed to have use cases within `anyOf` rule attribute definitions)

ttschampel:
By 2 cents:

- Members are currently used by Tagging PDP for computing rollup data policy
- I also think decision logic should be using these member values as part of the decision.  If a data policy with data attribute "someattr" is requested and its attribute definition includes members ["someotherattr"] and requesting entity has "someotherattr" as an entitlement then access should be granted.
- The use of subject mappings to directly entitling attribute values that have members should really not be used except for special cases (if ever?)

jakedoublev:
Does the above align with the architecture direction of the platform as you both see it @jrschumacher @ttschampel ?

PR #1258, opened 2024-08-06
fix(core): drop unused/deprecated resources table & add comments to policy DB

### LINKED ISSUE #1110: Policy: tables and columns of policy db schema should be commented
### Background

While we do have ERDs and documented migrations for the Platform Policy PostgreSQL database, there are no [comments](https://www.postgresql.org/docs/current/sql-comment.html) on tables or columns.

### Acceptance Criteria

1. New migration is added for the policy database that: 
    - adds comments to all policy tables 
    - adds comments to all policy table columns

### LINKED ISSUE #1216: Add comments to policy schema
The current schema lacks embedded comments within the database. We need to add these for better DBA experience.

jakedoublev:
Closes #1110

Ignores comments on `members` due to #984


service/policy/db/migrations/20240805000001_drop_deprecated_resources_table.sql
```diff
@@ -0,0 +1,26 @@
...
+    name VARCHAR NOT NULL,
+    namespace VARCHAR NOT NULL,
+    version INTEGER NOT NULL,
+    fqn VARCHAR,
+    labels JSONB,
+    description VARCHAR,
+    policytype VARCHAR NOT NULL,
+    resource JSONB
+);
+
```
jrschumacher, line 25: Could we move this to a distinct PR and update the VARCHAR to TEXT? They are the same thing but for old school DBAs it will remove the concern of specifying a length.

service/policy/db/migrations/20240805000001_drop_deprecated_resources_table.sql
```diff
@@ -0,0 +1,26 @@
...
+    id SERIAL PRIMARY KEY,
+    name VARCHAR NOT NULL,
+    namespace VARCHAR NOT NULL,
+    version INTEGER NOT NULL,
+    fqn VARCHAR,
+    labels JSONB,
+    description VARCHAR,
+    policytype VARCHAR NOT NULL,
+    resource JSONB
+);
```
jrschumacher, line 24: This is interesting. Not sure what the default behavior is for a migration the removes deprecated stuff but I feel like this might be unnecessary. :shrug:

jrschumacher (approved):

jrschumacher (approved):

service/policy/db/migrations/20240805000001_drop_deprecated_resources_table.sql
jakedoublev: I agree that it's unnecessary, but I added it to keep with the existing pattern so that "every up migration has a down migration". I don't think we've encountered this scenario yet. 🤷

jakedoublev: I think I'm going to leave it because, while it seems useless to add a down migration to add back an unused table, if someone else added a migration for something they believed was unused, I think I'd want them to add a down migration in case it was actually used unknowingly. It's probably best to just reinforce the practice. 🤔

Issue #1217, opened 2024-07-26
Move ERS call out of rego


elizabethhealy:
Currently, we call ERS from inside rego. 
Rego only supports http calls and we'd like the ability to hit ERS with grpc. Additionally this would remove some complexity from the rego input.

With this change, the new inputs to the rego would be a list of entity representations (returned from ERS) and the relevant subject mappings.

<img width="784" alt="Screenshot 2024-07-26 at 12 13 57 PM" src="https://github.com/user-attachments/assets/9768e767-f1d7-40a0-8117-09e4baa76be6">
<img width="763" alt="Screenshot 2024-07-26 at 12 21 37 PM" src="https://github.com/user-attachments/assets/f812ade1-6d68-42e4-925e-9b74bc055aed">

strantalis:
@elizabethhealy I wonder if we could not pass in subject mappings but instead inside of the builtin func call `MatchSubjctMappings` rpc.

elizabethhealy:
> MatchSubjctMappings

@strantalis has `MatchSubjctMappings` been implemented? i remember it was planned but hadn't heard anything about it

strantalis:
@elizabethhealy I think @jakedoublev implemented this a while back.

https://github.com/opentdf/platform/blob/5aef245720cb417a51622ffd1276f57163319c6d/service/policy/subjectmapping/subject_mapping.go#L163-L176

damorris25:
This is a good idea.

PR #1265, opened 2024-08-06
feat(authz): Remove external ers configuration from authorization

elizabethhealy:
We will configure the external ers via the service sdk instance 
resolves https://github.com/opentdf/platform/issues/1217


strantalis (dismissed):

jakedoublev (approved):
lgtm

Issue #1085, opened 2024-07-02
JWKS for KAS keys


pflynn-virtru:
All keys available for encryption via SDK clients should be exposed with a JWKS endpoint.

Each JWK should have a property that can be used to match an attribute to a key.

relates to:
- https://github.com/opentdf/platform/issues/717
- https://github.com/opentdf/platform/issues/335

strantalis:
@pflynn-virtru Why do we want to match an attribute to a key? Shouldn't we always just get the most up to date key from kas?

jrschumacher:
@strantalis are we still going to support this in the future?

PR #1267, opened 2024-08-06
feat(core): Lets service keys to be stored as jwk sets

dmihalcik-virtru:
- Adds jwk set output to init-keys for kas
- updates config to replace `KeyList` with a key set path
- Parses the jwk set to the existing format internally, which is a little ugly

Relates to https://github.com/opentdf/platform/issues/1085


dmihalcik-virtru:
closing as we reconsider key management utilities and admin / day 2 experience

PR #1268, opened 2024-08-06
fix(core): 1269 readme quickstart has incorrect steps

### LINKED ISSUE #1269: Quickstart documentation steps are incorrect
The current [Quickstart docs](https://github.com/opentdf/platform?tab=readme-ov-file#quick-start) have a few issues:

If you attempt to run the `docker compose` step without first generating dev keys/certs, you will get the following error:

image

This also leaves your `./keys` directory in a bad state (see below), and running the `./.github/scripts/init-temp-keys.sh` script won't fix this because any existing files don't seem to get recreated.  You have to delete the directory and rerun.

image

The other issue is that the `opentdf-example.yaml` config file uses a `keycloak` host in numerous places that isn't accessible by the go service when running `go run ./service start`.  You need to use `localhost`, which is fixed by copying the `opentdf-dev.yaml` file into your local `opentdf.yaml`.

In conclusion, the proper steps should be:
1. create dev keys/certs before anything else
2. docker compose up
3. copy `opentdf-dev.yaml` into local `opentdf.yaml`
4. run the keycloak provision command
5. start the platform service

ryanulit:
closes #1269

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,12 @@ To get started with the OpenTDF platform make sure you are running the same Go v
 <!-- markdownlint-disable MD034 github embedded sourcecode -->
 https://github.com/opentdf/platform/blob/main/service/go.mod#L3
 
+Generate development keys/certs for the platform infrastructure.
+
+```sh
+./.github/scripts/init-temp-keys.sh
+```
+
 Start the required infrastructure with [compose-spec](https://compose-spec.io).
 
 ```sh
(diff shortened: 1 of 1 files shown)


jakedoublev (commented):
Looks good to me, but I'd prefer @dmihalcik-virtru to also review for visibility.

ryanulit:
@dmihalcik-virtru does this look ok to you?

dmihalcik-virtru (approved):

PR #1270, opened 2024-08-06
feat(policy): 1256 resource mapping groups db support

### LINKED ISSUE #1256: Update database schema to support Resource Mapping groups
## Acceptance Criteria
- Add new migration for table storing groups of Resource Mappings
- Add new SQL queries for CRUD using `sqlc`
- Ensure no broken tests

ryanulit:
closes #1256 

- add migration for new `resource_mapping_groups` table and related ERD
- add basic SQL queries and `sqlc` generated functions (will be expanded on in follow-up PRs)

### DIFF
--- a/service/policy/db/migrations/20240806142109_add_resource_mapping_groups.sql
+++ b/service/policy/db/migrations/20240806142109_add_resource_mapping_groups.sql
@@ -8,7 +8,7 @@ CREATE TABLE IF NOT EXISTS resource_mapping_groups (
     UNIQUE(namespace_id, name)
 );
 
-ALTER TABLE resource_mappings ADD COLUMN group_id UUID REFERENCES resource_mapping_groups(id);
+ALTER TABLE resource_mappings ADD COLUMN group_id UUID REFERENCES resource_mapping_groups(id) ON DELETE SET NULL;
 
 -- +goose StatementEnd
 
--- a/service/policy/db/migrations/20240806142109_add_resource_mapping_groups.sql
+++ b/service/policy/db/migrations/20240806142109_add_resource_mapping_groups.sql
@@ -8,7 +8,14 @@ CREATE TABLE IF NOT EXISTS resource_mapping_groups (
     UNIQUE(namespace_id, name)
 );
 
+COMMENT ON TABLE resource_mapping_groups IS 'Table to store the groups of resource mappings by unique namespace and group name combinations';
+COMMENT ON COLUMN resource_mapping_groups.id IS 'Primary key for the table';
+COMMENT ON COLUMN resource_mapping_groups.namespace_id IS 'Foreign key to the namespace of the attribute';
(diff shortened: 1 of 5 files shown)


service/policy/db/migrations/20240806142109_add_resource_mapping_groups.sql
```diff
@@ -0,0 +1,29 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE TABLE IF NOT EXISTS resource_mapping_groups (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    namespace_id UUID NOT NULL REFERENCES attribute_namespaces(id),
```
jakedoublev, line 6:
```suggestion
    namespace_id UUID NOT NULL REFERENCES attribute_namespaces(id) ON CASCADE DELETE,
```

jakedoublev: if the namespace is deleted, we want to delete these groups

service/policy/db/migrations/20240806142109_add_resource_mapping_groups.sql
```diff
@@ -0,0 +1,29 @@
...
+);
+
+COMMENT ON TABLE resource_mapping_groups IS 'Table to store the groups of resource mappings by unique namespace and group name combinations';
+COMMENT ON COLUMN resource_mapping_groups.id IS 'Primary key for the table';
+COMMENT ON COLUMN resource_mapping_groups.namespace_id IS 'Foreign key to the namespace of the attribute';
+COMMENT ON COLUMN resource_mapping_groups.name IS 'Name for the group of resource mappings';
+
+ALTER TABLE resource_mappings ADD COLUMN group_id UUID REFERENCES resource_mapping_groups(id) ON DELETE SET NULL;
+
+COMMENT ON COLUMN resource_mappings.group_id IS 'Foreign key to the parent group of the resource mapping';
```
jakedoublev, line 18:
```suggestion
COMMENT ON COLUMN resource_mappings.group_id IS 'Foreign key to the parent group of the resource mapping (optional, a resource mapping may not be in a group)';
```

service/policy/db/query.sql
```diff
@@ -27,3 +29,27 @@ RETURNING id;
 -- name: DeleteKeyAccessServer :execrows
 DELETE FROM key_access_servers WHERE id = $1;
 
+---------------------------------------------------------------- 
+-- RESOURCE MAPPING GROUPS
+----------------------------------------------------------------
+
+-- name: GetResourceMappingGroup :one
```
jakedoublev, line 36: We support LIST of all policy objects, so you'll probably need one here for groups as well.

service/policy/db/migrations/20240806142109_add_resource_mapping_groups.sql
ryanulit: Good catch, and it should be `ON DELETE CASCADE`

Issue #1259, opened 2024-08-06
Slow response with GetEntitlement requests


jrschumacher:
When making a request to GetEntitlements the response is extremely slow. Additionally, the payload sent via the IPC exceeds the default 4Mb gRPC message size. A workaround was implemented in #1207 to enable setting a message size via configuration.

It was discovered that the root of the response time was split between [overlogging](https://github.com/opentdf/platform/blob/f275e25e4f986455fc536d2c93f7e9535f8519ab/service/authorization/authorization.go#L393) and non-performant [db queries](https://github.com/opentdf/platform/blob/cc15f25af2c3e839d7ad45283b7bd298a80e8728/service/policy/db/attribute_fqn.go#L188-L195) [respectively 1/3 and 2/3 of the cost].

## Acceptance criteria

- remove the excessive logging in favor of a payload size or count, etc
- investigate SQL issue and optimize performance

PR #1271, opened 2024-08-07
feat(core): improve entitlements performance 

suchak1:
### Context:

While developing COP, we found that GetEntitlements could take around 5.5 seconds to return a response for the federal dataset. This forced us to increase the server timeout, increase the grpc message size, and implement caches.

After investigating the latency, we determined the two main causes: [excess logs](https://github.com/opentdf/platform/blob/f275e25e4f986455fc536d2c93f7e9535f8519ab/service/authorization/authorization.go#L393) (33%) and [excess database queries](https://github.com/opentdf/platform/blob/cc15f25af2c3e839d7ad45283b7bd298a80e8728/service/policy/db/attribute_fqn.go#L188-L195) (66%). 

#### Primary
1. In the case of logging subject mappings, we now log their count instead of their content.
2. In the case of database calls, we now list attributes, list subject mappings, and match them based on their values. The database calls were O(n) time complexity because they were dependent on the number of attribute values. Now the database calls are constant time. We still loop through all the values to match their subject mappings; however we were already doing that in the `prepareValues` values function, so the new approach is strictly better (especially due to our ubiquitous use of maps).

#### Rego Query Optimization:
Yet if we simply match subject mappings and attribute values, the rego query becomes massive (65 mb). It takes 3 seconds to build (20%) and evaluate (80%). To optimize for not only time but also space, we remove unrelated values for each fqn/attribute pair in the rego input (unless the attribute rule is hierarchical). 

After all the optimizations, fetching entitlements using the federal dataset now takes about 125 ms. This is a **_latency reduction of 98%_**.

resolves: https://github.com/opentdf/platform/issues/1259


service/authorization/authorization.go
```diff
@@ -359,40 +360,84 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
 	return rsp, nil
 }
 
-func (as *AuthorizationService) GetEntitlements(ctx context.Context, req *authorization.GetEntitlementsRequest) (*authorization.GetEntitlementsResponse, error) {
-	as.logger.DebugContext(ctx, "getting entitlements")
-	request := attr.GetAttributeValuesByFqnsRequest{
-		WithValue: &policy.AttributeValueSelector{
-			WithSubjectMaps: true,
-		},
+func makeSubMapsByValLookup(subjectMappings []*policy.SubjectMapping) map[string][]*policy.SubjectMapping {
```
jrschumacher, line 363:
```suggestion
// makeSubMapsByValLookup creates a lookup map of subject mappings by attribute value ID.
func makeSubMapsByValLookup(subjectMappings []*policy.SubjectMapping) map[string][]*policy.SubjectMapping {
```

service/authorization/authorization.go
```diff
@@ -359,40 +360,84 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
-	if req.GetScope() == nil {
-		// TODO: Reomve and use MatchSubjectMappings instead later in the flow
-		listAttributeResp, err := as.sdk.Attributes.ListAttributes(ctx, &attr.ListAttributesRequest{})
-		if err != nil {
-			as.logger.ErrorContext(ctx, "failed to list attributes", slog.String("error", err.Error()))
-			return nil, status.Error(codes.Internal, "failed to list attributes")
+	return lookup
+}
+
+func updateValsWithSubMaps(values []*policy.Value, subMapsByVal map[string][]*policy.SubjectMapping) []*policy.Value {
```
jrschumacher, line 375:
```suggestion
// updateValsWithSubMaps updates the subject mappings of values using the lookup map.
func updateValsWithSubMaps(values []*policy.Value, subMapsByVal map[string][]*policy.SubjectMapping) []*policy.Value {
```

service/authorization/authorization.go
```diff
@@ -359,40 +360,84 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
-		var attributeFqns []string
-		for _, attr := range listAttributeResp.GetAttributes() {
-			for _, val := range attr.GetValues() {
-				attributeFqns = append(attributeFqns, val.GetFqn())
-			}
+	}
+	return values
+}
+
+func updateValsByFqnLookup(attribute *policy.Attribute, scopeMap map[string]bool, fqnAttrVals map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue) map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue {
```
jrschumacher, line 384:
```suggestion
// updateValsByFqnLookup updates the lookup map with attribute values by FQN.
func updateValsByFqnLookup(attribute *policy.Attribute, scopeMap map[string]bool, fqnAttrVals map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue) map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue {
```

service/authorization/authorization.go
```diff
@@ -359,40 +360,84 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
-	} else {
-		// get subject mappings
-		request.Fqns = req.GetScope().GetAttributeValueFqns()
+		a := &policy.Attribute{Rule: rule, Values: values}
+		fqnAttrVals[v.GetFqn()] = &attr.GetAttributeValuesByFqnsResponse_AttributeAndValue{Attribute: a, Value: v}
+	}
+	return fqnAttrVals
+}
+
+func makeValsByFqnsLookup(attrs []*policy.Attribute, subMapsByVal map[string][]*policy.SubjectMapping, scopeMap map[string]bool) map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue {
```
jrschumacher, line 400:
```suggestion
// makeValsByFqnsLookup creates a lookup map of attribute values by FQN.
func makeValsByFqnsLookup(attrs []*policy.Attribute, subMapsByVal map[string][]*policy.SubjectMapping, scopeMap map[string]bool) map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue {
```

jrschumacher (changes requested):
@suchak1 this looks good. Could we add some comments to make sure others understand what's happening?

jrschumacher (approved):
This looks good to me.

ttschampel (approved):

strantalis:
@suchak1 Would the MatchSubjectMappings rpc have helped here?

suchak1:
No, since the `MatchSubjectMappings` rpc just matches SubjectConditionSets with SubjectMappings and returns those SubjectMappings based on the provided SubjectConditionSets conditions.

service/authorization/authorization.go
```diff
@@ -359,40 +360,107 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
+	// add attribute value FQNs from scope to the map
+	for _, fqn := range scope.GetAttributeValueFqns() {
+		scopeMap[fqn] = true
+	}
+	return scopeMap
+}
+
+func (as *AuthorizationService) GetEntitlements(ctx context.Context, req *authorization.GetEntitlementsRequest) (*authorization.GetEntitlementsResponse, error) {
+	as.logger.DebugContext(ctx, "getting entitlements")
+	attrsRes, err := as.sdk.Attributes.ListAttributes(ctx, &attr.ListAttributesRequest{})
```
strantalis, line 443: @suchak1 Do we even need to make this call?

Subject mappings should return the value and its fqn but it seems like the sql query is not returning that field. Maybe this requires a follow up and larger discussion around what is actually returned in these calls.

service/authorization/authorization.go
```diff
@@ -359,40 +360,107 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
+	if err != nil {
+		as.logger.ErrorContext(ctx, "failed to list attributes", slog.String("error", err.Error()))
+		return nil, status.Error(codes.Internal, "failed to list attributes")
 	}
-	avf, err := as.sdk.Attributes.GetAttributeValuesByFqns(ctx, &request)
+	subMapsRes, err := as.sdk.SubjectMapping.ListSubjectMappings(ctx, &subjectmapping.ListSubjectMappingsRequest{})
```
jrschumacher, line 448: @suchak1 based on @strantalis' feedback would this provide any additional performance since it reduces the subject mappings to the scope via the DB request? https://github.com/opentdf/platform/blob/main/service/policy/db/subject_mappings.go#L766

service/authorization/authorization.go
jrschumacher: @strantalis I'd prefer to create an issue to add support for ListAttributes to allow returning values as well. That said, returning values for every attribute seems expensive. If we wanted this optimization, we should figure out how to reduce the number of attributes needed in the lookup.

service/authorization/authorization.go
suchak1: See [this answer](https://github.com/opentdf/platform/pull/1271#issuecomment-2276804350).

elizabethhealy: also we'd need to call MatchSubjectMappings from within the rego but rego doesnt doesnt currently support grpc

service/authorization/authorization.go
```diff
@@ -359,40 +360,107 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
+	// map keys will be attribute value IDs
+	lookup := make(map[string][]*policy.SubjectMapping)
+	for _, sm := range subjectMappings {
+		val := sm.GetAttributeValue()
+		id := val.GetId()
+		// if attribute value ID exists
+		if id != "" {
+			// append the subject mapping to the slice of subject mappings for the attribute value ID
+			lookup[id] = append(lookup[id], sm)
+		}
```
jakedoublev, line 374: is it an error case if `id == ""`, which looks like it would silently fail here, or is that intentional?

service/authorization/authorization.go
```diff
@@ -359,40 +360,107 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
-		for _, attr := range listAttributeResp.GetAttributes() {
-			for _, val := range attr.GetValues() {
-				attributeFqns = append(attributeFqns, val.GetFqn())
-			}
+	}
+	return values
+}
+
+// updateValsByFqnLookup updates the lookup map with attribute values by FQN.
+func updateValsByFqnLookup(attribute *policy.Attribute, scopeMap map[string]bool, fqnAttrVals map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue) map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue {
```
jakedoublev, line 392: Optional nit: `trimValsByFqnLookup`

service/authorization/authorization.go
```diff
@@ -359,40 +360,107 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
-		// TODO: Reomve and use MatchSubjectMappings instead later in the flow
-		listAttributeResp, err := as.sdk.Attributes.ListAttributes(ctx, &attr.ListAttributesRequest{})
-		if err != nil {
-			as.logger.ErrorContext(ctx, "failed to list attributes", slog.String("error", err.Error()))
-			return nil, status.Error(codes.Internal, "failed to list attributes")
+	return lookup
+}
+
+// updateValsWithSubMaps updates the subject mappings of values using the lookup map.
+func updateValsWithSubMaps(values []*policy.Value, subMapsByVal map[string][]*policy.SubjectMapping) []*policy.Value {
```
jakedoublev, line 380: Optional Nit: `populateValsWithSubmaps`

service/authorization/authorization.go
```diff
@@ -359,40 +360,107 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
+			// restore ALL attribute values if attribute rule is hierarchical
+			// This is key to honoring comprehensive hierarchy.
+			values = attribute.GetValues()
 		}
-		request.Fqns = attributeFqns
-	} else {
-		// get subject mappings
-		request.Fqns = req.GetScope().GetAttributeValueFqns()
+		// only clone relevant fields for attribute
+		a := &policy.Attribute{Rule: rule, Values: values}
```
jakedoublev, line 409: Optional Nit:
```suggestion
		trimmedAttr := &policy.Attribute{Rule: rule, Values: values}
```

service/authorization/authorization.go
```diff
@@ -359,40 +360,107 @@ func (as *AuthorizationService) GetDecisions(ctx context.Context, req *authoriza
...
+		attrs[i].Values = updateValsWithSubMaps(attrs[i].GetValues(), subMapsByVal)
+		// update the lookup map with attribute values by FQN
+		fqnAttrVals = updateValsByFqnLookup(attrs[i], scopeMap, fqnAttrVals)
+	}
+	return fqnAttrVals
+}
+
+// makeScopeMap creates a map of attribute value FQNs.
+func makeScopeMap(scope *authorization.ResourceAttribute) map[string]bool {
+	// if scope not defined, return nil pointer
```
jakedoublev, line 429: Optional nit:
```suggestion
	// if scope not defined, return nil map
```

service/authorization/authorization.go
jakedoublev: I'm a little concerned about `ListAttributes` becoming `GetAllPolicy` which seems like what would be needed for all of this in-memory restructuring. That said, we can tackle the selector enhancements to optionally support it if desired.

service/authorization/authorization.go
jakedoublev: Note `MatchSubjectMappings` needs to be updated regardless: https://github.com/opentdf/platform/issues/1290

service/authorization/authorization.go
strantalis: The sql query was just missing the join statement to get the fqn for the attribute value.

```
{
  "subject_mappings": [
    {
      "id": "7e1237d0-5445-4a8e-95f6-5c1e3bb4a125",
      "attribute_value": {
        "id": "01ea3af5-97e9-40d0-b148-2b061cfe53d3",
        "attribute": null,
        "value": "mar",
        "members": [],
        "grants": [],
        "fqn": "https://demo.com/attr/relto/value/mar",
        "active": true,
        "subject_mappings": [],
        "metadata": null
      },
      "subject_condition_set": {
        "id": "bb7145fa-5cc2-46eb-9d4d-b0d83bee284e",
        "subject_sets": [
          {
            "condition_groups": [
              {
                "conditions": [
                  {
                    "subject_external_selector_value": ".attributes.nationality[]",
                    "operator": "SUBJECT_MAPPING_OPERATOR_ENUM_IN",
                    "subject_external_values": [
                      "MAR"
                    ]
                  },
                  {
                    "subject_external_selector_value": ".clientId",
                    "operator": "SUBJECT_MAPPING_OPERATOR_ENUM_IN",
                    "subject_external_values": [
                      "dsp-outlook",
                      "dsp-sharepoint",
                      "shared",
                      "dsp-shared",
                      "opentdf",
                      "dsp-tdf-reader",
                      "dsp-tdf-reader-web",
                      "dsp-windows-filewatcher"
                    ]
                  }
                ],
                "boolean_operator": "CONDITION_BOOLEAN_TYPE_ENUM_OR"
              }
            ]
          }
        ],
        "metadata": {
          "created_at": "2024-08-08T17:06:31.155605Z",
          "updated_at": "2024-08-08T17:06:31.155605Z",
          "labels": {}
        }
      },
      "actions": [
        {
          "standard": "STANDARD_ACTION_DECRYPT"
        }
      ],
      "metadata": {
        "created_at": "2024-08-08T17:06:31.163358Z",
        "updated_at": "2024-08-08T17:06:31.163358Z",
        "labels": {}
      }
    },
......
```

service/authorization/authorization.go
strantalis: We now get the entity representation outside of rego. So once match subject mappings is updated we can call it before rego.

strantalis (dismissed):

service/authorization/authorization.go
suchak1: The id check also checks that the attribute value exists because of the function chaining. We could add a warning here in the future like `missing attribute value on subject mapping - db is potentially in a bad state`. I don't think it should necessarily error out the function since it's in a loop.

strantalis (approved):

PR #1276, opened 2024-08-07
chore(ci): Remove keytool step 🚧

dmihalcik-virtru:
- instead of a JKS encrypted with 'password', just store self-signed CA as a plain text pkcs12 file


dmihalcik-virtru:
@imdominicreed I think this should do what the previous step was doing but IDK how to validate it - do we have automated tests in this repo that verify that the self signed cert is respected?

PR #1278, opened 2024-08-07
chore(ci): re-run on description or title change

strantalis:
I ended up splitting out the pull request validation jobs into a separate workflows. It was getting confusing mixing code related checks and non code checks in one workflow.

resolves: https://github.com/opentdf/platform/issues/1279


jakedoublev (dismissed):

.github/workflows/pr-checks.yaml
```diff
@@ -0,0 +1,151 @@
...
+jobs:
+  pull-reqeust-checks:
+    needs:
+      - title-validation
+      - associated-issue
+    runs-on: ubuntu-22.04
+    if: always()
+    steps:
+      - if: contains(needs.*.result, 'failure')
+        run: echo "Failed due to ${{ contains(needs.*.result, 'failure') }}" && exit 1
```
dmihalcik-virtru, line 20: The reason we have the `ci` job in the other one is so we don't have to update admin settings to enable/disable branch protection; instead we have a single job listed as blocking CI. This workflow does not require such a marker job.

```suggestion
```

.github/workflows/checks.yaml
```diff
@@ -15,139 +19,6 @@ on:
       - checks_requested
 
 jobs:
-  pr:
```
dmihalcik-virtru, line 18: Perhaps we should add a direct call to [commitlint](https://github.com/conventional-changelog/commitlint) on `merge_group` events looking at all commits since `main` on the current branch. Ideally we'd find an action that supports checking both the PR desc/title and the generated commits themselves, but AFAICT there isn't one yet

dmihalcik-virtru (dismissed):
Minor note about removing an an unneeded job

.github/workflows/pr-checks.yaml
strantalis: Then I would need to set a required status check for each of those jobs I think.

jrschumacher (approved):

.github/workflows/checks.yaml
```diff
@@ -366,11 +237,9 @@ jobs:
       - image
       - integration
       - license
-      - pr
       - platform-xtest
-      - associated-issue
-    runs-on: ubuntu-latest
-    if: always()
+    runs-on: ubuntu-22.04
+    if: ${{ !cancelled() }}
```
dmihalcik-virtru, line 242: Yeah, I guess it is confusing to have this always be showing up as a red X everywhere, but since merges and PRs block on this we need to have it always run. If cancelled early this will count as skipped and therefore green
```suggestion
```

.github/workflows/pr-checks.yaml
```diff
@@ -0,0 +1,158 @@
...
+jobs:
+  pull-reqeust-checks:
+    needs:
+      - title-validation
+      - associated-issue
+    runs-on: ubuntu-22.04
+    if: ${{ !cancelled() }}
+    steps:
+      - if: contains(needs.*.result, 'failure')
+        run: echo "Failed due to ${{ contains(needs.*.result, 'failure') }}" && exit 1
```
dmihalcik-virtru, line 25:
```suggestion
```

dmihalcik-virtru (changes requested):

pflynn-virtru (approved):

.github/workflows/checks.yaml
strantalis: I added this condition based on the docs here and @el-virt feedback. 

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/expressions#always

.github/workflows/pr-checks.yaml
strantalis: Maybe I am missing something @dmihalcik-virtru but why don't we need this if we want it to be a required status check on a pull request?

.github/workflows/checks.yaml
strantalis: If we add always. If any of the jobs it depends on are skipped this will show up as passed when having `always()`

jrschumacher: Can we just create two jobs? One for sensitive stuff and one for non-sensitive. 

That way our workflow isn't so complicated. It feels wrong that we're making it hard to understand because we don't want to deal with terraform and gitops.

dmihalcik-virtru: It looks like cancelled jobs [have the status](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/contexts#needs-context) 'cancelled'. So let's go with this and in the end merge_group will catch any problems that fall through the PR review check

also, re: naming jobs and deciding which ones are 'required' - I guess this isn't an issue for you two since you are probably both admins, but there is no way that i've found for a non-admin to know or set required jobs. So it isn't an issue with gitops but visibility. Only admins may edit which jobs are required, and non-admins get surprised with a list of 'required' jobs after generating a PR.

image

.github/workflows/pr-checks.yaml
dmihalcik-virtru: yeah, this is fine if you don't want to make the `pr` job required which was what i thought you were going with. can you please fix the spelling?

.github/workflows/pr-checks.yaml
```diff
@@ -0,0 +1,158 @@
...
+      - reopened
+      - edited
+  merge_group:
+    branches:
+      - main
+    types:
+      - checks_requested
+
+jobs:
+  pull-reqeust-checks:
```
dmihalcik-virtru, line 17:
```suggestion
  pull-request-checks:
```

dmihalcik-virtru (changes requested):

biscoe916 (approved):

biscoe916 (approved):

PR #1280, opened 2024-08-07
feat(core): Add condition protos

elizabethhealy:
SPIKE: looking into have recursive condition sets


service/policy/conditions.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
+
+package policy;
+
+import "buf/validate/validate.proto";
+import "policy/objects.proto"; 
+
+message SingleCondition {
+  // a selector for a field value on a flattened Entity Representation (such as from idP/LDAP)
+  string subject_external_selector_value = 1;
```
jakedoublev, line 10: We should decouple these selector names from "subject" entity terminology, employing something like the below instead throughout:

```suggestion
  string external_selector_value = 1;
  repeated string external_selected_values = 3;
```

Issue #1295, opened 2024-08-09
KAS Grants to Attributes/Values should be able to be LISTed


jakedoublev:
### Background

We have a concept in policy protos of assignments of KASs to Attribute Definitions, and assignments to Attribute Values.

https://github.com/opentdf/platform/blob/main/service/policy/attributes/attributes.proto#L11-L23

There are administration and test needs to LIST the known KAS grants in platform policy.

### Acceptance Criteria

- [ ] new LIST protos/RPC `ListKeyAccessServerGrants` that returns:
    - [ ] NamespaceGrants with KAS info and Namespace info (id, ns fqn)
    - [ ] AttributeGrants with KAS info and Attribute info (id. attr fqn)
    - [ ] ValueGrants with KAS info and Value info (val id, val fqn)
- [x] new db queries in sqlc (see `query.sql` and [readme](https://github.com/opentdf/platform/tree/main/service/policy/db)) to support this list API
- [x] integration tests & fixtures

jakedoublev:
Blocked by #1296

jakedoublev:
Only partially completed

PR #1317, opened 2024-08-12
feat(core): policy support for LIST of kas grants (protos/db)

jakedoublev:
Resolves https://github.com/opentdf/platform/issues/1295 (partially)

- [x] new db queries in sqlc (see `query.sql` and [readme](https://github.com/opentdf/platform/tree/main/service/policy/db)) to support this list API
- [x] integration tests & fixtures


jakedoublev:
Service endpoints/RPCs require a release of `/protocol/go` as a dependency with the new proto gencode from this PR, and will therefore be in a follow-up PR

strantalis (approved):

PR #1328, opened 2024-08-13
fix(sdk): align sdk with platform modes

strantalis:
Resolves: https://github.com/opentdf/platform/issues/1360

### DIFF
--- a/sdk/options.go
+++ b/sdk/options.go
@@ -22,18 +22,19 @@ type config struct {
 	tokenExchange           *oauth.TokenExchangeInfo
 	tokenEndpoint           string
 	scopes                  []string
-	policyConn              *grpc.ClientConn
-	authorizationConn       *grpc.ClientConn
-	entityresolutionConn    *grpc.ClientConn
+	policyConn              *grpc.ClientConn // Deprecated
+	authorizationConn       *grpc.ClientConn // Deprecated
+	entityresolutionConn    *grpc.ClientConn // Deprecated
 	extraDialOptions        []grpc.DialOption
 	certExchange            *oauth.CertExchangeInfo
-	wellknownConn           *grpc.ClientConn
+	wellknownConn           *grpc.ClientConn // Deprecated
 	platformConfiguration   PlatformConfiguration
 	kasSessionKey           *ocrypto.RsaKeyPair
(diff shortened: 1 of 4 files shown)


sdk/options.go
```diff
@@ -22,18 +22,19 @@ type config struct {
 	tokenExchange           *oauth.TokenExchangeInfo
 	tokenEndpoint           string
 	scopes                  []string
-	policyConn              *grpc.ClientConn
-	authorizationConn       *grpc.ClientConn
-	entityresolutionConn    *grpc.ClientConn
+	policyConn              *grpc.ClientConn // Deprecated
```
jakedoublev, line 25: Question: does the deprecation notice work if you in-line it? I couldn't find a clear answer in the docs and it looks like all examples on the first two pages of code searches in GH are the line above: https://github.com/search?q=%22%2F%2F+deprecated%22+path%3A*.go&type=code&p=1

strantalis: I am honestly not sure. It was more for an internal reminder. I am really not sure if we should deprecate them or just rip them out.

jrschumacher: At this point, I think we need to defecate since we know there's external use.

sdk/options.go
```diff
@@ -99,18 +100,21 @@ func withCustomAccessTokenSource(a auth.AccessTokenSource) Option {
 	}
 }
 
+// Deprecated: Use WithCustomCoreConnection instead
 func WithCustomPolicyConnection(conn *grpc.ClientConn) Option {
```
jrschumacher, line 104: Should we just reroute this to the customcoreconnection?

strantalis: I am thinking about that. Does that get confusing and will that break anyone currently using multiple custom connections ?

jrschumacher: Yes, but if they are deprecated then the IDE should complain. A deployment may fail, but we can tell them that we no longer support that case. I'll leave it up to you.

strantalis: Updated

sdk/options.go
strantalis: I just removed this internal fields and now just set them to coreConn

jrschumacher (approved):

pflynn-virtru (approved):

PR #1318, opened 2024-08-12
fix(sdk): let value grants override attr grants

dmihalcik-virtru:
- If both a value and its attr have a grant assigned, only use the value grant.
- Closes #1329

PR #1346, opened 2024-08-14
feat(sdk): Load KAS keys from policy service

### LINKED ISSUE #1349: Use stored KAS keys from kas registry, when available
KASes can be all over the place and untrusted or flaky, or maybe not even on the internet. Instead, use the cached version of the key stored in the policy service's KAS registry, when found

dmihalcik-virtru:
- During encrypt autoconfigure, load any KAS keys found in the policy service's Kas Registry that are associated with grants from the new 'cached' field in the kas registry, if it is set
- These are then used to wrap any key access objects
- This wasn't previously used since the SDK requires the key identifiers and algorithm which were not present in the old 'public key' field
- Fixes #1349
- Adds a bats test that demonstrates loading a key from a KAS that is offline


service/policy/objects.proto
```diff
@@ -265,16 +265,36 @@ message KeyAccessServer {
...
+      id: "uri_format"
+      message: "URI must be a valid URL (e.g., 'https://demo.com/') followed by additional segments. Each segment must start and end with an alphanumeric character, can contain hyphens, alphanumeric characters, and slashes."
       expression: "this.matches('^https://[a-zA-Z0-9]([a-zA-Z0-9\\\\-]{0,61}[a-zA-Z0-9])?(\\\\.[a-zA-Z0-9]([a-zA-Z0-9\\\\-]{0,61}[a-zA-Z0-9])?)*(/.*)?$')"
     }];
 
     // public key - optional since can also be retrieved via url
     string local = 2;
+
+    // public key with additional information. Current preferred version
+    KasPublicKeySet cached = 3;
```
jakedoublev, line 298: This construct should replace local, not exist alongside it, correct? A base64-encoded public key (which was the prior value for local in this `oneof`) has no use going forward once this `KasPublicKeySet cached` construct is added?

jakedoublev:
Sorry, just realized those comments ^^ should be on https://github.com/opentdf/platform/pull/1347 instead. Resolving here and adding there.

service/policy/objects.proto
dmihalcik-virtru: it is a oneof so this provides backwards compatibility. I can mark it deprecated if you want

jakedoublev: Thanks - please mark it deprecated and reserved. 🙏  I'd prefer not to preserve backwards compatibility with an early API implementation that was never fully utilized or supported downstream.

sdk/granter.go
```diff
@@ -228,21 +228,52 @@ func NewGranterFromService(ctx context.Context, as attributes.AttributesServiceC
...
+		return "rsa:2048"
+	case policy.KasPublicKeyAlgEnum_KAS_PUBLIC_KEY_ALG_ENUM_UNSPECIFIED:
+		return ""
+	}
+	return ""
+}
+
+func storeKeysToCache(kases []*policy.KeyAccessServer, c *kasKeyCache) {
+	for _, kas := range kases {
+		if kas.GetPublicKey() == nil || kas.GetPublicKey().GetCached() == nil || kas.GetPublicKey().GetCached().GetKeys() == nil || len(kas.GetPublicKey().GetCached().GetKeys()) == 0 {
```
jakedoublev, line 257: The proto accessors should prevent nil panics here, so you could go all the way to `.GetKeys()` and just check nil/length there?

dmihalcik-virtru: yes, you are correct. this explains why the linter loves requiring the getter

dmihalcik-virtru:
> This looks good to me, but it looks like the pipeline is ❌ and it will probably have conflicts with #1318 ?

yes, I'll fix that up after merging this.

sdk/tdf_config.go
```diff
@@ -118,16 +117,16 @@ func WithDataAttributes(attributes ...string) TDFOption {
 
 // WithDataAttributeValues appends the given data attributes to the bound policy.
 // Unlike `WithDataAttributes`, this will not trigger an attribute definition lookup
-// during autoconfigure. That is, to use autoconfigure in an 'offline' context,
+// during  That is, to use autoconfigure in an 'offline' context,
```
jakedoublev, line 120: I think there's a missing word/phrase here

jakedoublev (dismissed):

sdk/tdf_config.go
dmihalcik-virtru: thx. added back

strantalis (approved):

PR #1347, opened 2024-08-14
feat(core): Adds key ids to kas registry

### LINKED ISSUE #1348: Store key identifiers with public key in KAS registry
We need to keep these values. Ideally this could be by adding JWK set support, but that might be too much work.

dmihalcik-virtru:
- Adds a new type to the oneof, which is quicker to work with than the full JWK Set spec
- Fixes #1348


service/policy/objects.proto
```diff
@@ -265,16 +265,36 @@ message KeyAccessServer {
...
 
+// A KAS public key and some associated metadata for further identifcation
+message KasPublicKey {
+  // x509 ASN.1 content in PEM envelope, usually
+  string pem = 1;
+
+  // A unique string identifier for this key
+  string kid = 2;
+
+  // A known algorithm type with any additional parameters encoded
```
jakedoublev, line 276: can we specify what kind of encoding any additional parameters should follow?

dmihalcik-virtru: Currently, this is functionally an enum with the values `ec:secp256r1` or `rsa:2048` - in the case EC + curve and bits, RSA with bits. I'll call those out explicitly

jrschumacher: If that's the case should we use enums in the proto

service/policy/objects.proto
```diff
@@ -265,16 +265,38 @@ message KeyAccessServer {
   common.Metadata metadata = 100;
 }
 
+// A KAS public key and some associated metadata for further identifcation
```
jrschumacher, line 268: Since we [only support two keys](https://github.com/opentdf/platform/blob/main/service/internal/security/crypto_provider.go#L8-L13) we should prevent users for uploading unsupported keys

```suggestion
enum KasPublicKeyAlgEnum {
  KAS_PUBLIC_KEY_ALG_ENUM_UNSPECIFIED = 0;
  KAS_PUBLIC_KEY_ALG_ENUM_RSA_2048 = 1;
  reserved 2 to 4;
  KAS_PUBLIC_KEY_ALG_ENUM_EC_SECP256R1 = 5;
  reserved 5 to 100;
}

// A KAS public key and some associated metadata for further identifcation
```

service/policy/objects.proto
```diff
@@ -265,16 +265,38 @@ message KeyAccessServer {
...
+  // x509 ASN.1 content in PEM envelope, usually
+  string pem = 1;
+
+  // A unique string identifier for this key
+  string kid = 2;
+
+  // A known algorithm type with any additional parameters encoded.
+  // To start, these may be `rsa:2048` for encrypting ZTDF files and 
+  // `ec:secp256r1` for nanoTDF, but more formats may be added as needed.
+  string alg = 3;
```
jrschumacher, line 279:
```suggestion
  KasPublicKeyAlgEnum alg = 3;
```

service/policy/objects.proto
dmihalcik-virtru: please no

dmihalcik-virtru: We are an open product and this is intended to be extensible and useful. by adding explicit values here we prevent ourselves from adding new types that are experimental or unsupported by opentdf

dmihalcik-virtru: Let's see how it works out.

dmihalcik-virtru: Eh it isn't that big of a change so I'll go ahead with this. In the end I do still want to replace this whole thing with a JWK Set, which will be definitely be safest to use strings. We don't want policy service to have too big of an opinion about these things because it will cause surprising breakages when round tripping through various non-compliant protocol buffer implementations.

https://github.com/opentdf/platform/pull/1346/commits/0a902c605a802a7fe6136212b130c1994d3c9530

pflynn-virtru (approved):

jrschumacher (approved):

PR #1350, opened 2024-08-14
wrapped h2 in CORS instead of h to enable CORS on GRPC

davidbkirk:
CORS issues from COP to platform when on different domains

@davidmmartin5 @danny-holloway

### DIFF
--- a/service/internal/server/server.go
+++ b/service/internal/server/server.go
@@ -196,9 +196,19 @@ func newHTTPServer(c Config, h http.Handler, a *auth.Authentication, g *grpc.Ser
 		l.Error("disabling authentication. this is deprecated and will be removed. if you are using an IdP without DPoP set `enforceDPoP = false`")
 	}
 
+	// Enable pprof
+	if c.EnablePprof {
+		h = pprofHandler(h)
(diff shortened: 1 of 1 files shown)


stoneb344:
CORS issue in OpenShift with requests to different subdomain:
Screenshot 2024-08-14 at 3 26 36 PM

jakedoublev:
Does setting the [configured CORS config](https://github.com/opentdf/platform/blob/main/opentdf-dev.yaml#L79-L80) resolve this?

davidbkirk:
@stoneb344  https://github.com/opentdf/platform/blob/2b7093136f6af1b6a86e613c095cefe403c9a06c/opentdf-example.yaml#L67-L92

stoneb344:
Setting those values resolved the CORS error. Thanks for pointing it out

stoneb344:
With `server.cors.enabled: true` and `server.cors.allowedorigins:` including an entry for the COP origin (https://cop-demox.apps.dsp.shp.virtru.us), we encounter a separate CORS issue regarding a preflight redirect:

Screenshot 2024-08-16 at 1 04 42 PM

PR #1352, opened 2024-08-14
feat(sdk): Lets SDK use PublicKeys from policy service

dmihalcik-virtru:
- During encrypt autoconfigure, load any KAS keys found in the policy service's Kas Registry that are associated with grants
- These are then used to wrap any key access objects
- Adds a bats test that demonstrates loading a key from a KAS that is offline


dmihalcik-virtru:
Wrong branch. Use https://github.com/opentdf/platform/pull/1346 instead

PR #1365, opened 2024-08-16
feat(sdk): public client and other enhancements to well-known SDK functionality

### LINKED ISSUE #1373: Unauthenticated Clients (CLI and PEPs) should be able to fetch well-known idp config values with SDK
The SDK should allow clients like `otdfctl` and PEPs to fetch well-known idP config values that drive an authentication flow.

The platform should also make an optional and configurable well-known `public client id` available for discovery.

jakedoublev:
Resolves #1373

This PR makes the following changes:
1. provisions a public client to keycloak with an allowlisted redirectUri of `http://localhost:9000` to support `otdfctl`
2. adds a few `idP` accessor methods to the SDK for reading the well-known configuration of the platform pre-auth
3. removes a few errors about the SDK being unauthenticated now that reading from the well-known pre-auth is a valid SDK use 
4. adds a `public_client_id` to the well-known idP config (with example configs updated) and a warning log about discovery if not provided in the config on startup
5. cannot remove `platform_issuer` as a top-level well-known config key that is a duplicate of `idp.issuer` because of this issue https://github.com/opentdf/java-sdk/issues/119

```json
{
  "configuration": {
    "health": {
      "endpoint": "/healthz"
    },
    "idp": {
      "authorization_endpoint": "http://localhost:8888/auth/realms/opentdf/protocol/openid-connect/auth",
      "id_token_signing_alg_values_supported": [
        "PS384",
        "RS384",
        "EdDSA",
        "ES384",
        "HS256",
        "HS512",
        "ES256",
        "RS256",
        "HS384",
        "ES512",
        "PS256",
        "PS512",
        "RS512"
      ],
      "issuer": "http://localhost:8888/auth/realms/opentdf",
      "jwks_uri": "http://localhost:8888/auth/realms/opentdf/protocol/openid-connect/certs",
      "public_client_id": "opentdf-public", // this is new
      "require_request_uri_registration": true,
      "response_types_supported": [
        "code",
        "none",
        "id_token",
        "token",
        "id_token token",
        "code id_token",
        "code token",
        "code id_token token"
      ],
      "subject_types_supported": [
        "public",
        "pairwise"
      ],
      "token_endpoint": "http://localhost:8888/auth/realms/opentdf/protocol/openid-connect/token"
    }
  }
}
```


sdk/options.go
```diff
@@ -93,7 +93,7 @@ func WithTokenEndpoint(tokenEndpoint string) Option {
 	}
 }
```
jakedoublev, line 95: maybe? 
```suggestion
func WithStandardOauthTokenSource(ts oauth.TokenSource) Option {
    return func(c *config) {
        c.standardOauthTokenSource = ts
    }
}
```

sdk/platformconfig.go
```diff
@@ -2,7 +2,7 @@ package sdk
 
 import "log/slog"
 
-func (s SDK) PlatformIssuer() string {
+func (s SDK) getIdpConfig() map[string]interface{} {
 	// This check is needed if we want to fetch platform configuration over ipc
 	if s.config.platformConfiguration == nil {
```
jakedoublev, line 7: @strantalis does this still look right after the merge conflict?

jakedoublev:
Cannot deprecate `platform_issuer` at top level without doing this first: https://github.com/opentdf/java-sdk/issues/119

lib/fixtures/keycloak.go
```diff
@@ -305,6 +306,21 @@ func SetupKeycloak(ctx context.Context, kcConnectParams KeycloakConnectParams) e
...
+		Name:            gocloak.StringP(opentdfPublicClientID),
+		PublicClient:    gocloak.BoolP(true),
+		ProtocolMappers: &protocolMappers,
+		// TODO: take via CLI arguments?
+		// Note: otdfctl CLI auth code flow runs from localhost:9000)
+		RedirectURIs: &[]string{"http://localhost:9000/*"},
+	}, nil, nil)
+	if err != nil {
+		return err
+	}
```
jrschumacher, line 322: If we're hardcoding the RedirectURIs then I think this should be a `otdfctlPublicClient` like `enable_otdfctl_client` or something like that.

opentdf-example.yaml
```diff
@@ -27,6 +27,7 @@ server:
   auth:
     enabled: true
     enforceDPoP: false
+    publicclientid: "opentdf-public"
```
jrschumacher, line 30: Can we use underscores to make this more legible.

sdk/options.go
```diff
@@ -16,6 +16,8 @@ type Option func(*config)
 
 // Internal config struct for building SDK options.
 type config struct {
+	// Platform configuration structure is subject to change. Consume via accessor methods.
+	PlatformConfiguration   PlatformConfiguration
```
jrschumacher, line 20: If this is the case, should we lowercase `platformConfiguration`?

jrschumacher:
```suggestion
	platformConfiguration   PlatformConfiguration
```

sdk/options.go
```diff
@@ -164,7 +165,7 @@ func WithCustomWellknownConnection(conn *grpc.ClientConn) Option {
 // Use this option with caution, as it may lead to unexpected behavior
 func WithPlatformConfiguration(platformConfiguration PlatformConfiguration) Option {
 	return func(c *config) {
-		c.platformConfiguration = platformConfiguration
+		c.PlatformConfiguration = platformConfiguration
```
jrschumacher, line 168:
```suggestion
		c.platformConfiguration = platformConfiguration
```

jrschumacher (changes requested):

sdk/options.go
jakedoublev: This was from feedback from @strantalis to put these accessors a level beyond the top-level SDK methods. Exposing the `PlatformConfiguration` means the below:

```golang
// if exposed:
client := sdk.New()
// client.PlatformConfiguration is map[string]interface{}
client.PlatformConfiguration.Issuer() // gives the well-known issuer and same APIs for well-known config accessors

// if unexposed
client := sdk.New()
client.PlatformIssuer() // a new top-level method for every well-known accessor
```

lib/fixtures/keycloak.go
jakedoublev: I think this public client may also be used by varied PEPs and is not guaranteed to be `otdfctl` only. Note this block of code is only used in the `go build -o platform && ./platform provision keycloak` platform binary command flow (which maybe we should just deprecate rather than continuing to support both?). The `./platform provision keycloak-from-config` redirect_uris are in YAML and not hardcoded: https://github.com/opentdf/platform/pull/1365/files#diff-2d149f4c96fefaec806f5f07b2f6b804903ac1b5f0b90a54cfebe53c088119c0R74

jakedoublev: TODO: Make `provision keycloak` the primary command that reads from the YAML and deprecate `keycloak-from-config`.

jakedoublev:
Work on this uncovered https://github.com/opentdf/platform/issues/1384

jrschumacher (approved):

PR #1366, opened 2024-08-16
fix(core): cleanup sensitive info being logged from configuration

strantalis:
Should Resolve: #751

In this pull request, I add the masq package to provide developers the ability to mask struct fields by applying the tag masq:"secret".

I also removed logging of the services field from the config object. This change forces anyone adding additional services to log their own service-specific configuration separately. The main reason for this is that there are too many unknowns regarding the understanding of the config structure. However, if developers leverage this logger and apply the struct tags, they can redact sensitive fields.

For more information, visit https://github.com/m-mizutani/masq.

Example Loading Config Log

>time=2024-08-16T12:12:18.497Z level=DEBUG msg="config loaded" config.dev_mode=true config.db.host=localhost config.db.port=5432 config.db.database=opentdf config.db.user=postgres config.db.password=[REDACTED] config.db.sslmode=prefer config.db.schema=opentdf config.db.runMigrations=true config.db.verifyConnection=true config.logger="{Level:trace Output:stdout Type:text}" config.mode=[all] config.sdk_config="{Endpoint: Plaintext:false ClientID: ClientSecret:[REDACTED]}" config.server.auth="{Enabled:false PublicRoutes:[] AuthNConfig:{EnforceDPoP:false Issuer:https://virtru-keylcoak.ngrok.pizza/auth/realms/opentdf Audience:http://localhost:8080 Policy:{Default: RoleClaim: RoleMap:map[] Csv: Model:} CacheRefresh:15m DPoPSkew:1h0m0s TokenSkew:1m0s}}" config.server.grpc="{ReflectionEnabled:true MaxCallRecvMsgSizeBytes:4194304 MaxCallSendMsgSizeBytes:4194304}" config.server.cryptoProvider="{Type:standard StandardConfig:{Keys:[{Algorithm:rsa:2048 KID:r1 Private:kas-private.pem Certificate:kas-cert.pem Usage: Purpose:} {Algorithm:ec:secp256r1 KID:e1 Private:kas-ec-private.pem Certificate:kas-ec-cert.pem Usage: Purpose:}] RSAKeys:map[] ECKeys:map[]}}" config.server.tls="{Enabled:false Cert: Key:}" config.server.cors="{Enabled:false AllowedOrigins:[*] AllowedMethods:[GET POST PATCH PUT DELETE OPTIONS] AllowedHeaders:[ACCEPT Authorization Content-Type X-CSRF-Token X-Request-ID] ExposedHeaders:[Link] AllowCredentials:true MaxAge:3600}" config.server.port=8080 config.server.host="" config.server.enablePprof=false

Example Service Log

>time=2024-08-16T12:12:18.676Z level=DEBUG msg="entity_resolution configuration" namespace=entityresolution config="{URL:http://localhost:8888/auth Realm:opentdf ClientID:tdf-entity-resolution ClientSecret:[REDACTED] LegacyKeycloak:true SubGroups:false InferID:{From:{ClientID:false Email:true Username:true}}}"


jrschumacher (dismissed):
This looks like a good start but I think we need a little more to prevent accidental leaks. Areas of concern:

- copying config to an interface and printing
- marshaling and printing
- extra config props that are not mapped to a struct

We can address those at another time.

go.work
```diff
@@ -1,10 +1,12 @@
-go 1.21
+go 1.22
```
pflynn-virtru, line 1: this strongly affects downstream

dmihalcik-virtru: Yeah it looks like masq requires 1.22 for no real reason?

strantalis: I reverted that but not sure what it will break if a dep requires it? `tidy` must have updated that.

strantalis:
Because we detected performance issues adding masq. I have decided to remove it for the time being. Instead we just use the standard `LogValue` method to remove sensitive content when logging configuration.

jrschumacher:
Specifically, masq used reflection. When logging large objects like request bodies this caused severe performance issues.

jrschumacher (changes requested):

jrschumacher (approved):

PR #1385, opened 2024-08-19
fix(sdk): 🔒 During read, limits TDF Manifest to 10MB 

dmihalcik-virtru:
This commit limits TDF Manifest sizes to 10MB or smaller.  Prior to this fix, a protocol attack is possible to define a large manifest size which will result in a large memory allocation (even if the manifest transmitted is substantially smaller).

This commit also includes the fuzz testing that was used to discover this flaw.

In `reader` the Manifest load uses the function `ReadAllFileData`.  This function is documented to only be used on small files, however to accomplish that a call to `ReadFileSize` beforehand would be necessary.

Because the call to check the defined size is likely to be missed, this commit updates `ReadAllFileData` to instead accept the maximum size as an argument, which is validated before allocating the byte slice.

Fixes https://github.com/opentdf/platform/issues/1391


jentfoo (approved):

dmihalcik-virtru:
@jentfoo Can you take a look and make sure I haven't introduced any issues with my fixups?

Thanks!

.github/workflows/checks.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
           args: --out-format=colored-line-number
       - if: matrix.directory == 'service'
         run: .github/scripts/init-temp-keys.sh
-      - run: go test ./... -short
+      - run: go test ./... -short -fuzztime=30s
```
jentfoo, line 73: I don't think we want to introduce fuzzing as part of our normal unit test workflow (beyond the fuzz seeds, which will be validated without any changes here).

We should keep unit tests running fast and be deterministic.  Unfortunately fuzzing is neither, it's very easy to have one time failures and it often needs to run for a long period to be valuable.

I would like to introduce regular automated fuzzing, but I think we should do that as future work.  When we do introduce it we need to consider:
* How do we securely report the failed test payload, since it may not be easy to recreate (or even impossible)
* What cadence makes sense (cost vs benefit balance)

Since this project is open source one option we may consider to meet these needs: https://github.com/google/oss-fuzz

But in any case, I rather do these changes separate from the fix here.

sdk/fuzz_test.go
```diff
@@ -0,0 +1,150 @@
...
+	f.Add(writeBytes(pb.writePolicyBody))
+
+	f.Fuzz(func(t *testing.T, data []byte) {
+		pb = &PolicyBody{}
+		err := pb.readPolicyBody(bytes.NewReader(data))
+		if err != nil {
+			assert.Zerof(t, *pb, "unexpected %v", *pb)
+			return
+		}
+		switch pb.mode {
```
jentfoo, line 122: I am somewhat cautious to put this type of validation into a fuzz test.  If we add a new mode it may not be obvious this test needs to be updated, in part because the test likely wont fail immediately (needing a random mode that now has been added).

In fuzzing we also want to keep validation somewhat minimal so that fuzz iterations can be as fast as possible.  Usually I defer to only the default validation which is:
* It does not panic
* It returns quickly (I believe 10 seconds is the default)
* There is no out of memory condition hit (unfortunately this is system dependent, so you may want to consider adding additional process limits if on a system with a lot of ram)

Instead I think this type of validation is better as a unit test, where an invalid mode is used and validated that the `undefined policy mode` error is returned.

sdk/resource_locator.go
```diff
@@ -48,28 +48,11 @@ func NewResourceLocator(url string) (*ResourceLocator, error) {
...
-		return rl, err
-	}
-
-	l := oneByte[0]
-	body := make([]byte, l)
-	_, err = reader.Read(body)
+	err := rl.readResourceLocator(reader)
 	if err != nil {
-		return rl, err
+		return nil, err
```
jentfoo, line 53: No concerns at all, just highlighting this is a behavior change (I was trying to match the existing behavior as close as I can).

sdk/fuzz_test.go
dmihalcik-virtru: fair enough. It is interesting that when not run with `-fuzz`, the default behavior is to run the tests on the seeds - basically turning the fuzz test into a table driven test, in golang parlance. For more complex 'expected error' tests, it may make sense to share the seeds from the table driven tests that exercise error conditions explicitly

jentfoo (approved):

jentfoo:
Thank you for the help @dmihalcik-virtru 🙌

strantalis (approved):

PR #1394, opened 2024-08-19
feat(sdk): support oauth2 tokensource with option

### LINKED ISSUE #1307: SDK should support `WithOIDCAccessToken` option
### Background

Currently, the [otdfctl CLI](https://github.com/opentdf/otdfctl/pull/144) is only able to authenticate users through the client credentials flow, which is itself a limitation of the SDK (relying on the [WithClientCredentials option](https://github.com/opentdf/platform/blob/cfab3ad8a72f3a2f1a28ccca988459ddcdcbd7f6/sdk/options.go#L74)). There is also another supported option [WithTokenExchange](https://github.com/opentdf/platform/blob/cfab3ad8a72f3a2f1a28ccca988459ddcdcbd7f6/sdk/options.go#L120-L129). However, the CLI and other clients should not be required to necessarily utilize on a token exchange flow when they may be driven by an entity directly (as in the case of a CLI) and there are cases where the client should be of lower trust.

### Acceptance Criteria

1. SDK authenticates gRPC requests with a passed in OIDC Access Token
2. Unit tests

jakedoublev:
Resolves #1307


sdk/sdk.go
```diff
@@ -210,11 +210,30 @@ func SanitizePlatformEndpoint(e string) (string, error) {
...
 		return c.customAccessTokenSource, nil
 	}
 
+	if c.oauthAccessTokenSource != nil {
+		if c.dpopKey == nil {
+			rsaKeyPair, err := ocrypto.NewRSAKeyPair(dpopKeySize)
+			if err != nil {
+				return nil, fmt.Errorf("could not generate RSA Key: %w", err)
+			}
+			c.dpopKey = &rsaKeyPair
```
jrschumacher, line 232:
```suggestion
			var err error
			c.dpopKey, err = buildRSAKeyPair(dpopKeySize)
			if err != nil {
				return nil, err
			}
```

jrschumacher (approved):
Reuse function?

sdk/idp_oauth_access_token_source.go
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// AccessToken use a pointer receiver so that the token state is shared
+func (t *OAuthAccessTokenSource) AccessToken(_ context.Context, _ *http.Client) (auth.AccessToken, error) { // must satisfy auth.AccessTokenSource interface
+	tok, err := t.source.Token()
+	if err != nil {
+		return "", fmt.Errorf("error getting access token: %w", err)
+	}
+
+	if tok.Expiry.Before(time.Now()) {
```
jentfoo, line 50:
```suggestion
	if !tok.Valid() {
```

It looks like the valid check does some additional checks (token has been set, and the expiration has been set as well).  The expiration check also allows a small time buffer.

sdk/sdk.go
```diff
@@ -233,10 +228,16 @@ func buildIDPTokenSource(c *config) (auth.AccessTokenSource, error) {
 		c.dpopKey = &rsaKeyPair
 	}
 
+	if c.certExchange != nil && c.tokenExchange != nil {
```
jentfoo, line 231: Why did you move this condition down?  It seems ideal to fail before potentially generating a new key

jentfoo (commented):
Some initial comments, I assume this is a WIP since unit tests are missing.

sdk/sdk.go
jakedoublev: That was an oversight. Agreed on generating the key pair only when needed.

jrschumacher (approved):

sdk/idp_oauth_access_token_source_test.go
```diff
@@ -0,0 +1,97 @@
...
+	// Interface checks
+	tok, err := tokenSource.AccessToken(context.Background(), nil)
+	require.NoError(t, err)
+	assert.Equal(t, tok, auth.AccessToken(mockToken))
+	made, err := tokenSource.MakeToken(func(jwk.Key) ([]byte, error) { return []byte(mockToken), nil })
+	require.NoError(t, err)
+	assert.Equal(t, made, []byte(mockToken))
+}
+
+func TestNewOAuthAccessTokenSource_ExpiredToken(t *testing.T) {
```
jentfoo, line 45: It could be good to add a test for the empty string `""` token as well.  Right now it should be the same as the expired condition, but it will be a good case for the future when the TODO is implemented for refreshing the token.

jentfoo (approved):
LGTM, thank you for the updates!  One comment to consider for the future

sdk/idp_oauth_access_token_source_test.go
jakedoublev: Thanks, and I'm not sure if we do want the SDK concerned with refreshing access tokens or if it's work an SDK consumer should do on their own.

Issue #1406, opened 2024-08-20
Service module panics due to regex pattern introduced in Go 1.22


jakedoublev:
https://github.com/opentdf/otdfctl/actions/runs/10479662480/job/29025717434?pr=293 
The service module will currently panic as we unknowingly [introduced a regex capturing group](https://github.com/golang/go/issues/64108) only introduced in 1.22. It looks like it was missed in CI because we use 1.22.5 in the [pipeline checks](https://github.com/opentdf/platform/blob/main/.github/workflows/checks.yaml#L96) but 1.21 in our go.work and [service go.mod](https://github.com/opentdf/platform/blob/main/service/go.mod).

PR #1407, opened 2024-08-20
fix(core): bump go version to 1.22

jakedoublev:
Resolves https://github.com/opentdf/platform/issues/1406

### DIFF
--- a/go.work
+++ b/go.work
@@ -1,4 +1,4 @@
-go 1.21
+go 1.22
 
 use (
 	./examples
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -40,7 +40,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7
         with:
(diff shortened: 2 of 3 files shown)


go.work
```diff
@@ -1,4 +1,4 @@
-go 1.21
+go 1.22
```
jentfoo, line 1: This was recently discussed as having downstream effects: https://github.com/opentdf/platform/pull/1366#discussion_r1719940378

cc @pflynn-virtru and @dmihalcik-virtru

strantalis: I think we have no choice to force users to upgrade. We introduced a breaking change in our regex that is only supported in 1.22.

strantalis (approved):

go.work
jakedoublev: I'm not extremely familiar with the `go.work` workspace setup intricacies, but it looks like it forces `1.22` if a module in the workspace requires it: 
Screenshot 2024-08-20 at 3 31 32 PM

Unfortunately, the breaking change necessitates this or we get a panic on builds from `1.21` (see #1406).

PR #1412, opened 2024-08-21
feat(policy): 1398 add metadata support to Resource Mapping Groups

### LINKED ISSUE #1398: Update Resource Mapping Groups policy object to support the `Metadata` property
Update DB logic returning Resource Mapping Group objects to ensure they support `Metadata` property operations

ryanulit:
closes #1398 

Implements new migration to add `resource_mapping_groups.metadata` column and updates protos, sqlc queries and DB logic to use it


service/policy/db/migrations/20240821000000_add_res_map_groups_metadata_field.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE resource_mapping_groups ADD COLUMN metadata JSONB;
```
jakedoublev, line 4: Can we please add an existence check to each?

```suggestion
ALTER TABLE IF EXISTS resource_mapping_groups ADD COLUMN metadata JSONB;
```

service/policy/db/resource_mapping.go
```diff
@@ -48,17 +54,29 @@ func (c PolicyDBClient) GetResourceMappingGroup(ctx context.Context, id string)
...
 	return &policy.ResourceMappingGroup{
 		Id:          rmGroup.ID,
 		NamespaceId: rmGroup.NamespaceID,
 		Name:        rmGroup.Name,
+		Metadata:    metadata,
 	}, nil
 }
 
 func (c PolicyDBClient) CreateResourceMappingGroup(ctx context.Context, r *resourcemapping.CreateResourceMappingGroupRequest) (*policy.ResourceMappingGroup, error) {
+	metadataBytes, _, err := db.MarshalCreateMetadata(r.GetMetadata())
```
jakedoublev, line 71: It looks like this would marshal JSONB with something like `{"labels":{}}` and an empty map when the metadata was empty on create. Perhaps we should check that behavior, and if that's what it indeed does, check if it's nil and pass nil bytes in that case?

service/policy/db/migrations/20240821000000_add_res_map_groups_metadata_field.sql
ryanulit: Done

jakedoublev:
This looks good to me. ✅ when release freeze is ended.

jakedoublev (approved):

PR #1418, opened 2024-08-21
feat(policy): 1370 add audit support for Resource Mapping Groups

### LINKED ISSUE #1370: Add audit support for Resource Mapping Groups
Audit logs should be created for Create/Update/Read operations utilizing the existing approach already in-use by the Resource Mapping object operations within the `ResourceMappingService`

Examples:
https://github.com/opentdf/platform/blob/5b9e2cfe3607a4412550c6f5f3f3f699d918734a/service/policy/resourcemapping/resource_mapping.go#L20
https://github.com/opentdf/platform/blob/5b9e2cfe3607a4412550c6f5f3f3f699d918734a/service/policy/resourcemapping/resource_mapping.go#L54
https://github.com/opentdf/platform/blob/5b9e2cfe3607a4412550c6f5f3f3f699d918734a/service/policy/resourcemapping/resource_mapping.go#L59

ryanulit:
closes #1370


service/policy/resourcemapping/resource_mapping.go
```diff
@@ -75,24 +84,59 @@ func (s ResourceMappingService) CreateResourceMappingGroup(ctx context.Context,
...
+
+	item, err := s.dbClient.UpdateResourceMappingGroup(ctx, id, req)
 	if err != nil {
+		s.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
 		return nil, db.StatusifyError(err, db.ErrTextUpdateFailed, slog.String("id", id), slog.String("resourceMappingGroup", req.String()))
 	}
 
+	// UpdateResourceMappingGroup only returns the ID of the updated resource mapping group
+	// so we need to fetch the updated resource mapping to compute the audit diff
+	// todo: would be easier to just return the whole group on update
```
ryanulit, line 107: Bringing this up as a topic for discussion on the PR, but would it not make sense to just return the entire object on update and/or create?  The postgres queries could be updated to `RETURNING <all_columns_of_table>`.

jakedoublev: Either one is fine. I'm not very concerned about performance on write if we keep the roundtrip because read throughout policy is so much more common (though RM Groups are only read by PDPs). 

In general, db queries can be inclusive or just `id`, but the RPC behavior should be preserved that Update just provides back the updated ID for consistency with the current to avoid confusion with before/after state. Create returns everything and Delete appears to be inconsistent so can be either just the ID or everything.

jakedoublev: addressing in a separate PR to match existing conventions and documented ADR

ryanulit: Decision has been made to only return what is absolutely necessary according to the recently merged ADR docs: https://github.com/opentdf/platform/tree/main/service/policy/adr

jakedoublev (approved):

PR #1422, opened 2024-08-22
feat(policy): 1421 tech debt migrate Resource Mappings object queries to sqlc

### LINKED ISSUE #1421: Migrate Resource Mapping object queries to sqlc

ryanulit:
closes #1421 

also implements:
- architecture decisions per [adr readme
](https://github.com/opentdf/platform/blob/main/service/policy/adr/0000-current-state.md)
- recent audit improvements


service/policy/db/utils.go
```diff
@@ -27,9 +30,35 @@ var createSuffix = "RETURNING id, " + constructMetadata("", false)
 func unmarshalMetadata(metadataJSON []byte, m *common.Metadata, logger *logger.Logger) error {
 	if metadataJSON != nil {
 		if err := protojson.Unmarshal(metadataJSON, m); err != nil {
-			logger.Error("could not unmarshal metadata", slog.String("error", err.Error()))
+			logger.Error("failed to unmarshal metadata", slog.String("error", err.Error()), slog.String("metadataJSON", string(metadataJSON)))
```
jakedoublev, line 33: Up the chain, `db.StatusifyError` will log any error that occurred throughout policy CRUD before responding to the client request, so it may be better here to wrap these bits in a format string and return it instead of logging multiple times for the same issue.
```suggestion
return fmt.Errorf("failed to unmarshal metadataJSON [%s]: %w", string(metadataJSON), err)
```

Same comment below with the `unmarshalAttributeValue`.

service/policy/db/utils.go
```diff
@@ -27,9 +30,35 @@ var createSuffix = "RETURNING id, " + constructMetadata("", false)
...
+	if attributeValueJSON != nil {
+		if err := protojson.Unmarshal(attributeValueJSON, av); err != nil {
+			logger.Error("failed to unmarshal attribute value", slog.String("error", err.Error()), slog.String("attributeValueJSON", string(attributeValueJSON)))
+			return err
+		}
+	}
+	return nil
+}
+
+func pgtypeUUID(s string) pgtype.UUID {
```
jakedoublev, line 50: 🚀

jakedoublev (dismissed):

jakedoublev:
It looks like the test in `otdfctl` is failing as a result of these changes, but it's a change on the CLI side that's needed. We should do a GET before DELETE for helpful CLI output so the test passes instead of relying on the DELETE db call to populate the full context of what was deleted from the RPC response.

https://github.com/opentdf/otdfctl/blob/c7eafb986a89e14e2baeea84cab68fe748cea51f/cmd/policy-resourceMappings.go#L134-L146
https://github.com/opentdf/otdfctl/blob/main/pkg/handlers/resourceMappings.go#L68

service/policy/db/utils.go
ryanulit: I like that idea and in general I think we should always just return errors and let calls further up the chain handle logging.  Also has the benefit of not having to pass the pointer to the logger instance around everywhere.

jakedoublev (approved):

PR #1443, opened 2024-08-22
fix(core): policy db should use pool connection hook to set search_path

### LINKED ISSUE #1442: Policy: intermittent table not found error where schema in search_path is not used in queries
There are reports of various errors when the db schema is not utilized in SQL queries when doing lookups on policy tables.

jakedoublev:
Resolves #1442

Instead of executing the `SET search_path` on the first connection found when a pool is instantiated, we should utilize the `AfterConnect` [PGX hook](https://github.com/jackc/pgx/blob/4f7e19d67df4d411ac1c19373390021a2f23aa07/pgxtest/pgxtest.go#L37) which exposes a connection immediately after it's opened and before it's utilized.

### DIFF
--- a/service/pkg/db/db.go
+++ b/service/pkg/db/db.go
@@ -158,13 +158,6 @@ func New(ctx context.Context, config Config, logCfg logger.Config, o ...OptsFunc
 		}
 	}
 
-	// Set the Client search_path to the schema
-	q := fmt.Sprintf("SET search_path TO %s", config.Schema)
-	if _, err := c.Pgx.Exec(ctx, q); err != nil {
-		return nil, fmt.Errorf("failed to SET search_path for db Client schema to [%s]: %w", config.Schema, err)
-	}
-
-	slog.Info("successfully set database client search_path", slog.String("schema", config.Schema))
 	return &c, nil
 }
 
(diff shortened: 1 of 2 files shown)


jakedoublev:
Tested these changes under simulated flakiness with:
`while true; do docker restart <postgresql container id>; sleep 30; done`

Every time, I got a log:
`level=INFO msg="successfully set database client search_path" schema=opentdf_policy`

I never had a failed policy LIST attributes call.

Running the same test on `main`, I reproduced the error reported: `time=2024-08-22T16:07:39.361-07:00 level=ERROR msg="ERROR: relation \"attribute_definitions\" does not exist (SQLSTATE 42P01)" error="ERROR: relation \"attribute_definitions\" does not exist (SQLSTATE 42P01)"`

service/pkg/db/db.go
```diff
@@ -185,7 +178,21 @@ func (c Config) buildConfig() (*pgxpool.Config, error) {
...
+		return nil, fmt.Errorf("failed to parse pgx config: %w", err)
+	}
+	// Configure the search_path schema immediately on connection opening
+	parsed.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
+		_, err := conn.Exec(ctx, fmt.Sprintf("SET search_path TO %s", c.Schema))
+		if err != nil {
+			slog.Error("failed to set database client search_path", slog.String("schema", c.Schema), slog.String("error", err.Error()))
+			return err
+		}
+		slog.Info("successfully set database client search_path", slog.String("schema", c.Schema))
```
strantalis, line 192: Should we set this to debug?

strantalis (dismissed):

service/pkg/db/db.go
jakedoublev: Now that we can reliably reproduce and verify it's not an issue, I'm good with that. Good call. Will fix now.

strantalis (approved):

PR #1446, opened 2024-08-23
fix(authz): entitlements fqn casing

### LINKED ISSUE #1359: TDF Policy Data Attributes are case sensitive
TDF Policy Data Attributes seem to be case sensitive and do not get normalized to lower case when doing a keyRewrap, and authorization of getDecision. 

If the TDF Policy Data Attributes is created with cases (example being: `https://example.com/attr/Caps/value/Bad`), the request will get denied even if the user is entitled with `https://example.com/attr/caps/value/bad`.

Steps to Reproduce:

1. Encrypt a TDF with an attribute that contains cases like `https://example.com/attr/Caps/value/Bad`. 
2. Entitle a user with that attribute cases by creating a subject mapping, shouldn't matter I believe, and should be normalized down. Tested with OTDFCTL.
3. Decrypt the TDF and the request hitting Platform has the attributes with cases. (I am on an older version of SDK) so it could be possible that the new version of SDK cleans this up. But server-side should catch this in case a user is doing RPC call rather than using the sdk.

The request should be a denied by Platform.

suchak1:
Closes #1359

### DIFF
--- a/service/authorization/authorization.go
+++ b/service/authorization/authorization.go
@@ -434,7 +434,7 @@ func makeScopeMap(scope *authorization.ResourceAttribute) map[string]bool {
 	scopeMap := make(map[string]bool)
 	// add attribute value FQNs from scope to the map
 	for _, fqn := range scope.GetAttributeValueFqns() {
-		scopeMap[fqn] = true
+		scopeMap[strings.ToLower(fqn)] = true
 	}
 	return scopeMap
 }
--- a/service/authorization/authorization.go
+++ b/service/authorization/authorization.go
@@ -425,8 +425,8 @@ func makeValsByFqnsLookup(attrs []*policy.Attribute, subMapsByVal map[string][]*
 	return fqnAttrVals
 }
 
-// makeScopeMap creates a map of attribute value FQNs.
-func makeScopeMap(scope *authorization.ResourceAttribute) map[string]bool {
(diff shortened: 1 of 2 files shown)


jakedoublev:
✅ from me. I was able to test successfully with `otdfctl` and an encrypt/decrypt flow where the encryption was with a mixed case attribute and there were valid entitlements.

ttschampel (commented):
Can we add a unit test for this case sensitive flow?

service/authorization/authorization.go
```diff
@@ -425,16 +425,16 @@ func makeValsByFqnsLookup(attrs []*policy.Attribute, subMapsByVal map[string][]*
 	return fqnAttrVals
 }
 
-// makeScopeMap creates a map of attribute value FQNs.
-func makeScopeMap(scope *authorization.ResourceAttribute) map[string]bool {
+// MakeScopeMap creates a map of attribute value FQNs.
+func MakeScopeMap(scope *authorization.ResourceAttribute) map[string]bool {
```
strantalis, line 429: @suchak1  is there a reason for this to be public?

suchak1: reverted

jakedoublev (approved):

strantalis (approved):

PR #1466, opened 2024-08-27
fix(core): Add NanoTDF KID padding removal and update logging level

### LINKED ISSUE #1467: NanoTDF KID padding rewrap fix
Different SDKs implement NanoTDF KID differently - make KAS rewrap more robust - trim from and back if not to spec for 2, 8, 32 bytes

pflynn-virtru:
Introduced padding removal logic in the `GetIdentifier` method of `ResourceLocator` to ensure cleaned identifiers. Also changed log level from Info to Debug in `nanoTDFRewrap` for more appropriate logging detail.

resolves #1467

### DIFF
--- a/sdk/resource_locator.go
+++ b/sdk/resource_locator.go
@@ -150,6 +150,7 @@ func (rl *ResourceLocator) setURLWithIdentifier(url string, identifier string) e
 }
 
 // GetIdentifier - identifier is returned if the correct protocol enum is set else error
+// padding is removed unlike rl.identifier direct access
 func (rl ResourceLocator) GetIdentifier() (string, error) {
 	// read the identifier if it exists
 	switch rl.protocol & 0xf0 {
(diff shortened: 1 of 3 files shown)


sdk/resource_locator.go
```diff
@@ -159,7 +160,10 @@ func (rl ResourceLocator) GetIdentifier() (string, error) {
 		if rl.identifier == "" {
 			return "", fmt.Errorf("no resource locator identifer: %d", rl.protocol)
 		}
-		return rl.identifier, nil
+		// remove padding
+		cleanedIdentifier := strings.TrimRight(rl.identifier, "\x00")
+		cleanedIdentifier = strings.TrimLeft(cleanedIdentifier, "\x00")
```
dmihalcik-virtru, line 165: We should probably decide if we want to pad left or right and just do one, but I'm fine with this for now.

dmihalcik-virtru (approved):

sdk/resource_locator.go
pflynn-virtru: agreed.  rewrap is more robust (within reason and with security considerations) and the SDK should be more strict and uniform

PR #1472, opened 2024-08-27
fix(sdk): Fix possible panic for `AttributeValueFQN.Prefix()`

### LINKED ISSUE #1468: Panic possible in `granter` with possible invalid name and value inputs
It was discovered through fuzzing that there are several cases where a `NewAttributeValueFQN` can pass without error but subsequent actions on it could result in a panic.

jentfoo:
This commit adds additional checks to validate that Attribute Name / Values are not empty or contain a slash.  This fixes a panic condition where `NewAttributeValueFQN` would return without error but `Prefix()` would then fail when the prefix is attempted to be parsed by `NewAttributeNameFQN`.

This commit includes unit testing, and the fuzz testing that initially discovered this issue.  See there for possibly dangerous values.

An alternative or additional solution that we may want to consider:
If `Prefix()` is a common or frequent operation we could proactively invoke the `NewAttributeNameFQN` from within `NewAttributeValueFQN`.  That way any future validation added to `NewAttributeNameFQN` would be proactively checked rather than needing `Prefix()` to be invoked to be tested.

Also included are some minor doc changes, spelling fixes, and some other fuzz testing that was co-located with these changes.

PR closes #1468


sdk/granter.go
```diff
@@ -27,16 +27,28 @@ const (
...
+	for i, part := range parts {
+		if part == "" {
+			return fmt.Errorf("%w: empty url path parts are not allowed", ErrInvalid)
+		} else if i > 1 && // skip first two parts as they will have protocol slashes
+			strings.Contains(part, "/") {
+			return fmt.Errorf("%w: slash not allowed in name or values", ErrInvalid)
+		}
+	}
+	return nil
+}
```
jrschumacher, line 50: @jentfoo is the SDK panicking or the platform? 

When we add namespace grants we'll just want to make sure the namespace has a valid policy entry to ensure we can get the appropriate public key. Otherwise it's just a string comparison to derive the foreign key which will resolve to the KAS grant.

Also, we technically prescribe the `https://<ns>/attr/<attr-name>/value/<value-name>` as the format, but there's equally no reason why someone wouldn't want to use urn or similar.

jrschumacher (commented):
Really awesome to see these contributions!! Thank you.

sdk/granter.go
jentfoo: > is the SDK panicking or the platform?

I am still learning the code and how it's invoked.  I labeled the PR as `sdk` because all direct invocations are from within the SDK, but I don't know if this also effects the platform.

jrschumacher: I'll reach out tomorrow and show you our integration tests maybe you can help us identify how to do fuzz testing there 😄

dmihalcik-virtru (approved):

PR #1475, opened 2024-08-27
fix(core): only strip nano kids to the right

### LINKED ISSUE #1506: nanotdf kid: be consistent about padding and length
Strip padding from the right (c-style strings)

dmihalcik-virtru:
- we pad to the right, not the left
- resolves #1506


pflynn-virtru (approved):

dmihalcik-virtru:
@strantalis I need approval from someone with ownership

strantalis (approved):

PR #1512, opened 2024-09-04
feat(sdk): Add namesapce grants to key splitting

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/1516


sdk/granter_test.go
```diff
@@ -552,3 +596,115 @@ func TestReasonerSpecificity(t *testing.T) {
 		})
 	}
 }
+
+func TestReasonerSpecificityWithNamespaces(t *testing.T) {
```
strantalis, line 600: Can we update the table test message to make it clear what is happening and what the expected output should be.

elizabethhealy:
successful xtest run https://github.com/opentdf/tests/actions/runs/10724372814/job/29739869061?pr=189

dmihalcik-virtru (approved):

sdk/granter_test.go
strantalis: @elizabethhealy I should have provided an example but I was thinking like `expect namespace kas when no value or definition grant`

strantalis (changes requested):

sdk/granter_test.go
elizabethhealy: updated

strantalis (approved):

PR #1522, opened 2024-09-05
chore(core): migrate to latest buf schema

dmihalcik-virtru:


elizabethhealy (commented):
can you update the two buf doc gen yamls as well 

https://github.com/opentdf/platform/blob/main/buf.gen.grpc.docs.yaml

https://github.com/opentdf/platform/blob/main/buf.gen.openapi.docs.yaml

elizabethhealy (approved):

PR #1532, opened 2024-09-06
fix(sdk)!: CreateNanoTDF not using functional options

### LINKED ISSUE #1042: `CreateNanoTDF` not using functional options
Originally when we designed `CreateTDF` we decided to leverage the functional options pattern. For consistency we need to follow that patter with NanoTDF as well. 

https://github.com/opentdf/platform/blob/7dea6407322b5e625ee2810dfcf407c010d9996f/sdk/nanotdf.go#L660

sujankota:
Resolves #1042

### DIFF
--- a/examples/cmd/encrypt.go
+++ b/examples/cmd/encrypt.go
@@ -102,18 +102,10 @@ func encrypt(cmd *cobra.Command, args []string) error {
 		}
 		cmd.Println(string(manifestJSON))
 	} else {
-		nanoTDFConfig, err := client.NewNanoTDFConfig()
-		if err != nil {
-			return err
-		}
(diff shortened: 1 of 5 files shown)


strantalis (approved):

dmihalcik-virtru:
IMO we should add a new function, e.g. `CreateNanoTDFOpts()`, instead of modifying the existing function; this will allow us to deprecate the current behavior and upgrade without breaking anyone downstream

dmihalcik-virtru:
See suggestion: https://github.com/opentdf/platform/pull/1534

dmihalcik-virtru:
Let's rethink this as part of a unified *TDF API, that can do both nano and z/base TDF together

PR #1534, opened 2024-09-09
chore(sdk): Only do a minor bump;

dmihalcik-virtru:
deprecate don't delete

### DIFF
--- a/sdk/nanotdf.go
+++ b/sdk/nanotdf.go
@@ -662,12 +662,17 @@ func NewNanoTDFHeaderFromReader(reader io.Reader) (NanoTDFHeader, uint32, error)
 // ============================================================================================================
 
 // CreateNanoTDF - reads plain text from the given reader and saves it to the writer, subject to the given options
-func (s SDK) CreateNanoTDF(writer io.Writer, reader io.Reader, opts ...NanoTDFOption) (uint32, error) {
+func (s SDK) CreateNanoTDFOptions(writer io.Writer, reader io.Reader, opts ...NanoTDFOption) (uint32, error) {
 	config, err := newNanoTDFConfig(opts...)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


jrschumacher:
@dmihalcik-virtru this isn't great DX. If we don't want to use `client.CreateNanoTDF()` then we should use `client.CreateTDF()` and support all options through abstractions. I believe this will satisfy your concern about deprecation and solve for the specific need for a uniform and developer friendly SDK.

Here are some stories for **encrypt**:

- As a developer, I want to use only one function to create TDFs. I don't know what ZTDF and NanoTDF is, I just want a TDF and I want the intelligence of the SDK to give me the best TDF. 
- As a developer, I know that I am building for optimization and I want to use NanoTDF, but I want to make sure that I get an error if I try to add an option that is invalid for NanoTDF.
- As a developer, I know that I am building for richness of data and I want to use ZTDF, but I want to make sure that I get an error if I try and add an option that is invalid for ZTDF.

Equally, here are some stories for **decrypt**:

- As a developer, I have a TDF and I want it decrypted. I don't care what the format is, I just want what's inside.
- As a developer, I am streaming large data and I know that this must be ZTDF. I want a stream output that I can control.

dmihalcik-virtru:
yeah, we need to let CreateTDF take a Reader instead of a SeekableReader, too

dmihalcik-virtru:
Also, additional thoughts/use cases:

1. During decrypt, I want to only decrypt a specific set of bytes (random access decrypt)
2. I want to do a lot of similar encrypt (or decrypt) operations and want to shave performance down a lot - e.g. maybe building the options themselves becomes a bottleneck if we aren't careful

dmihalcik-virtru:
I'll put together some thoughts about how best to design these APIs

jrschumacher:
Sounds good. Do you think it's fair to push this issue back to todo to get better grooming so we can identify all the usecases we want to cover?

Issue #952, opened 2024-06-07
Support per component logging configuration


ttschampel:
It would be nice to support per component log levels (e.g. like [istio](https://istio.io/latest/docs/ops/diagnostic-tools/component-logging/)) so a user can target debug to specific areas of interest.

Also, a log format that contains the component would be more informative when reading/debugging the logs

strantalis:
@ttschampel Part of this issue is solved but I don't think it has been fully rolled out yet. We are currently create a new child logger with the namespace name of the service. Maybe namespace should be changed to something else because we use that for policy as well. 

https://github.com/opentdf/platform/blob/main/service/pkg/server/services.go#L137

cc: @pbacon-blaber

PR #1537, opened 2024-09-10
feat(core): Ability to add namespace level loggers

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/952
behavior: if a proper log_level is provided within a ns config, it will create a new logger for that ns using that config and the ns will use that logger instead of the default. if no log_level is provided for the ns or the log_level is not valid, then it will use the default logger

### DIFF
--- a/opentdf-dev.yaml
+++ b/opentdf-dev.yaml
@@ -22,6 +22,10 @@ services:
         alg: rsa:2048
         legacy: true
   entityresolution:
+    logger:
+      level: info
+      type: text
+      output: stdout
     url: http://localhost:8888/auth
     clientid: 'tdf-entity-resolution'
     clientsecret: 'secret'
--- a/opentdf-dev.yaml
+++ b/opentdf-dev.yaml
@@ -22,10 +22,7 @@ services:
         alg: rsa:2048
         legacy: true
   entityresolution:
-    logger:
-      level: info
-      type: text
-      output: stdout
+    log_level: info
     url: http://localhost:8888/auth
     clientid: 'tdf-entity-resolution'
     clientsecret: 'secret'
--- a/service/pkg/server/services.go
+++ b/service/pkg/server/services.go
@@ -121,7 +121,7 @@ func startServices(ctx context.Context, cfg config.Config, otdf *server.OpenTDFS
 			continue
 		}
 
-		var svcLogger *logging.Logger
+		var svcLogger *logging.Logger = logger.With("namespace", ns)
 		extractedLogLevel, err := tryExtractingServiceLoggerConfig(cfg.Services[ns])
 
 		// If ns has log_level in config, create new logger with that level
@@ -136,8 +136,6 @@ func startServices(ctx context.Context, cfg config.Config, otdf *server.OpenTDFS
 					svcLogger = newSvcLogger.With("namespace", ns)
 				}
 			}
-		} else {
-			svcLogger = logger.With("namespace", ns)
 		}
 
 		var svcDBClient *db.Client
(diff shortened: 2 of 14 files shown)


service/pkg/server/services.go
```diff
@@ -136,9 +139,19 @@ func startServices(ctx context.Context, cfg config.Config, otdf *server.OpenTDFS
...
 			}
 
+			// If the service provides its own logging config, create a new logger for that service using that config
+			if extractedLogger, err := tryExtractingServiceLoggerConfig(cfg.Services[svc.Namespace]); err == nil {
+				slog.Debug("configuring logger")
+				newSvcLogger, err := logging.NewLogger(extractedLogger)
+				if err != nil {
+					return fmt.Errorf("could not start logger: %w", err)
+				}
+				thisSvcLogger = newSvcLogger.With("namespace", ns)
```
dmihalcik-virtru, line 149: My understanding was that we'd want a `service` field that was the svc.Namespace value for all services, not just ones that have a custom configuration

elizabethhealy: theres a line above that mimics this for the default service logger https://github.com/opentdf/platform/blob/8240645270e1cd25001c6f40d4276093c6e3cb23/service/pkg/server/services.go#L123
the line here just does the same but for the new service specific logger

dmihalcik-virtru:
Would this make more sense to be configurable at the handler level? I'm fine with this approach, but by using separate loggers entirely, instead of `With` to generate a logger chain, seems like an anti-pattern but I can't quite figure out why

opentdf-dev.yaml
```diff
@@ -22,6 +22,10 @@ services:
         alg: rsa:2048
         legacy: true
   entityresolution:
+    logger:
+      level: info
+      type: text
```
dmihalcik-virtru, line 27: Do we want to customize anything other than level?

elizabethhealy:
@dmihalcik-virtru but when using With() the resulting logger uses the same handler, so if the default log level/type is different from a services configuration we'd have to create a new handler anyway (even if we only wanted to change the level, slog doesnt have a way to edit the loggers handler config so we have to make a new one), also slog doesn't give you the ability to set the handler of a logger you need to create a new one for the new handler

opentdf-dev.yaml
strantalis: @elizabethhealy @dmihalcik-virtru 
I think maybe to start we only expose `level`. I don't think we have a use case for different logger types or outputs. 

Can we just simplify this field to `log_level` maybe?

service/pkg/server/services.go
strantalis: I think this logic might need to be defined outside the service loop because for example `policy` contains N number of services.

- attributes
- namespaces,
- subjectmappings
- ...

service/pkg/server/services.go
```diff
@@ -179,10 +196,22 @@ func startServices(ctx context.Context, cfg config.Config, otdf *server.OpenTDFS
 	return nil
 }
 
+func tryExtractingServiceLoggerConfig(cfg serviceregistry.ServiceConfig) (string, error) {
```
dmihalcik-virtru, line 199:
```suggestion
func extractServiceLoggerConfig(cfg serviceregistry.ServiceConfig) (string, error) {
```
image

service/pkg/server/services.go
```diff
@@ -120,11 +121,27 @@ func startServices(ctx context.Context, cfg config.Config, otdf *server.OpenTDFS
 			continue
 		}
 
-		svcLogger := logger.With("namespace", ns)
+		var svcLogger *logging.Logger
+		extractedLogLevel, err := tryExtractingServiceLoggerConfig(cfg.Services[ns])
```
dmihalcik-virtru, line 125:
```suggestion
		extractedLogLevel, err := extractServiceLoggerConfig(cfg.Services[ns])
```

dmihalcik-virtru (approved):

dmihalcik-virtru (approved):

PR #1540, opened 2024-09-17
feat(ci): Get traffic metrics for OpenTDF repos 

cassandrabailey293:
this PR runs a workflow once a week to get Github repo traffic metrics for each of the opentdf public repos. it puts it in a gcp bucket where bigquery can be integrated. Goal is to be able to visualize the data over time to see views, referral sources, clones, etc. can be expanded in the future if we want any more info. 

https://github.com/opentdf/platform/actions/runs/10904129901
^ this was a successful run of the workflow.

### DIFF
--- a/.github/scripts/metrics.js
+++ b/.github/scripts/metrics.js
@@ -1,0 +1,49 @@
+async function getMetrics(github) {   
+    // Views -- last 14 days
+    const views = await github.rest.repos.getViews({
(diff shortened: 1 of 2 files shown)


cassandrabailey293:
fixes https://github.com/opentdf/platform/issues/1541

elizabethhealy (approved):

PR #1565, opened 2024-09-20
DO NOT SUBMIT: Collect manifest size stat and some code to plot it

dmihalcik-virtru:


dmihalcik-virtru:
Useful and used to determine that 10MiB is more than enough space for 187 GiB files at 2 meg segments

Issue #1510, opened 2024-09-04
CI: pipeline should skip most tests if only updates were `.md` files


jakedoublev:
### Background

The pipeline currently runs Go tests on every module, integration tests, xtests, build tests, etc if any files were changed. However, changes to documentation alone should not trigger these tests.

### Acceptance Criteria

1. add a [paths-ignore](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) directive to the [checks.yaml](https://github.com/opentdf/platform/blob/main/.github/workflows/checks.yaml) to skip `**.*.md`

jakedoublev:
related to #5448

PR #1582, opened 2024-09-26
refactor(sdk): Update to newer deps

dmihalcik-virtru:
Fixes https://github.com/opentdf/platform/issues/1584

PR #1587, opened 2024-09-30
fix(ci): Don't run tests for markdown changes

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/1510

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -25,10 +25,12 @@ on:
       - main
     types:
       - checks_requested
-    paths-ignore:
-      - '**/*.md'
-      - LICENSE
-      - CODEOWNERS
+    paths:
+      - '**'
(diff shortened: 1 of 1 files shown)


elizabethhealy:
https://github.com/opentdf/platform/pull/1582 fixes the ci vuln failure

dmihalcik-virtru (commented):
The `/docs` file is generated code so probably *should* be checked, at least to make sure that has not been manually modified, but maybe only on merge_group events would be fine

Perhaps something like:

```yaml
   paths:
      - '**'
      - '!**/*.md'
      - 'docs/**'
     - !LICENSE
     - !CODEOWNERS
```

See: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

dmihalcik-virtru (approved):

PR #1588, opened 2024-09-30
feat(policy): 1552 implement recommended changes to audit process

ryanulit:


service/logger/audit/logger_test.go
```diff
@@ -84,13 +84,6 @@ var policyCRUDParams = PolicyEventParams{
 	ActionType: ActionTypeUpdate,
 	ObjectID:   "test-object-id",
 	ObjectType: ObjectTypeKeyObject,
-
```
ryanulit, line 87: Removing the diff logic test here because it is actually created by code in the policy.go file, so these tests don't even need to test that logic

service/logger/audit/policy.go
```diff
@@ -57,15 +47,61 @@ func CreatePolicyEvent(ctx context.Context, isSuccess bool, params PolicyEventPa
...
 		},
 		RequestID: auditDataFromContext.RequestID,
 		Timestamp: time.Now().Format(time.RFC3339),
-	}, nil
+	}
+
+	if params.Original != nil {
+		auditEventOriginal, err := marshallProtoToAuditObject(params.Original)
+		if err != nil {
+			return nil, err
```
jakedoublev, line 62:
```suggestion
			return nil, fmt.Errorf("failed to marshal original audit object: %w", err)
```

service/logger/audit/policy.go
```diff
@@ -57,15 +47,61 @@ func CreatePolicyEvent(ctx context.Context, isSuccess bool, params PolicyEventPa
...
+		}
+		auditEvent.Original = auditEventOriginal
+
+		if params.Updated != nil {
+			auditEventUpdated, err := marshallProtoToAuditObject(params.Updated)
+			if err != nil {
+				return nil, err
```
jakedoublev, line 69:
```suggestion
			return nil, fmt.Errorf("failed to marshal updated audit object: %w", err)
```

service/logger/audit/policy.go
ryanulit: Done

service/logger/audit/policy.go
ryanulit: Done

jakedoublev (approved):

Issue #1358, opened 2024-08-15
Policy: KAS registry public key `local` field should be reserved


jakedoublev:
Once https://github.com/opentdf/platform/pull/1347 is merged, we should reserve the unused `local` field that has not been respected to date and is now replaced with the cached key work in #1347 supporting key ids.

PR #1590, opened 2024-10-01
feat(policy): DSP-51 - deprecate PublicKey local field

suchak1:
resolves: https://github.com/opentdf/platform/issues/1358


service/policy/objects.proto
```diff
@@ -308,11 +308,6 @@ message PublicKey {
       expression: "this.matches('^https://[a-zA-Z0-9]([a-zA-Z0-9\\\\-]{0,61}[a-zA-Z0-9])?(\\\\.[a-zA-Z0-9]([a-zA-Z0-9\\\\-]{0,61}[a-zA-Z0-9])?)*(/.*)?$')"
     }];
 
-    // public key; PEM of RSA public key; prefer `cached`
```
dmihalcik-virtru, line 311: You'll need to reserve this number:

https://protobuf.dev/programming-guides/proto3/#reserved-field-numbers

dmihalcik-virtru (changes requested):

suchak1:
reserved fields are not supported in oneOf
https://github.com/protocolbuffers/protobuf/issues/5338
https://github.com/protobufjs/protobuf.js/issues/1008

service/policy/objects.proto
suchak1: @dmihalcik-virtru Could you take another look at this?

dmihalcik-virtru (dismissed):

service/integration/attribute_fqns_test.go
```diff
@@ -244,8 +244,14 @@ func (s *AttributeFqnSuite) TestGetAttributeByFqn_WithKeyAccessGrants_Definition
 	localKAS, err := s.db.PolicyClient.CreateKeyAccessServer(s.ctx, &kasregistry.CreateKeyAccessServerRequest{
 		Uri: "https://example.org/kas2",
 		PublicKey: &policy.PublicKey{
```
jakedoublev, line 246: This looks great, @suchak1. Can you please add assertions to 1-2 tests in this suite and those in `kas_registry_test` that the key found in the KAS or Grant was the same as the 'cached' key used to create?

It looks like we had assertions around those for the local keys previously but none covering assertion that a KAS created with a cached key finds the same cached key on lookup (which tests our unmarshaling/marshaling of the proto to and from the db is sound).

suchak1: added assertions to both test files

jakedoublev (approved):

strantalis (approved):

PR #1600, opened 2024-10-01
feat(core): Unify NanoTDF and ZTDF APIs

sujankota:


strantalis:
There has been discussion around aligning nanotdf more with the tdf spec. I wonder if we should wait for that before trying to unify the two into a single interface.

PR #1601, opened 2024-10-02
feat(core): Add connection to optional remote ers

elizabethhealy:


service/pkg/server/start.go
```diff
@@ -197,6 +207,17 @@ func Start(f ...StartOptions) error {
 	return nil
 }
 
+func getNewGRPCConn(url string) (*grpc.ClientConn, error) {
+	defaultOptions := []grpc.DialOption{
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+	}
```
elizabethhealy, line 213: TODO: support dial options in the yaml config

opentdf-example-remote-ers.yaml
```diff
@@ -0,0 +1,111 @@
+mode: all
+modeoptions:
+  remoteERSUrl: http://localhost:8181
```
strantalis, line 3: What are your thoughts if we moved this configuration underneath `SDKConfig`. Trying to think how we could layer in other options for our internal products as well. 

https://github.com/opentdf/platform/blob/c36624cbefd8643cad6f01603046fa060ae06724/service/internal/config/config.go#L45-L58

Issue #1566, opened 2024-09-20
difficulty parsing audit logs


strantalis:
When we log audit logs they are being logged as a string in the msg field. We should move to logging the audit object as attributes with slog. 

https://github.com/opentdf/platform/blob/main/service/logger/audit/logger.go#L78

https://github.com/opentdf/platform/blob/main/service/logger/audit/logger.go#L101

https://github.com/opentdf/platform/blob/main/service/logger/audit/logger.go#L116

This might require us to update the audit struct to implement the LogValuer interface here.

https://github.com/opentdf/platform/blob/main/service/logger/audit/utils.go#L1

Acceptance Criteria:
- Audit logs should be parsable if the platform is logging in json / text format.

jakedoublev:
Not caused by, but related to #1552

PR #1612, opened 2024-10-04
fix(core): log audit object as json

sujankota:
Fixes https://github.com/opentdf/platform/issues/1566


service/logger/audit/utils.go
```diff
@@ -35,39 +50,81 @@ type auditEventObject struct {
...
 	Assertions  []string `json:"assertions,omitempty"`
 	Attrs       []string `json:"attrs,omitempty"`
 	Permissions []string `json:"permissions,omitempty"`
 }
 
+func (e eventObjectAttributes) LogValue() slog.Value {
+	return slog.GroupValue(
+		slog.String("assertions", strings.Join(e.Assertions, ",")),
+		slog.String("attrs", strings.Join(e.Attrs, ",")),
+		slog.String("permissions", strings.Join(e.Permissions, ",")))
```
dmihalcik-virtru, line 72: This won't escape the strings (e.g. commas will cause problems if we want to parse it back). Since the String valuer will escape the quotes, this shouldn't be too much of an issue, but maybe you will be better served with

```suggestion
		slog.Any("assertions", e.Assertions),
		slog.Any("attrs", e.Attrs),
		slog.Any("permissions", e.Permissions))
```

dmihalcik-virtru (approved):

service/logger/audit/logger.go
```diff
@@ -93,12 +87,7 @@ func (a *Logger) policyCrudBase(ctx context.Context, isSuccess bool, eventParams
 		return
 	}
 
-	auditEventJSONBytes, err := json.Marshal(auditEvent)
-	if err != nil {
-		a.logger.ErrorContext(ctx, "error marshalling policy attribute audit event", "err", err)
-	}
-
-	a.logger.Log(ctx, LevelAudit, string(auditEventJSONBytes))
+	a.logger.Log(ctx, LevelAudit, "policy curd", "audit", *auditEvent)
```
dmihalcik-virtru, line 90:
```suggestion
	a.logger.Log(ctx, LevelAudit, "policy crud", "audit", *auditEvent)
```

dmihalcik-virtru (approved):

dmihalcik-virtru:
It looks like you'll need to update `logger_test.go`

pflynn-virtru (approved):

strantalis (approved):

Issue #1564, opened 2024-09-20
Create custom errors for assertion failures, root signature failure, and kas rewrap failures


elizabethhealy:
expose errors in sdk so users are able to catch specific failures

PR #1613, opened 2024-10-07
feat(sdk): Expose error types for signature, assertion, and kas failures

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/1564

### DIFF
--- a/sdk/tdf.go
+++ b/sdk/tdf.go
@@ -105,7 +105,7 @@ func (s SDK) CreateTDFContext(ctx context.Context, writer io.Writer, reader io.R
 	}
 
 	if inputSize > maxFileSizeSupported {
-		return nil, ErrFileTooLarge
+		return nil, errFileTooLarge
 	}
 
 	_, err = reader.Seek(0, io.SeekStart)
@@ -323,7 +323,7 @@ func (r *Reader) Manifest() Manifest {
 func (s SDK) prepareManifest(ctx context.Context, t *TDFObject, tdfConfig TDFConfig) error { //nolint:funlen,gocognit // Better readability keeping it as is
 	manifest := Manifest{}
 	if len(tdfConfig.splitPlan) == 0 && len(tdfConfig.kasInfoList) == 0 {
-		return fmt.Errorf("%w: no key access template specified or inferred", ErrInvalidKasInfo)
+		return fmt.Errorf("%w: no key access template specified or inferred", errInvalidKasInfo)
 	}
 
 	manifest.EncryptionInformation.KeyAccessType = kSplitKeyType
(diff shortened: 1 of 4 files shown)


sdk/tdferrors.go
```diff
@@ -0,0 +1,37 @@
...
+	ErrRootSignatureFailure    = fmt.Errorf("[%w] tdf: issue verifying root signature", ErrTampered)
+)
+
+// Custom error struct for Assertion errors
+type ErrAssertionFailure struct { //nolint:errname // match naming of existing errors
+	ID string
+}
+
+func (e ErrAssertionFailure) Error() string {
+	return fmt.Errorf("[%w] tdf: issue verifying assertions, id: %s", ErrTampered, e.ID).Error()
```
strantalis, line 32: @elizabethhealy 
An assertion failure will not always indicate the assertion has been modified. There is the case where a pep or user passes in the wrong key when verifying assertions. Could you see if there a way to distinguish those two types of errors.

sdk/tdf.go
```diff
@@ -588,7 +576,7 @@ func (r *Reader) WriteTo(writer io.Writer) (int64, error) {
 		}
 
 		if int64(len(readBuf)) != seg.EncryptedSize {
-			return totalBytes, errTDFReaderFailed
+			return totalBytes, ErrTDFReaderFailed
```
strantalis, line 579: I think maybe we need an encrypted segment size mismatch here and in a few other places.

sdk/tdferrors.go
```diff
@@ -0,0 +1,37 @@
...
+	errFileTooLarge     = errors.New("tdf: can't create tdf larger than 64gb")
+	errWriteFailed      = errors.New("tdf: io.writer fail to write all bytes")
+	errInvalidKasInfo   = errors.New("tdf: kas information is missing")
+	errKasPubKeyMissing = errors.New("tdf: kas public key is missing")
+	errRewrapForbidden  = errors.New("tdf: rewrap request 403")
+	// Exposed tamper detection errors, Catch all possible tamper errors with errors.Is(ErrTampered)
+	ErrTampered                = errors.New("tamper detected")
+	ErrRootSigValidation       = fmt.Errorf("[%w] tdf: failed integrity check on root signature", ErrTampered)
+	ErrSegSizeMismatch         = fmt.Errorf("[%w] tdf: mismatch encrypted segment size in manifest", ErrTampered)
+	ErrTDFReaderFailed         = fmt.Errorf("[%w] tdf: fail to read bytes from TDFReader", ErrTampered)
```
strantalis, line 18: I don't think a read failure would indicate tampering.

sdk/tdferrors.go
```diff
@@ -0,0 +1,37 @@
...
+	ErrSegSigValidation        = fmt.Errorf("[%w] tdf: failed integrity check on segment hash", ErrTampered)
+	ErrTDFPayloadReadFail      = fmt.Errorf("[%w] tdf: fail to read payload from tdf", ErrTampered)
```
strantalis, line 20: Same here not sure if this indicates tampering.

sdk/tdferrors.go
elizabethhealy: ya i think this one was because of the way it was used when the segment size didnt match the buf len (see above comment) but if we switch those to ErrSegSizeMismatch as suggested then this can be removed from the tamper errors

sdk/tdferrors.go
elizabethhealy: i forget the reasoning for this one, @dmihalcik-virtru ?

dmihalcik-virtru: These indicate that the segment pointers are invalid, e.g. out of sequence, overlapping, or out of range. Likely indicative of a corrupt file in some fashion, although I suppose there could be some technically 'readable' files that have this property, they will not be a valid TDF file.

sujankota (approved):

PR #1619, opened 2024-10-07
feat(policy): 1618 update KAS CRUD to align with ADR decisions

### LINKED ISSUE #1618: Policy: Update KAS service CRUD to align with ADR decisions
Update KAS CRUD logic in service and DB layers to adhere to the decisions made in https://github.com/opentdf/platform/blob/main/service/policy/adr/0000-current-state.md

ryanulit:
fix #1618


jakedoublev (approved):
🚀

PR #1622, opened 2024-10-08
feat(policy): 1580 Resource Mappings GET/LIST should provide attribute value FQNs in response

### LINKED ISSUE #1580: Policy: resource mappings GET/LIST should provide attribute value FQNs in response
### Background

LIST of resource mappings is a known service to service API for PDPs and should therefore provide the FQN in response of mapped attribute values to avoid PEPs/PDPs receiving the value id and having to do a roundtrip.

### Acceptance Criteria

1. resource mappings return fqns of attribute values they map to in response
2. integration tests

ryanulit:
fix #1580


jakedoublev (approved):
lgtm

PR #1626, opened 2024-10-09
fix: Simplify request ID extraction from context for AUDIT

### LINKED ISSUE #1489: `requestID` missing on audit logs
In order for us to trace audit events through the system we add a `requestID` field for easier traceability. I noticed while debugging that some audit events have a zeroed out id. 

```json
"requestId": "00000000-0000-0000-0000-000000000000",
```

Acceptance Criteria:
- Make sure audit events have a an id and are being propagated through the system
- Tests are written to validate expected output of the `requestID` field

pflynn-virtru:
Refactored the code to directly fetch the request ID as a UUID type instead of first parsing it from a string. This change eliminates unnecessary error handling and simplifies logic for extracting the request ID from the context.z

Resolves #1489

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -68,7 +68,7 @@ linters-settings:
             reason: "gofrs' package is not go module"
   gosec:
     excludes:
-      - G115 # integer overflow protection; 
+      - G115 # integer overflow protection;
 
   govet:
     # Enable all analyzers.
@@ -138,7 +138,7 @@ linters:
     - errname # checks that sentinel errors are prefixed with the Err and error types are suffixed with the Error
     - errorlint # finds code that will cause problems with the error wrapping scheme introduced in Go 1.13
     - exhaustive # checks exhaustiveness of enum switch statements
-    - exportloopref # checks for pointers to enclosing loop variables
+    # - exportloopref # checks for pointers to enclosing loop variables Since Go1.22 (loopvar) this linter is no longer relevant. Replaced by copyloopvar."
     # - fatcontext
     - forbidigo # forbids identifiers
     - forcetypeassert # finds forced type assertions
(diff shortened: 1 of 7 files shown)


pbacon-blaber (approved):
LGTM pending tests

strantalis (approved):

.golangci.yaml
```diff
@@ -138,7 +138,7 @@ linters:
     - errname # checks that sentinel errors are prefixed with the Err and error types are suffixed with the Error
     - errorlint # finds code that will cause problems with the error wrapping scheme introduced in Go 1.13
     - exhaustive # checks exhaustiveness of enum switch statements
-    - exportloopref # checks for pointers to enclosing loop variables
```
dmihalcik-virtru, line 141: We still target 1.21, so this should still be enabled

service/logger/audit/utils.go
```diff
@@ -68,14 +68,12 @@ type ContextData struct {
 	ActorID   string
 }
 
-// Gets relevant audit data from the context object.
+// GetAuditDataFromContext Gets relevant audit data from the context object.
```
dmihalcik-virtru, line 71:
```suggestion
// GetAuditDataFromContext gets relevant audit data from the context object
```

dmihalcik-virtru (changes requested):

.golangci.yaml
pflynn-virtru: Ci is `1.22` so the lint setup fails.  `copyloopvar` is commented out because it requires `1.22`.  
Two options:
1. uncomment `copyloopvar`
2. move CI to 1.21

dmihalcik-virtru (approved):

Issue #788, opened 2024-05-13
Support claims entity type in opa/ERS


elizabethhealy:
Enable the claims entity type in opa. It should not call ERS when the entity type is claims but rather should evaluate the subject mappings on the claims themselves. (I believe this is the expected behavior for this flow.)
It should be assumed claims is a type that can be parsed/queried by jq.

elizabethhealy:
should help solve https://github.com/opentdf/platform/issues/785

jrschumacher:
> It should be assumed claims is a type that can be parsed/queried by jq.
Do we need full jq support, or could we keep it simple and only support object dot-notation? I have no opinion if we use a jq compliant library under the hood, but limiting what we advertise will reduce complexity and cost less to maintain.

elizabethhealy:
@jrschumacher im fine either way, i think dot notation is probably simpler and easier to support, or maybe some variation of dot notation that also supports like like item1.item2[*]

jrschumacher:
@elizabethhealy is this completed or should we reopen?

elizabethhealy:
we should reopen, i think there was another ticket issue linked to this as well, ill find it

elizabethhealy:
relates to https://github.com/opentdf/platform/issues/785

Issue #785, opened 2024-05-13
Enable Authorization based on OIDC claims alone without the need for Entity Resolution Service


jrschumacher:
The addition of Entity Resolution Service offers the ability to fetch custom data from an IdP or some other source when making an Authorization request. The challenge with this is that it increases the necessary requirements to get OpenTDF up and running.

Currently, ERS only supports Keycloak and any additional needs will need to be developed and deployed apart from the platform. OpenTDF has no plans to add or maintain support for other IdP or data sources. 

This enhancement focuses on reducing the complexity of starting the OpenTDF service by supporting any IdP (that meets our requirements) without any custom code as well as reducing any custom integration with Keycloak which requires an API key to fetch additional data.

## Acceptance Criteria

- should be able to specify a token property to validate
- should fail authorization if property is not found in token

strantalis:
Had this written this morning but never hit create.

https://github.com/opentdf/platform/issues/793

pflynn-virtru:
The TDF specification addresses this with  `tdf_claims`.  See https://github.com/opentdf/spec/blob/2a95f6f434ae241df1d2371b33c2b3c564e5ee67/protocol/README.md?plain=1#L15

Would this functionality address this issue?  (Note this is v1 behavior and will need to be ported to v2)

PR #1630, opened 2024-10-10
feat(authz): JWT ERS that just returns claims

elizabethhealy:
resolves https://github.com/opentdf/platform/issues/788
resolves https://github.com/opentdf/platform/issues/785

Simple entity resolution service that just looks at claims

CreateEntityChainFromJwt will return an entity of type claims with the claims being the contents of the jwt
ResolveEntities returns the entity representation if not type claims, and if type claims then it just returns the claims

Mode in the config to dictate which ERS to run -- default is keycloak


pflynn-virtru (approved):

service/entityresolution/claims/claims_entity_resolution.go
```diff
@@ -0,0 +1,141 @@
...
+	// Wrap the struct in an *anypb.Any message
+	anyClaims, err := anypb.New(structClaims)
+	if err != nil {
+		return nil, fmt.Errorf("error wrapping in anypb.Any: %w", err)
+	}
+
+	entities = append(entities, &authorization.Entity{
+		EntityType: &authorization.Entity_Claims{Claims: anyClaims},
+		Id:         "jwtentity-claims",
+		Category:   authorization.Entity_CATEGORY_SUBJECT,
```
elizabethhealy, line 125: entity in chain is a subject, should we also include an environmental entity? (similar to how we do client id in keycloak ers?)

PR #1631, opened 2024-10-11
chore(core): Use buf format on .protos

dmihalcik-virtru:


dmihalcik-virtru:
I'm not sure if this is a good idea, since it `buf format` produces different results from running the builtin (i guess?) format command in vscode.

PR #1634, opened 2024-10-14
fix: allow standard users to get authorization decisions

mkleene:
Lets `standard` users retrieve authorization decisions. [Does not allow these users to retrieve entitlements](https://github.com/opentdf/platform/blob/e3ed0b5ce6039000c9e3c574d3d6ce2931781235/service/authorization/authorization.proto#L289-L305). Implements discussion from https://github.com/virtru-corp/data-security-platform/pull/403.

Resolves https://github.com/opentdf/platform/issues/1645.

### DIFF
--- a/service/internal/auth/casbin.go
+++ b/service/internal/auth/casbin.go
@@ -89,6 +89,8 @@ p,	role:admin,		/v1/token/authorization,						  				        *,      allow
 p,	role:standard,		policy.*,																read,			allow
 p,	role:standard,		kasregistry.*,													read,			allow
 p,	role:standard,    kas.AccessService/Rewrap, 			           *,			allow
+p,      role:standard,          kas.AuthorizationService/GetDecisions,             *,                   allow
+p,      role:standard,          kas.AuthorizationService/GetDecisionsByToken,      *,                   allow
 ## HTTP routes
 p,	role:standard,		/attributes*,														read,			allow
 p,	role:standard,		/namespaces*,														read,			allow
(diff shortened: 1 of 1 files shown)


jakedoublev:
Is this a blocker for you @mkleene? There's been recent discussion about adding a Casbin role specific to services that are privileged and distinct from the `standard` user role to avoid opening up some routes too widely.

mkleene:
> Is this a blocker for you @mkleene? There's been recent discussion about adding a Casbin role specific to services that are privileged and distinct from the `standard` user role to avoid opening up some routes too widely.

Yes, it is. Gateway's decrypt mode needs to be able to determine if a user has access to content.

mkleene:
> Is this a blocker for you @mkleene? There's been recent discussion about adding a Casbin role specific to services that are privileged and distinct from the `standard` user role to avoid opening up some routes too widely.

Although I think that this agrees with @jrschumacher 's thoughts here: https://github.com/virtru-corp/data-security-platform/pull/403, where `GetEntitlements` is a bit more locked down but these two methods are not.

pflynn-virtru (approved):

service/internal/auth/casbin.go
```diff
@@ -89,6 +89,8 @@ p,	role:admin,		/v1/token/authorization,						  				        *,      allow
 p,	role:standard,		policy.*,																read,			allow
 p,	role:standard,		kasregistry.*,													read,			allow
 p,	role:standard,    kas.AccessService/Rewrap, 			           *,			allow
+p,      role:standard,          kas.AuthorizationService/GetDecisions,             *,                   allow
+p,      role:standard,          kas.AuthorizationService/GetDecisionsByToken,      *,                   allow
```
jrschumacher, line 93:
```suggestion
p,      role:standard,          kas.AuthorizationService/GetDecisions,             read,                   allow
p,      role:standard,          kas.AuthorizationService/GetDecisionsByToken,      read,                   allow
```

jrschumacher: @mkleene just aligning it to the `Get` prefix to be scoped in case there is ever introduction of `SetDecision` we don't accidentally expose it.

jrschumacher (approved):

jrschumacher (approved):

PR #1635, opened 2024-10-14
fix(authz): Move logs containing subject mappings to trace level

### LINKED ISSUE #1503: Authorization service is logging the subject mappings during an AccessPDP check
There is somewhere in the code that is logging all subject mappings when log level is set to debug. Debug log level should still be performant even when debugging an issue. With the log being so large, this could potentially decrease performance as subject mappings are added.

Find in the code where this log statement is taking place and remove it or change it to trace level log.

## Acceptance Criteria:

- Remove the excessive logging of subject mappings

elizabethhealy:
resolves #1503

### DIFF
--- a/service/internal/access/pdp.go
+++ b/service/internal/access/pdp.go
@@ -43,7 +43,7 @@ func (pdp *Pdp) DetermineAccess(
 	}
 
 	// Unlike with Values, there should only be *one* Attribute Definition per FQN (e.g "https://namespace.org/attr/MyAttr")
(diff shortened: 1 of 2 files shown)


jakedoublev:
Thanks for this, @elizabethhealy. Is there capacity to check if this is really a warning/error case or something we should bypass entirely?

https://github.com/opentdf/platform/pull/1635/files#diff-a1a2a7e3fed8bf6e543f348e2130da8b24f6ceb39cddc98f0aabc1f17906f818R570-R573

pflynn-virtru (approved):

PR #1636, opened 2024-10-14
feat(policy): add optional name field to registered KASes in policy

### LINKED ISSUE #153: Consider Adding a "name", or "commonName" column to the "KeyAccessServer" Table for more uniform UX
Currently, our KeyAccessServer schema consists of the following columns:

- uuid | "id"
- varchar | "uri"
- jsonb | "public_key"
- jsonb | "metadata"

And the proposed method for adding a "common name" for the KAS server being registered is to add a metadata object in the form of:
```
{
 "labels": {"name": "SOME_NAME" }
}
```

However, my concern is that this will lead to messy, and unreliable code, when attempting to display the KAS Name in say, a dropdown list or table in the browser (or other UI), as the name label, is not guarenteed to be present on the DB record. 

The solution would be to add a varchar "name" column to the "KeyAccessServer" table, so that we, as well as any future developers working on the DSP, can always rely on that field being present, and the customer has a simple way to differentiate between their registered KeyServers (without having to memorize the UUIDs)

jakedoublev:
Partially resolves #153


service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -24,6 +24,17 @@ message CreateKeyAccessServerRequest {
   string uri = 1 [(buf.validate.field).required = true];
   PublicKey public_key = 2 [(buf.validate.field).required = true];
 
+  // Optional
+  string name = 20 [
```
ryanulit, line 28: Would be good to add unit tests for these validations if time permits

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -71,10 +91,11 @@ message KeyAccessServerGrants {
 // GET request to the specific policy object.
 message ListKeyAccessServerGrantsRequest {
   // Optional
-  // Filter LIST by either ID or URI of a registered Key Access Server.
-  // If neither is provided, grants from all registered KASs to policy attribute objects are returned.
+  // Filter LIST by either ID, URI, or common name of a registered Key Access Server.
+  // If none are provided, grants from all registered KASs to policy attribute objects are returned.
   string kas_id = 1;
   string kas_uri = 2;
+  string kas_name = 3;
```
ryanulit, line 98: Assuming a follow-up PR will be adding this column to the SQL query?

ryanulit (approved):
couple questions but LGTM otherwise

jakedoublev:
Moving back to `draft` to keep on hold until after the release freeze.

ryanulit (approved):

ryanulit (approved):

PR #1655, opened 2024-10-15
chore(ci): Add initial CodeQL Workflow

jentfoo:
I want to explore using CodeQL to assist as part of our static analysis strategy.  Because this tool is available for free to open source projects, `opentdf` is a good place to start this testing.

This PR adds an initial configuration which will scan the codebase once a week.  These results can be ignored for the time being.  Instead the security team will review the results and make sure we tune away any noise first (or help in opening PR's to address any valid issues found).

[SEC-4162]: https://virtru.atlassian.net/browse/SEC-4162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,48 @@
+name: "CodeQL"
+
+on:
+  schedule:
+    - cron: '0 13 * * 1' # At 1:00 PM UTC every Monday
+  pull_request:
+    paths:
+      - '.github/workflows/codeql.yaml'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'go' ]
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version-file: go.work
+      if: ${{ matrix.language == 'go' }}
+
+    - name: Initialize the CodeQL tools for scanning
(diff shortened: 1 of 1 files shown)


jakedoublev:
Thanks for this @jentfoo We already have a [nightly checks job for long-running tasks](https://github.com/opentdf/platform/blob/main/.github/workflows/nightly-checks.yaml) and I'm wondering if we should add this job to that GHA. Is there any pro/con you see to running this nightly instead of weekly and grouping it within those checks?

jrschumacher:
@jentfoo were trying to keep non-OpenTDF issues out of the OpenTDF repo so there is full transparency. Any issues with removing the back link?

jentfoo:
@jrschumacher removed

@jakedoublev I considered adding it to the nightly tasks, but I did not see an obvious advantage to doing so.  I initially had this as a draft trying to confirm that the auto build is sufficient, and it seems initially that it is.

I would prefer to keep this independent.  Although this is starting as a weekly run, if we do decide to stick with CodeQL long term, I would like to transition it to run as a PR check so that feedback can be promptly provided.  Currently it seems to finish similar to other PR checks (~1min total), but I want to make sure results are tuned before it becomes a PR level check.

Let me know if you have any additional thoughts, but that would be my primary argument for keeping this isolated.

Thank you both for the quick look!

jakedoublev:
This looks good to me, but I'm unsure if we're in a code freeze for the release at present. This GHA update would have no impact on the release as there are no code changes, but it would make reverting or adding recent commits to `main` more challenging, so I will hold off on ✅ and defer to @opentdf/architecture .

jrschumacher:
I'm approving this since it's just CI

jrschumacher (approved):

Issue #1499, opened 2024-09-03
Policy: MatchSubjectMappings should be operator agnostic


jakedoublev:
### Background

At time of original implementation, there were two logical operators within Subject Condition Sets' Conditions: `IN` and `NOT_IN`. It was[ documented at the time](https://github.com/opentdf/platform/blob/main/service/policy/db/subject_mappings.go#L736-L747) that there were known performance concerns around a platform being driven to give granular access instead of ABAC. For example, if an admin utilized a specific widely used selector within the entitlement conditions of policy, like an entity representation field `.emailAddress` which every person entity would likely have, every condition would need to be resolved for every user.

As a result, the database query was given matching logic to check `IN/NOT_IN` presence along with the field when identifying matched subject mappings. This reduced the performance concern because a platform admin's poor use of selectable fields could return _every subject mapping_ for _every request_. The idea was that this policy API could return _all possible matches for an entity_ but would not do the logic of applying any logic on the conditions to determine true entitlement (leaving that to Authorization Service). 

However, maturing of the platform led to the additional operator being added `IN_CONTAINS`, and it is known there may be likely future operators `STARTS_WITH` and `ENDS_WITH`. 

These operators are not supported by the [current API implementation](https://github.com/opentdf/platform/blob/main/service/policy/db/subject_mappings.go#L701-L708). This has not been a problem because `MatchSubjectMappings` is not used internally within the platform and no consumers of the platform have reported this behavior as an issue.

Within Authorization Service, the operators mean:
- `NOT_IN`: entity _has_ the selector field on their entity representation but it DOES NOT resolve to one of the specified values in the list
- `IN`: entity _has_ the selector field on their entity representation AND it resolves to one of the values in the list
- `IN_CONTAINS`: entity _has_ the selector field on their entity representation AND its value contains one of the substrings specified in the list

In all cases above, including `NOT_IN`, the entity representation logically resolved contains the selector field to be applied by the operator. In other words, given an entity representation for `entityA` containing solely `{"id":"myUniqueId"}` and a (simplified) condition of `{"selectorField":".group","operator":"NOT_IN","values":["anyGroup"]}`, the entity does not resolve to `true` under the condition alone because `group` is not present. NOT_IN means the field is there the values are not present rather than _the field is not there so the values are not present_.

### Acceptance Criteria
1. `MatchSubjectMappings` is driven by _purely_ entity selector fields and _not_ values
2. Documentation in the database code and in the proto comments clarifies this behavior
[...]

PR #1657, opened 2024-10-16
feat(policy): 1651 move GetAttributesByValueFqns RPC request validation to protovalidate

ryanulit:
fix #1651

PR #1658, opened 2024-10-16
fix(policy): make MatchSubjectMappings operator agnostic

suchak1:
resolves: https://github.com/opentdf/platform/issues/1499


service/policy/db/subject_mappings.go
```diff
@@ -554,13 +549,8 @@ func selectMatchedSubjectMappingsSQL(subjectProperties []*policy.SubjectProperty
...
 
 		hasField := "each_condition->>'subject_external_selector_value' = '" + sp.GetExternalSelectorValue() + "'"
-		hasValue := "(each_condition->>'subject_external_values')::jsonb @> '[\"" + sp.GetExternalValue() + "\"]'::jsonb"
-		hasInOperator := "each_condition->>'operator' = 'SUBJECT_MAPPING_OPERATOR_ENUM_IN'"
-		hasNotInOperator := "each_condition->>'operator' = 'SUBJECT_MAPPING_OPERATOR_ENUM_NOT_IN'"
-		// Parses the json and matches the row if either of the following conditions are met:
-		where += "((" + hasField + " AND " + hasValue + " AND " + hasInOperator + ")" +
-			" OR " +
-			"(" + hasField + " AND NOT " + hasValue + " AND " + hasNotInOperator + "))"
+		// Parses the json and matches the row if the selector exists:
```
jakedoublev, line 552: I didn't touch this RPC to avoid merge conflicts, but perhaps you could remove this check and move it to the proto validation as part of this effort? https://github.com/opentdf/platform/pull/1658/files#diff-79178ab55805aac9e744330a45936ecde9e39baef7e8838764209f44b6a3f862R535 Similar to the work just done in GetAttributesByValueFQNs: https://github.com/opentdf/platform/pull/1657

jakedoublev: I also think we can ignore a check for values now that this route only drives based off selectors (explicitly _not_ values after your current work here): https://github.com/opentdf/platform/pull/1658/files#diff-79178ab55805aac9e744330a45936ecde9e39baef7e8838764209f44b6a3f862R542

suchak1: done

service/policy/db/query.sql
```diff
@@ -691,6 +691,31 @@ LEFT JOIN subject_condition_set scs ON scs.id = sm.subject_condition_set_id
 WHERE sm.id = $1
 GROUP BY av.id, sm.id, scs.id;
 
+-- name: MatchSubjectMappings :many
+SELECT
+    sm.id,
+    sm.actions,
+    JSON_STRIP_NULLS(JSON_BUILD_OBJECT('labels', sm.metadata -> 'labels', 'created_at', sm.created_at, 'updated_at', sm.updated_at)) AS metadata,
```
jakedoublev, line 698: Not a blocker, but we can probably forgo metadata for SMs and SCSs in this service to service API as they were probably only there originally due to reusing Squirrel-managed queries.

service/policy/db/subject_mappings.go
```diff
@@ -87,97 +83,6 @@ func unmarshalActionsProto(actionsJSON []byte, actions *[]*policy.Action) error
 	return nil
 }
 
-func subjectMappingSelect() sq.SelectBuilder {
```
jakedoublev, line 90: 🚀

service/policy/db/query.sql
jakedoublev: Would you mind please opening an issue if you don't want to capture within this current work?

jakedoublev (dismissed):

service/policy/db/query.sql
suchak1: https://github.com/opentdf/platform/issues/1673

service/policy/objects.proto
```diff
@@ -220,7 +220,7 @@ message SubjectConditionSet {
    and the external_value is the value selected by the external_selector_value on that Entity Representation (Subject Context). These mirror the Condition.
 */
 message SubjectProperty {
-  string external_selector_value = 1 [(buf.validate.field).required = true];
```
jakedoublev, line 223: I'm working on the `otdfctl` implementation of this API and realized the protovalidate below on line 224 that a field `external_value` must be specified breaks our desired `MatchSubjectMappingsRequest` implementation where only selectors are included for consideration (not values). The CLI must add a dummy value like `ExternalValue: "abc"` for a proto message to pass validation.

It looks like the proto message `SubjectProperty` is only used for this API, so perhaps we should remove this validation in this PR so the API can be called without extra work on clients for it to compile.

jakedoublev (approved):

jakedoublev (approved):

jakedoublev (approved):

PR #1679, opened 2024-10-23
feat(policy): 1603 policy improve upsertattrfqn

### LINKED ISSUE #1603: Policy: improve attribute FQN re-indexing
### Background

FQNs are utilized within TDFs and are the identifiers of attributes outside of the policy administration/database. There is a concept in the platform of attribute FQN reindexing. FQNs in the lookup table can get out of sync if:

1. an update was made to an attribute object (namespace name, definition name, value value) via RPC but there was a connection issue and the upsert to replace the FQNs containing the change failed
2. a manual database change was made altering any of the same policy objects

A functionality was added to the platform binary to [reindex FQNs](https://github.com/opentdf/platform/blob/main/service/cmd/policy.go#L29) and there is an open issue to add it to [otdfctl](https://github.com/opentdf/otdfctl/issues/157). 

However, some potential issues have become clear over time.

1. It should be possible to run a re-index on an interval

Because reindexing is actually crucial for accessing TDFs containing the intended updated state of the FQN, consistency could be guaranteed at said interval.

2. The `upsert` queries have room for optimization and efficiency

The `upsert` db query for a namespace only updates the `https://<namespace>` FQN, the definition `upsert` query only updates the rows matching `https://<namespace>/attr/<attribute definition name>`, and so on. 

It is not a problem because the CLI reindex command runs every query for every item, but optimization will be more important as at-scale deployments of the platform are realized and the volume of items in the database goes up.

An improvement was recently unblocked by the[ SQLC migration](https://github.com/opentdf/platform/pull/1445) [see comments](https://github.com/opentdf/platform/pull/1445#discussion_r1784773293).

### Acceptance Criteria

1. FQN reindexing happens on a configurable interval (jobs/workflows approach?)
2. FQN namespace `upsert` query overwrites the new namespace name in every FQN built from the namespace ID/name being upserted
3. FQN definition `upsert` query overwrites the new namespace name in every FQN built from the definition ID/name being upserted
4. FQN value `upsert` query overwrites the new namespace name in every FQN built from the value ID/name being upserted

ryanulit:
partially fix #1603 

- refactors FQN upserts to use a batch SQL query and removes `upsertAttrFqn` function in favor of calling SQLC generated functions directly
- simplifies `AttrFqnReindex` function used by CLI
- refactors Namespaces/Attributes/Values Create and UnsafeUpdate DB layer methods to return result from Get method directly (allowing code reuse and FQN retrieval after upsert)


service/policy/db/attribute_values.go
```diff
@@ -33,23 +33,12 @@ func (c PolicyDBClient) CreateAttributeValue(ctx context.Context, attributeID st
...
 
 	// Update FQN
-	fqn := c.upsertAttrFqn(ctx, attrFqnUpsertOptions{valueID: createdID})
-	if fqn != "" {
-		c.logger.Debug("created new attribute value FQN",
-			slog.String("value_id", createdID),
-			slog.String("value", value),
-			slog.String("fqn", fqn),
-		)
+	_, err = c.Queries.UpsertAttributeValueFqn(ctx, createdID)
```
ryanulit, line 36: The response is ignored at the moment, but we could log the rows returned if necessary?  Format of the returned rows looks like:

image

We could also just return the list of FQNs that were upserted...

jakedoublev (dismissed):

service/policy/db/namespaces.go
```diff
@@ -157,27 +153,12 @@ func (c PolicyDBClient) UnsafeUpdateNamespace(ctx context.Context, id string, na
 	}
 
 	// Update all FQNs that may contain the namespace name
-	nsFqn := c.upsertAttrFqn(ctx, attrFqnUpsertOptions{namespaceID: id})
-	c.logger.Debug("upserted fqn for unsafely updated namespace", slog.Any("fqn", nsFqn))
-
-	attrs, err := c.ListAttributes(ctx, StateAny, id)
+	_, err = c.Queries.UpsertAttributeNamespaceFqn(ctx, id)
 	if err != nil {
 		return nil, err
```
jakedoublev, line 158:
```suggestion
		return nil, db.WrapIfKnownInvalidQueryErr(err)
```

service/policy/db/namespaces.go
```diff
@@ -94,16 +94,12 @@ func (c PolicyDBClient) CreateNamespace(ctx context.Context, r *namespaces.Creat
 	}
 
 	// Update FQN
-	fqn := c.upsertAttrFqn(ctx, attrFqnUpsertOptions{namespaceID: createdID})
-	c.logger.Debug("upserted fqn for created namespace", slog.Any("fqn", fqn))
+	_, err = c.Queries.UpsertAttributeNamespaceFqn(ctx, createdID)
+	if err != nil {
+		return nil, err
```
jakedoublev, line 99:
```suggestion
		return nil, db.WrapIfKnownInvalidQueryErr(err)
```

service/policy/db/attribute_values.go
jakedoublev: This is fine for now without further enhancement.

jakedoublev (approved):

PR #1681, opened 2024-10-24
fix(sdk): Stops including binding in assertion hashes

sujankota:


dmihalcik-virtru (approved):

strantalis (approved):

jentfoo:
Is there any back story on why this needs to be removed from the assertion?

strantalis:
@jentfoo 
> Is there any back story on why this needs to be removed from the assertion?

Because the way the assertion is defined in the spec is that it includes the signature. So there is a "chicken and egg" problem with the binding object. 

https://github.com/opentdf/spec/pull/37/files#diff-9889d13f2cc03fa0cb9708ff10a6399fdc54b3800591c65daf4fe288bdd788b1R189-R197

jentfoo (approved):
Approving after DM conversation.  Summary:

It's weird to remove aspects of the assertions before checking the signing, but I accept that this is necessary given the opentdf spec.  I think we also need to be careful that existing assertions will continue to have hashes match (I believe that to be true, but should be validated).

PR #1684, opened 2024-10-24
feat(policy): 1659 implement transaction support in policy service

### LINKED ISSUE #1659: Policy: Research and implement basic transaction functionality for policy with SQLC
We’ve identified some issues with the lack of transactions across the system and policy management. We need to ensure the data is reliable and the reqs to the db are minimal. 

## Acceptance Criteria
- Identify the process of adding transactions with SQLC / pgx
- Document the process and publish to the platform documentation

ryanulit:
fix #1659


ryanulit:
This PR will be left as a draft for the foreseeable future as an example of a simple transactions implementation.  At the moment there is not enough time to dedicate to the effort, as there are higher priorities such as the config service.  We will need to go through every service to be sure that this approach is feasible.

PR #1714, opened 2024-11-04
fix: NanoTDF secure key from debug logging and iv conflict risk

jentfoo:
This change removes two debug logs where the nanoTDF key was encoded and logged.

In addition this fixes a small 1 in 16 million chance that an all zero IV could be used on the data (resulting in a key + iv reuse when compared to the policy).

### DIFF
--- a/sdk/nanotdf.go
+++ b/sdk/nanotdf.go
@@ -423,9 +423,6 @@ func writeNanoTDFHeader(writer io.Writer, config NanoTDFConfig) ([]byte, uint32,
 		return nil, 0, fmt.Errorf("ocrypto.CalculateHKDF failed:%w", err)
 	}
 
-	encoded := ocrypto.Base64Encode(symmetricKey)
-	slog.Debug("writeNanoTDFHeader", slog.String("symmetricKey", string(encoded)))
-
 	aesGcm, err := ocrypto.NewAESGcm(symmetricKey)
 	if err != nil {
 		return nil, 0, fmt.Errorf("ocrypto.NewAESGcm failed:%w", err)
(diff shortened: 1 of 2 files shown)


pflynn-virtru (approved):

sdk/nanotdf.go
```diff
@@ -652,6 +649,33 @@ func NewNanoTDFHeaderFromReader(reader io.Reader) (NanoTDFHeader, uint32, error)
...
+		}
+		ivPadded = append(ivPadded, iv...)
+		for _, b := range ivPadded {
+			if b != 0 {
+				return ivPadded, nil
+			}
+		}
+		// all zero IV, this is extremely rare so should be able to be addressed in the next loop
+		if loopCount >= loopCountLimit { // crazy, there must be an issue with the constants
+			return nil, errors.New("nonZeroPaddedIV loop exceeded limit")
```
jrschumacher, line 673: Should we expose this error or a similar one as fatal? This way developers can test for `error.Is(sdk.ErrNanoFatalZeroPaddedIv)`

jentfoo: It's hard to imagine a condition where this error case is hit.  That said, I put the loop limit in because a tight loop would likely be harder to understand if a bug does develop here.

Personally I don't think it's worth elevating, but let me know your preference.

jrschumacher: That's fair.

jrschumacher (approved):

sdk/nanotdf_test.go
```diff
@@ -241,6 +241,23 @@ func NotTestCreateNanoTDF(t *testing.T) {
 	}
 }
 
+func TestNonZeroRandomPaddedIV(t *testing.T) {
+	iv, err := nonZeroRandomPaddedIV()
+	require.NoError(t, err)
+	require.NotNil(t, iv)
+	assert.Equal(t, ocrypto.GcmStandardNonceSize, len(iv))
```
jrschumacher, line 248:
```suggestion
	assert.Len(t, ocrypto.GcmStandardNonceSize, iv)
```

jrschumacher (approved):

PR #1715, opened 2024-11-04
feat: update service registry in preperation for connectrpc migration

strantalis:
Depends On: #1709 

To transition to ConnectRPC and register service handlers effectively, we need to utilize generics. The reason for this is that the generated ConnectRPC code for creating handlers enforces strong typing:

`NewXServiceHandler(svc XServiceHandler, opts ...connect.HandlerOption)`

For reference, see an example here: 

https://github.com/opentdf/platform/blob/0ef65d410a8e1bc8b82f52b6a4f0f469a2f7f4fe/protocol/go/policy/kasregistry/kasregistryconnect/key_access_server_registry.connect.go#L190C51-L190C88

By leveraging generics, we can maintain type safety while also keeping interceptors centrally managed. This approach ensures that no individual service can inadvertently modify the interceptor list, helping maintain consistent security and functionality across all services.


pflynn-virtru (approved):

service/entityresolution/entityresolution.go
```diff
@@ -12,25 +12,35 @@ type ERSConfig struct {
...
+					panic(err)
+				}
+				if inputConfig.Mode == ClaimsMode {
+					claimsSVC, claimsHandler := claims.RegisterClaimsERS(srp.Config, srp.Logger)
+					return &EntityResolution{EntityResolutionServiceServer: claimsSVC}, claimsHandler
+				}
 
-			// Default to keyclaok ERS
-			return keycloak.RegisterKeycloakERS(srp.Config, srp.Logger)
+				// Default to keyclaok ERS
```
jakedoublev, line 40:
```suggestion
				// Default to keycloak ERS
```

service/pkg/serviceregistry/serviceregistry.go
```diff
@@ -75,21 +64,77 @@ type DBRegister struct {
...
+	RegisterFunc    RegisterFunc[S]
+	httpHandlerFunc HandlerServer
+	// ConnectRPCServiceHandler is the function that will be called to register the service with the
+	ConnectRPCFunc func(S, ...connect.HandlerOption) (string, http.Handler)
+	// DB is optional and used to register the service with a database
+	DB DBRegister
+}
+
+func (s Service[S]) GetNamespace() string {
+	return s.Namespace
```
jakedoublev, line 109: should we lowercase these fields that are now exposed via getter?

jakedoublev (dismissed):

service/pkg/serviceregistry/serviceregistry.go
strantalis: We still have to be able to set those values when defining a service 

https://github.com/opentdf/platform/pull/1715/files#diff-40830a88cb3098fbde9b40c1554e8468642d6ab11646fbdd2daa971e5d379bc6R27-R30

jakedoublev (approved):

jrschumacher (approved):

Issue #1524, opened 2024-09-05
ADR: Enable support for connect-rpc clients


dmihalcik-virtru:
Options:

1. Directly switch to connect-rpc This requires changing all 68 rpc call sites to use connect api, about a dozen registration calls, and four or five interceptors. Also will break our current in-process gateway, so we'd have to figure out how to use an external grpc-gateway or migrate existing REST clients to connect-web or similar.
2. Use https://github.com/connectrpc/vanguard-go; probably about the same amount of work on platform but seems to offer a smoother transition for clients, supporting REST, google grpc, and connect rpc from a single server

jrschumacher:
@dmihalcik-virtru updated description to add some resources

strantalis:
## Proposal: Transition to Connect RPC Protocol and Deprecate REST Endpoints

Our platform currently supports both gRPC and gRPC-Gateway to provide RPC and RESTful interfaces for our APIs. While this dual support aims to offer flexibility, it has led to increased complexity and maintenance overhead that is unsustainable for our small team. This proposal outlines the rationale for transitioning to the Connect RPC protocol and deprecating the REST endpoints generated via gRPC-Gateway.

### Current Challenges

- Duplication of Endpoints: We have approximately 65 RPC methods, resulting in around 130 endpoints when considering both RPC and REST interfaces. This duplication doubles the effort required for development, testing, and maintenance.
- Maintenance Overhead: The small size of our team makes it challenging to support and thoroughly test both sets of endpoints. This has already led to issues such as:
  - https://github.com/opentdf/platform/issues/1553
  - https://github.com/opentdf/platform/issues/1573
- Scalability Concerns: As the number of RPC methods grows, the burden of maintaining equivalent REST endpoints will increase proportionally, exacerbating the existing issues.
- Lack of HTTP/1.1 support in our SDK's

### Advantages of Moving to Connect RPC

- Unified Protocol Support: Connect RPC provides support for both HTTP/1.1 and HTTP/2, ensuring broader compatibility without the need for separate REST endpoints.
- Simplified Development: By focusing on a single protocol, we reduce complexity in our codebase, making it easier to develop new features and fix bugs.
- Consistent API Experience: Clients can interact with our APIs in a consistent manner, reducing confusion and potential errors caused by differing implementations between RPC and REST endpoints.
[...]

PR #1733, opened 2024-11-07
feat: backend migration to connect-rpc

strantalis:
### Proposed Changes

Implementation of the following adr https://github.com/opentdf/platform/issues/1524

The breaking changes are from changing the Idempontency Level. This is required if we would like to support GET request on some of our services. 

https://connectrpc.com/docs/go/get-requests-and-caching

### Checklist


pflynn-virtru (dismissed):

service/internal/server/server.go
```diff
@@ -195,25 +197,51 @@ func NewOpenTDFServer(config Config, logger *logger.Logger) (*OpenTDFServer, err
...
+	code        int
+	wroteHeader bool
+}
+
+func (rw *grpcGatewayResponseWriter) Header() http.Header {
+	return rw.w.Header()
+}
+
+func (rw *grpcGatewayResponseWriter) WriteHeader(statusCode int) {
+	if rw.wroteHeader == false {
```
jrschumacher, line 212:
```suggestion
	if !rw.wroteHeader {
```

service/internal/server/server.go
```diff
@@ -104,13 +101,19 @@ type CORSConfig struct {
 	MaxAge           int      `mapstructure:"maxage" json:"maxage" default:"3600"`
 }
```
jrschumacher, line 103:
```suggestion
const GRPCGatewayDeprecationDate = fmt.Sprintf("@%d", time.Date(2025, time.March, 25, 0, 0, 0, 0, time.UTC).Unix())

```

service/internal/server/server.go
```diff
@@ -195,25 +197,51 @@ func NewOpenTDFServer(config Config, logger *logger.Logger) (*OpenTDFServer, err
...
+	wroteHeader bool
+}
+
+func (rw *grpcGatewayResponseWriter) Header() http.Header {
+	return rw.w.Header()
+}
+
+func (rw *grpcGatewayResponseWriter) WriteHeader(statusCode int) {
+	if rw.wroteHeader == false {
+		rw.w.Header().Set("Deprecation", fmt.Sprintf("@%d", time.Date(2025, time.March, 25, 0, 0, 0, 0, time.UTC).Unix()))
```
jrschumacher, line 213:
```suggestion
		rw.w.Header().Set("Deprecation", GRPCGatewayDeprecationDate)
```

service/internal/server/server.go
```diff
@@ -230,37 +258,60 @@ func newHTTPServer(c Config, h http.Handler, a *auth.Authentication, g *grpc.Ser
...
+		// Apply CORS to connectRPC and extra handlers
+		connectRPC = corsHandler.Handler(connectRPC)
+		grpcGateway = corsHandler.Handler(grpcGateway)
+	}
+
+	// Add authN interceptor to extra handlers
+	if c.Auth.Enabled {
+		grpcGateway = a.MuxHandler(grpcGateway)
+	} else {
+		l.Error("disabling authentication. this is deprecated and will be removed. if you are using an IdP without DPoP set `enforceDPoP = false`")
```
jrschumacher, line 272: We should consider using hmac for a minimally safe no-oidc system or just push deprecation of this.

service/internal/server/server.go
strantalis: You can't do a constant like this in go.

service/internal/server/server.go
strantalis: It is currently nice for development but a simple hmac token setup would also make development easy.

pflynn-virtru (approved):

PR #1740, opened 2024-11-12
feat(authz): Remove org-admin role, move privileges to admin role

elizabethhealy:
### Proposed Changes

* dont need a separate org-admin role, remove it from keycloak setup and casbin config

### Checklist


pflynn-virtru (dismissed):

jrschumacher (changes requested):
@elizabethhealy and I discussed updating this to use even more generic policy for admin. This way we don't have to keep tweaking the policy for each addition unless we want to expose it to standard entities.

service/internal/auth/casbin.go
```diff
@@ -46,70 +44,49 @@ var defaultPolicy = `
...
-p,	role:org-admin,		/resource-mappings*,										*,			allow
-p,	role:org-admin,		/key-access-servers*,										*,			allow
-p,	role:org-admin, 	/kas/v2/rewrap,						  		        *,      allow
-p,	role:org-admin,		/unsafe*,										            *,			allow
-p,	role:org-admin,		/v1/entitlements,						  				        *,      allow
-p,	role:org-admin,		/v1/authorization,						  				        *,      allow
-p,	role:org-admin,		/v1/token/authorization,						  				        *,      allow
 
 # Role: Admin
 ## gRPC routes
```
jrschumacher, line 49:
```suggestion
## gRPC and HTTP routes
```

service/internal/auth/casbin.go
```diff
@@ -46,70 +44,49 @@ var defaultPolicy = `
...
 p,	role:admin,		/resource-mappings*,												*,			allow
 p,	role:admin,		/key-access-servers*,												*,			allow
 p,	role:admin,		/kas/v2/rewrap,						  				        *,      allow
+p,	role:admin,		/unsafe*,										            *,			allow
 p,	role:admin,		/v1/entitlements,						  				        *,      allow
 p,	role:admin,		/v1/authorization,						  				        *,      allow
 p,	role:admin,		/v1/token/authorization,						  				        *,      allow
-
+p,	role:admin,		/entityresolution/entitychain,							*,  	allow
+p,	role:admin,		/entityresolution/resolve,							*,  	allow
```
jrschumacher, line 64:
```suggestion
```

service/internal/auth/casbin.go
```diff
@@ -46,70 +44,49 @@ var defaultPolicy = `
...
 
 # Public routes
 ## gRPC routes
 ## for ERS, right now we don't care about requester role, just that a valid jwt is provided when the OPA engine calls (enforced in the ERS itself, not casbin)
-p,	role:unknown,			entityresolution.EntityResolutionService.ResolveEntities,					write,		allow
 p,	role:unknown,     kas.AccessService/Rewrap, 			                                  *,	  allow
 ## HTTP routes
 ## for ERS, right now we don't care about requester role, just that a valid jwt is provided when the OPA engine calls (enforced in the ERS itself, not casbin)
-p,	role:unknown,			/entityresolution/resolve,							  write,		allow
 p,	role:unknown,		  /kas/v2/rewrap,													  *,		allow
```
jrschumacher, line 90: I feel like we're missing the KAS public key routes. Also, does rewrap actually support no authorization? Is that even a usecase?

elizabethhealy: rewrap does need the token, but i guess it doesnt care about the role which i think is what this section is, not really public, casbin still requires them to provide tokens

service/internal/auth/casbin.go
elizabethhealy: :face_palm:

service/internal/auth/casbin.go
elizabethhealy: ah ok heres where the truly public routes are defined https://github.com/opentdf/platform/blob/88496cdb36d015fddffa2383c742350e3ed373b3/service/internal/auth/authn.go#L45-L57

jrschumacher (approved):

PR #1764, opened 2024-11-15
chore(ci): swap out kc chainguard image to dockerhub image

strantalis:
### Proposed Changes

I think the chainguard keycloak image is now private so our tests are breaking.

https://github.com/opentdf/platform/actions/runs/11859704681/job/33053273929

### Checklist

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -8,7 +8,7 @@ services:
     - ./keys/localhost.key:/etc/x509/tls/localhost.key
     - ./keys/ca.jks:/truststore/truststore.jks
     # This is kc 24.0.1 with opentdf protocol mapper on board
-    image: cgr.dev/chainguard/keycloak@sha256:37895558d2e0e93ffff75da5900f9ae7e79ec6d1c390b18b2ecea6cee45ec26f
+    image: keycloak/keycloak:25.0
     restart: always
     command:
     - "start-dev"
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -8,7 +8,7 @@ services:
     - ./keys/localhost.key:/etc/x509/tls/localhost.key
     - ./keys/ca.jks:/truststore/truststore.jks
     # This is kc 24.0.1 with opentdf protocol mapper on board
-    image: cgr.dev/chainguard/keycloak@sha256:37895558d2e0e93ffff75da5900f9ae7e79ec6d1c390b18b2ecea6cee45ec26f
+    image: keycloak/keycloak:25.0
```
jakedoublev, line 11: should we go with 25.0.6? https://hub.docker.com/layers/keycloak/keycloak/25.0.6/images/sha256-89aae522be5c945670620f61bdabed612835fb151de8d5aa7091d093e1718a30?context=explore

strantalis: 25.0 is always the latest on version 25. If you look at the digest sha they are the same .

jakedoublev: Right, I guess I was thinking if you pull the latest 25 locally and I have a cached 25 locally so we get different versions, then KC has a regression and we experience a bug, we wouldn't think to look at our Keycloak images without the specified patch version.

jakedoublev (approved):

PR #1776, opened 2024-11-20
fix: properly set casbin authz policy

strantalis:
### Proposed Changes

When calling the SetPolicy function it was overwrittng what ever was defined within the yaml configuration. This pr addresses this by changing the current functionality to overwrite the default policy while still allowing the policy in configuration to take precedence.   

The pull request also introduces two new fields `UsernameClaim` and `GroupsClaim` aligning more with openid based claim names and not keycloak realm roles. The authz policy will pull the username and list of groups and run them through the casbin enforce function now. A user can now scope a policy to a username. 

```
p, strantalis, policy.attributes.*, read, allow
```

### Checklist


docs/configuration.md
```diff
@@ -245,19 +246,34 @@ server:
...
         standard: opentdf-standard
         admin: opentdf-admin
 
+      ## Policy that will extend the builtin policy.
+      extension: |
+        p, role:admin, *, *, allow
+        p, role:standard, policy:attributes, read, allow
+        p, role:standard, policy:subject-mappings, read, allow
+        g, opentdf-admin, role:admin
+        g, alice@opentdf.io, role:standard
```
elizabethhealy, line 273: dumb question: what is the p vs g in the first column?

service/internal/auth/casbin.go
```diff
@@ -44,6 +34,9 @@ var defaultPolicy = `
 # delete - delete the resource
 # unsafe - unsafe actions
 
+## Grouping Statements - Maps users/groups to roles
+g, opentdf-admin, role:admin
+g, opentdf-standard, role:standard
```
elizabethhealy, line 39: i like the policy extending strategy instead of overriding, though what about the situation where the users wants to use these roles in their idp to mean different things? is it possible to change these lines when passing in your own policy? or is the default policy/groupings always there

strantalis: Yeah I added the ability to extend so they could do something like this.

```yaml
policy:
  extension: |
        g, my-group, role:admin
        g, casbin-user, role:standard
```

Or they could define their own roles

```yaml
policy:
  extension: |
        g, my-group, role:custom
        p, role:custom, policy.attributes.*, read, allow
```

service/internal/auth/casbin.go
```diff
@@ -12,221 +11,117 @@ import (
...
-		isDefaultRoleClaim = true
-		c.RoleClaim = defaultRoleClaim
+	if c.RoleMap != nil {
+		for k, v := range c.RoleMap {
+			c.Csv = strings.Join([]string{
+				c.Csv,
+				strings.Join([]string{"g", v, fmt.Sprintf("role:%s", k)}, ", "),
+			}, "\n")
+		}
+	}
```
jrschumacher, line 72: @strantalis let's remove this

PR #1778, opened 2024-11-21
feat(policy): 1500 Attribute create with Values (one RPC Call) should employ a db transaction

ryanulit:
### Proposed Changes

*

### Checklist

### DIFF
--- a/service/policy/db/policy.go
+++ b/service/policy/db/policy.go
@@ -1,6 +1,9 @@
 package db
 
 import (
+	"context"
+	"fmt"
+
 	"github.com/opentdf/platform/protocol/go/common"
 	"github.com/opentdf/platform/service/logger"
 	"github.com/opentdf/platform/service/pkg/db"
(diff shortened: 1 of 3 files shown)


service/policy/db/policy.go
```diff
@@ -31,6 +34,29 @@ func NewClient(c *db.Client, logger *logger.Logger, configuredListLimitMax, conf
...
+
+	txClient := &PolicyDBClient{c.Client, c.logger, c.Queries.WithTx(tx), c.listCfg}
+
+	err = query(txClient)
+	if err != nil {
+		c.logger.WarnContext(ctx, "error during DB transaction, rolling back")
+
+		if rollbackErr := tx.Rollback(ctx); rollbackErr != nil {
+			// this should never happen, but if it does, we want to know about it
+			return fmt.Errorf("failed to rollback DB transaction for [%w]: %w", err, rollbackErr)
```
jakedoublev, line 51: Do you think there is value in making this a named error in the db package (i.e. in `service/pkg/db`: `var ErrRollbackFailed = errors.New("ErrRollbackFailed: transaction failed with failed rollback")`? We can then still join it with these two wrapped errors that way but more easily test for it with `errors.Is` downstream.

ryanulit: I debated this, but ultimately we would want to know the actual postgres error that caused the rollback or failed commit and have that returned from `db.StatusifyError()` in the Connect response message, instead of a vague "rollback failed" or "commit failed" message.  And we wouldn't lose the rollback or commit errors, as they would still get logged internally by the `slog.Error` catch-all in `db.StatusifyError()`.

ryanulit: Per our slack huddle discussion, I've gone ahead and added the named errors for possible future usage.

PR #1797, opened 2024-11-28
fix(authz): handle pagination in authz service

jakedoublev:
### Proposed Changes

* Handle pagination limit/offset in `authorization.GetEntitlements`
* If quantity of attributes or subject mappings exceeds default "list" API limit as configured, we need to retrieve all to make accurate entitlement decision

### Checklist


elizabethhealy (dismissed):
lgmt, can we add some tests or make a follow-up ticket to do so?

jakedoublev:
@elizabethhealy does the unit test added in the most recent commit with policy mocks look sufficient to you?

elizabethhealy (approved):

PR #1810, opened 2024-12-03
fix(core): expose rest based healthcheck

strantalis:
### Proposed Changes

Missed this while migrating to connect-rpc but accidentally removed the REST based healthcheck which causes kubernetes deployments to break. 

### Checklist

### DIFF
--- a/service/health/health.go
+++ b/service/health/health.go
@@ -60,7 +60,7 @@ func NewRegistration() *serviceregistry.Service[grpchealth.Checker] {
 						}
 					})
 					if err != nil {
-						srp.Logger.Error("failed to register healthz endpoint", slog.String("error", err.Error()))
+						panic(errors.Join(errors.New("failed to register healthz endpoint"), err))
 					}
 
 					return nil
--- a/service/health/health.go
+++ b/service/health/health.go
@@ -37,7 +37,7 @@ func NewRegistration() *serviceregistry.Service[grpchealth.Checker] {
 				}
 				hs := HealthService{logger: srp.Logger}
 				return hs, func(_ context.Context, mux *runtime.ServeMux) error {
-					mux.HandlePath(http.MethodGet, "/healthz", func(w http.ResponseWriter, r *http.Request, _ map[string]string) {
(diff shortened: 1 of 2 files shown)


jrschumacher (approved):

dmihalcik-virtru (approved):

pflynn-virtru (approved):

service/health/health.go
```diff
@@ -31,7 +35,36 @@ func NewRegistration() *serviceregistry.Service[grpchealth.Checker] {
...
+				return hs, func(_ context.Context, mux *runtime.ServeMux) error {
+					err := mux.HandlePath(http.MethodGet, "/healthz", func(w http.ResponseWriter, r *http.Request, _ map[string]string) { //nolint:contextcheck // check is not relevant here
+						resp, err := hs.Check(context.Background(), &grpchealth.CheckRequest{
+							Service: r.URL.Query().Get("service"),
+						})
+						if err != nil {
+							http.Error(w, err.Error(), http.StatusInternalServerError)
+							return
+						}
+						status := map[string]interface{}{"status": strings.ToUpper(resp.Status.String())}
```
dmihalcik-virtru, line 48: Should we include the service name here if it is set?

service/health/health.go
```diff
@@ -31,7 +35,36 @@ func NewRegistration() *serviceregistry.Service[grpchealth.Checker] {
...
+
+							return
+						}
+						w.WriteHeader(http.StatusOK)
+						if err := json.NewEncoder(w).Encode(status); err != nil {
+							srp.Logger.Error("failed to encode health status", slog.String("error", err.Error()))
+						}
+					})
+					if err != nil {
+						srp.Logger.Error("failed to register healthz endpoint", slog.String("error", err.Error()))
```
dmihalcik-virtru, line 63: Shouldn't we panic here?

dmihalcik-virtru (dismissed):

service/health/health.go
strantalis: The `grpchealth.CheckResponse` only has `Status` built in.

mkleene (approved):

PR #1812, opened 2024-12-04
feat(core): Expose context authn methods

elizabethhealy:
### Proposed Changes

* enable setting context info and retrieving token/jwk from context from additional services added on platform extension

### Checklist

### DIFF
--- a/service/pkg/auth/context_auth.go
+++ b/service/pkg/auth/context_auth.go
@@ -8,11 +8,11 @@ import (
 	"github.com/opentdf/platform/service/logger"
 )
 
-const (
-	authnContextKey = authContextKey("dpop-jwk")
+var (
+	authnContextKey = authContextKey{}
 )
 
-type authContextKey string
+type authContextKey struct{}
 
 type authContext struct {
 	key         jwk.Key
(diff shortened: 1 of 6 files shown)


service/pkg/auth/context_auth.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+	"github.com/opentdf/platform/service/logger"
+)
+
+const (
+	authnContextKey = authContextKey("dpop-jwk")
+)
+
+type authContextKey string
```
strantalis, line 15: Can we take this time to update the context key to a struct. Its better practice and reduces the risk of this key getting overridden.

`type authContextKey struct{}`

elizabethhealy: i think the only advantage of using a struct over a custom typed string it that youre not allocating as much memory. "Users of WithValue should define their own types for keys. To avoid allocating when assigning to an interface{}, context keys often have concrete type struct{}" https://pkg.go.dev/context#WithValue
wouldnt authContextKey("dpop-jwk") and authContextKey{"key":""dpop-jwk"} have a similar risk of getting overrriden

elizabethhealy: updated to use struct

strantalis (approved):

PR #1832, opened 2024-12-20
fix(core): GetDecisions should handle empty string and non-existent attributes

elizabethhealy:
### Proposed Changes

* if a resouce attribute fqn list only has "", DENY
* if a resource attribute fqn list has "" and other attributes, ignore the ""
* if a resource attribute fqn list has a non existent attribute, DENY

### Checklist

### DIFF
--- a/service/authorization/authorization.go
+++ b/service/authorization/authorization.go
@@ -29,9 +29,13 @@ import (
 	"github.com/opentdf/platform/service/pkg/db"
 	"github.com/opentdf/platform/service/pkg/serviceregistry"
 	"github.com/opentdf/platform/service/policies"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 const EntityIDPrefix string = "entity_idx_"
+
+var ErrEmptyStringAttribute = errors.New("resource attributes must have at least one attribute value fqn")
 
 type AuthorizationService struct { //nolint:revive // AuthorizationService is a valid name for this struct
 	sdk    *otdf.SDK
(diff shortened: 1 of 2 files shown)


pflynn-virtru (approved):

service/authorization/authorization.go
```diff
@@ -287,6 +295,11 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
 
 	for raIdx, ra := range dr.GetResourceAttributes() {
 		for ecIdx, ec := range dr.GetEntityChains() {
+			// check if we already have a decision for this entity chain
+			responseIdx := (raIdx * len(dr.GetEntityChains())) + ecIdx
+			if response[responseIdx] != nil {
```
jentfoo, line 300: It's not obvious to me what condition would result in a duplicate.  As long as an overflow does not occur, and `dr.GetEntityChains()` remains constant, I don't believe a duplicate responseIdx is possible.

Is there a condition I am missing?

jentfoo (commented):
Looks reasonable, but I think the added complexity in the commented for loop can be removed.

service/authorization/authorization.go
elizabethhealy: there is a block earlier in the function which sets decisions to deny if specific errors are returned when retrieving attribute definitions https://github.com/opentdf/platform/pull/1832/files#diff-2c91e289704441218433186b7c84b34608c4dc1eeb9e82dfde5968552ab29cc5R233

jentfoo: That makes sense, thank you!

jentfoo (approved):

PR #1833, opened 2024-12-24
MIC-1436: User can decrypt TDF files created with FileWatcher2.0.8 and older.

mustyantsev:
### Proposed Changes

*We need decrypt old FW tdf files. Old FW tdf files do not support SID and assertions are represented as JSON object

### Checklist

- [x ] I have added or updated unit tests

### DIFF
--- a/sdk/assertion.go
+++ b/sdk/assertion.go
@@ -123,6 +123,49 @@ func (a Assertion) GetHash() ([]byte, error) {
 	return ocrypto.SHA256AsHex(transformedJSON), nil
 }
 
+type FlexibleValue struct {
+	AsString *string
+	AsObject map[string]interface{}
+}
+
+func (fv FlexibleValue) MarshalJSON() ([]byte, error) {
+	if fv.AsObject != nil {
+		objAsJSON, err := json.Marshal(fv.AsObject)
+		if err != nil {
+			return nil, err
+		}
+		return json.Marshal(string(objAsJSON))
+	}
+
+	if fv.AsString != nil {
+		return json.Marshal(*fv.AsString)
+	}
+
+	return json.Marshal(nil)
+}
+
+func (fv *FlexibleValue) UnmarshalJSON(data []byte) error {
+	// Try to unmarshal as a raw string
+	var strValue string
+	if err := json.Unmarshal(data, &strValue); err == nil {
+		var temp map[string]interface{}
+		if json.Unmarshal([]byte(strValue), &temp) == nil {
+			fv.AsObject = temp
+		} else {
+			fv.AsString = &strValue
+		}
+		return nil
+	}
+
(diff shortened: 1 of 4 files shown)


sujankota (dismissed):

sdk/assertion.go
```diff
@@ -131,7 +174,7 @@ type Statement struct {
 	// Schema describes the schema of the payload. (e.g. tdf)
 	Schema string `json:"schema,omitempty" validate:"required"`
 	// Value is the payload of the assertion.
-	Value string `json:"value,omitempty"  validate:"required"`
+	Value FlexibleValue `json:"value,omitempty"  validate:"required"`
```
dmihalcik-virtru, line 177: Hmm this will break downstream code that creates AssertionConfig objects. I think you need to make Statement marshallable/unmarshallable and handle value there instead of creating a new type here.

dmihalcik-virtru (changes requested):

sdk/tdf.go
```diff
@@ -863,6 +863,10 @@ func (r *Reader) doPayloadKeyUnwrap(ctx context.Context) error { //nolint:gocogn
 
 			unencryptedMetadata = metaData
 		}
+
+		if r.manifest.EncryptionInformation.KeyAccessObjs[0].SplitID == "" {
```
dmihalcik-virtru, line 867: If this is for treating `sid`-less multi-kao files as 'or', you should instead remove the `mixedSplits` processing.

mustyantsev: @dmihalcik-virtru is that what you meant (**remove the mixedSplits**)?

dmihalcik-virtru: I think you can also get rid of the `if mixedSplits && ...` block below and remove the variable entirely

dmihalcik-virtru: If removing it is causing a regression, you can add this back and I'll take a look after this gets merged in

sdk/assertion_test.go
```diff
@@ -33,3 +34,63 @@ func TestTDFWithAssertion(t *testing.T) {
 
 	assert.Equal(t, "4a447a13c5a32730d20bdf7feecb9ffe16649bc731914b574d80035a3927f860", string(hashOfAssertion))
 }
+
+func TestTDFWithAssertionJsonObject(t *testing.T) {
```
elizabethhealy, line 38: nit: can we add another test where the assertion value is a string to make sure the new json.unmarshal behaves correctly in both cases

elizabethhealy (dismissed):

sdk/assertion_test.go
mkleene: will add a test on a later pull

elizabethhealy (approved):

dmihalcik-virtru (approved):

PR #1834, opened 2025-01-03
feat(core): Deprecate server.CryptoProvider for kas.keyring

dmihalcik-virtru:
### Proposed Changes

* Replace `access.CryptoProvider` with `recrypt.Provider`, with simplified interface for implementation of 3rd party crypto layers
* refactors code to use it
* supports for old config formats, adds more tests around them
* TK (later): Expose this to service constructor somehow? How can I pass a live object to a packed-in service?

### Checklist


jentfoo (dismissed):

examples/cmd/encrypt.go
```diff
@@ -51,19 +51,6 @@ func encrypt(cmd *cobra.Command, args []string) error {
...
 	in := strings.NewReader(plainText)
 
-	opts := []sdk.Option{
-		sdk.WithInsecurePlaintextConn(),
-		sdk.WithClientCredentials("opentdf-sdk", "secret", nil),
-	}
-
-	if noKIDInKAO {
-		opts = append(opts, sdk.WithNoKIDInKAO())
-	}
```
elizabethhealy, line 61: should we leave this if in in case someone uses the no-kid-in-nano command flag

service/pkg/server/services.go
```diff
@@ -167,8 +167,20 @@ func startServices(ctx context.Context, cfg config.Config, otdf *server.OpenTDFS
 				}
 			}
 
+			svcConfig := cfg.Services[ns]
+			if ns == "kas" {
+				logger.Debug("updating kas key configuration", slog.String("namespace", ns), slog.Any("legacyConfig", cfg.Server.CryptoConfig2024))
+				// Upgrade the the kas configuration, if there is a legacy `CryptoProvider` configuration
+				// present in the otdf server config.
+				if cfg.Server.CryptoConfig2024 != nil {
```
pflynn-virtru, line 175: `cfg.Server.CryptoConfig2024` is this the legacy config renamed.  should the new config be named `cfg.Server.CryptoConfig2025`.  I have also seen append `_V1` 
is this versioned in the yaml?

pflynn-virtru (dismissed):

service/pkg/server/services.go
dmihalcik-virtru: There is no new config under `Server`; after this it is moved entirely to the `services.kas` field.

The type names are not exposed in the YAML config file, and the names of the fields can/will be overridden with tags (e.g. `mapstructure:"config2025") as needed. The config struct types and their fields need to be exported (capitalized) for use with the `mapstructure` library, but we have as of yet not treated them as part of the API and subject to standard versioning considerations. However, we may want to do that before marking this whole library as 1.x.

I did a grep through our internal code and was unable to find references to the moved/removed/renamed types

dmihalcik-virtru:
We are moving more and more to allowing integrators to customize the configuration, so I don't want to do something so normative and fixed

PR #1835, opened 2025-01-03
feat(core): adds bulk rewrap to sdk and service

imdominicreed:
### Proposed Changes

* Creates a new Rewrap API to support Bulk Rewrap Operations

### Checklist

### Testing Instructions

- Tests already check for backwards compatibility.
- Adds Bulk Benchmark to example for testing.
- Adds Bulk Round Trip tests.


service/kas/kas.proto
```diff
@@ -48,11 +81,32 @@ message RewrapRequest {
...
+
+message RewrapResult {
+	string policy_id = 1;
+	repeated KAORewrapResult results = 2;
+}
+
+
 message RewrapResponse {
+	// Deprecated
   map<string, google.protobuf.Value> metadata = 1;
```
dmihalcik-virtru, line 102: Reserve, don't bother deprecating, field ids that aren't being set anymore.
Are they still being set? I don't see how the round trip requests would work if not...

dmihalcik-virtru (changes requested):

service/kas/kas.proto
```diff
@@ -48,11 +81,32 @@ message RewrapRequest {
...
+	string status = 3;
+	oneof result { 
+		bytes kas_wrapped_key = 4;
+		string error = 5;
+	}
+}
+
+message RewrapResult {
+	string policy_id = 1;
+	repeated KAORewrapResult results = 2;
```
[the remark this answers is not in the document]
imdominicreed: i was following the spec for this, i am totally okay with this change, if we wanted a spec change.

service/kas/kas.proto
```diff
@@ -31,6 +31,39 @@ message LegacyPublicKeyRequest {
...
+  string body = 2;
+}
+
+message KeyAccessObjectRequest {
+  string key_access_object_id = 1;
+  KeyAccess key_access_object = 2;
+}
+
+message RewrapRequestBody {
+  repeated KeyAccessObjectRequest key_access_object_requests = 1;
```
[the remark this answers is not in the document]
imdominicreed: i was following the spec for this, i am totally okay with this change, if we wanted a spec change.

service/kas/kas.proto
```diff
@@ -31,6 +31,39 @@ message LegacyPublicKeyRequest {
...
+  PolicyRequest policy = 2;
```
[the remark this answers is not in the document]
imdominicreed: since the id in the policy request is a required field similar to the kao id. however the actual policy body is a little more unstructured.

service/kas/kas.proto
```diff
@@ -31,6 +31,39 @@ message LegacyPublicKeyRequest {
...
+  string protocol = 3;
+  string key_type = 4;
+  string kas_url = 5;
+  string kid = 6;
+  string split_id = 7;
+  bytes wrapped_key = 8;
+  bytes header = 9;
+  string algorithm = 10;
+}
+message PolicyRequest {
```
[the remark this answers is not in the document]
imdominicreed: i was following the spec for this. i believe this is so its easier to handle errors with a specific policy id.

service/kas/kas.proto
imdominicreed: fixed, but how do i fix the buf breaking action?

sdk/bulk.go
```diff
@@ -0,0 +1,138 @@
...
+
+type BulkDecryptOption func(request *BulkDecryptRequest)
+
+func WithTDFs(tdfs ...*BulkTDF) BulkDecryptOption {
+	return func(request *BulkDecryptRequest) {
+		request.AppendTDFs(tdfs...)
+	}
+}
+
+func (s SDK) CreateBulkRewrapRequest(options ...BulkDecryptOption) *BulkDecryptRequest {
```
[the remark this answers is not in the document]
dmihalcik-virtru: Unless an application needs it, don't export it. 
```suggestion
func (s SDK) createBulkRewrapRequest(options ...BulkDecryptOption) *BulkDecryptRequest {
```

service/kas/kas.proto
dmihalcik-virtru: Options:

1. The 'correct' way: leave these fields defined (and the implicitly generated `MetadataEntry` message type from the map), and fill them out when possible / requested, to allow 'old' clients. Mark them as deprecated `[deprecated=true]`.
2. The 'satisficing' way: Mark them as `[deprecated=true]` and don't fill them out. AFAICT they are never used in the clients, so this should be fine. I can only find references to 
3. Disable [`breaking`](https://buf.build/docs/configuration/v2/buf-yaml/#use_1) for this PR then add it back immediately after merge.

I think `2` should be fine. I haven't been able to find any references to `metadata` or `schema_version` in the clients. 


However both SessionPublicKey and entity key are both referenced a lot. IIUC these are still used and should not be deprecated?

imdominicreed: deprecated metadata, schema_version, and entity_wrapped_key. since they are all part of the old API and eventually should probably no longer be used in the upcoming versions.

sdk/bulk.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	if len(errList) != 0 {
+		return BulkDecryptionErrors(errList)
+	}
+
+	return nil
+}
+
+func (b *BulkDecryptRequest) AppendTDFs(tdfs ...*BulkTDF) {
```
dmihalcik-virtru, line 140:
```suggestion
func (b *BulkDecryptRequest) appendTDFs(tdfs ...*BulkTDF) {
```

sdk/kas_client.go
```diff
@@ -20,25 +22,21 @@ const (
...
-	Algorithm       string    `json:"algorithm,omitempty"`
-	ClientPublicKey string    `json:"clientPublicKey"`
-	SchemaVersion   string    `json:"schemaVersion,omitempty"`
+type KAOResult struct {
+	SymmetricKey      []byte
+	Error             error
+	KeyAccessObjectID string
+}
+
+type Decryptor interface {
```
dmihalcik-virtru, line 37: Only export when there is a specific need to use the type in callers. This is a package internal type and should not be exported
```suggestion
type decryptor interface {
```

sdk/kas_client.go
```diff
@@ -20,25 +22,21 @@ const (
...
 }
 
-// once the backend moves over we should use the same type that the golang backend uses here
-type rewrapRequestBody struct {
-	KeyAccess       KeyAccess `json:"keyAccess"`
-	Policy          string    `json:"policy,omitempty"`
-	Algorithm       string    `json:"algorithm,omitempty"`
-	ClientPublicKey string    `json:"clientPublicKey"`
-	SchemaVersion   string    `json:"schemaVersion,omitempty"`
+type KAOResult struct {
```
dmihalcik-virtru, line 31: Don't export this type as it is not exposed by any exported types or functions. 
```suggestion
type kaoResult struct {
```

Also, you are probably better off using a refactor option in your IDE than accepting this as it will just break stuff

sdk/kas_client_test.go
```diff
@@ -6,6 +6,9 @@ import (
 	"net/http"
 	"testing"
 
+	kaspb "github.com/opentdf/platform/protocol/go/kas"
+	"google.golang.org/protobuf/encoding/protojson"
+
```
dmihalcik-virtru, line 11:
```suggestion
```

service/kas/access/rewrap.go
```diff
@@ -2,6 +2,10 @@ package access
 
 import (
 	"bytes"
+
+	"google.golang.org/protobuf/encoding/protojson"
```
dmihalcik-virtru, line 6: Regroup these and move this down to the 'non-runtime' package import block below

service/kas/kas.proto
```diff
@@ -31,6 +31,38 @@ message LegacyPublicKeyRequest {
   string algorithm = 1;
 }
 
+message KeyAccess {
+  string encrypted_metadata = 1;
+  bytes policy_binding = 2;
+  string protocol = 3;
+  string key_type = 4;
+  string kas_url = 5;
```
dmihalcik-virtru, line 39:
```suggestion
  string key_type = 4 [json_name = "type"];
  string kas_url = 5 [json_name = "url"];
```

service/kas/kas.proto
```diff
@@ -31,6 +31,38 @@ message LegacyPublicKeyRequest {
...
+  string kid = 6;
+  string split_id = 7;
```
dmihalcik-virtru, line 41: This field and several others are ignored by KAS at the moment, but we may want to model them properly anyway.

```suggestion
  string split_id = 7 [json_name = "sid"];
```

service/kas/kas.proto
```diff
@@ -31,6 +31,38 @@ message LegacyPublicKeyRequest {
   string algorithm = 1;
 }
 
+message KeyAccess {
+  string encrypted_metadata = 1;
+  bytes policy_binding = 2;
```
dmihalcik-virtru, line 36: I'd say this should be an inner type PolicyBinding to match the TDF spec for KAOs, but in the **Request Body Example** it is a string? So IDK.

https://github.com/opentdf/spec/blob/main/schema/tdf/KeyAccessObject.md

service/kas/kas.proto
```diff
@@ -31,6 +31,38 @@ message LegacyPublicKeyRequest {
...
+  string protocol = 3;
+  string key_type = 4;
+  string kas_url = 5;
+  string kid = 6;
+  string split_id = 7;
+  bytes wrapped_key = 8;
+  bytes header = 9;
+  string algorithm = 10;
```
dmihalcik-virtru, line 44:
```suggestion
  string algorithm = 10 [json_name = "alg"];
```

Hah this is not currently encoded in the stored KAOs or listed in the spec, so we don't know what to name it. I'd suggest we add it, using `alg` to match the name in the `policyBinding`. We currently can infer it from kid, but that might not always be the case. In practice we will also want several algorithms:

- the one used to encrypt the wrapped key - can be inferred from kid
- to verify the policy binding - explicit in the `policyBinding` object
- to wrap/encapsulate the response - I guess this is up to the server now?
- others? E.g. encryptedMetadata is AES with the unwrappedKey value (which might not be the payload DEK but instead a fragment of the DEK or, eventually, some other value?)

service/kas/kas.proto
```diff
@@ -31,6 +31,38 @@ message LegacyPublicKeyRequest {
...
+message KeyAccess {
+  string encrypted_metadata = 1;
+  bytes policy_binding = 2;
+  string protocol = 3;
+  string key_type = 4;
+  string kas_url = 5;
+  string kid = 6;
+  string split_id = 7;
+  bytes wrapped_key = 8;
+  bytes header = 9;
```
dmihalcik-virtru, line 43: AFAICT if nano, this value is set but not the others? Maybe add some documentation around this

dmihalcik-virtru (changes requested):

service/kas/kas.proto
imdominicreed: I added this within PolicyBinding message

service/kas/kas.proto
imdominicreed: Created PolicyBinding message to match spec.

sdk/bulk.go
```diff
@@ -0,0 +1,145 @@
...
+		return decryptor, nil
+	case Standard:
+		return s.createTDF3DecryptHandler(tdf.Writer, tdf.Reader)
+	case Invalid:
+	}
+	return nil, fmt.Errorf("unknown tdf type: %s", tdfType)
+}
+
+// BulkDecrypt
+func (s SDK) BulkDecrypt(ctx context.Context, opts ...BulkDecryptOption) error {
```
strantalis, line 70: There is a side of me that feels like the consumer of the sdk should decide when to actually decrypt payload. There might be a use case where someone would want to access encrypted metadata but not decrypt the full payload. Or maybe they want to do a random read to see if its actually the correct payload they need before actually decrypting it.

strantalis (commented):
@imdominicreed Can we make sure to add comments to the code. For example adding comments to the `WithX` options and anything that exposed from the sdk.

sdk/bulk.go
imdominicreed: i agree, although i think this feature should be seperate PR. NanoTDF lib would take some refactoring to get this feature working. Futhermore, I think as Bulk Go API will eventually need some more thoughtful features to be added into it, i.e. accessing manifests, collections, and encrypted metadata.  

Originally this is why I had `BulkDecryptRequest`.

biscoe916: I agree with you both, I think we should have the option to defer the decrypt portion of the operation, but I think that can come in a future ticket.

patmantru (dismissed):

sdk/bulk.go
```diff
@@ -0,0 +1,150 @@
...
+
+type BulkDecryptRequest struct {
+	TDFs    []*BulkTDF
+	TDFType TdfType
+}
+
+// BulkErrors List of Errors that Failed during Bulk Decryption
+type BulkErrors []error
+
+func (b BulkErrors) Error() string {
```
dmihalcik-virtru, line 27: Optional? I don't think this really is an Unwrap? But it might make sense to implement [the unwrap interface](https://pkg.go.dev/errors#pkg-overview) anyway

```suggestion
func (b BulkErrors) Unwrap() []error {
	return b
}

func (b BulkErrors) Error() string {
```

dmihalcik-virtru (dismissed):

biscoe916 (approved):

Issue #1485, opened 2024-08-30
ADR: Enhancement to the KeyAccessServer Table


jrschumacher:
# Enhance Key Access Server Table

## Context and Problem Statement

Currently, the Key Access Server (KAS) stores a jsonb representation of the public_key. This design presents a limitation, as it implies that a Key Access Server can have only a single public key associated with it. However, this is not the case.

In our proto definitions, we define the PublicKey field within the KeyAccessServer proto message:

https://github.com/opentdf/platform/blob/069f939923cb3570c1e62453f68022a0b9c3e544/service/policy/objects.proto#L269

The PublicKey message can take on one of the following forms:

https://github.com/opentdf/platform/blob/069f939923cb3570c1e62453f68022a0b9c3e544/service/policy/objects.proto#L302-L318

This allows for either storing a reference to a remote public key or defining a public key via a KasPublicKeySet, which is essentially a repeated list of KasPublicKey entries:

https://github.com/opentdf/platform/blob/069f939923cb3570c1e62453f68022a0b9c3e544/service/policy/objects.proto#L284-L295

### Potential Issues:

1. **Key Rotation Confusion**: As an admin, I could list multiple RSA keys, including old rotated keys, which might create confusion about which key is active.
2. **Lack of Key Selection Mechanism**: There is currently no way for our SDK to identify which key should be used as the correct or active key.

Additionally, there is a mismatch between how we assign the KAS to a namespace, attribute, or attribute value. This assignment would work if the KAS were intended to operate with only a single key at a time. However, this does not align with how KAS was originally designed.

### Proposed Solution:

Instead of associating a KAS directly with a namespace, attribute, or attribute value, we should use Grants to map a specific Key to the policy resources. This would better reflect the original design intent and allow for more flexible future key management.

## Decision Drivers

* Support key rotations
* Ensure strict values
* Allow Key to `namespace`, `attribute`, or `value` mapping

## Considered Options

* Store multiple keys within the `public_key` jsonb typed column
* Create an auxiliary table to store the keys with a many:1 relationship with the KeyAccessServer table

## Decision Outcome

Chosen option: "{title of option}"

### Store multiple keys within the `public_key` jsonb typed column

* 🟢 Existing table structure and no migration needed
* 🔴 With JSONB the data does not have a schema to conform to and data structure migrations are hard
* 🔴 No indication from a high-level which key is primary without decoding the JSONB
* 🔴 Row needs to be updated "frequently" to capture key rotations and identify the current key
[...]

PR #1836, opened 2025-01-06
feat(policy): adds new public keys table

strantalis:
### Proposed Changes

Implements the following ADR as discussed here https://github.com/opentdf/platform/issues/1485. 

- Adds new `public_keys` table along with mappings pivot tables for attribute values, definitions and namespaces.
  - On create key if there is a previous key with the same alg it will make previous key inactive while mapping the the new key to the same namespaces, attribute definitions and values. 
  - On first mapping public key field `was_mapped` gets set to `true`
- Adds new CRUD RPC's to manage public keys and their associations
  - Updates only allow metadata to be updated considering public keys immutable
- Returns public keys on `GetAttributeValueByFQN` rpc
- Adds new ENUM values
  - `KAS_PUBLIC_KEY_ALG_ENUM_RSA_4096`
  - `KAS_PUBLIC_KEY_ALG_ENUM_EC_SECP384R1`
  - `KAS_PUBLIC_KEY_ALG_ENUM_EC_SECP521R1`

```mermaid
erDiagram

    key_access_server {
        uuid       id                PK
        varchar    uri               UK
        varchar    name              UK "new optional name column"
        jsonb      public_key
        jsonb      metadata
    }

    public_keys {
        uuid        id                      PK 
        boolean     is_active         
        boolean     was_used          
        uuid        key_access_server_id    FK
        varchar(36) key_id
        varchar(50) alg                     "algorithm"
        constraint  unique_key              UK  "enforces unique key_id and algorithm per KAS (key_access_server_id, key_id, alg)"
        constraint  unique_active_key       UK  "enforce only one active key per KAS per algorithm"
        text        public_key
        jsonb       metadata
    }

    attribute_namespace_public_key_map {
        uuid namespace_id  FK
        uuid public_key_id FK
    }

    attribute_definition_public_key_map {
        uuid attribute_definition_id FK
        uuid public_key_id           FK
    }

    attribute_value_public_key_map {
        uuid attribute_value_id FK
        uuid public_key_id      FK
    }

    key_access_server 1 -- 1+ public_keys : "has"
    public_keys 1 -- 1+ attribute_namespace_public_key_map : "maps"
    public_keys 1 -- 1+ attribute_definition_public_key_map : "maps"
    public_keys 1 -- 1+ attribute_value_public_key_map : "maps"
```

### Checklist


service/policy/objects.proto
```diff
@@ -262,6 +307,21 @@ message KeyAccessServer {
   common.Metadata metadata = 100;
 }
 
+message Key {
+  string id = 1;
+
+  bool is_active = 2;
```
jakedoublev, line 313: Might be better to use `google.protobuf.BoolValue`here so you can differentiate between set and false, set and true, and unset? https://protobuf.dev/reference/protobuf/google.protobuf/#bool-value This is particularly useful when marshaling/unmarshaling to JSON as an unset field value will become false (Go zero value) without this rather than skipped in the JSON.

service/policy/objects.proto
```diff
@@ -101,37 +111,45 @@ message Action {
...
-   with permitted actions TRANSMIT and DECRYPT
+   Example: Subjects in sets 1 and 2 are entitled attribute value
+   http://wwww.example.org/attr/example/value/one with permitted actions
+   TRANSMIT and DECRYPT
    {
     "id": "someid",
     "attribute_value": {example_one_attribute_value...},
-    "subject_condition_set": {"subject_sets":[{subject_set_1},{subject_set_2}]...},
-    "actions": [{"standard": "STANDARD_ACTION_DECRYPT"}", {"standard": "STANDARD_ACTION_TRANSMIT"}]
+    "subject_condition_set":
```
jakedoublev, line 149: I think we've been steadily removing examples from proto comments throughout, and it could be a good time to strip these examples off.

service/policy/objects.proto
jakedoublev: Same with `was_used`?

service/policy/objects.proto
```diff
@@ -262,6 +307,21 @@ message KeyAccessServer {
   common.Metadata metadata = 100;
 }
 
+message Key {
+  string id = 1;
```
jakedoublev, line 311: Can we add a comment that this is not the `kid` but the row ID in the database?

jakedoublev, line 311:
```suggestion
  // the database record ID, not the key ID (`kid`)
  string id = 1;
```

Should we add a comment that this is not the `kid` but the row ID in the database?

service/policy/objects.proto
```diff
@@ -279,15 +339,13 @@ message KasPublicKey {
...
+  // To start, these may be `rsa:2048` for encrypting ZTDF files and
   // `ec:secp256r1` for nanoTDF, but more formats may be added as needed.
   KasPublicKeyAlgEnum alg = 3;
 }
 
 // A list of known KAS public keys
-message KasPublicKeySet {
-  repeated KasPublicKey keys = 1;
-}
+message KasPublicKeySet { repeated KasPublicKey keys = 1; }
```
jakedoublev, line 348: Looking at this makes me wonder - do we want to get rid of the concept of a public key set alongside the deprecation of grants? If so, can we document that?

service/policy/objects.proto
```diff
@@ -247,11 +286,17 @@ message ResourceMapping {
...
-      message: "URI must be a valid URL (e.g., 'https://demo.com/') followed by additional segments. Each segment must start and end with an alphanumeric character, can contain hyphens, alphanumeric characters, and slashes.",
-      expression: "this.matches('^https?://[a-zA-Z0-9]([a-zA-Z0-9\\\\-]{0,61}[a-zA-Z0-9])?(\\\\.[a-zA-Z0-9]([a-zA-Z0-9\\\\-]{0,61}[a-zA-Z0-9])?)*(:[0-9]+)?(/.*)?$')"
-  }];
+  string uri = 2 [ (buf.validate.field).cel = {
+    id : "uri_format",
+    message : "URI must be a valid URL (e.g., 'https://demo.com/') followed by "
+              "additional segments. Each segment must start and end with an "
+              "alphanumeric character, can contain hyphens, alphanumeric "
+              "characters, and slashes.",
+    expression : "this.matches('^https?://"
```
jakedoublev, line 295: Was splitting up the regex like this a requirement from a new linter? I am not sure the formatting improvement is worth the readability impact to the regex. 🤔

service/policy/unsafe/unsafe.proto
```diff
@@ -186,6 +186,19 @@ message UnsafeDeleteAttributeValueResponse {
   policy.Value value = 1;
 }
 
+// WARNING!!
+message UnsafeDeletePublicKeyRequest {
```
jakedoublev, line 190: If delete is unsafe, is update also unsafe?

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -160,36 +230,61 @@ message ListKeyAccessServerGrantsResponse {
...
+  rpc CreateKey(CreateKeyRequest) returns (CreateKeyResponse) {}
+
+  rpc GetKey(GetKeyRequest) returns (GetKeyResponse) {}
+
+  rpc ListKeys(ListKeysRequest) returns (ListKeysResponse) {}
+
+  rpc UpdateKey(UpdateKeyRequest) returns (UpdateKeyResponse) {}
+
+  // DeleteKey is a soft delete, marking the key as inactive.
+  rpc DeleteKey(DeleteKeyRequest) returns (DeleteKeyResponse) {}
```
jakedoublev, line 289: Can we please update this to `DeactivateKey` like the other safe deletion/deactivations of policy objects that are not deletions (because the unique constraints remain enforced by the db when inactive to prevent data leakage)?

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -160,36 +230,61 @@ message ListKeyAccessServerGrantsResponse {
...
-    option (google.api.http) = {get: "/key-access-servers/grants"};
+  rpc ListKeyAccessServerGrants(ListKeyAccessServerGrantsRequest)
+      returns (ListKeyAccessServerGrantsResponse) {
+    option (google.api.http) = {
+      get : "/key-access-servers/grants"
+    };
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+
+  rpc CreateKey(CreateKeyRequest) returns (CreateKeyResponse) {}
```
jakedoublev, line 280: We're abandoning the http gateway intentionally because of Connect, correct?

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -114,39 +116,107 @@ message KeyAccessServerGrants {
...
-// employ selectors for grants to specific policy objects or build the attribute tree relation.
-// If grants to a known namespace, attribute, or value are needed, use the respective
-// GET request to the specific policy object.
+  policy.PageResponse pagination = 10;
+}
+
+message UpdateKeyRequest {
+  // Required
+  string id = 1;
+  bool active = 2;
```
jakedoublev, line 160: Is the supported update functionality strictly deactivation?

service/integration/attributes_test.go
```diff
@@ -1267,6 +1267,18 @@ func (s *AttributesSuite) Test_RemoveKeyAccessServerFromAttribute_Returns_Succes
 	s.Equal(aKas, resp)
 }
 
+func (s *AttributesSuite) Test_GetAttribute_Returns_Only_Active_PublicKeys() {
```
jakedoublev, line 1270: I see why this makes sense, but I wonder how you're envisioning policy administration of these associations once created. For example, if I'm a CLI user getting a unique violation error because I'm trying to create a key that already existed but was deactivated, how do I find that key? I think in my mind `GetAttribute` could maybe give everything in every state, but List APIs all take active state filters in, and the lookup GetByFQN is always strictly active for all policy it returns.

service/policy/db/migrations/20241125220354_keys_table.sql
```diff
@@ -0,0 +1,233 @@
...
+        is_active boolean NOT NULL DEFAULT TRUE,
+        was_used boolean NOT NULL DEFAULT FALSE,
+        key_access_server_id uuid NOT NULL REFERENCES key_access_servers (id),
+        key_id varchar(36) NOT NULL,
+        alg varchar(50) NOT NULL,
+        public_key text NOT NULL,
+        metadata jsonb,
+        created_at timestamp,
+        updated_at timestamp,
+        UNIQUE (key_access_server_id, key_id, alg),
```
jakedoublev, line 15: Is the intended behavior here that we want to allow a single KAS to have multiple keys of different algorithms with the same ID? 

```
KAS_A + kid_123 + rsa:2048
KAS_A + kid_123 + secp:256
```

jakedoublev: If so, or if the desired intention is different and I am misunderstanding, can we please comment in the migration that this is the case?

service/policy/db/models.go
```diff
@@ -55,6 +55,21 @@ func (ns NullAttributeDefinitionRule) Value() (driver.Value, error) {
 	return string(ns.AttributeDefinitionRule), nil
 }
 
+type ActiveDefinitionPublicKeysView struct {
```
jakedoublev, line 58: With the database migrated, would you please run `make policy-erd-gen` to generate the updated ERD?

service/logger/audit/constants.go
```diff
@@ -52,6 +53,7 @@ const (
 	ActionTypeUpdate
 	ActionTypeDelete
 	ActionTypeRewrap
+	ActionTypeSoftDelete
```
jakedoublev, line 56: We've considered [deactivations historically to be updates for audit purposes](https://github.com/opentdf/platform/blob/9e7d5b972f9e55752b79e3191b7625b7389b1b2d/service/policy/attributes/attributes.go#L165). Could you please align all deactivation audit calls to this new soft delete or update the key soft deletes to use the existing `ActionTypeUpdate`?

service/policy/db/key_access_server_registry.go
```diff
@@ -259,3 +261,206 @@ func (c PolicyDBClient) ListKeyAccessServerGrants(ctx context.Context, r *kasreg
...
+		Alg:               key.GetAlg().String(),
+		PublicKey:         key.GetPem(),
+		Metadata:          metadataJSON,
+	})
+	if err != nil {
+		return nil, db.WrapIfKnownInvalidQueryErr(err)
+	}
+
+	// Get freshly created key
+	ck, err := c.GetPublicKey(ctx, &kasregistry.GetKeyRequest{Id: id})
```
jakedoublev, line 286: We've been trying to avoid doing immediate GETs during CREATEs to prevent data consistency issues in a system with multiple DB replicas (planning in advance). Would you mind employing either:
- a transaction https://github.com/opentdf/platform/blob/9e7d5b972f9e55752b79e3191b7625b7389b1b2d/service/policy/db/policy.go#L37
- populating everything in the response from the `RETURNING` SQL and request?

service/policy/db/migrations/20241125220354_keys_table.sql
```diff
@@ -0,0 +1,233 @@
...
+    public_keys AS ky
+    INNER JOIN attribute_definition_public_key_map AS km ON ky.id = km.key_id
+    LEFT JOIN key_access_servers AS kas ON ky.key_access_server_id = kas.id
+WHERE
+    ky.is_active = TRUE
+GROUP BY
+    km.definition_id;
+
+-- Trigger for attribute_value_key_map
+CREATE TRIGGER trigger_update_was_used_value
```
jakedoublev, line 169: Can you please help me understand the tradeoff you chose keeping this permanent state in sync with a column and transactions instead of doing the JOINs as needed when checking if a key is assigned to a policy object?

service/policy/db/migrations/20241125220354_keys_table.sql
```diff
@@ -0,0 +1,233 @@
+-- +goose Up
```
jakedoublev, line 1: Could you please add a markdown file `service/policy/db/migrations/20241125220354_keys_table.md` that links to the ADR so we have that historical link between the migration and the relational change?

jakedoublev (commented):
This is a lot of work. Thanks for tackling this @strantalis.

service/policy/kasregistry/key_access_server_registry.proto
jakedoublev: Should we make the paths clearer between deactivation, reactivation, and a true update (change of KID, alg, or PEM) even if those are all unsafe?

service/policy/unsafe/unsafe.proto
strantalis: Update only allows modification if `is_active`. Not sure if thats an unsafe action though.

service/policy/kasregistry/key_access_server_registry.proto
strantalis: Yeah we are and I didn't want to add more to be removed.

service/policy/kasregistry/key_access_server_registry.proto
strantalis: I think adding deactivation and reactivation makes sense. 

@cassandrabailey293 @jp-ayyappan When managing public keys should we allow modification of that resource or should we force an admin to create a new key so they can retain it for audit purposes in the database.

service/policy/objects.proto
strantalis: I must have a linter locally that did this.

service/policy/objects.proto
strantalis: Yeah that should go away. I believe I only leveraged `KasPublicKey`. I will go through and add deprecation comments to that and kas grants.

service/policy/db/key_access_server_registry.go
strantalis: wrapped it in a transaction

service/policy/unsafe/unsafe.proto
strantalis: I removed the modification of is_active. I am thinking keys should be immutable.

jakedoublev:
I am comfortable approving this now but would like to give others a chance to review as well.

service/integration/kas_registry_test.go
```diff
@@ -1013,6 +1016,357 @@ func (s *KasRegistrySuite) Test_ListKeyAccessServerGrants_Offset_Succeeds() {
...
+func (s *KasRegistrySuite) Test_Create_Public_Key() {
+	// The initial rsa2048 public key is created in the fixture and should be active
+	kID := s.f.GetPublicKey("key_1").ID
+	r1, err := s.db.PolicyClient.GetPublicKey(s.ctx, &kasregistry.GetKeyRequest{Id: kID})
+	s.Require().NoError(err)
+	s.NotNil(r1)
+	s.True(r1.GetKey().GetIsActive().GetValue())
+
+	kasID := s.f.GetKasRegistryKey("key_access_server_1").ID
+	kasRegistry := &kasregistry.CreateKeyRequest{
```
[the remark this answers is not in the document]
strantalis: I updated everything to include PublicKey in the name for consistency.

service/integration/kas_registry_test.go
```diff
@@ -1013,6 +1016,357 @@ func (s *KasRegistrySuite) Test_ListKeyAccessServerGrants_Offset_Succeeds() {
...
+		s.False(slices.ContainsFunc(r.GetKeys(), func(key *policy.Key) bool {
+			return key.GetId() == kID
+		}))
+	}
+}
+
+func (s *KasRegistrySuite) Test_Create_Pulblic_Key_Unique_Constraint() {
+	// We can't have a duplicate public keys with the same (key_access_server_id, kid, alg) set
+	kasID := s.f.GetKasRegistryKey("key_access_server_1").ID
+	kasRegistry := &kasregistry.CreateKeyRequest{
```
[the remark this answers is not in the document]
strantalis: Yeah I think this is a good callout. Test could easily be executed in a different order which would break the tests.

service/policy/db/key_access_server_registry.go
```diff
@@ -259,3 +262,264 @@ func (c PolicyDBClient) ListKeyAccessServerGrants(ctx context.Context, r *kasreg
...
+	// Validate kas_id is uuid if set
+	if r.GetKasId() != "" {
+		if _, err := uuid.Parse(r.GetKasId()); err != nil {
+			return nil, db.StatusifyError(err, db.ErrEnumValueInvalid.Error())
+		}
+	}
+
+	params := listPublicKeysParams{
+		KasID: r.GetKasId(),
+		// KasUri:  r.GetKasUri(),
```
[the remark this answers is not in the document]
strantalis: What are your thoughts on this? I never circled back on it but thought it would be useful. Maybe its something we keep but implement in the future?

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -87,109 +95,251 @@ message UpdateKeyAccessServerRequest {
...
+  repeated Key keys = 1; 
+
+  policy.PageResponse pagination = 10;
+}
+
+message ListPublicKeyMappingRequest {
+  // Optional
+  string kas_id = 1;
+  // Optional
+  string kas_name = 2;
```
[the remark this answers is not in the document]
strantalis: I commented them out for now.

service/policy/db/key_access_server_registry.go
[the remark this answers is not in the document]
jakedoublev: That sounds good to me. I know we've had feedback that KASes are known by URIs and common names that are known to be unique, so allowing filtering by those instead of the ID of a KAS is a nicer UX. That said, it'd be good to merge this and extend as needed rather than expanding scope further.

jakedoublev (approved):

PR #1846, opened 2025-01-08
fix(core): Return deny decision for empty entity chains

elizabethhealy:
### Proposed Changes

* deny for entity chain with no entities
* permit for resource attribute with no fqns

### Checklist

### DIFF
--- a/service/authorization/authorization_test.go
+++ b/service/authorization/authorization_test.go
@@ -1786,6 +1786,29 @@ func Test_GetDecisions_Empty_EC_RA(t *testing.T) {
 			numDecisions:   1,
 			decisionResult: authorization.DecisionResponse_DECISION_PERMIT,
 		},
+		{
+			name: "Entity Chain with empty entity list and Resource attribute with empty fqn list",
+			req: connect.Request[authorization.GetDecisionsRequest]{
+				Msg: &authorization.GetDecisionsRequest{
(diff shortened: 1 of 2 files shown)


dmihalcik-virtru (approved):

service/authorization/authorization_test.go
```diff
@@ -1633,3 +1627,200 @@ func Test_GetDecisionsAllOf_Pass_EC_RA_Length_Mismatch(t *testing.T) {
...
+	}
+
+	///////////// Run tests /////////////////////
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			resp, err := as.GetDecisions(ctxb, &tc.req)
+			require.NoError(t, err)
+			assert.NotNil(t, resp)
+			assert.Len(t, resp.Msg.GetDecisionResponses(), tc.numDecisions)
+			if tc.decisionResult != authorization.DecisionResponse_DECISION_UNSPECIFIED {
```
dmihalcik-virtru, line 1821: This seems like an unnecessary branch

elizabethhealy: there are test cases i want to test with no decision responses returned, thus `resp.Msg.GetDecisionResponses()[0]` in the line below would give an index error, i only want to run the assert below if the test case expects there to be a decision result in the list
though i could change this if to `if tc.numDecisions != 0` instead

PR #1853, opened 2025-01-09
chore(docs): How to think about errors

dmihalcik-virtru:
### Proposed Changes

* Adds some guidance for go developers when they add code that changes how errors are created, returned, or expected to be handled.

### Checklist


Contributing-errors.md
```diff
@@ -0,0 +1,152 @@
...
+it would make sense to say “Unsupported curve, use  SECP11R2”.
+This exception exists because the user is tweaking the curve.
+
+Expect that people have not read the documentation and don’t know the spec.
+
+> Applications can use the `errors.Is` and `As` methods to trigger appropriate behavior in response to an error condition.
+
+#### *actionable* - errors should be actionable to enable the developer to fix their own problem
+
+People like feedback [when it's actionable](https://fortune.com/2023/10/09/analyzed-2-years-performance-reviews-13000-workers-proof-low-quality-feedback-driving-employee-retention-down-careers-snyder-yen/) otherwise they look for alternatives.
```
patmantru, line 123: paywalled article

Contributing-errors.md
```diff
@@ -0,0 +1,163 @@
...
+- command line [status codes](https://tldp.org/LDP/abs/html/exitcodes.html)
+
+This document focuses on the first category, go `error` typed return values.
+
+### Concerns for Golang Library Developers
+
+- When a library does not provide explicitly exported and documented error types and constants,
+  application developers might depend on `strings.Contains()`
+  which bind their applications to the error message and prohibit improving error language
+  - Prefer using sentinel values for expected conditions
```
mkleene, line 22: an example of this might be nice, especially if we have an example in our code

mkleene (dismissed):
This is really good. The only thing that I can think of that might make it a bit clearer is a couple of code examples.

Contributing-errors.md
```diff
@@ -0,0 +1,163 @@
...
+
+A good illustration of this is when English speakers use libraries like Ant Design (Alibaba backed frontend design system).
+The value of Ant Design as a product excuses some of the grammar and language barriers for native English speakers,
+but anecdotally, it seems that English-as-a-second-language (ESL) speakers (who also don’t know Chinese) tend to favor English-driven tooling.
+
+Errors should use proper grammar and spelling.
+Usefulness and accuracy of errors is reenforced by the quality of the grammar and spelling of errors.
+This is especially true when we are striving for international usage while not having the capacity for internationalization.
+
+Errors should be idiomatic with the language.
```
jakedoublev, line 108: Can we specify a noun -> adjective structure in error naming? For example, we would prefer `ErrCredentialsMissing` and `ErrCredentialsInvalid` instead of `ErrMissingCredentials` or `ErrInvalidCredentials`. This will help with grouping similar errors to reduce cognitive load.

Contributing-errors.md
dmihalcik-virtru: Looking at the code we don't do a good job of exporting or naming things. I think the only good example is the `sdk.ErrInvalid`, but even that is a little wrong (it should be `sdk.ErrReadInvalid` or `sdk.ErrTDFInvalid` following the rules below, since the SDK might also have invalid policy related requests).

jakedoublev (approved):

PR #1860, opened 2025-01-15
feat(core): Adds ErrInvalidPerSchema

dmihalcik-virtru:
### Proposed Changes

# `IsValid` enhancements:

- Allows some ztdf manifest fields to be set to `null`, to allow for an improperly coded client that was emitting them with these values instead of leaving them undefined.
- These were all optional, except `assertion.appliesToState`, which I guess we can assume has a default? Not sure but I guess it is `encrypted`?
- Adds some tests to the example `isvalid` command
- Adds a new exported sentinel error, ErrInvalidPerSchema. This can be used to check for 'only kinda' invalid files, that might still load using the Read method.
- Adds a properly failing manifest test that is as the comment describes

### Checklist


pflynn-virtru (approved):

elizabethhealy (approved):
lgtm, only comment is would it be better to have the isvalid tests at the unit level instead of from examples? or does isvalidtdf require a connection to kas that would require it to run with the rest of the example commands

PR #1887, opened 2025-01-28
fix: cleanup kas public key create error messages

strantalis:
### Proposed Changes

As I was working on the cli the error messages I originally implemented felt too generic. Here is another stab at making them more clear when create a new public key fails.

### Checklist


dmihalcik-virtru (approved):
you don't have to commit go.work.sum

As far as content goes, this will help track debugging so I like it, but I'm not sure how much use it will be to otdfctl?

jakedoublev (approved):

strantalis:
@dmihalcik-virtru 
> you don't have to commit go.work.sum
> 

Are you typically ignoring any changes that automatically happen locally to it? 

> As far as content goes, this will help track debugging so I like it, but I'm not sure how much use it will be to otdfctl?

Right now the cli doesn't do anything special and just returns the error message it gets back from an rpc call so I think it will help a little.

PR #1889, opened 2025-01-30
fix: add pagination to list public key mappings response

strantalis:
### Proposed Changes

Missed adding pagination response to the list public key mappings response. Also includes additional tests for limiting the result set and also testing of filters by kas id and public key id. 

### Checklist


elizabethhealy (approved):

service/policy/db/key_access_server_registry.go
```diff
@@ -426,15 +427,27 @@ func (c PolicyDBClient) ListPublicKeyMappings(ctx context.Context, r *kasregistr
...
 		if err != nil {
 			return nil, db.WrapIfKnownInvalidQueryErr(err)
 		}
 		mappings[i] = pkm
 	}
 
+	var total int32
+	var nextOffset int32
+	if len(listRows) > 0 {
+		total = int32(listRows[0].Total)
```
pflynn-virtru, line 440: is `[0]`right?

strantalis: I took this from the previous pattern of list operations. But I believe its right as every row will have the total count included in it. 

@jakedoublev Am I off on the thinking here?

jakedoublev: That's accurate thinking @strantalis (https://github.com/opentdf/platform/pull/1889/files#diff-92033a056a1e64fa5b759cf500f2644f7e69e51fb26fa231dbe947bac600bf3fR987). The count is found in advance as a CTE and returned in every result row. There's probably a better way to do it, but this is optimal on count time and leads to a more maintainable query than gathering a count and all list results as JSON with a giant `JSONB_BUILD_OBJECT`.

jakedoublev (approved):

PR #1901, opened 2025-02-03
feat: add ability to retrieve policy resources by id or name

strantalis:
### Proposed Changes

The pull request introduces the concept of an `identifier` on a set of `Get` rpc's opening up the ability get a resource by more than it's `id`. 

On the new `ListPublicKeys` and `ListPublicKeyMappings` rpc's a new `oneof` field was added so a user could filter by kas id, name or uri. This presents as a proto breaking change but because we haven't released these new features it is okay to bypass this check. 

### Checklist


service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -158,13 +169,14 @@ message ListPublicKeysResponse {
...
 
 message ListPublicKeyMappingRequest {
-  // Optional
-  string kas_id = 1;
-  // Future filter by fields
-  // // Optional
-  // string kas_name = 2;
-  // // Optional
-  // string kas_uri = 3;
+  oneof identifier {
```
strantalis, line 172: I am not concerned about these breaking changes as we haven't released this new functionality yet.

service/policy/attributes/attributes.go
```diff
@@ -96,9 +96,17 @@ func (s *AttributesService) GetAttribute(ctx context.Context,
 ) (*connect.Response[attributes.GetAttributeResponse], error) {
 	rsp := &attributes.GetAttributeResponse{}
 
-	item, err := s.dbClient.GetAttribute(ctx, req.Msg.GetId())
+	var identifier any
+
+	if req.Msg.GetId() != "" { //nolint:staticcheck // Id can still be used until removed
+		identifier = req.Msg.GetId() //nolint:staticcheck // Id can still be used until removed
```
jrschumacher, line 102: I don't understand why we need this if it's only running in the platform. Is this related to ConnectRPC and legacy gRPC requests?

service/policy/attributes/attributes.proto
```diff
@@ -86,10 +86,17 @@ message ListAttributesResponse {
 }
 
 message GetAttributeRequest {
-  // Required
+  // Deprecated
   string id = 1 [
```
jakedoublev, line 90: I'd like to discuss this breaking change and the pros/cons. An alternative could be allowing either `id` or `fqn` without `oneof`, then logically preferring `id` in the event both were provided.

service/policy/attributes/attributes_test.go
```diff
@@ -412,6 +428,14 @@ func TestGetAttributeValueRequest(t *testing.T) {
 	}
 	err = getValidator().Validate(req)
 	require.NoError(t, err)
+
+	req = &attributes.GetAttributeValueRequest{
+		Identifier: &attributes.GetAttributeValueRequest_ValueId{
```
jakedoublev, line 433: should we do an FQN test here?

service/policy/db/attribute_values.go
```diff
@@ -41,8 +41,27 @@ func (c PolicyDBClient) CreateAttributeValue(ctx context.Context, attributeID st
...
+		av, err = c.Queries.GetAttributeValueById(ctx, i.ValueId)
+	case *attributes.GetAttributeValueRequest_Fqn:
+		var fv GetAttributeValueByFqnRow
+		fv, err = c.Queries.GetAttributeValueByFqn(ctx, i.Fqn)
+		// Same struct fields allow for struct casting
+		av = GetAttributeValueByIdRow(fv)
+	case string:
+		av, err = c.Queries.GetAttributeValueById(ctx, i)
+	default:
+		// Hopefully this will never happen
```
jakedoublev, line 61: If we keep a switch for the oneof / deprecation, can we defined a typed error for this then wrap it with the unknown type?

```golang
return nil, fmt.Errorf("type: %T, %w", ErrTypeIdentifierInvalid")
```

service/policy/attributes/attributes.go
strantalis: No this is because I moved the fields to a oneof in the proto definition. So the id field would be removed in a later release. 

The intention of the oneof was so a user could only set either x_id or fqn/name. But after syncing with @jakedoublev it might not be worth the extra effort to move to the oneof.

service/policy/attributes/attributes.proto
strantalis: This was discussed with @jakedoublev and @jrschumacher with a decision to move forward with the use of oneof

service/policy/db/query.sql
```diff
@@ -87,13 +87,15 @@ OFFSET @offset_;
...
-    name,
+    kas.id,
+    kas.uri, 
+    kas.public_key, 
+    kas.name,
     JSON_STRIP_NULLS(JSON_BUILD_OBJECT('labels', metadata -> 'labels', 'created_at', created_at, 'updated_at', updated_at)) AS metadata
-FROM key_access_servers
-WHERE id = $1;
+FROM key_access_servers AS kas
+WHERE (sqlc.narg('id')::uuid IS NULL OR kas.id = sqlc.narg('id')::uuid)
```
dmihalcik-virtru, line 96: Why not use the `@` macro here instead? I think we are using it elsewhere already

```suggestion
WHERE (@id::uuid IS NULL OR kas.id = @id::uuid)
```

service/go.mod
```diff
@@ -120,7 +120,7 @@ require (
 	github.com/opencontainers/image-spec v1.1.0 // indirect
 	github.com/opentracing/opentracing-go v1.2.0 // indirect
 	github.com/pelletier/go-toml/v2 v2.1.1 // indirect
-	github.com/pkg/errors v0.9.1 // indirect
+	github.com/pkg/errors v0.9.1
```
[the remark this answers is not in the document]
strantalis: I don't think this was copilot. A `go mod tidy` removes that `// indirect`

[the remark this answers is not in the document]
strantalis: Actually this is a good catch. It was either co-pilot or the auto adding of an import that added the wrong errors pkg

service/policy/objects.proto
```diff
@@ -328,15 +328,31 @@ enum KasPublicKeyAlgEnum {
 // A KAS public key and some associated metadata for further identifcation
 message KasPublicKey {
   // x509 ASN.1 content in PEM envelope, usually
-  string pem = 1;
+  string pem = 1 [
+     (buf.validate.field).string = {
+        min_len: 1,
+        max_len: 2048
```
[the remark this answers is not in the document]
strantalis: yeah I think this makes sense

service/policy/objects.proto
```diff
@@ -328,15 +328,31 @@ enum KasPublicKeyAlgEnum {
...
+      }
+  ];
 
   // A unique string identifier for this key
-  string kid = 2;
+  string kid = 2 [
+    (buf.validate.field).string = {
+      min_len : 1,
+      max_len : 32
```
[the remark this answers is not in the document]
strantalis: Isn't a `max_len` going to be problematic in the future if we start to leverage these for key splitting in nanotdf. Or do you see us marking a key for use with nanotdf or tdf?

service/policy/db/query.sql
strantalis: That is a shortcut for `sqlc.arg()` not `sqlc.narg()` which are nullable parameters. Allowing nullable parameters allows us to query by different parameters.

dmihalcik-virtru (approved):

jrschumacher (approved):
Approving to satisfy requirement. We will be addressing codeowners issues.

PR #1903, opened 2025-02-04
feat(sdk): DSP-417 - Code flow port configurable

c-r33d:
### Proposed Changes

1. Allow code flow port to be configurable.
2. Default to 9000 if not specified.
3. Expose code flow port in SDK through PlatformConfiguration.

### Checklist


docs/Configuring.md
```diff
@@ -79,6 +88,7 @@ Root level key `server`
 | `auth.dpopskew`         | The amount of time drift allowed between when the client generated a dpop proof and the server time.          | `1h`    | OPENTDF_SERVER_AUTH                  |
 | `auth.skew`             | The amount of time drift allowed between a tokens `exp` claim and the server time.                            | `1m`    | OPENTDF_SERVER_AUTH_SKEW             |
 | `auth.public_client_id` | The oidc client id. This is leveraged by otdfctl.                                                             |         | OPENTDF_SERVER_AUTH_PUBLIC_CLIENT_ID |
+| `auth.code_flow_port`   | The port used to perform oidc code flow authorization. This is leveraged by otdfctl.                          | `9000`  | OPENTDF_SERVER_AUTH_CODE_FLOW_PORT   |
```
jakedoublev, line 91: We should probably associate this new value's name with the `public_client`. I think we should also forgo calling it out for the code flow specifically because it's intended to be allowlisted in the idP as a valid `redirectUri` but not specifically coupled to one Oauth flow.

Lastly, I think consuming PEPs like `otdfctl` may choose to default to 9000, but it seems like a potential footgun to default the config value as it could result in a case where the `.well-known` reports that there is a valid local port that is not in fact valid or allowlisted.

```suggestion
| `auth.public_client_local_port`   | The localhost port added to the idP as a valid redirectUri for the public_client_id. This is leveraged by otdfctl.                          |   | OPENTDF_SERVER_AUTH_PUBLIC_CLIENT_LOCAL_PORT   |
```

Does all that make sense, or do you disagree with any of it?

c-r33d: Yes, I believe the changes make sense. 

I debated on making the port default to 9000 for that reason, but I believe it's probably better to let the consumer decide how they want to handle a missing value and not have our code make assumptions on how the admin setup the platform.

strantalis:
@c-r33d I need to understand this change better. The platform today has no involvement with this type of flow. Is the port needed for the otdfctl redirect?

c-r33d:
> @c-r33d I need to understand this change better. The platform today has no involvement with this type of flow. Is the port needed for the otdfctl redirect?

@strantalis 
Yes, it is for the redirect url that is passed as a parameter during a [code flow login](https://github.com/opentdf/otdfctl/blob/main/pkg/auth/auth.go#L235). 

An example of how the redirectUrl is added to the **opentdf-public** client is [here](https://github.com/opentdf/platform/blob/main/service/cmd/keycloak_data.yaml#L79-L80).

Without the redirectUrl being added to the client the code flow logic will be denied. We decided that it would be nice to have the port configurable and exposed via the well-known/configuration, assuming other clients will use the same flow. (Public client/Public client port).

service/internal/auth/config.go
```diff
@@ -17,14 +17,15 @@ type Config struct {
...
-	PublicClientID string        `mapstructure:"public_client_id" json:"public_client_id,omitempty"`
+	EnforceDPoP           bool          `mapstructure:"enforceDPoP" json:"enforceDPoP" default:"false"`
+	Issuer                string        `mapstructure:"issuer" json:"issuer"`
+	Audience              string        `mapstructure:"audience" json:"audience"`
+	Policy                PolicyConfig  `mapstructure:"policy" json:"policy"`
+	CacheRefresh          string        `mapstructure:"cache_refresh_interval"`
+	DPoPSkew              time.Duration `mapstructure:"dpopskew" default:"1h"`
+	TokenSkew             time.Duration `mapstructure:"skew" default:"1m"`
+	PublicClientID        string        `mapstructure:"public_client_id" json:"public_client_id,omitempty"`
+	PublicClientLocalPort string        `mapstructure:"public_client_local_port" json:"public_client_local_port,omitempty"`
```
dmihalcik-virtru, line 28: This is a client feature, so it appears odd to include it in the service configuration. Does OTDFCTL not have a separate configuration system?

jakedoublev: There's a few pieces there, but here are some general thoughts.

### otdfctl CLI

The `otdfctl` CLI is a PEP and administrative tool that is platform-agnostic. Through the use of profiles or flags/arguments, it can communicate with `n` number of local/hosted platforms connected to as many idPs.

As a result, it relies on the platform's `.well-known` config to provide the `public_client_id` to authenticate CLI users through (with the authz endpoint and token endpoint). It's very thin by design.

### Public Client

The `public_client_id` was added to provide a general client (when configured) for "public" OAuth flows (no client secret).

1. browser flows - a browser-based PEP (likely served behind the running platform) needs to authenticate users, it can defer to the `public_client_id` in the `.well-known` to authenticate users, and it should already assume its origin is allowListed for redirects as it cannot change its origin
2. native flows - a native application (comprised of static assets and a process on an operating system) is likely running on a custom scheme `myapp://` and to authenticate more seamlessly could start up a localhost process on a port to facilitate auth
3. automations - as an automation is a process in itself, a second process must be started to facilitate an auth flow (i.e. a localhost process with a port)
4. CLI - as the CLI is a process, a second process must be started to facilitate an auth flow (i.e. a localhost process with a port)

If we elected not to add this port to the `.well-known` config alongside the `public_client_id`, we could either:

1. remove the public client as well and require all PEPs to have foreknowledge of clients when configured
2. try to allowlist all local ports for the public client if all idPs allow this `http://localhost*`
3. maintain the current assumption of a single port hardcoded and allowlisted for local server processes

PR #1905, opened 2025-02-04
fix(core): Requires unique kids

dmihalcik-virtru:
### Proposed Changes

- When loading KAS keys, validate that key identifiers are unique
- This will be required as we move to a key ID - first (instead of alg first) method of selecting keys
- While here, moved some of the 'legacy conversion' code into more testable functions and generally improved testing and handling/detection of error conditions, including increasing test coverage of loading/displaying EC keys

### Checklist

### DIFF
--- a/service/internal/security/standard_crypto.go
+++ b/service/internal/security/standard_crypto.go
@@ -220,6 +220,7 @@ func loadDeprecatedKeys(rsaKeys map[string]StandardKeyInfo, ecKeys map[string]St
 
 	return &StandardCrypto{
 		keysByAlg: keysByAlg,
+		keysByID:  keysByID,
 	}, nil
 }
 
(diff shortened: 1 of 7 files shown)


service/internal/security/standard_crypto.go
```diff
@@ -241,11 +251,7 @@ func (s StandardCrypto) RSAPublicKey(kid string) (string, error) {
 }
 
 func (s StandardCrypto) ECCertificate(kid string) (string, error) {
-	ecKeys, ok := s.keys[AlgorithmECP256R1]
-	if !ok || len(ecKeys) == 0 {
-		return "", ErrCertNotFound
-	}
-	k, ok := ecKeys[kid]
+	k, ok := s.keysByID[AlgorithmECP256R1]
```
jentfoo, line 254: This looks like a bug, should this be lookup by id?

service/internal/security/standard_crypto.go
```diff
@@ -257,11 +263,7 @@ func (s StandardCrypto) ECCertificate(kid string) (string, error) {
 }
 
 func (s StandardCrypto) ECPublicKey(kid string) (string, error) {
-	ecKeys, ok := s.keys[AlgorithmECP256R1]
-	if !ok || len(ecKeys) == 0 {
-		return "", ErrCertNotFound
-	}
-	k, ok := ecKeys[kid]
+	k, ok := s.keysByID[AlgorithmECP256R1]
```
jentfoo, line 266: bug?

service/internal/security/standard_crypto.go
dmihalcik-virtru: yeah and a good argument for more unit test coverage

mkleene (approved):
looks good, the only thing I'd say suggest is that we could be a bit more fussy with the types and have a `map[string]rsaKeyLIst` or something that that it's clear that we will never fail to pull an RSA or ECC key out of the map.

jentfoo (approved):

PR #1910, opened 2025-02-06
fix: Improve http.Client usage for security and performance

### LINKED ISSUE #1891: Improve http.Client usage
We should improve our usage with `http.Client`.  We don't currently reuse http clients, or define them in a consistent way in our Go code.  This leads to several potential concerns:
* A lack of reuse prevents the ability to gain potential performance gains through connection pooling.  SSL handshakes are notably slow, we likely are able to achieve real performance gains by being able to reuse already established connections.
* A lack of timeout configuration could be a point of DoS risk.
* A lack of redirect limitations could result in a trusted source initiating a blind SSRF.

I plan to address this by creating a public function for returning a well configured http.Client from the `sdk`.  Then utilize this client from within the SDK, platform, and backend broadly.  If anyone has feedback on this design let me know in Slack or once I submit the PR.

jentfoo:
### Proposed Changes
This change switches us from maintaining a tls config which we then on-demand initialize an `http.Client` with to instead maintain and reuse an `http.Client` instance. This enables us to utilize the connection pooling which occurs within the `http.Transport` to reduce ssl handshakes and thus reduce latency.

In addition this change provides us a central place to configure out `http.Client` (`httputil`). Allowing us to easily set configuration options to reduce the security risks of using an unconfigured `http.Client`. Notably timeouts to reduce DoS risks, and control around following redirects to prevent blind SSRF's.

This PR 3/4 addresses #1891.  After it's merged a small change will be made in the `service` to fully utilize the API being introduced in the SDK here.

### Checklist

### DIFF
--- a/sdk/httputil/http.go
+++ b/sdk/httputil/http.go
@@ -11,7 +11,7 @@ const (
 	// defaults to match DefaultTransport - defined to satisfy lint
 	maxIdleConns          = 100
 	idleConnTimeout       = 90 * time.Second
-	tLSHandshakeTimeout   = 10 * time.Second
+	tlsHandshakeTimeout   = 10 * time.Second
 	expectContinueTimeout = 1 * time.Second
 )
 
@@ -44,7 +44,7 @@ func SafeHTTPClientWithTLSConfig(cfg *tls.Config) *http.Client {
 		ForceAttemptHTTP2:     true,
 		MaxIdleConns:          maxIdleConns,
 		IdleConnTimeout:       idleConnTimeout,
-		TLSHandshakeTimeout:   tLSHandshakeTimeout,
+		TLSHandshakeTimeout:   tlsHandshakeTimeout,
 		ExpectContinueTimeout: expectContinueTimeout,
 	})
 }
--- a/sdk/httputil/http.go
+++ b/sdk/httputil/http.go
@@ -1,0 +1,64 @@
+package httputil
+
+import (
+	"crypto/tls"
+	"net/http"
+	"time"
+)
+
+const (
+	defaultTimeout = 120 * time.Second
+	// defaults to match DefaultTransport - defined to satisfy lint
+	maxIdleConns          = 100
+	idleConnTimeout       = 90 * time.Second
+	tLSHandshakeTimeout   = 10 * time.Second
+	expectContinueTimeout = 1 * time.Second
+)
+
+var preventRedirectCheck = func(_ *http.Request, _ []*http.Request) error {
+	return http.ErrUseLastResponse // Prevent following redirects
+}
+
+var safeHTTPClient = &http.Client{
+	Transport:     http.DefaultTransport,
+	Timeout:       defaultTimeout,
+	CheckRedirect: preventRedirectCheck,
+}
+
+// SafeHTTPClient returns a default http client which has sensible timeouts, won't follow redirects, and enables idle
+// connection pooling.
+func SafeHTTPClient() *http.Client {
+	return safeHTTPClient
+}
+
+// SafeHTTPClientWithTLSConfig returns a http client which has sensible timeouts, won't follow redirects, and if
+// specified a http.Transport with the tls.Config provided.
+func SafeHTTPClientWithTLSConfig(cfg *tls.Config) *http.Client {
+	if cfg == nil {
+		return safeHTTPClient
+	}
+	return SafeHTTPClientWithTransport(&http.Transport{
+		TLSClientConfig: cfg,
+		// config below matches DefaultTransport
+		Proxy:                 http.ProxyFromEnvironment,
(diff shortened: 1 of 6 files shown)


go.work.sum
```diff
@@ -2043,6 +2043,7 @@ golang.org/x/net v0.20.0/go.mod h1:z8BVo6PvndSri0LbOE3hAn0apkU+1YvI6E70E9jsnvY=
 golang.org/x/net v0.23.0/go.mod h1:JKghWKKOSdJwpW2GEx0Ja7fmaKnMsbu+MWVZTokSYmg=
 golang.org/x/net v0.25.0/go.mod h1:JkAGAh7GEvH74S6FOH42FLoXpXbE/aqXSrIQjXgsiwM=
 golang.org/x/net v0.26.0/go.mod h1:5YKkiSynbBIh3p6iOc/vibscux0x38BZDkn8sCUPxHE=
+golang.org/x/net v0.28.0/go.mod h1:yqtgsTWOOnlGLG9GFRrK3++bGOUEkNBoHZc8MEDWPNg=
```
jentfoo, line 2046: It's unclear to me why this was not committed?  It might be worth a CI check to ensure the go.* is not mutated on build.

sdk/httputil/http.go
```diff
@@ -0,0 +1,64 @@
...
+	defaultTimeout = 120 * time.Second
+	// defaults to match DefaultTransport - defined to satisfy lint
+	maxIdleConns          = 100
+	idleConnTimeout       = 90 * time.Second
+	tLSHandshakeTimeout   = 10 * time.Second
+	expectContinueTimeout = 1 * time.Second
+)
+
+var preventRedirectCheck = func(_ *http.Request, _ []*http.Request) error {
+	return http.ErrUseLastResponse // Prevent following redirects
```
jentfoo, line 19: Completely preventing redirects may not be realistic.  However I am hopeful, because determining what IDOR's are allowed can be challenging.

I am proposing this initially, but it need more testing. If we are uncomfortable with this, maybe we can add some logging or metrics that we could use to help validate the safety here?

jrschumacher: Is this something where an end-user might want to support following redirects? Or maybe an allowlist of redirect endpoints? The latter seems more complicated than adding `sdk.WithUnsafeFollowRedirects()`

sdk/httputil/http.go
```diff
@@ -0,0 +1,64 @@
...
+	return SafeHTTPClientWithTransport(&http.Transport{
+		TLSClientConfig: cfg,
+		// config below matches DefaultTransport
+		Proxy:                 http.ProxyFromEnvironment,
+		ForceAttemptHTTP2:     true,
+		MaxIdleConns:          maxIdleConns,
+		IdleConnTimeout:       idleConnTimeout,
+		TLSHandshakeTimeout:   tLSHandshakeTimeout,
+		ExpectContinueTimeout: expectContinueTimeout,
+	})
```
jrschumacher, line 49: We might want to expose something in the sdk which will allow developers to change these defaults (since they are defaults). This way we can support a variety of network environments, especially over remote areas with high-latency.

go.work.sum
dmihalcik-virtru: We probably should add this file to .gitignore. This file is very noisy, as it is dependent on the union of all transitive deps for the repository and the patch version of go used to generate it.

This and the .work file aren't used downstream. The release CI deletes this file and recreates it as needed. Both are checked in as a convenience for developers. Previously, we had developers run a shell script to initialize the workspace, but that proved too difficult for people onboarding.

sdk/httputil/http.go
jentfoo: > Is this something where an end-user might want to support

Not unless we add an allow list / allowed IDOR configuration.  Without that configuration this feature is insecure and could be used to make internal requests with the help of a redirect.

Since configuration adds complexity, I think we should only add it if we find it's necessary. There should not be a large need for this, so it would be good to understand that need first.

sdk/httputil/http.go
jentfoo: I don't think most users will need to concern themselves with this.  But if they have a long running process with spare memory it's conceivable that someone may want to change it.  However I don't believe it should be necessary, and even without precise tuning this will be significantly better than what we have today.

I guess generally I prefer to just keep configuration simple and defaults sane, and only bring in the config complexity when a clear use case is understood.  It's easy to expand, but hard to retract.

That said, no strong preference here since tuning these defaults wont increase risk to our users. If you would like them to be configurable should I just add those values to the `config` struct?  Let me know your preference.

dmihalcik-virtru (approved):

dmihalcik-virtru (approved):

strantalis (approved):

PR #1914, opened 2025-02-11
fix(sdk): Fix compatibility between bulk and non-bulk rewrap

dmihalcik-virtru:
### Proposed Changes

* Fixes for version skew in the rewrap introduced by the bulk change
* Pre-bulk services' responses are now interpretable from responses from post-bulk sdks
* Post-bulk SDKs now fill out necessary fields for pre-bulk services
* Post-bulk services have slightly better parsing behavior

### Checklist


strantalis (commented):
Are there any unit / integration tests we can update or add to verify this functionality?

dmihalcik-virtru:
> Are there any unit / integration tests we can update or add to verify this functionality?

I've added a few test cases in `TestUpgradeRewrapRequestV1Suite`

sdk/kas_client.go
```diff
@@ -72,9 +76,38 @@ func (k *KASClient) makeRewrapRequest(ctx context.Context, requests []*kas.Unsig
 		return nil, fmt.Errorf("error making rewrap request: %w", err)
 	}
 
+	upgradeRewrapResponseV1(response, requests)
```
elizabethhealy, line 79: if we convert v1 responses to v2, is that still compatible with clients expecting v1 responses?

elizabethhealy (approved):

sdk/kas_client.go
dmihalcik-virtru: They will ignore unknown fields.

PR #1925, opened 2025-02-18
feat: add option to resolve groups for users in Keycloak

dschmidt:
### Proposed Changes

Optionally load groups when resolving users from keycloak.
This allows specifying condition sets based on group membership.

I can add tests if you are interested in this patch at all.

### Checklist


jrschumacher:
Thanks for the contribution! We'll review and add some comments. It will require some tests to ensure we can merge.

PR #1934, opened 2025-02-24
chore(ci): Updates to golangci-lint 1.64

dmihalcik-virtru:
### Proposed Changes

- Updates config to use latest recommended linters
- Fixes new linter errors and warnings, most notably moving to the now standard `slices` package instead of the `exp` one

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,9 +17,9 @@ all: toolcheck clean build lint license test
 toolcheck:
 	@echo "Checking for required tools..."
 	@which buf > /dev/null || (echo "buf not found, please install it from https://docs.buf.build/installation" && exit 1)
-	@which golangci-lint > /dev/null || (echo "golangci-lint not found, run  'go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.61.0'" && exit 1)
+	@which golangci-lint > /dev/null || (echo "golangci-lint not found, run  'go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.5'" && exit 1)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -19,7 +19,7 @@ toolcheck:
 	@which buf > /dev/null || (echo "buf not found, please install it from https://docs.buf.build/installation" && exit 1)
 	@which golangci-lint > /dev/null || (echo "golangci-lint not found, run  'go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.61.0'" && exit 1)
 	@which protoc-gen-doc > /dev/null || (echo "protoc-gen-doc not found, run 'go install github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc@v1.5.1'" && exit 1)
-	@golangci-lint --version | grep "version v\?1.6[123]" > /dev/null || (echo "golangci-lint version must be v1.61 or later [$$(golangci-lint --version)]" && exit 1)
+	@golangci-lint --version | grep "version v\?1.6[1234]" > /dev/null || (echo "golangci-lint version must be v1.61 or later [$$(golangci-lint --version)]" && exit 1)
```
jakedoublev, line 22: Do we want to bump this up a little so our local testing more closely aligns with CI?

jakedoublev (approved):

pflynn-virtru (dismissed):

Makefile
dmihalcik-virtru: I don't want to block people developing locally if they already have a workflow set up, but yes, it will be confusing when CI fails on issues that aren't found locally, so I've updated it here. Notably, the `tenv` deprecation (and revive changes) will mean any older version will miss several new lint types

elizabethhealy (approved):

PR #1936, opened 2025-02-24
feat: add custom json marshaler for sdk.Statement

dschmidt:
### Proposed Changes

* add a custom json Marshaler for sdk.Statement that works similar to the  Unmarshaler and handles json structured values

If you agree with this change, I'm happy to add tests.

I assume it's a good idea to try to decode the value to make sure it's actual json. I just wasn't sure whether we should return an error if not or if we should just use the value as string.
Opinions?

### Checklist

### DIFF
--- a/sdk/assertion.go
+++ b/sdk/assertion.go
@@ -108,6 +108,16 @@ func (a Assertion) GetHash() ([]byte, error) {
 	// Remove the binding key
 	delete(jsonObject, "binding")
 
+	// Deep patch: fix "value" inside "statement" if it is a string
+	if statement, ok := jsonObject["statement"].(map[string]interface{}); ok {
+		if valueStr, ok := statement["value"].(string); ok {
+			var valueObj interface{}
+			if err := json.Unmarshal([]byte(valueStr), &valueObj); err == nil {
+				statement["value"] = valueObj // replace the string with parsed object
+			}
+		}
+	}
+
 	// Marshal the map back to JSON
 	assertionJSON, err = json.Marshal(jsonObject)
 	if err != nil {
--- a/sdk/assertion.go
+++ b/sdk/assertion.go
@@ -123,6 +123,32 @@ func (a Assertion) GetHash() ([]byte, error) {
 	return ocrypto.SHA256AsHex(transformedJSON), nil
 }
 
+func (s *Statement) MarshalJSON() ([]byte, error) {
+	// Define a custom struct for serialization
+	type Alias Statement
+	aux := &struct {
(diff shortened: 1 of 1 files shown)


sdk/assertion.go
```diff
@@ -123,6 +123,32 @@ func (a Assertion) GetHash() ([]byte, error) {
...
+		var tmp interface{}
+		// Attempt to decode Value to validate it's actual json
+		if err := json.Unmarshal(raw, &tmp); err == nil {
+			aux.Value = raw
+		} else {
+			aux.Value = s.Value
+		}
+	} else {
+		aux.Value = s.Value
+	}
```
jrschumacher, line 147: I reviewed our implementation of assertions compared to our spec, and it looks like we missed the enum of the `statement.format`. I've raised this with our architects, and we're going to review the alignment of the ZTDF/BaseTDF spec with the official SDKs (Go, Java, and Web).

I'd suggest implementing the marshaling within your PEP to support this in the short term. Something like:

```go
type MyStatement struct {
  opentdf.Statement
}

func (s *MyStatement) MarshalJSON() ([]byte, error) {
  // ...
}
```

jrschumacher: Alternatively you should be able to just marshal the data since we have the json tags on the `Statement`.

dschmidt: This doesn't work.

`AssertionConfig.Statement` would need to be an interface, so I could have a custom `MarshalJSON` implementation be called.
I cannot assign `MyStatement` to `AssertionConfig` - if I wrap that in `MyAssertionConfig`and adjust the statement type there, I cannot use it in `WithAssertions` .. and I cannot add it to TDFConfig so it's not used in `CreateTDFContext`.

I don't think it makes sense to go down this road

dschmidt:
I've updated the PR with a change to Assertion.GetHash()` because after an update of the SDK the verification of the assertion signing failed.

Tbh it feels even messier now than before.
Why don't we just make `Statement.Value` a `json.RawMessage`?
That should simplify everything a lot...

dschmidt:
@jrschumacher ping? :)

jrschumacher:
@dschmidt we're reviewing this with some additional changes to assertions that are coming. We should be ready for a decision this week.

dschmidt:
Alright, sounds perfect!

Happy this isn't forgotten :)

pflynn-virtru:
Draft PR here: https://github.com/opentdf/platform/pull/2687

There is a WIP ADR.  It will hold the decision on this refactor.

We are heavily considering keeping the `Statement.Value` a string in all SDK implementations.

Issue #1935, opened 2025-02-24
SDK: should have better DX when initialized with disconnected platform


jakedoublev:
### Background

At present, calling `sdk.New()` with a platform endpoint that does not resolve to a connected platform resolves to a potentially misleading error: `ErrPlatformConfigFailed` ([see logic](https://github.com/opentdf/platform/blob/939c71603b1f12e8994632a5b3871d7ebbbaba37/sdk/sdk.go#L123-L146)).

This is misleading because it could mask legitimate platform errors:
1. config not being served but platform running
2. config served at wrong location but platform running
3. logic to get config is bad, but platform running

It would be better to test connection to a running platform during a remote connection (non-IPC) SDK instantiation within the `sdk.New` logic, and to return a more specific error `ErrPlatformEndpointNotFound` when the provided platformEndpoint does not resolve to a running platform.

This will clarify platform running states to PEPs, who are currently expected to write their own logic to detect `ErrPlatformConfigFailed` as the key error for an invalid or disconnected platform endpoint to provide the user with helpful information about checking the platform endpoint value or platform state ([see otdfctl](https://github.com/opentdf/otdfctl/blob/60a18f7174e47e34d43683d01522fa2009dca742/pkg/auth/auth.go#L312-L314)).

### Acceptance Criteria

1. A small check is added before fetching the platform configuration within sdk.New to ensure the platform is actually reachable
2. If it is not reachable, a more helpful named error is provided to scope the issue to the platform likely not running at the provided endpoint

jakedoublev:
In consultation with @dmihalcik-virtru, decided to implement via:
1. `WithConnectionValidation` constructor option
2. exposing a health check validation function from the sdk module for client consumption

If clients are using `WithPlatformConfiguration` at present to bypass the current connectivity check, we will not break them today with the intention to refactor code paths between IPC, remotely connected, and offline scenarios when the offline effort is picked up.

PR #1937, opened 2025-02-24
feat(sdk): sdk.New should validate platform connectivity and provide precise error

jakedoublev:
### Proposed Changes

* Validates healthy platform is available within `sdk.New` when triggered
* Provides precise error if error connecting to provided `platformEndpoint` or status is unhealthy
* Exposes validation function from SDK module for DX

Resolves https://github.com/opentdf/platform/issues/1935

### Checklist

### DIFF
--- a/sdk/sdk.go
+++ b/sdk/sdk.go
@@ -32,6 +32,7 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
+	healthpb "google.golang.org/grpc/health/grpc_health_v1"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


sdk/sdk.go
```diff
@@ -112,12 +114,21 @@ func New(platformEndpoint string, opts ...Option) (*SDK, error) {
...
+	// IF IPC is disabled we build a validated healthy connection to the platform
 	if !cfg.ipc {
 		platformEndpoint, err = SanitizePlatformEndpoint(platformEndpoint)
 		if err != nil {
 			return nil, fmt.Errorf("%w [%v]: %w", ErrPlatformEndpointMalformed, platformEndpoint, err)
 		}
+		
+		// TODO: to support an offline SDK, we would need to remove the connection requirement
+		// For now we will only skip for test purposes
+		if !cfg.testSkipValidatePlatformConnectivity {
```
dmihalcik-virtru, line 126: Why not add a `WithConnectionValidation` option to the constructor? This would keep the current behavior of the SDK object (arguably this is a semver breaking change) and allow you to add the feature in otdfctl or other app integrators as needed

jakedoublev: I could also move this connectivity check inside the case where `WithPlatformConfiguration` was not utilized, but it feels like that option is actually driving a code path implementing something like `WithUnvalidatedPlatform`. 🤷

jentfoo (approved):

dmihalcik-virtru (approved):

PR #1960, opened 2025-03-04
chore: enable policy-bot evaluation

strantalis:
### Proposed Changes

For more information on policy-bot spec https://github.com/palantir/policy-bot?tab=readme-ov-file#policyyml-specification

### Checklist


eastokes (dismissed):
👍 
Looks good to me, I'm not totally up to speed on the spec nuances but makes sense at a quick review

jrschumacher (changes requested):
@strantalis could we annotate this `.policy.yml` document to help communicate what the sections mean and where to go for further information.

jentfoo (changes requested):
NODTM with the policy, but please add `.policy.yml` to the CODEOWNERS.  This config can be very sensitive, so we should ensure it's protected.

strantalis:
@jentfoo 
> NODTM with the policy, but please add `.policy.yml` to the CODEOWNERS. This config can be very sensitive, so we should ensure it's protected.

So I think the goal would be to flatten the codeowners and have this rule enforce an approval from architecture or security if `.policy.yml` is changed. 

https://github.com/opentdf/platform/pull/1960/files#diff-2b2c40c447b8cc3c757b3526d74d04fd8d43e3f17c7ad4f9e6cdfa9bc1fc80d3R136-R152

strantalis:
> @strantalis could we annotate this `.policy.yml` document to help communicate what the sections mean and where to go for further information.

@jrschumacher I added a link to the configuration. Are you looking for more details than what is put in the descriptions?

dmihalcik-virtru:
IMO we should simplify/reduce CODEOWNERs to avoid duplication of effort, but it doesn't seem to me that policy-bot will give us the kind of close integration with tooling we want to fully replace it. Also, I worry about reliability and visibility about using a private hosted service, especially on a public repo

strantalis:
@dmihalcik-virtru 

> IMO we should simplify/reduce CODEOWNERs to avoid duplication of effort, but it doesn't seem to me that policy-bot will give us the kind of close integration with tooling we want to fully replace it. Also, I worry about reliability and visibility about using a private hosted service, especially on a public repo

I think we would need to flatten the codeowners but I didn't want to change that until we felt comfortable with using policy-bot. We can also make the service public once we feel comfortable enough with using it.

This shouldn't block any pull requests until we set a required status check.

pnancarrow:
> but it doesn't seem to me that policy-bot will give us the kind of close integration with tooling we want to fully replace it.

what is policy bot lacking that would give you the close integration you are looking for? Github doesn't natively support much of what can be done with policy bot, so it doesn't seem there is a better option.

jentfoo:
@strantalis I am not sure if putting this restriction in policy bot alone is sufficient.  Unless I am mistaken, I believe the referenced policy config will be the one in the PR. The concern is that someone could modify the config as part of their PR to bypass or remove restrictions like these.

I know the config can be referenced from a different repo, and that would address my concerns. Let me know if you know of anything else, but CODEOWNERS seems a straight forward solution. Additionally it's already pretty standard, I would prefer to keep ownership defined in CODEOWNERS.

jentfoo (approved):
@strantalis and @jrschumacher, I had access to update the fork branch.  I added the proposed change to CODEOWNERS, and with that change I feel this looks good to start with 🚀

jrschumacher (approved):

PR #1961, opened 2025-03-04
fix(core): Updates ec-wrapped to newer salt

dmihalcik-virtru:
### Proposed Changes

*

### Checklist


lib/ocrypto/asym_decryption.go
```diff
@@ -107,8 +107,11 @@ type ECDecryptor struct {
 }
 
 func NewECDecryptor(sk *ecdh.PrivateKey) (ECDecryptor, error) {
-	// TK Make these reasonable? IIRC salt should be longer, info maybe a parameters?
-	salt := []byte("salt")
+	// TK Move salt and info out of library, into API option functions
+	digest := sha256.New()
+	digest.Write([]byte("TDF"))
```
sujankota, line 112: Do you think we should define a constant value for use everywhere?

sujankota (approved):

lib/ocrypto/asym_decryption.go
dmihalcik-virtru: yeah, my hope was to move it out of ocrypto into the application (service/kas or sdk) but I didn't think of a clean way to do it.  I've got a couple of local branches where I'm trying different APIs

biscoe916 (approved):

PR #1966, opened 2025-03-06
fix(core): Fixes merge fail in bulk logic

dmihalcik-virtru:
### Proposed Changes

* Fixes some code in bulk rewrap that wasn't properly assigning the errors - or returning the correct error code - to the 'response' objects. 
* This most notably was causing failures with mixed EC and RSA wrapping requests, and for some 400s to be 500s.

### Checklist


mkleene (approved):
Looks OK and it seems safer, if anything. Is it important to guarantee this behavior with a test or does that seem like overkill?

dmihalcik-virtru:
I'll look into refactoring it to be more testable and see if can put together it in a quick followup

PR #1997, opened 2025-03-20
fix(core): refine key management schema

strantalis:
### Proposed Changes

This is the schema changes required to support key management. 

```mermaid
erDiagram
key_access_server {
  uuid id
  varchar uri
  varchar name
  varchar source_type
}

key_access_server_keys {
  uuid id
  uuid key_access_server_id
}

asym_keys {
  uuid id
  varchar key_id
  varchar algorithm
  varchar key_status
  varchar key_mode
  jsonb   public_key_ctx
  jsonb   private_key_ctx
  date    expiration
  uuid    provider_config_id
  jsonb   metadata
  timestamp created_at
  timestamp updated_at
}

sym_keys {
  uuid id
  varchar key
  varchar key_id
  varchar key_status
  varchar key_mode
  uuid provider_config_id
  jsonb metadata
  timestamp created_at
  timestamp updated_at
}

provider_configuration {
  uuid id
  varchar provider_type
  jsonb config_json
  jsonb metadata
  timestamp created_at
  timestamp updated_at
}

namespace_public_key_mappings {
  uuid namespace_id
  uuid kas_key_id
}

definition_public_key_mappings {
  uuid definition_id
  uuid kas_key_id
}

value_public_key_mappings {
  uuid value_id
  uuid kas_key_id
}

key_access_server ||--o{ key_access_server_keys : has
key_access_server_keys ||--|| asym_keys : inherits
asym_keys }o--|| provider_configuration : uses
sym_keys }o--|| provider_configuration : uses
asym_keys ||--o{ namespace_public_key_mappings : maps_to
asym_keys ||--o{ definition_public_key_mappings : maps_to
asym_keys ||--o{ value_public_key_mappings : maps_to

```

### Checklist


service/integration/attribute_values_test.go
```diff
@@ -967,17 +967,17 @@ func (s *AttributeValuesSuite) Test_RemoveKeyAccessServerFromValue_Returns_Succe
...
-	s.Require().NoError(err)
-	s.NotNil(v)
-
-	// ensure only active public keys are returned
-	s.NotEmpty(v.GetKeys())
-	for _, k := range v.GetKeys() {
-		s.True(k.GetIsActive().GetValue())
-	}
-}
+// func (s *AttributeValuesSuite) Test_GetAttributeValue_Returns_Only_Active_PublicKeys() {
```
jakedoublev, line 970: Should we remove or skip these commented out tests rather than comment out?

service/policy/db/migrations/20250319000000_key_management_updates.sql
```diff
@@ -0,0 +1,302 @@
...
+
+DROP TABLE IF EXISTS attribute_namespace_public_key_map;
+
+DROP TABLE IF EXISTS attribute_definition_public_key_map;
+
+DROP TABLE IF EXISTS attribute_value_public_key_map;
+
+DROP TABLE IF EXISTS public_keys;
+
+-- DROP FUNCTION IF EXISTS update_was_mapped;
```
jakedoublev, line 27: Did we still need this?

service/policy/db/migrations/20250319000000_key_management_updates.sql
```diff
@@ -0,0 +1,302 @@
...
+ALTER TABLE key_access_servers
+ADD COLUMN IF NOT EXISTS source_type VARCHAR;
```
jakedoublev, line 29: Remind me if `source_type` is a postgres enum, a proto-enforced enum, or wide open?

service/policy/db/migrations/20250319000000_key_management_updates.sql
```diff
@@ -0,0 +1,302 @@
...
+
+CREATE TABLE IF NOT EXISTS
+    provider_config (
+        id UUID DEFAULT gen_random_uuid () CONSTRAINT provider_config_pkey PRIMARY KEY,
+        provider_name VARCHAR(255) NOT NULL,
+        config jsonb NOT NULL,
```
jakedoublev, line 35:
```suggestion
        config JSONB NOT NULL,
```

service/policy/db/migrations/20250319000000_key_management_updates.sql
```diff
@@ -0,0 +1,302 @@
...
+        created_at TIMESTAMP DEFAULT NOW(),
+        updated_at TIMESTAMP DEFAULT NOW(),
+        metadata jsonb
```
jakedoublev, line 38:
```suggestion
        metadata JSONB
```

service/policy/db/migrations/20250319000000_key_management_updates.sql
```diff
@@ -0,0 +1,302 @@
...
+COMMENT ON COLUMN provider_config.created_at IS 'Timestamp when the provider configuration was created';
+
+COMMENT ON COLUMN provider_config.updated_at IS 'Timestamp when the provider configuration was last updated';
+
+COMMENT ON COLUMN provider_config.metadata IS 'Additional metadata for the provider configuration';
+
+CREATE TABLE IF NOT EXISTS
+    asym_key (
+        id UUID DEFAULT gen_random_uuid () CONSTRAINT asym_key_pkey PRIMARY KEY,
+        key_id VARCHAR(36) NOT NULL, -- UNIQUE????
```
jakedoublev, line 58: Unique makes sense to me, and aligns with the comment on the row we're adding below.

service/policy/db/migrations/20250319000000_key_management_updates.sql
```diff
@@ -0,0 +1,302 @@
...
+        key_algorithm INT NOT NULL,
+        key_status INT NOT NULL,
+        key_mode INT NOT NULL,
+        public_key_ctx jsonb,
+        private_key_ctx jsonb,
+        expiration TIMESTAMP,
+        provider_config_id UUID CONSTRAINT asym_key_provider_config_fk REFERENCES provider_config (id),
+        metadata jsonb,
```
jakedoublev, line 66:
```suggestion
        public_key_ctx JSONB,
        private_key_ctx JSONB,
        expiration TIMESTAMP,
        provider_config_id UUID CONSTRAINT asym_key_provider_config_fk REFERENCES provider_config (id),
        metadata JSONB,
```

jakedoublev (approved):

service/integration/attribute_values_test.go
strantalis: I removed them. If we need to reference them they are always in the git history.

service/policy/db/migrations/20250319000000_key_management_updates.sql
strantalis: Yeah we should still drop that function. I was probably running into issues and commented it out. Adding ti back.

service/policy/db/migrations/20250319000000_key_management_updates.sql
strantalis: So I should take a look at what we have done with other enums but was thinking we should only define enums in one place either proto's or sql.  In this case I picked proto's

service/policy/db/migrations/20250319000000_key_management_updates.sql
strantalis: but does it need to be unique between all KAS servers?

jakedoublev: I also want to defer to you and your vision here, but I think being more restrictive up front is easier to walk back then being flexible now and trying to restrict later, so whichever you think is best I think would be good.

strantalis: I think i should add a unique constraint on key_id + algorithm. Does that make sense. I originally had something like that in the public keys table.

strantalis: Discussed offline and there will be a unique constraint on kas and key_id.

PR #1999, opened 2025-03-21
fix: Allow parsing IPs as hostnames

mkleene:
### Proposed Changes

* Also rely more on parsing using `net` and `url` libraries as opposed to regex

### Checklist

### DIFF
--- a/sdk/sdk.go
+++ b/sdk/sdk.go
@@ -12,7 +12,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
-	"regexp"
+	"strings"
 
 	"github.com/opentdf/platform/lib/ocrypto"
 	"github.com/opentdf/platform/protocol/go/authorization"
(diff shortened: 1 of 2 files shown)


dmihalcik-virtru (dismissed):

sdk/sdk.go
```diff
@@ -207,25 +207,33 @@ func New(platformEndpoint string, opts ...Option) (*SDK, error) {
...
+		newE := "https://" + e
+		u, err = url.ParseRequestURI(newE)
+		if err != nil {
+			return "", errors.Join(fmt.Errorf("cannot parse platform endpoint [%s]", newE), err)
+		}
 	}
 
-	if !regexp.MustCompile(`^(https?:\/\/)?(([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}(:\d+)?|(localhost)(:\d+)?)\/?$`).MatchString(e) {
-		return "", errors.New("platform endpoint is not valid")
+	if u.Host == "" {
```
elizabethhealy, line 223: nit: can you move this inside the if above? looks like this will only be true if the above block is entered

mkleene: good call, fixed

elizabethhealy (approved):

PR #2003, opened 2025-03-25
chore(ci): add label workflow

strantalis:
### Proposed Changes

Adding label workflow which has one job at this time  and that is to label pull requests with `external-contributor` if  the user opening the pull request is not a member of the org. 

This workflow keys off of `author_associations` from the pull request context.  

Best [link](https://docs.github.com/en/graphql/reference/enums#commentauthorassociation) that I could find to describe each association 

Also adds the following actions
- https://github.com/CodelyTV/pr-size-labeler
- https://github.com/actions/labeler

### Checklist

### DIFF
--- a/.github/workflows/label.yaml
+++ b/.github/workflows/label.yaml
@@ -1,0 +1,82 @@
+name: Label Pull Requests
+
+on:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - unlabeled
+
+permissions:
+  pull-requests: write
+
+jobs:
+  external-contributor:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Author Association and Label PR
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const pr = context.payload.pull_request;
+            if (!pr) {
+              console.log("Could not get PR from context");
+              return;
+            }
+
+            // Define associations considered "internal"
+            const internalAssociations = ["MEMBER"];
+
+            // Label to add if the author is external
+            const externalLabel = "external-contributor";
+
+            const authorAssociation = pr.author_association;
+            const isExternal = !internalAssociations.includes(authorAssociation);
+
+            const prLabels = pr.labels.map(label => label.name);
+            const hasExternalLabel = prLabels.includes(externalLabel);
+
+            console.log(`Event: ${context.eventName}, Action: ${context.payload.action}`);
+            console.log(`Author: ${pr.user.login}, Association: ${authorAssociation}, Is External: ${isExternal}`);
+            console.log(`Current PR Labels: ${prLabels.join(', ')}`);
+
+            // Logic for 'unlabeled' event: only re-add if *our* label was removed and author is still external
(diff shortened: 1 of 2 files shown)


.github/workflows/label.yaml
```diff
@@ -0,0 +1,82 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check Author Association and Label PR
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const pr = context.payload.pull_request;
+            if (!pr) {
+              console.log("Could not get PR from context");
```
dmihalcik-virtru, line 24: This should probably fail and indicate an error, right? Maybe we want a label like 'UNLABELABLE' or something? But i guess without the PR there is nothing to label?

.github/workflows/label.yaml
```diff
@@ -0,0 +1,82 @@
...
+              } else if (isExternal && hasExternalLabel) {
+                 console.log(`Author is external, but label "${externalLabel}" is already present.`);
+              } else {
+                console.log(`Author association "${authorAssociation}" is internal or label already present. No label added.`);
+              }
+            }
+            // Optional: Add logic for 'synchronize' if you uncomment it in the 'on:' section
+            // else if (context.payload.action === 'synchronize') { ... }
+            else {
+              console.log(`Unhandled action type: ${context.payload.action}. No labeling action taken.`);
```
dmihalcik-virtru, line 81: Maybe error log messages such as this one should step summaries or otherwise easier to access?

.github/workflows/label.yaml
```diff
@@ -0,0 +1,82 @@
...
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  issue_number: pr.number,
+                  labels: [externalLabel]
+                });
+                console.log(`Label "${externalLabel}" added successfully.`);
+              } else if (isExternal && hasExternalLabel) {
+                 console.log(`Author is external, but label "${externalLabel}" is already present.`);
+              } else {
+                console.log(`Author association "${authorAssociation}" is internal or label already present. No label added.`);
```
dmihalcik-virtru, line 75: You already handle the case where the label is present above. You probably want to remove the label if it is labelled as external, but maybe just note that it is?
```suggestion
                if (hasExternalLabel) {
                  console.log(`Author association "${authorAssociation}" is internal. However, external contributor label is present.`);
                } else {
                  console.log(`Author association "${authorAssociation}" is internal. No label added.`);
                }
```

dmihalcik-virtru (dismissed):

.github/workflows/label.yaml
strantalis: I switched to core.error which I think will add an annotation.

.github/workflows/label.yaml
strantalis: I switched to core.warning if someone how we get into a situation the author is internal but the pr as a external label contributor.

dmihalcik-virtru (dismissed):

jrschumacher (dismissed):

dmihalcik-virtru (approved):

jrschumacher (approved):

PR #2005, opened 2025-03-25
fix(core)!: update GRPC Gateway to use IPC

### LINKED ISSUE #2004: Replace GRPCGateway loopback call with IPC call
When adding ConnectRPC support #1524 deprecated support of GRPCGateway was enabled through a loopback call. This worked fine when running the server with servicemesh TLS termination, but when running with TLS termination at the server the loopback request would fail due to service name mismatch.

To resolve this, we need to update the GRPCGateway registration to use the in-process connection.

jrschumacher:
Closes #2004
DSPX-593

### Proposed Changes

* Bind GRPCGateway to the ConnectRPC IPC connection

### Checklist

### Testing Instructions

CleanShot 2025-03-25 at 18 21 38


service/internal/auth/authn.go
```diff
@@ -222,6 +223,13 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 			return
 		}
 
+		md, ok := metadata.FromIncomingContext(ctxWithJWK)
+		if !ok {
+			md = metadata.New(nil)
+		}
+		md.Append("access_token", ctxAuth.GetRawAccessTokenFromContext(ctxWithJWK, nil))
+		ctxWithJWK = metadata.NewIncomingContext(ctxWithJWK, md)
```
dmihalcik-virtru, line 231: This seems unnecessary now? Are you referring to it later?

dmihalcik-virtru (approved):
One suggstion would be to call this `fix(core)!`, as it introduces a breaking change to the gateway function. I think this will cause the version number for service to tick over to 0.5.0

jrschumacher:
That's fine by me but I didn't realize the breaking change would tick the minor. I thought it would bump the major.

service/internal/auth/authn.go
jrschumacher: Honestly I'd rather punt and refactor wholly later when we are not under a time crunch.

strantalis (approved):

PR #2012, opened 2025-03-31
feat(policy): add key management to kas registry service

strantalis:
feat(policy): add base key management proto's


service/policy/objects.proto
```diff
@@ -367,18 +360,87 @@ message PublicKey {
...
     // public key with additional information. Current preferred version
     KasPublicKeySet cached = 3;
   }
 }
+
+/*
+   Key Management
+*/
+
+message ProviderConfig {
```
jakedoublev, line 385: KeyProviderConfig?

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -351,6 +351,96 @@ message ListKeyAccessServerGrantsResponse {
...
+  common.MetadataUpdateEnum metadata_update_behavior = 101;
+}
+
+message UpdateKeyResponse {
+  AsymmetricKey key = 1;
+}
+
+message RotateKeyRequest {
+  // Required (Current Active Key ID)
+  string active_key_id = 1 [(buf.validate.field).string.uuid = true];
```
jakedoublev, line 428: Is this the `kid` field or the row `id`?

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -351,6 +351,96 @@ message ListKeyAccessServerGrantsResponse {
...
+
+  // Required New Key Information
+  Algorithm key_algorithm = 2 [(buf.validate.field).enum.defined_only = true];
+  KeyMode key_mode = 3 [(buf.validate.field).enum.defined_only = true];
+
+  // Optional
+  bytes private_key_ctx = 4; // Specific structure based on key provider implementation
```
jakedoublev, line 435: Confirming no `public_key_ctx` needed here?

jakedoublev (commented):
There's a lot here, so if you'd prefer a follow-on, that's totally fine, but it'd be great to see unit tests for all the proto validation rules in the write RPCs sooner rather than later.

service/policy/kasregistry/key_access_server_registry.proto
strantalis: the kid. I feel like this is going to get confusing. Do you think `active_kid` is better?

service/policy/kasregistry/key_access_server_registry.proto
strantalis: We need it. I was about to say we would derive the public key from the private key but if its encrypted that won't be the case.

service/policy/objects.proto
strantalis: I will update this. I think this makes sense now that I moved it to objects.proto to remove the circular dependencies.

service/policy/kasregistry/key_access_server_registry.proto
jakedoublev: Yes, that's clearer. I also think we should capture in the proto comments that the `AsymmetricKey` fields that `id` is a uuid and `key_id` is a name something like `my-kas-key-123` that could be a uuid but is admin-defined.

strantalis:
> There's a lot here, so if you'd prefer a follow-on, that's totally fine, but it'd be great to see unit tests for all the proto validation rules in the write RPCs sooner rather than later.

We are going to start working on the implementation tomorrow. Was hoping by breaking things down reviews might be more manageable.

jakedoublev (approved):

PR #2016, opened 2025-03-31
fix(sdk): Allow reuse of session key

dmihalcik-virtru:
### Proposed Changes

- This was dropped during merging in the support for EC session keys

### Checklist


pflynn-virtru (approved):

dmihalcik-virtru:
Looks like npmjs was returning 404s on camelcase? It seems to be restored so I'm retrying

PR #2020, opened 2025-03-31
feat(policy): add enhanced standard/custom actions protos

jakedoublev:
* Actions need to be truly considered within ABAC entitlement decisioning
* Actions are a precursor to obligations
* Custom actions should be CRUDable
* Standard actions should support mutable metadata key/value store common throughout policy
* bump /protocol/go to 1.24 to handle underlying protovalidate changes


service/policy/actions/actions.proto
```diff
@@ -0,0 +1,116 @@
...
+        expression: "this.matches('^[a-zA-Z0-9](?:[a-zA-Z0-9_-]*[a-zA-Z0-9])?$')"
+      }
+    ];
+  }
+}
+message GetActionResponse {
+  policy.Action action = 1;
+}
+
+message ListActionsRequest {
```
jakedoublev, line 38: specify which are standard (protos) and which are custom

ryanulit (dismissed):

ryanulit (approved):

PR #2029, opened 2025-04-01
chore(ci): Improve benchmark output

dmihalcik-virtru:
### Proposed Changes

* Include bulk rewrap benchmark in CI
* lets the output be markdown and included in step summaries
* Adds a comment to PRs that includes benchmark results

This doesn't yet include any performance regression analysis

### Checklist

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -250,18 +250,13 @@ jobs:
         uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea #v7.0.1
         with:
           script: |
+            const commentBody = `${ toJSON(env.BENCHMARK_METRICS_OUTPUT) }\n\n${ toJSON(env.BENCHMARK_BULK_OUTPUT) }`;
             github.rest.issues.createComment({
               owner: context.repo.owner,
               repo: context.repo.repo,
               issue_number: context.issue.number,
-              const commentBody = `${ toJSON(env.BENCHMARK_METRICS_OUTPUT) }\n\n${ toJSON(env.BENCHMARK_BULK_OUTPUT) }`;
-              github.rest.issues.createComment({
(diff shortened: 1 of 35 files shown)


.github/workflows/checks.yaml
```diff
@@ -218,12 +218,44 @@ jobs:
...
+
+          {
+            echo "BENCHMARK_METRICS_OUTPUT<<EOO";
+            echo "$OUTPUT";
+            echo "EOO"
+          } >> "$GITHUB_ENV"
+
+      - name: save benchmark results as a comment
+        if: github.event_name == 'pull_request'
+        uses: actions/github-script@v7
```
strantalis, line 250: We need to start pinning to commit shas going forward.

elizabethhealy (dismissed):
lgtm, we probably need a follow up to define thresholds for failure, like if theres a X% drop in throughput

strantalis (commented):
@dmihalcik-virtru I think you need to allow a few permissions on the benchmark job so you can comment on the pull request. 

I think you might need `issue: write` and `pull-request: write`

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

strantalis:
> @dmihalcik-virtru I think you need to allow a few permissions on the benchmark job so you can comment on the pull request.
> 
> I think you might need `issue: write` and `pull-request: write`
> 
> https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

Actually the default permissions on the org are permissive. We need to change that going forward as well.

dmihalcik-virtru:
> Update github-script to use commit sha of latest v7 release.

Done, and pinned all the other actions I could find

strantalis (approved):

elizabethhealy (approved):

PR #2037, opened 2025-04-01
feat(policy): actions crud service endpoints and proto validation

jakedoublev:


service/policy/actions/actions.proto
```diff
@@ -49,8 +49,8 @@ message ListActionsResponse {
 message CreateActionRequest {
   // Required
   string name = 1 [
-    (buf.validate.field).string.max_len = 253,
     (buf.validate.field).required = true,
+    (buf.validate.field).string.max_len = 253,
```
jakedoublev, line 53:
```suggestion
    (buf.validate.field).string.max_len = 253,
    (buf.validate.field).required = true,
```

service/policy/actions/actions_test.go
```diff
@@ -0,0 +1,273 @@
...
+
+// Actions proto validation
+
+type ActionSuite struct {
+	suite.Suite
+	v protovalidate.Validator
+}
+
+// Set up the test environment
+func (s *ActionSuite) SetupSuite() {
```
ryanulit, line 49: 🔥

service/policy/actions/actions_test.go
```diff
@@ -0,0 +1,273 @@
...
+				Name: n,
+			}
+			err := s.v.Validate(req)
+			s.Require().Error(err)
+			s.Require().Contains(err.Error(), errMessageRequiredActionName)
+		})
+	}
+
+	// no name
+	req := &actions.CreateActionRequest{
```
ryanulit, line 75: Nit: another way of doing this would be adding the expected err message and test name to a testcases struct: https://github.com/opentdf/platform/pull/2066/files#diff-7308d2e883c56c9eed1f3372e0289a3238fe9b5996998b38d26453a34814f7abR80-R167

ryanulit (changes requested):
LGTM, just one comment about the service proto name

service/policy/actions/actions.go
```diff
@@ -0,0 +1,87 @@
...
+
+		return nil
+	}
+}
+
+func NewRegistration(ns string, dbRegister serviceregistry.DBRegister) *serviceregistry.Service[actionsconnect.ActionServiceHandler] {
+	actionsSvc := new(ActionsService)
+	onUpdateConfigHook := OnConfigUpdate(actionsSvc)
+
+	return &serviceregistry.Service[actionsconnect.ActionServiceHandler]{
```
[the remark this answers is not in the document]
jakedoublev: Good catch. However, changing the proto would now unfortunately be a breaking change and policy objects differ in plurality among their service names, so I will update this service code to be singular instead to align with the proto:

```shell
jake.vanvorhis@jakevanvorhis platform % grpcurl -plaintext localhost:8080 list
authorization.AuthorizationService
entityresolution.EntityResolutionService
grpc.health.v1.Health
kas.AccessService
policy.attributes.AttributesService
policy.kasregistry.KeyAccessServerRegistryService
policy.namespaces.NamespaceService
policy.resourcemapping.ResourceMappingService
policy.subjectmapping.SubjectMappingService
policy.unsafe.UnsafeService
wellknownconfiguration.WellKnownService
```

service/policy/actions/actions_test.go
jakedoublev: Thanks, and agreed. I opted to iterate through the slice of names that fail due to the invalid name format validation and keep other tests separate as they differ if the name in the request proto is optional or not (identifier name in GET and updated name in UPDATE are optional).

jakedoublev: However, I did push up some a new commit to clean up some unit test variable names.

service/policy/actions/actions.go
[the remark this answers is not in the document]
ryanulit: I'd argue that Auth'n and EntityResolution as singular make sense, but any actual object should be plural (which you could also argue is just a REST idea). The ResourceMappingService being singular was my mistake.

But all that being said, updating the Go struct is fine with me.

ryanulit (approved):

PR #2045, opened 2025-04-03
feat(sdk): CreateTDF option to run with specific target schema version

elizabethhealy:
### Proposed Changes

* Supports creating tdfs with hex encoded hashes and without the schema version in the manifest

### Checklist

### DIFF
--- a/sdk/tdf_config_test.go
+++ b/sdk/tdf_config_test.go
@@ -95,3 +95,21 @@ func TestWithAssertions(t *testing.T) {
 	assert.Equal(t, id1, cfg.assertions[0].ID)
 	assert.Equal(t, id2, cfg.assertions[1].ID)
 }
+
+func TestWithTargetMode(t *testing.T) {
+	cfg := makeConfig(t, WithTargetMode("0.0.0"))
+	assert.True(t, cfg.useHex)
+	assert.True(t, cfg.excludeVersionFromManifest)
+
+	cfg = makeConfig(t, WithTargetMode("v0.0.0"))
+	assert.True(t, cfg.useHex)
+	assert.True(t, cfg.excludeVersionFromManifest)
+
+	cfg = makeConfig(t, WithTargetMode("4.3.0"))
+	assert.False(t, cfg.useHex)
+	assert.False(t, cfg.excludeVersionFromManifest)
+
+	cfg = makeConfig(t, WithTargetMode("v4.3.1"))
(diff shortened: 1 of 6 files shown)


sdk/tdf_config_test.go
```diff
@@ -95,3 +95,21 @@ func TestWithAssertions(t *testing.T) {
 	assert.Equal(t, id1, cfg.assertions[0].ID)
 	assert.Equal(t, id2, cfg.assertions[1].ID)
 }
+
+func TestWithTargetMode(t *testing.T) {
+	cfg := makeConfig(t, WithTargetMode("0.0.0"))
```
dmihalcik-virtru, line 100: Add an empty target mode test, and maybe some without the patch version?

Also, I'd suggest doing this with a table test, just to make the output slightly more debuggable

elizabethhealy: added, right now i have it just default to false if the input is empty "", but if an invalid semver is provided then it will error, does this make sense? or do you think we should also error for ""?

dmihalcik-virtru (dismissed):

sdk/tdf_config_test.go
dmihalcik-virtru: I think there might be scenarios where people want to set this value to the default explicitly, if only for convenience, so I think this is good.

sdk/tdf_test.go
```diff
@@ -612,9 +679,13 @@ func (s *TDFSuite) Test_TDFWithAssertion() {
...
-			tdfObj, err := s.sdk.CreateTDF(fileWriter, bufReader,
-				WithKasInformation(kasURLs...),
-				WithAssertions(test.assertions...))
+			createOptions := []TDFOption{WithKasInformation(kasURLs...),
+				WithAssertions(test.assertions...)}
+			if test.useHex {
+				createOptions = append(createOptions, WithTargetMode("0.0.0"))
+			}
+
+			tdfObj, err := s.sdk.CreateTDF(fileWriter, bufReader, createOptions...)
```
strantalis, line 688: Can we add asserting that the version is what we expect and also that the a segment and rootSignature is hex encoded.

strantalis (changes requested):

sdk/tdf_test.go
elizabethhealy: updated

strantalis (dismissed):

strantalis (approved):

PR #2051, opened 2025-04-07
feat(policy): KAS Keys

c-r33d:
### Proposed Changes

1.) Add crud for adding keys to Key Access Server table

### Checklist


service/policy/db/key_access_server_registry.go
```diff
@@ -296,3 +297,305 @@ func (c PolicyDBClient) ListKeyAccessServerGrants(ctx context.Context, r *kasreg
...
+		return nil, fmt.Errorf("cannot create a new key when an active key already exists with algorithm %s", r.GetKeyAlgorithm().String())
+	}
+
+	// PrivateCTX should be > 0 len
+	if len(privateCtx) < 1 {
+		return nil, fmt.Errorf("private key context is required")
+	}
+
+	// Verify that publicKey is in the request if key is mode local
+	if mode == int32(policy.KeyMode_KEY_MODE_LOCAL) && len(pubCtx) < 1 {
```
strantalis, line 339: We will always want a public key. Otherwise the platform will need to grab it from aws kms when we map the key to a value or namespace.

service/policy/db/key_access_server_registry.go
```diff
@@ -296,3 +297,305 @@ func (c PolicyDBClient) ListKeyAccessServerGrants(ctx context.Context, r *kasreg
...
+func (c PolicyDBClient) UpdateKey(ctx context.Context, r *kasregistry.UpdateKeyRequest) (*policy.AsymmetricKey, error) {
+	keyID := r.GetId()
+	if !pgtypeUUID(keyID).Valid {
+		return nil, db.ErrUUIDInvalid
+	}
+
+	// Check if trying to update to unspecified key status
+	if r.GetKeyStatus() == policy.KeyStatus_KEY_STATUS_UNSPECIFIED && r.GetMetadata() == nil && r.GetMetadataUpdateBehavior() == common.MetadataUpdateEnum_METADATA_UPDATE_ENUM_UNSPECIFIED {
+		return nil, fmt.Errorf("cannot update key status to unspecified")
+	}
```
strantalis, line 457: Can we add this via a proto option 

https://buf.build/docs/reference/protovalidate/rules/enum_rules/#in

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -365,7 +365,7 @@ message CreateKeyRequest {
   // Required
   KeyMode key_mode = 4 [(buf.validate.field).enum.defined_only = true]; // The mode of the key (e.g., local or external)
   // Required
-  bytes private_key_ctx = 5; // Context or additional data specific to the private key, based on the key provider implementation
+  bytes private_key_ctx = 5 [(buf.validate.field).required = true]; // Context or additional data specific to the private key, based on the key provider implementation
```
strantalis, line 368: I think this needs to be optional in the case of a user importing just another KAS's public key. We might want to make the public key required.

service/policy/db/key_access_server_registry.go
c-r33d: I think since the value is optional there is no way to differentiate between that value being updated and metadata via protobuf validate. Meaning it is acceptable that a user is only updating the metadata value, and the KeyStatus would be the default value (i.e. 0, in this case).

We could not throw the error but I think this helps give informative errors to the end user. What do you think?

strantalis (approved):

PR #2055, opened 2025-04-08
feat(policy): DSPX-898 NDR database schema

ryanulit:
### Proposed Changes

> [!IMPORTANT]
> For historical purposes: `Non Data Resource` was the original name of this object, but per discussion below we landed on `Registered Resource` 

* Create new database migrations for NDR schema

### Checklist


service/policy/db/migrations/20250407000000_add_non_data_resources.sql
```diff
@@ -0,0 +1,51 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE TABLE IF NOT EXISTS non_data_resource_groups (
+  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+  name VARCHAR NOT NULL UNIQUE,
```
jakedoublev, line 6: Can you please add support in each of these tables for the typical policy object metadata key value labels here consistent across policy primitives?

service/policy/db/migrations/20250407000000_add_non_data_resources.sql
```diff
@@ -0,0 +1,51 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE TABLE IF NOT EXISTS non_data_resource_groups (
```
jakedoublev, line 4: We have `groups:values` here, and some other ideas for the parent are `class`, `type`, `definition` (which is used in attributes but could be confusing with no rule), `category` (which is used for entity types). Any of those sound better to you?

service/policy/db/migrations/20250407000000_add_non_data_resources.sql
```diff
@@ -0,0 +1,51 @@
...
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+COMMENT ON TABLE non_data_resource_groups IS 'Table to store non-data resource groups';
+COMMENT ON COLUMN non_data_resource_groups.id IS 'Primary key for the table';
+COMMENT ON COLUMN non_data_resource_groups.name IS 'Name for the non-data resource group';
+COMMENT ON COLUMN non_data_resource_groups.created_at IS 'Timestamp when the record was created';
+COMMENT ON COLUMN non_data_resource_groups.updated_at IS 'Timestamp when the record was last updated';
+
+CREATE TABLE IF NOT EXISTS non_data_resource_values (
```
jakedoublev, line 16: Can we add cascade deletion to this migration so if the parent is deleted, all its values are deleted?

ryanulit: good catch, added

service/policy/db/migrations/20250407000000_add_non_data_resources.sql
ryanulit: Done

service/policy/db/migrations/20250407000000_add_non_data_resources.sql
ryanulit: What if we just went with:

`non_data_resources`  and `non_data_resource_values`

As that would be almost consistent with `attribute_definitions` and `attribute_values` and gets rid of the annoying `groups` suffix.

ttschampel: I like non_data_resources and non_data_resource_values

service/policy/db/migrations/20250407000000_add_non_data_resources.sql
```diff
@@ -0,0 +1,55 @@
...
+COMMENT ON COLUMN non_data_resource_groups.id IS 'Primary key for the table';
+COMMENT ON COLUMN non_data_resource_groups.name IS 'Name for the non data resource group';
+COMMENT ON COLUMN non_data_resource_groups.metadata IS 'Metadata for the non data resource group (see protos for structure)';
+COMMENT ON COLUMN non_data_resource_groups.created_at IS 'Timestamp when the record was created';
+COMMENT ON COLUMN non_data_resource_groups.updated_at IS 'Timestamp when the record was last updated';
+
+CREATE TABLE IF NOT EXISTS non_data_resource_values (
+  id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+  non_data_resource_group_id UUID NOT NULL REFERENCES non_data_resource_groups(id) ON DELETE CASCADE,
+  value VARCHAR NOT NULL,
```
jakedoublev, line 21: I was thinking about this last night and we should make `value` unique under the `non_data_resource_group_id` so you only have one `network : abc` at a time.

ryanulit: Ah right good call, that's the one case I forgot about

ryanulit: Done

service/policy/db/migrations/20250407000000_add_non_data_resources.sql
jakedoublev: @ttschampel what are your thoughts on `registered_resources` instead of `non_data_resources` to describe the objects by what they are instead of by what they are not?

ttschampel: I'm good with this

jakedoublev (approved):
🎉

PR #2057, opened 2025-04-08
fix(sdk): Allow tdfs with explicit 4.2.2 specver

dmihalcik-virtru:
### Proposed Changes

* javascript sdk sets this in compatiblity mode

### Checklist


mkleene:
@dmihalcik-virtru should I add this to the Java SDK?

elizabethhealy:
this is fine to add but sdks released between 3.26 and now won't be able to decrypt something  with schema 4.22 and hex, if a schema version is present it will assume its hexless 
javascript should not add a schema version if target < 4.30

PR #2064, opened 2025-04-11
fix(core): bump to go 1.24 and bump service proto module dependencies

jakedoublev:
Buf has [new proto validation features](https://buf.build/docs/protovalidate/schemas/predefined-rules/#example-case) we want to take advantage of, and we need to bump the underlying service deps in advance of utilizing those features in proto gencode. As one module requires a minimum of `1.23.4`, we are bumping to `1.24.0` to avoid pinning to frequently-changing patch versions.


elizabethhealy (dismissed):

go.work
```diff
@@ -1,4 +1,4 @@
-go 1.23.0
+go 1.23.4
```
pflynn-virtru, line 1: is this bump required?  `.0` was specified because it affects downstream

jentfoo: Starting in go 1.23.0 the final patch digit needs to be supplied, otherwise `go mod tidy` will result in differences in the go mod which causes the CI to fail.

That said, this version does not need to be updated.  The toolchain will always pull in the latest patch version.

jakedoublev:
```shell
% service % go mod tidy -go=1.23.0
go: github.com/bufbuild/protovalidate-go@v0.9.3 requires go@1.23.4, but 1.23.0 is requested
```
Unfortunately 1.23.4 is required for these dependencies.

pflynn-virtru (approved):

strantalis (approved):

elizabethhealy (approved):

PR #2071, opened 2025-04-15
feat(policy): DSPX-901 NDR database crud

ryanulit:
### Proposed Changes

* add db crud

### Checklist


service/policy/db/registered_resources.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	return nil
+}
+
+///
+/// Registered Resources
+///
+
+func (c PolicyDBClient) CreateRegisteredResource(ctx context.Context, r *registeredresources.CreateRegisteredResourceRequest) (*policy.RegisteredResource, error) {
```
jakedoublev, line 42: Just a reminder to wrap this in a transaction from the service layer.

service/integration/registered_resources_test.go
```diff
@@ -0,0 +1,851 @@
...
+	existingResOnly := s.f.GetRegisteredResourceKey("res_only")
+
+	list, err := s.db.PolicyClient.ListRegisteredResources(s.ctx, &registeredresources.ListRegisteredResourcesRequest{})
+	s.Require().NoError(err)
+	s.NotNil(list)
+
+	foundCount := 0
+
+	for _, r := range list.GetResources() {
+		if r.GetId() == existingRes.ID {
```
jakedoublev, line 175: Could this test find the same `existingRes` twice and pass incorrectly here?

service/internal/fixtures/policy_fixtures.yaml
```diff
@@ -615,3 +615,49 @@ namespace_key_map:
...
+registered_resources:
+  metadata:
+    table_name: registered_resources
+    columns:
+      - id
+      - name
+  data:
+    res_with_values:
+      id: f3a1b2c4-5d6e-7f89-0a1b-2c3d4e5f6789
+      name: res1
```
jakedoublev, line 632: Totally not required, but I like the pattern you replicated below where the key is the same as the name/value so in test logs you know more quickly which fixture to look at.

jakedoublev (approved):

service/policy/db/query.sql.go
```diff
@@ -3791,6 +3964,169 @@ func (q *Queries) listPublicKeys(ctx context.Context, arg listPublicKeysParams)
 	return items, nil
 }
 
+const listRegisteredResourceValues = `-- name: listRegisteredResourceValues :many
+WITH counted AS (
+    SELECT COUNT(id) AS total
+    FROM registered_resources
```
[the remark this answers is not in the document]
ryanulit: Good catch, and it also needs the WHERE clause from the actual query as well.  Made both fixes in my latest commit.

service/internal/fixtures/policy_fixtures.yaml
ryanulit: Done

jakedoublev (dismissed):

service/integration/registered_resources_test.go
ryanulit: The `existingRes` and `existingResOnly` records would only be in the results list once each, so that shouldn't happen. The real issue though is that I'm relying on the order of the values retrieved, which is flaky since there is no strict ordering when building the JSON for the values. Going to refactor the GET/LIST tests.

service/integration/registered_resources_test.go
```diff
@@ -0,0 +1,918 @@
...
+		},
+	})
+	s.Require().Error(err)
+	s.Require().ErrorIs(err, db.ErrNotFound)
+	s.Nil(got)
+
+	// verify resource values deleted
+	gotValues := created.GetValues()
+
+	// gotValue1, err := s.db.PolicyClient.GetRegisteredResourceValue(s.ctx, gotValues[0].GetId())
```
jakedoublev, line 443:
```suggestion
```

jakedoublev (dismissed):

jakedoublev (approved):

PR #2085, opened 2025-04-18
feat(sdk): Add a KAS allowlist

elizabethhealy:
### Proposed Changes

* adds a concept of an allowlist for kas urls during decrypt, if a url from a kao is not allowed then error
* the allowlist takes in url formatted strings and reformats to `scheme://host:port` and adds that to a set, if no scheme provided, will default to https
* the default contents of the kas allow list is the list of kases from the key access registry + the platform url
* this also adds tdf reader options: withKasAllowlist and WithIgnoreAllowlist -- withKasAllowlist lets you manually specify the kases in your allowlist (not using the registry), and WithIgnoreAllowlist lets you ignore the allowlist entirely
* currently we are not implementing a cache and just creating the allowlist on each request (expected slow down with queries to key access registry service)

Other additions to support the above:
* the concept of nano reader options - we have nano options on encrypt but not decrypt, add nano reader config and reader options similar to tdf3
* ability to pass in decrypt options from bulk decrypt -- previously bulk decrypt did not support decrypt options, ive added ability to specify either tdf3 or nano decrypt options which are passed down to the decrypter

Things to note:
* the platform url passed into the sdk should now include the scheme, otherwise the allowlist will default to https causing decrypt failures if youre using the platform kas on http

### Checklist


sdk/tdf.go
```diff
@@ -664,6 +666,31 @@ func (s SDK) LoadTDF(reader io.ReadSeeker, opts ...TDFReaderOption) (*Reader, er
...
+			}
+			config.kasAllowlist = AllowList{}
+			for _, kas := range kases.GetKeyAccessServers() {
+				err = config.kasAllowlist.Add(kas.GetUri())
+				if err != nil {
+					return nil, fmt.Errorf("kasAllowlist.Add failed: %w", err)
+				}
+			}
+			// grpc target does not have a scheme
+			err = config.kasAllowlist.Add("http://" + s.conn.Target())
```
dmihalcik-virtru, line 684: Should we default to `https`?

elizabethhealy: this is basically just adding a scheme so the url parsing works correctly, its removed when its actually added to the allowlist, it doesnt matter what the scheme is here

sdk/bulk.go
```diff
@@ -82,9 +95,49 @@ func (s SDK) BulkDecrypt(ctx context.Context, opts ...BulkDecryptOption) error {
...
+				// if no kasAllowlist is set, we get the allowlist from the registry
+				allowlist, err := allowListFromKASRegistry(ctx, s.KeyAccessServerRegistry, s.conn.Target())
+				if err != nil {
+					return fmt.Errorf("failed to get allowlist from registry: %w", err)
+				}
+				bulkReq.TDF3DecryptOptions = append(
+					bulkReq.TDF3DecryptOptions,
+					withKasAllowlist(allowlist),
+				)
+			}
```
dmihalcik-virtru, line 133: Seems duplicative? 

Also, instead of building a dummy, you could have a second phase for the options, like defaults, bulk defaults (what is the dummy here), then the actual stuff.

But yeah this is good enough and we can optimize later, esp. when if we (hopefully) get the 'unified' API out like we did for web-sdk

sdk/bulk.go
```diff
@@ -82,9 +95,45 @@ func (s SDK) BulkDecrypt(ctx context.Context, opts ...BulkDecryptOption) error {
 	tdfDecryptors := make(map[string]decryptor)
 	policyTDF := make(map[string]*BulkTDF)
 
+	switch bulkReq.TDFType {
+	case Nano:
+		dummy := &NanoTDFReaderConfig{}
+		for _, opt := range bulkReq.NanoTDFDecryptOptions {
+			_ = opt(dummy)
+		}
+		if !dummy.ignoreAllowList && dummy.kasAllowlist == nil {
```
jentfoo: Instead of checking `nil` it may be more robust to instead check `len(..) == 0`, that way empty slices are also handled.

sdk/tdf_config.go
```diff
@@ -275,6 +277,68 @@ type TDFReaderConfig struct {
...
+	// if port is empty, default to 443.
+	if port == "" {
+		return parsedURL.Hostname(), nil
+	}
+
+	return net.JoinHostPort(parsedURL.Hostname(), port), nil
+}
+
+func newAllowList(kasList []string) (AllowList, error) {
+	allowList := make(AllowList)
```
jentfoo, line 307:
```suggestion
	allowList := make(AllowList, len(kasList))
```

sdk/tdf_config.go
```diff
@@ -275,6 +277,68 @@ type TDFReaderConfig struct {
...
+	for _, kasURL := range kasList {
+		err := allowList.Add(kasURL)
+		if err != nil {
+			return nil, fmt.Errorf("error adding kas url(%s) to allowlist: %w", kasURL, err)
+		}
+	}
+	return allowList, nil
+}
+
+func (a AllowList) IsAllowed(kasURL string) bool {
```
jentfoo, line 317: Can we add a check here that the `kasURL` protocol is `https`, and add a unit test that `ftp` or `http` urls wont be allowed?

sdk/bulk.go
jakedoublev: Today I Learned: https://go.dev/play/p/3sq57Tmcy7C

sdk/tdf_config.go
elizabethhealy: for all of our opentdf platform instances we use http, a hardcoded check would block testing. but we could possibly add an option to be more strict when making decisions

sdk/nanotdf.go
```diff
@@ -1017,13 +1030,30 @@ func (n *NanoTDFDecryptHandler) Decrypt(_ context.Context, result []kaoResult) (
...
+	if handler.config.kasAllowlist == nil && !handler.config.ignoreAllowList {
+		if s.KeyAccessServerRegistry != nil {
+			// retrieve the registered kases if not provided
+			allowList, err := allowListFromKASRegistry(ctx, s.KeyAccessServerRegistry, s.conn.Target())
+			if err != nil {
+				return 0, fmt.Errorf("allowListFromKASRegistry failed: %w", err)
+			}
+			handler.config.kasAllowlist = allowList
+		} else {
+			slog.Warn("No KAS allowlist provided and no KeyAccessServerRegistry available, ignoring allowlist")
```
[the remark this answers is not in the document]
elizabethhealy: both this and the option to specify an allowlist instead of using the kases in the registry are included to provide more flexibility for use cases where strict enforcement isnt desired or feasable, mainly testing and development in our case.
its definitely more of a "proceed at your own risk" kind of option, and would not be recommended for use in any of the peps in prod envs.

sdk/tdf_config.go
jentfoo: If we can't enforce https at this point I think we should avoid stripping the protocol from the allow list.  And if no protocol is specified, default to https.  Then at this point we can validate the expected protocol matches.

sdk/nanotdf.go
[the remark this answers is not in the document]
jentfoo: This does not appear to be an option though, but rather a fallback behavior that could occur due the registry being unavailable.  Because an option to ignore does exist, I believe this should fail deny, rather than fail in a way that bypasses the security control.

sdk/tdf_config.go
strantalis: @jentfoo Even if someone specified `ftp` our client wouldn't use that protocol. Just trying to understand why matching on the protocol is important in this case.

jentfoo: I am not so worried about ftp, it was just another example of possible confusion.  I am more worried about `http` being used when the service should have cert validation.  MitM is possible with http, not just from a clear text perspective, but also from the ability to impersonate the service.

For example the exploit would still be possible if:
* The attacker specified an http kas server that is otherwise on an allowed domain
* The attacker manipulates the host file or performs another DNS attack to now allow the client to talk to the attacker controlled KAS

sdk/nanotdf.go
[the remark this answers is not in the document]
elizabethhealy: ah youre right! i forgot i did this. we should remove this and just fail if the kas registry isnt available, assuming the ignoreallowlist and provided allowlist options arent set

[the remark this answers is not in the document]
elizabethhealy: updated, ty for catching this!
sorry i originally thought this comment was pointed elsewhere in the code

sdk/tdf_config.go
elizabethhealy: updated to require matching schemes as well

sdk/tdf.go
```diff
@@ -647,6 +649,26 @@ func createPolicyObject(attributes []AttributeValueFQN) (PolicyObject, error) {
 	return policyObj, nil
 }
 
+func allowListFromKASRegistry(ctx context.Context, kasRegistryClient kasregistry.KeyAccessServerRegistryServiceClient, platformURL string) (AllowList, error) {
+	kases, err := kasRegistryClient.ListKeyAccessServers(ctx, &kasregistry.ListKeyAccessServersRequest{})
```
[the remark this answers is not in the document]
jentfoo: @elizabethhealy and @jakedoublev Let me know your thoughts here still.  What do we expect customers to utilize?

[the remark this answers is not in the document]
jakedoublev: Sorry, didn't see this until now. I agree that the KAS registry is the most secure store of truth for valid KASes and should be the . I also think an unknown number of deployment scenarios means this could be a footgun, because the same KAS could be known to 4 different PEP deployments as `['http://mykas.interal.cluster.url:9000','https://mykas.com','https://platform.com/kas','https://subdomain.platform.com/kas']`, so allowing some clientside autonomy in the decision is helpful for the scenarios we don't have foreknowledge about. I thought we were aligned on that, but perhaps I misinterpreted.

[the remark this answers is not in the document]
jrschumacher: With Go and Java this would only be a risk if they either accepted unsanitized user endpoint or allowed injection on compilation. I think this is outside our scope of responsibility as an open platform.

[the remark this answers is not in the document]
jentfoo: I am fine with the described use case.  If we thought this was only to be used for testing we could consider a log message for safety, but if we think there is good use to support both simultaneously I am fine with it.  Thank you!

examples/cmd/encrypt.go
```diff
@@ -99,15 +99,19 @@ func encrypt(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	baseKasUrl := platformEndpoint
+	if !strings.HasPrefix(baseKasUrl, "http://") && !strings.HasPrefix(baseKasUrl, "https://") {
+		baseKasUrl = fmt.Sprintf("http://%s", baseKasUrl)
```
jentfoo, line 104:
```suggestion
		baseKasUrl = fmt.Sprintf("https://%s", baseKasUrl)
```

jentfoo: We should require http to be explicit, and default https

sdk/tdf_config.go
```diff
@@ -275,6 +281,81 @@ type TDFReaderConfig struct {
...
+	if parsedURL.Hostname() == "" {
+		return "", fmt.Errorf("no host parsed from url: %s", kasURL)
+	}
+
+	// Default to "https" if no scheme is provided
+	if parsedURL.Scheme == "" {
+		parsedURL.Scheme = schemeHTTPS
+	}
+
+	// Default to port 443 if no port is provided
```
jentfoo, line 308: I am unsure what the goal of this is, it seems like if the port is not set we should let the client choose the default based on the protocol (http vs https / 80 vs 443).  Why always default 443 if not set?

sdk/tdf_config.go
```diff
@@ -275,6 +281,81 @@ type TDFReaderConfig struct {
...
+	}
+	parsedURL, err := url.Parse(kasURL)
+	if err != nil {
+		return "", fmt.Errorf("cannot parse kas url(%s): %w", kasURL, err)
+	}
+	if parsedURL.Hostname() == "" {
+		return "", fmt.Errorf("no host parsed from url: %s", kasURL)
+	}
+
+	// Default to "https" if no scheme is provided
```
jentfoo, line 303: Seems impossible?

jentfoo (commented):
Looks generally pretty good, remaining comments are relatively minor

sdk/tdf.go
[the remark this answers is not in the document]
jakedoublev: I think logs around things like these are definitely a good improvement. I think to add those, we would want to take in a logger on initialization so that we can log where the SDK consumer intends rather than assuming stdout is valid (which could break our CLI implementation and others that may utilize standardized outputs, like JSON and/or for piping).

sdk/tdf_config.go
elizabethhealy: if the platform url passed in on sdk initialization doesnt have a scheme this will be triggered

examples/cmd/encrypt.go
elizabethhealy: this is just for the example which we run for testing and development locally (usually without tls)

sdk/tdf_config.go
elizabethhealy: the comment is incorrect, it only adds 443 if the scheme is https and there is no port provided. ill update the comment to reflect the logic

examples/cmd/encrypt.go
jentfoo: As long as the example wont be referenced for future implementations, I am ok with that.

jentfoo (approved):

jentfoo (dismissed):

strantalis (approved):

Issue #1652, opened 2024-10-15
feat(policy): PolicyDBClient struct should not expose SQLC query generated functions


ryanulit:
### Background

Due to the PolicyDbClient struct being composed with the SQLC generated Queries struct, all generated functions are exposed to the service layer, allowing it to call generated methods and bypass the DB layer.

...more tbd....

jrschumacher:
Let's just go with lowercasing the leading character. Seems to be the simplest solution.

PR #2110, opened 2025-04-24
feat(policy): key management crud

c-r33d:
### Checklist


service/policy/db/migrations/20250429000000_keys_table.sql
```diff
@@ -2,16 +2,16 @@
 -- +goose StatementBegin
 CREATE TABLE IF NOT EXISTS
```
jakedoublev, line 3: I think there might be unintended side effects of updating an existing migration that consumers of the latest platform will already have run and have present in their goose migrations table. I think it would be safer to capture the `CASCADE` statement clauses added in a separate migration.

service/policy/db/migrations/20250429000002_key_management_update_timestamp.md
```diff
@@ -0,0 +1,240 @@
+```mermaid
```
jakedoublev, line 1: Same comment about this migration. With the entire schema, it's pretty difficult to see what changed and why. Would you please adjust these markdown files alongside the sql migrations to follow examples like in these other migraitons? https://github.com/opentdf/platform/blob/main/service/policy/db/migrations/20241125220354_keys_table.md or https://github.com/opentdf/platform/blob/main/service/policy/db/migrations/20240402000000_preserve_value_order.md

jakedoublev: I'd bet an LLM could do a great job taking the SQL migration and building a summary and mermaid diagram to capture.

service/policy/db/migrations/20250429000003_kas_source_type_update.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE IF EXISTS key_access_servers
+    ALTER COLUMN source_type TYPE INT USING source_type::INTEGER;
```
jakedoublev, line 4: Is this use of an integer here because we're storing the proto enum directly? Not sure if it's worth the LOE to you, but when we store proto enums in the DB, I've generally found it to be easier to transform to their string representations so a DBA or developer can inspect the db and read columns directly instead of holding the cognitive load of integer enum representations in their head. Food for thought.

service/policy/db/query.sql
```diff
@@ -108,12 +115,142 @@ SET
...
 WHERE id = $1;
 
 -- name: DeleteKeyAccessServer :execrows
 DELETE FROM key_access_servers WHERE id = $1;
 
+
+-----------------------------------------------------------------
+-- Key Access Server Keys
+------------------------------------------------------------------
+-- name: CreateKey :one
```
jakedoublev, line 129: Can you please lowercase all of these queries to align with the open issue we have to internalize them? https://github.com/opentdf/platform/issues/1652
```suggestion
-- name: createKey :one
```

service/policy/db/query.sql
```diff
@@ -332,15 +469,40 @@ WITH target_definition AS (
...
+            JSONB_AGG(
+                DISTINCT JSONB_BUILD_OBJECT(
+                    'id', kask.id,
+                    'key_id', kask.key_id,
+                    'key_status', kask.key_status,
+                    'key_mode', kask.key_mode,
+                    'key_algorithm', kask.key_algorithm,
+                    'private_key_ctx', ENCODE(kask.private_key_ctx::TEXT::BYTEA, 'base64'),
+                    'public_key_ctx', ENCODE(kask.public_key_ctx::TEXT::BYTEA, 'base64'),
+                    'metadata', JSON_STRIP_NULLS(JSON_BUILD_OBJECT('labels', kask.metadata -> 'labels', 'created_at', kask.created_at, 'updated_at', kask.updated_at)),
```
jakedoublev, line 489: I would advise removing metadata from this query as it adds additional complexity, and the query is designed for service to service RPCs where metadata is irrelevant (as it's an administrative construct and fully mutable).

service/policy/db/query.sql
```diff
@@ -358,17 +520,42 @@ namespaces AS (
...
```
jakedoublev, line 542: You can forgo metadata throughout all updated clauses in this query, per the comment above.

service/policy/db/query.sql
```diff
@@ -578,15 +824,40 @@ SELECT
...
+                'key_mode', kask.key_mode,
+                'key_algorithm', kask.key_algorithm,
+                'private_key_ctx', ENCODE(kask.private_key_ctx::TEXT::BYTEA, 'base64'),
+                'public_key_ctx', ENCODE(kask.public_key_ctx::TEXT::BYTEA, 'base64'),
+                'metadata', JSON_STRIP_NULLS(JSON_BUILD_OBJECT('labels', kask.metadata -> 'labels', 'created_at', kask.created_at, 'updated_at', kask.updated_at)),
+                'provider_config', JSON_STRIP_NULLS(JSON_BUILD_OBJECT(
+                    'id', pc.id,
+                    'name', pc.provider_name,
+                    'config_json', ENCODE(pc.config::TEXT::BYTEA, 'base64'),
+                    'metadata', JSON_STRIP_NULLS(JSON_BUILD_OBJECT('labels', pc.metadata -> 'labels', 'created_at', pc.created_at, 'updated_at', pc.updated_at))
```
jakedoublev, line 849: I would also advise forgoing key and provider metadata in a GetAttribute call/query.

service/policy/db/query.sql
```diff
@@ -775,16 +1055,41 @@ SELECT
...
+        JSONB_AGG(
+            DISTINCT JSONB_BUILD_OBJECT(
+                'id', kask.id,
+                'key_id', kask.key_id,
+                'key_status', kask.key_status,
+                'key_mode', kask.key_mode,
+                'key_algorithm', kask.key_algorithm,
+                'private_key_ctx', ENCODE(kask.private_key_ctx::TEXT::BYTEA, 'base64'),
+                'public_key_ctx', ENCODE(kask.public_key_ctx::TEXT::BYTEA, 'base64'),
+                'metadata', JSON_STRIP_NULLS(JSON_BUILD_OBJECT('labels', kask.metadata -> 'labels', 'created_at', kask.created_at, 'updated_at', kask.updated_at)),
```
jakedoublev, line 1075: Same here - no need for key/provider metadata in GetNamespace.

service/policy/db/query.sql
```diff
@@ -503,10 +715,35 @@ LEFT JOIN (
...
+                'provider_config', JSON_STRIP_NULLS(JSON_BUILD_OBJECT(
+                    'id', pc.id,
+                    'name', pc.provider_name,
+                    'config_json', ENCODE(pc.config::TEXT::BYTEA, 'base64'),
+                    'metadata', JSON_STRIP_NULLS(JSON_BUILD_OBJECT('labels', pc.metadata -> 'labels', 'created_at', pc.created_at, 'updated_at', pc.updated_at))
```
jakedoublev, line 735: I would advise removing the key/provider metadata from this query as well - we don't provide value or namespace metadata on GetAttribute and I think it makes sense to keep that consistent.

service/policy/kasregistry/key_access_server_registry.go
```diff
@@ -212,3 +212,104 @@ func (s KeyAccessServerRegistry) ListKeyAccessServerGrants(ctx context.Context,
...
+	}
+
+	resp, err := s.dbClient.CreateKey(ctx, r.Msg)
+	if err != nil {
+		s.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
+		return nil, db.StatusifyError(err, db.ErrTextCreationFailed, slog.String("keyAccessServer Keys", r.Msg.GetKasId()), slog.String("key id", r.Msg.GetKeyId()))
+	}
+
+	auditParams.ObjectID = resp.GetKey().GetKeyId()
+	auditParams.Original = resp.GetKey() // should we be logging this it will have the wrapped KEK and pub key?
```
jakedoublev, line 231: Maybe not, but we should look at other industry tools for logging behaviors and formalize the decision before merging this.

service/policy/kasregistry/key_access_server_registry.go
```diff
@@ -212,3 +212,104 @@ func (s KeyAccessServerRegistry) ListKeyAccessServerGrants(ctx context.Context,
...
+	switch i := r.Msg.GetIdentifier().(type) {
+	case *kasr.GetKeyRequest_Id:
+		s.logger.Debug("Getting keyAccessServer key by ID", slog.String("ID", i.Id))
+	case *kasr.GetKeyRequest_KeyId:
+		s.logger.Debug("Getting keyAccessServer by KeyId", slog.String("Name", i.KeyId))
+	default:
+		return nil, connect.NewError(connect.CodeInvalidArgument, nil)
+	}
+
+	auditParams := audit.PolicyEventParams{
```
jakedoublev, line 282: We haven't historically logged to audit for `read` events like Get/List. Are keys different, or should we continue to preserve audit logs for `write` events?

service/policy/kasregistry/key_access_server_registry.go
```diff
@@ -212,3 +212,104 @@ func (s KeyAccessServerRegistry) ListKeyAccessServerGrants(ctx context.Context,
...
+	}
+
+	updated, err := s.dbClient.UpdateKey(ctx, req.Msg)
+	if err != nil {
+		s.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
+		return nil, db.StatusifyError(err, db.ErrTextUpdateFailed, slog.String("keyAccessServer Keys", req.Msg.GetId()))
+	}
+
+	auditParams.Original = original
+	auditParams.Updated = updated
```
jakedoublev, line 262: Same question here about audit behavior.

service/policy/keymanagement/key_management.go
```diff
@@ -0,0 +1,173 @@
...
+	policydb "github.com/opentdf/platform/service/policy/db"
+)
+
+type Service struct {
+	dbClient policydb.PolicyDBClient
+	logger   *logger.Logger
+	config   *policyconfig.Config
+}
+
+func NewRegistration(ns string, dbRegister serviceregistry.DBRegister) *serviceregistry.Service[keyMgmtConnect.KeyManagementServiceHandler] {
```
jakedoublev, line 24: Can you please follow the update hook implementation throughout our policy services, which was probably introduced after this work began? 
```suggestion

func OnConfigUpdate(svc *Service) serviceregistry.OnConfigUpdateHook {
	return func(_ context.Context, cfg config.ServiceConfig) error {
		sharedCfg, err := policyconfig.GetSharedPolicyConfig(cfg)
		if err != nil {
			return fmt.Errorf("failed to get shared policy config: %w", err)
		}
		svc.config = sharedCfg
		svc.dbClient = policydb.NewClient(svc.dbClient.Client, svc.logger, int32(sharedCfg.ListRequestLimitMax), int32(sharedCfg.ListRequestLimitDefault))

		svc.logger.Info("key management service config reloaded")

		return nil
	}
}

func NewRegistration(ns string, dbRegister serviceregistry.DBRegister) *serviceregistry.Service[keyMgmtConnect.KeyManagementServiceHandler] {
```

service/policy/keymanagement/key_management.go
```diff
@@ -0,0 +1,173 @@
...
+	switch req := req.Msg.GetIdentifier().(type) {
+	case *keyMgmtProto.GetProviderConfigRequest_Id:
+		ksvc.logger.Debug("Getting Provider config by ID", slog.String("ID", req.Id))
+	case *keyMgmtProto.GetProviderConfigRequest_Name:
+		ksvc.logger.Debug("Getting Provider config by Name", slog.String("Name", req.Name))
+	default:
+		return nil, connect.NewError(connect.CodeInvalidArgument, nil)
+	}
+
+	auditParams := audit.PolicyEventParams{
```
jakedoublev, line 85: You may disagree, but I don't think we need audit logs for `read` events here, as we don't utilize them throughout the rest of the policy services.

service/policy/keymanagement/key_management.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		ksvc.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
+		return nil, db.StatusifyError(err, db.ErrTextGetRetrievalFailed, slog.String("id", providerConfigID))
+	}
+
+	pc, err := ksvc.dbClient.UpdateProviderConfig(ctx, req.Msg)
+	if err != nil {
+		ksvc.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
+		return nil, db.StatusifyError(err, db.ErrTextUpdateFailed, slog.String("keyManagementService", req.Msg.GetId()))
+	}
```
jakedoublev, line 140: I would advise using a transaction for this in case customers employ db replicas so you don't unintentionally introduce a data consistency problem. Here's an example: https://github.com/opentdf/platform/blob/main/service/policy/subjectmapping/subject_mapping.go#L86

service/policy/namespaces/namespaces.go
```diff
@@ -257,10 +257,44 @@ func (ns NamespacesService) RemoveKeyAccessServerFromNamespace(ctx context.Conte
...
-func (ns NamespacesService) AssignKeyToNamespace(context.Context, *connect.Request[namespaces.AssignKeyToNamespaceRequest]) (*connect.Response[namespaces.AssignKeyToNamespaceResponse], error) {
-	return nil, connect.NewError(connect.CodeUnimplemented, nil)
+func (ns NamespacesService) AssignPublicKeyToNamespace(ctx context.Context, r *connect.Request[namespaces.AssignPublicKeyToNamespaceRequest]) (*connect.Response[namespaces.AssignPublicKeyToNamespaceResponse], error) {
+	rsp := &namespaces.AssignPublicKeyToNamespaceResponse{}
+
+	key := r.Msg.GetNamespaceKey()
+	auditParams := audit.PolicyEventParams{
+		ActionType: audit.ActionTypeCreate,
+		ObjectType: audit.ObjectTypeKasAttributeNamespaceKeyAssignment,
+		ObjectID:   fmt.Sprintf("%s-%s", key.GetNamespaceId(), key.GetKeyId()),
```
jakedoublev, line 267: Should we consider a different delimiter here as UUIDs already contain hyphens?

service/integration/keymanagement_test.go
```diff
@@ -0,0 +1,304 @@
...
+	s.Nil(pc)
+}
+
+func (s *KeyManagementSuite) Test_GetProviderConfig_InvalidName_Fails() {
+	pc, err := s.db.PolicyClient.GetProviderConfig(s.ctx, &keymanagement.GetProviderConfigRequest_Name{
+		Name: "",
+	})
+	s.Require().Error(err)
+	s.Nil(pc)
+}
```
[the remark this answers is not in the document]
c-r33d: Yep, moved them to the protovalidation layer. Really one was already a duplicate

service/policy/db/migrations/20250429000003_kas_source_type_update.sql
c-r33d: I did this since it is consistent with the rest of the design. Might be good to come back and migrate everything to use the string version of the enum. I will remove the migration for the source type, but keep the others for now.

c-r33d: https://virtru.atlassian.net/browse/DSPX-1022

service/policy/kasregistry/key_access_server_registry.go
c-r33d: I do think for reading asymmetric keys it should be audited

service/policy/db/migrations/20250429000000_keys_table.sql
jakedoublev: Right now it's largely up to the implementing developer to not mess up a migration, so any automated migration testing would be a really nice improvement.

service/policy/keymanagement/key_management.go
jakedoublev: Totally accurate, but the Get above could be out of sync with the Update. There's no guarantee the service handler would hit the same db replica for both roundtrips. In this case, the only downside is a theoretical data inconsistency in audit logs, so I'm not too concerned about merging this as-is, but just something to think about.

jakedoublev:
Have we verified all audit logs forgo sensitive key information? Just to triple-check.

c-r33d:
> Have we verified all audit logs forgo sensitive key information? Just to triple-check.

I have triple checked audit logs and think they are good. One thing we might need to consider is what happens when the json is invalid for the private key ctx, so the db sends an error. I need to double-check if it will print out the entire json response. If so, I will need to handle statusfying db errors around keys differently.

jakedoublev (approved):

PR #2113, opened 2025-04-25
fix(ci): Update expired ca and certs in oauth unit tests

elizabethhealy:
### Proposed Changes

* regen ca and certs for oauth unit tests

### Checklist


dmihalcik-virtru (approved):
nit: probably ca-2025 or something would be better but this is fine

PR #2124, opened 2025-04-28
feat(authz): authz v2, ers v2 protos and gencode for ABAC with actions & registered resource 

jakedoublev:
### Proposed Changes

Versioned v2 RPCs for authorization service:

1. logically resolve actions permitted for each attribute value within the GetEntitlements/GetDecision responses, as these are part of the desired ABAC decisioning
2. support `registered_resources` (formerly NDR) in auth service requests
3. separate individual, multi resource, and bulk decision requests structurally

### Checklist


strantalis:
@jakedoublev Do we need to discuss how we want to version our protos. I think there are probably a couple different strategies like what you did but also making it package level like `policy.v1`. 

https://buf.build/docs/best-practices/style-guide/
https://buf.build/docs/lint/rules/#package_version_suffix

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,160 @@
...
+  // ephemeral id for tracking between request and response
+  string ephemeral_id = 1;
+  // Standard entity types supported by the platform
+  oneof entity_type {
+    // one of the entity options must be set
+    string email_address = 2;
+    string user_name = 3;
+    string remote_claims_url = 4;
+    string uuid = 5;
+    google.protobuf.Any claims = 6;
```
jakedoublev, line 26: I wonder if we can remove these and push any use cases into `EntityCustom`

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,161 @@
...
+  // ephemeral id for tracking between request and response
+  string ephemeral_id = 1;
+  repeated Entity entities = 2;
+}
+
+// Can the entity (PE or NPE) represented by each entity chain take the provided action on the set of resources
+message DecisionRequest {
+  EntityChain entity = 1;
+  policy.Action action = 2;
+  Resource resource = 3;
```
jakedoublev, line 55:
```suggestion
  repeated Resource resources = 3;
```

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,161 @@
...
+  // ephemeral id for tracking between request and response
+  string ephemeral_id = 1;
+  // Standard entity types supported by the platform
+  oneof entity_type {
+    // one of the entity options must be set
+    string email_address = 2;
+    string user_name = 3;
+    string remote_claims_url = 4;
+    string uuid = 5;
+    google.protobuf.Any claims = 6;
```
jakedoublev, line 26:
```suggestion
```

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,226 @@
...
+    message: "action.name must be provided"
+    expression: "has(this.action.name)"
+  };
+}
+message GetDecisionResponse {
+  // ephemeral id for tracking between request and response
+  string ephemeral_entity_chain_id = 1;
+  // ephemeral id for tracking between request and response
+  string ephemeral_resource_id = 2;
+  // name on action is required
```
ryanulit, line 98: Since this is a response object, I don't think this comment makes sense here.

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// Can the entity (PE or NPE) access?
+// 1. one entity chain
+// 2. one action
+// 3. one resource
+message GetDecisionRequest {
+  authorization.EntityChain entity = 1 [(buf.validate.field).required = true];
+  // name on action is required
+  policy.Action action = 2 [(buf.validate.field).required = true];
```
ryanulit, line 86: Any reason not to just use `string action_name` instead of requiring the full object? Could the RPC logic just look it up or will the consumer always have the full action ready to pass in?

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,226 @@
...
+  repeated GetDecisionMultiResourceResponse decision_responses = 1;
+}
+
+// What can the entity do, broken down as permitted actions on attribute value FQNs
+//
+// Note: the v1 API parameter 'scope' has been dropped, and it is recommended to use
+// GetDecision if the resource is known
+message GetEntitlementsRequest {
+  // list of requested entities
+  // todo: should this be an entity chain?
```
ryanulit, line 155: Will not deciding this now cause a need for a v3 service?

ryanulit (changes requested):

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,224 @@
...
+  // the decision result
+  Decision decision = 3;
+}
+
+// Can the entity (PE or NPE) access?
+// 1. one entity chain
+// 2. one action
+// 3. one resource
+message GetDecisionRequest {
+  authorization.EntityChain entity = 1 [(buf.validate.field).required = true];
```
elizabethhealy, line 84: nit: can the field be named entity_chain or similar? just "entity" may be confusing if there are multiple entities in the chain

elizabethhealy, line 84: does using the old entity chain message also use the old entity message?

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,224 @@
...
+  // name on action is required
+  policy.Action action = 2 [(buf.validate.field).required = true];
+  Resource resource = 3 [(buf.validate.field).required = true];
```
elizabethhealy, line 87: if a decision request only has a single resource, do we need the ephemeral_resource_id in the ResourceDecision on line 74 since theres only one? same thing for the entity chain ephemeral id in the decision response i guess - if were only doing a 1x1 evaluation do we need ids?

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,224 @@
...
+}
+message GetDecisionMultiResourceResponse {
+  // ephemeral id for tracking between request and response
+  string ephemeral_entity_chain_id = 1;
+  // name on action is to be relied upon
+  policy.Action action = 2;
+  // convenience flag indicating global resource decisions result (permit/deny)
+  google.protobuf.BoolValue all_permitted = 3;
+  // individual resource decisions
+  repeated ResourceDecision resource_decisions = 4;
```
elizabethhealy, line 135: i guess this is where you need the ephemeral resource ids in the decision responses, when you have many of them. but do you need the ephemeral_entity_chain_id?

service/authorization/authorization.proto
```diff
@@ -2,15 +2,16 @@ syntax = "proto3";
...
 import "policy/objects.proto";
 
-
 message Token {
-  string id = 1; // ephemeral id for tracking between request and response
-  string jwt = 2; // the token
+  // ephemeral id for tracking between request and response
+  string id = 1;
+  // the token
+  string jwt = 2 [(buf.validate.field).required = true];
```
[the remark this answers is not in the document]
jakedoublev: Yes, protovalidation tests will be handled separately as they can't be merged in simultaneously when the `protocol/go` module will contain new unreleased proto gencode.

service/authorization/v2/authorization.proto
jakedoublev: This is a good thought, and I debated this, but I think it's wise to use the `policy.Action` message to avoid a breaking change if referencing actions by IDs starts to make more sense down the road. If we went with `string action_name` now, we couldn't deprecate or walk that back as easily later as simply removing the validation rule and adding lookup by action ID logic.

service/entityresolution/entity_resolution.proto
```diff
@@ -24,10 +25,11 @@ import "google/api/annotations.proto";
 
 */
 message ResolveEntitiesRequest {
-    repeated authorization.Entity entities = 1;
+  // either v1 or v2 authorization entities are permitted, but not both
```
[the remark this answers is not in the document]
jakedoublev: @elizabethhealy asks if we should move to a v2 ers as well and keep the authorization.v2.Entity changes

service/authorization/v2/authorization.proto
jakedoublev: I'm still not sure if this should be a single Entity or a chain, but the idea of a chain is that it's something ERS builds

jakedoublev: allow a chain or a token we resolve into a chain within auth service

service/authorization/v2/authorization.proto
jakedoublev: clean up the response to omit what is just repeating the request fields right back

service/authorization/v2/authorization.proto
jakedoublev: needed for multi-resource decisioning, and bulk RPCs

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,161 @@
...
+  EntityReference entity = 1 [(buf.validate.field).required = true];
+  // optional parameter to return all entitled values for attribute definitions with hierarchy rules, propagating
+  // down the hierarchical values instead of returning solely the value that is directly entitled
+  optional bool with_comprehensive_hierarchy = 2;
+}
+message GetEntitlementsResponse {
+  repeated EntityEntitlements entitlements = 1;
+}
+
+service AuthorizationService {
```
jakedoublev, line 155: Removed `ByToken` RPCs since we're already going to be switching on the per entity type in each RPC to accommodate registered resources, so might as well just capture token in the same logic.

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,161 @@
...
+  DECISION_UNSPECIFIED = 0;
+  DECISION_DENY = 1;
+  DECISION_PERMIT = 2;
+  // DECISION_OBLIGATED = 3;
+}
+
+// The EntityReference is the actor in an entitlement or decision request - the PE, NPE, or PE+NPE being authorized.
+// The abstraction houses the distinct entity types, PE and/or NPE combinations, or a registered
+// resource value being treated as an entity in entitlement/authorization decisioning.
+message EntityReference {
```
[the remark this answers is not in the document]
jakedoublev: `SourceEntity` or `EntitySource` may be an option as well (probably the latter)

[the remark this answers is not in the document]
jakedoublev: `EntityIdentifier` with `oneof identifier`

service/entityresolution/entity_resolution.proto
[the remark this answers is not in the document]
jakedoublev: We are moving to v2 ERS RPCs handling entity proto messages broken out of authorization service.

service/authorization/v2/authorization.proto
jakedoublev: This has been updated.

service/authorization/v2/authorization.proto
```diff
@@ -0,0 +1,161 @@
...
+    // chain of one or more entities
+    entity.EntityChain entity_chain = 1 [(buf.validate.field).cel = {
+      id: "entity_chain_required"
+      message: "entities must be provided"
+      expression: "has(this.entities) && this.entities.size() > 0"
+    }];
+
+    // fully qualified name of the registered resource value stored in platform policy, where in
+    // this case the resource acts as and represents a single entity for authorization/entitlement decisioning
+    string registered_resource_value_fqn = 2 [(buf.validate.field).string.min_len = 1];
```
ryanulit, line 33:
```suggestion
    string registered_resource_value_fqn = 2 [
      (buf.validate.field).string = {
        min_len: 1
        uri: true
      }
    ];
```

service/authorization/authorization.proto
ryanulit, line 1: Nit: just undo this change to avoid extra git history

service/entityresolution/entity_resolution.proto
ryanulit, line 1: Nit: same as other comment

ryanulit (changes requested):

ryanulit (approved):

jrschumacher (approved):

strantalis (approved):

PR #2131, opened 2025-04-29
feat(policy): DSPX-1018 NDR retrieval by FQN support

ryanulit:
### Proposed Changes

* update service/db logic to support retrieval by FQN(s)
* refactors some proto field names to be more consistent with other protos (breaking changes per buf cli, but not actually breaking because no one is using this functionality yet) 

### Checklist


service/policy/db/registered_resources.go
```diff
@@ -277,6 +281,56 @@ func (c PolicyDBClient) GetRegisteredResourceValue(ctx context.Context, r *regis
...
+		if err != nil {
+			return nil, err
+		}
+
+		rv, err := c.Queries.getRegisteredResourceValue(ctx, getRegisteredResourceValueParams{
+			Name:  parsed.Name,
+			Value: parsed.Value,
+		})
+		if err != nil {
+			if errors.Is(err, pgx.ErrNoRows) {
```
ryanulit, line 302: Is there a better way to do this? Move the pgx import to the `service/pkg/db/errors.go` file so that this DB client code isn't dependent on the provider implementation? Or change the SQL query to use `LIMIT 1` or something (if possible)?

service/pkg/util/fqn.go
```diff
@@ -11,15 +11,25 @@ type FullyQualifiedResourceMappingGroup struct {
 	GroupName string
 }
 
+type FullyQualifiedRegisteredResourceValue struct {
+	Fqn   string
+	Name  string
+	Value string
+}
+
 // protovalidate already validates the FQN format in the service request
```
ryanulit, line 20: This is actually only true for the Resource Mapping Group FQNs, as all other attribute/value FQNs only validate that a proper URI is provided. Should we only care that a proper URI is provided in the request protos?

service/pkg/util/fqn_test.go
```diff
@@ -0,0 +1,60 @@
...
+	require.Equal(t, "namespace.com", parsed.Namespace)
+	require.Equal(t, "group_name", parsed.GroupName)
+}
+
+func TestParseResourceMappingGroupFqn_Invalid_Fails(t *testing.T) {
+	invalidFQNs := []string{
+		"",
+		"invalid",
+		"https://namespace.com",
+		"https://namespace.com/resm",
```
jakedoublev, line 24:
```suggestion
		"https://namespace.com",
		"http://namespace.com/resm/group_name",
		"somethinghttps://namespace.com/resm/group_name",
		"https://namespace.com/resm",
```

service/pkg/util/fqn_test.go
```diff
@@ -0,0 +1,60 @@
...
+	require.Equal(t, "valid", parsed.Name)
+	require.Equal(t, "test", parsed.Value)
+}
+
+func TestParseRegisteredResourceValueFqn_Invalid_Fails(t *testing.T) {
+	invalidFQNs := []string{
+		"",
+		"invalid",
+		"https://reg_res",
+		"https://reg_res/invalid",
```
jakedoublev, line 50:
```suggestion
		"https://reg_res",
		"https://reg_res/invalid",
		"http://reg_res/test/value/something",
		"somethinghttps://reg_res/test/value/something",
```

service/policy/db/registered_resources.go
jakedoublev: I'd suggest trying to refactor the SQL query. I think you could do it with a single roundtrip to surface the error?

Something like this: 
```sql
WITH name_value_pairs AS (
    SELECT unnest(:names::text[]) AS name, 
           unnest(:values::text[]) AS value
)
SELECT 
    rr.id AS resource_id, 
    rr.name AS resource_name,
    rrv.id AS value_id,
    rrv.value
FROM registered_resources rr
JOIN registered_resource_values rrv ON rr.id = rrv.registered_resource_id
JOIN name_value_pairs nvp ON rr.name = nvp.name AND rrv.value = nvp.value;
```

service/pkg/util/fqn.go
ryanulit: Per discussion with @jakedoublev, we decided sticking to the protovalidate `string.uri` validation rather than individual FQN regexes is the best path forward. Then the service/db code can handle parsing them for accuracy.

service/pkg/util/fqn_test.go
ryanulit: done

service/pkg/util/fqn_test.go
ryanulit: done

jakedoublev (dismissed):

strantalis (approved):

strantalis (approved):

PR #2132, opened 2025-04-29
fix(core): Improves errors when under heavy load

dmihalcik-virtru:
### Proposed Changes

* the nanotdf benchmark is causing errors that seem to be due to keycloak faiiling?
* or maybe just connectRPC failing?
* Either way, the error should be a 500 not a 403
* and the sdk isn't properly unwrapping the new response format to read it

### Checklist


service/kas/access/rewrap.go
```diff
@@ -716,7 +716,7 @@ func (p *Provider) nanoTDFRewrap(ctx context.Context, requests []*kaspb.Unsigned
 
 	pdpAccessResults, accessErr := p.canAccess(ctx, tok, policies)
 	if accessErr != nil {
```
jentfoo, line 718: Should there be a type check (errors.Is) to determine 403 vs 500?

dmihalcik-virtru: It looks like it is failing with `codes.Internal`

See definition here: https://pkg.go.dev/google.golang.org/grpc@v1.72.0/codes#Code

I could do a check with `status.Code` and filter by that, but it seems like we probably should be doing something smarter. Also, the fact that this is even happening at all is worth investigating further. I'm not convinced the problem isn't caused by the client itself canceling requests due to bugs in how the benchmark is implemented or the client is used, for example. I mostly just want to clean up the `benchmark` comment for now and make the error at least semi-reasonable (with this change, it at least reflects that the server response contains some kind of error, instead of failing to properly parse the response at all and failing when trying to parse an empty field)

strantalis (approved):

PR #2140, opened 2025-05-01
chore(ci): adds more context to autobump module version pr's

strantalis:
### Proposed Changes

*

### Checklist


dmihalcik-virtru:
It looks like now that the dependabot config is fixed, so we can probalby instead remove all of this custom code, in favor of just suggesting people click the 'check for updates' button on:

https://github.com/opentdf/platform/network/updates/21141832/jobs

This will generate better looking PRs than even this PR produces, and probably be more reliable/less maintenance.

Issue #1209, opened 2024-07-25
ADR: Resource Mapping enhancements to accommodate PEP needs


jrschumacher:
# Resource Mapping enhancements to accommodate PEP needs 

## Context and Problem Statement

As resource mappings have been explored by PEPs, it has been determined they lack some functionality. This ADR will cover a variety of changes to the resource mappings. The key changes include:

- needing FQN to group resource mappings
	- assign resource mappings to namespace
	- give resource mappings a unique name
- RPCs
  - get resource mapping by fqns
  - get resource mappings by attribute fqn
  - unsafe behaviors
- reversal functionality

### Resource Mapping FQN

The resource mapping FQN is focused on the ability to reference a unique mapping by a unique string that is portable. With the latest addition of policy import/export and the desire of our users to subscribe to external policy bundles we need to ensure there is uniqueness that is meaningful and can be transported.

To support this the suggested FQN structure is `https://<namespace>/<prefix>/<unique-name>`. This will align with our existing attribute structure of `https://demo.com/attr/my_attr/val/my_val`

#### Assign to namespace

Assigning the resource mapping to a namespace will require adding a new foreign key to the namespace table. This relationship will be a `1:*` between namespace and resource mapping.

**Migration:** each resource mapping will need to be tied to a namespace. If there is one namespace in the existing policy, then we will automatically assign resource mappings to those namespaces. If there are more than one, we will need to recommend that users migrate to a single namespace and manually rectify the situation later.

Going forward, every resource mapping **MUST** have a namespace.

#### Assign a unique name

Assigning a unique name to resource mappings will require that each resource mapping will get a unique name.

**Migration:** automatically name the resource mappings based on the known data between the terms and the attributes. Users can change these names using unsafe services or by directly manipulating the database.

### Proto & RPC

Protos will need to be updated to support the new properties of namespace and name.

#### Get Resource Mapping By FQNs

This RPC will primarily be used by PEPs.

We need the ability to one or more resource mapping by FQN. This means we should be able to fetch a collection of resource mappings based on an array of FQNs provided. These FQNs do not need to be associated to one or another.

If the FQN does not exist, we should indicate that the single FQN was not found and return the rest of the values. The request should be considered successful in general.

If no FQNs are found then the request should be considered a failure.

#### Get Resource Mappings by Attribute FQN

This RPC will primarily be used by admins.

Admins will likely want to know which resource mappings are associated to an attribute FQN. This will only support a single Attribute FQN.

### SDK

[...]

PR #2148, opened 2025-05-04
feat(core): introduce FQN utility library

### LINKED ISSUE #515: Policy: a shared library used throughout the platform for FQN construction, de-construction, and groupings
### Background

There is a need to look up attributes (namespaces, definitions, values) by FQNs, and therefore a need to build, store, validate, and tear them down. https://github.com/opentdf/platform/issues/95 added the functionality to store FQNs in the database for lookup FQN -> Policy Attribute Objects, but there are needs in Access PDP, test functions, and the policy services to dynamically build FQNs going from Policy Attribute Objects -> FQNs. This should be common functionality 

We should also expose this functionality on the SDK for use in PEPs without a separate import.

### Acceptance Criteria

1. A shared suite of functions is built and consumed throughout to build and tear apart FQNs
2. Unit tests thoroughly cover them
3. Existing helper funcs and `fmt.Sprintf` calls are replaced by the shared functions
4. The policy `GetAttributesByValueFQNs` request (consumed by Auth service requests) successfully populates FQNs for all policy objects in response without doing joins with the FQNs table, deferring to composite keys and foreign keys to preserve relational integrity

jakedoublev:
### Goals
* allow all PEPs, services, and SDK to use same validation logic, reducing drift
* open the door to URN support

### Proposed Changes
* new `policyidentifier` library
* new FQN parsers and builder functions (with tests) for attributes
* moves service/pkg/util FQN parsers to library, and updates imports
* once this is merged, we can pull it in as a dependency in the `service` module and deprecate our `fqnBuilder` implementations as well as `service/pkg/util/fqn.go`

Closes #515

### Checklist

- [X] I have added or updated unit tests


lib/fqn/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/opentdf/platform/lib/fqn
+
+go 1.23
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/stretchr/testify v1.10.0 // indirect
```
dmihalcik-virtru, line 8: Shouldn't this be direct?

jakedoublev: Yes - good eye. I forgot to `go mod tidy` this new module. Pushed up a new commit for that.

go.work
```diff
@@ -7,6 +7,7 @@ use (
 	./lib/fixtures
 	./lib/flattening
 	./lib/ocrypto
+	// ./lib/fqn
```
dmihalcik-virtru, line 10:
```suggestion
	./lib/fqn
```

lib/fqn/fqn.go
```diff
@@ -0,0 +1,250 @@
+package util
```
dmihalcik-virtru, line 1:
```suggestion
package fqn
```

lib/fqn/fqn.go
```diff
@@ -0,0 +1,250 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+type FullyQualifiedResourceMappingGroup struct {
```
dmihalcik-virtru, line 10:
```suggestion
type Group struct {
```

lib/fqn/fqn.go
```diff
@@ -0,0 +1,250 @@
...
+	Fqn       string
+	Namespace string
+	GroupName string
+}
+
+type FullyQualifiedRegisteredResourceValue struct {
```
dmihalcik-virtru, line 16:
```suggestion
type RegisteredResourceValue struct {
```

lib/fqn/fqn.go
```diff
@@ -0,0 +1,250 @@
...
+	Fqn   string
+	Name  string
+	Value string
+}
+
+// Structs and regexes for attribute FQNs
+type FullyQualifiedAttribute struct {
```
dmihalcik-virtru, line 23:
```suggestion
type AttributeValue struct {
```

lib/fqn/fqn_test.go
```diff
@@ -0,0 +1,608 @@
+package util
```
dmihalcik-virtru, line 1:
```suggestion
package fqn_test
```

or just `fqn`

lib/fqn/fqn.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+
+		parsed.Namespace = ns
+		return parsed, nil
+	}
+
+	return nil, errors.New("error: invalid attribute FQN format, must be https://<namespace>, https://<namespace>/attr/<name>, or https://<namespace>/attr/<name>/value/<value>")
+}
+
+func AttributeFqnBuilder(namespace, name, value string) (string, error) {
```
dmihalcik-virtru, line 223:
```suggestion
func AttributeBuilder(namespace, name, value string) (string, error) {
```

Also shouldn't this return AttributeValue or something other than `string`?

go.work
ryanulit, line 10: Is this commented just for testing purposes?

lib/fqn/fqn.go
```diff
@@ -0,0 +1,250 @@
...
+	// Only match exactly https:// followed by a namespace with no forward slashes
+	namespaceOnlyRegex = regexp.MustCompile(
+		`^https:\/\/(?<namespace>[^\/]+)$`,
+	)
+
+	// protovalidate already validates the FQN format in the service request
+	// for parsing purposes, we can just look for any non-whitespace characters
+	// e.g. should be in format of "https://<namespace>/resm/<group name>"
+	resourceMappingGroupFqnRegex = regexp.MustCompile(
+		`^https:\/\/(?<namespace>\S+)\/resm\/(?<groupName>\S+)$`,
```
ryanulit, line 55: If the other regexes are going to use `[^\/]+` to match on a value with any character but `/`, then might as well make it consistent in this and the registered resources regexes.

go.work
jakedoublev: @ryanulit @dmihalcik-virtru It looks like the image build action changes when this is added to the `go.work`. I'm guessing it's because the module is not yet merged?

lib/identifier/resource_mapping_group.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+type FullyQualifiedResourceMappingGroup struct {
+	Namespace string
+	GroupName string
```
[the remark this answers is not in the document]
jakedoublev: https://github.com/opentdf/platform/issues/1209 It is a relation to hold sets of resource mappings so you can group them, for example by tagging vendor.

lib/identifier/identifier.go
```diff
@@ -0,0 +1,77 @@
...
+type FullyQualified interface {
+	// FQN builds the FQN identifier string with the object names/values.
+	FQN() string
+	// Validate checks if the names/values are valid according to the regex rules.
+	Validate() error
+}
+
+// Parse parses an identifier (FQN) string into a specific type of FullyQualified object
+// and validates the overall structure along with each name/value of the object fields.
+func Parse[T FullyQualified](identifier string) (T, error) {
```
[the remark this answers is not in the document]
jakedoublev: This is nice because if we support URNs, or drop the schema requirement from FQNs, then the `Parse` function can figure out what kind of identifier was passed in rather than having clients do their own regex checks to see what they have like we do in our SDK files and various unit tests.

lib/identifier/attribute.go
```diff
@@ -0,0 +1,151 @@
+package identifier
```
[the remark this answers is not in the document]
jakedoublev: qualifier? I think the first API is actually better DX when we support more than FQNs, like URNs, or updates to current FQNs are realized.

```golang
// http scheme (currently not allowed by policy services)
a, err := identifier.Parse[*identifier.FullyQualifiedAttribute](
    "http://name.space/attr/name/value/val"
)
// https scheme (current state allowed by policy services)
a, err := identifier.Parse[*identifier.FullyQualifiedAttribute](
   "https://name.space/attr/name/value/val"
)
// urn
a, err := identifier.Parse[*identifier.FullyQualifiedAttribute](
   "urn:name.space:attr:name:value:val"
)
```

```golang
// in the caller, such as SDK or tests
attr := manifest.policy.attributes[0] // "http://name.space/attr/name/value/val"
// as a caller, I have to check if I'm looking at an fqn and hope I'm in sync with validations of services
isFQN := validFQNregex.test(attr)
if isFQN {
    a, err := fqn.ParseAttribute(attr)
} else {
    a, err := urn.ParseAttribute(attr)
}
```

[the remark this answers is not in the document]
jakedoublev: Perhaps `lib/policyidentifier`?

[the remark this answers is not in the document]
dmihalcik-virtru: I guess that is fine, if a bit long. Are you planning to re-export these types? Then arguably the lib/* folder isn't the place for them, as these are generally marked as 'internal' in the READMEs

go.work
jakedoublev: Updated CI to fix and pushed up a commit including this in go.work

lib/fqn/fqn.go
jakedoublev: pushed up a commit to address https://github.com/opentdf/platform/pull/2148/commits/ec49870770cd0095060de4cfecf15f21c081b33d

lib/policyidentifier/attribute.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	// If not a value FQN, try to match against the attribute definition pattern
+	defMatches := attributeDefinitionFQNRegex.FindStringSubmatch(fqn)
+	if len(defMatches) > 0 {
+		namespaceIdx := attributeDefinitionFQNRegex.SubexpIndex("namespace")
+		nameIdx := attributeDefinitionFQNRegex.SubexpIndex("name")
+
+		if len(defMatches) <= namespaceIdx || len(defMatches) <= nameIdx {
+			return nil, errors.New("error: valid attribute definition FQN format https://<namespace>/attr/<name> must be provided")
```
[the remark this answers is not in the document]
jakedoublev: 👍

dmihalcik-virtru (dismissed):
A good start. We may want to do some more work to canonicalize things to avoid the many problems http(s) URLs often have, and make sure we canonicalize the URL content via appropriate algorithms (hostname unicode normalization and ascii conversion, URL safe characters in the path), but that can be done later (I'm pretty sure we aren't doing it now outside of may be the web sdk?)

ryanulit (approved):

PR #2180, opened 2025-05-06
feat(policy): Rotate Keys

c-r33d:
### Proposed Changes

1.) Add the ability to rotate kas keys.
2.) Return an informative response including all the resources that were effected by the rotation.
3.) Audit rotated resources.

### Checklist


service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -493,20 +494,32 @@ message RotateKeyRequest {
...
     // Common metadata fields
-    common.Metadata metadata = 100;
+    common.MetadataMutable metadata = 100;
   }
 }
 
+/*
+* All resources that were rotated as part of the key rotation process
+*/
+message RotatedResources {
```
strantalis, line 510: Maybe we swap out `string` for a new message so we can expand fields later. 
```proto
message X {
  string id = 1;
  string fqn = 2;
}
```

strantalis (approved):

PR #2191, opened 2025-05-07
feat(policy): DSPX-1057 registered resource action attribute values

ryanulit:
### Proposed Changes

*

### Checklist


service/policy/db/schema_erd.md
```diff
@@ -122,7 +122,7 @@ erDiagram
         jsonb metadata "Metadata for the KAS (see protos for structure)"
         character_varying name UK "Optional common name of the KAS"
         jsonb public_key "Public key of the KAS (see protos for structure/options)"
-        integer source_type 
+        character_varying source_type
```
ryanulit, line 125: This was changed when running the make script, is it correct?

service/policy/objects.proto
```diff
@@ -430,6 +430,19 @@ message RegisteredResourceValue {
 
   RegisteredResource resource = 3;
 
+  repeated RegisteredResourceActionAttributeValue action_attribute_values = 4;
+
+  // Common metadata
+  common.Metadata metadata = 100;
+}
+
+message RegisteredResourceActionAttributeValue {
```
ryanulit, line 439: Move this inside of `RegisteredResourceValue` to make it clear this is only used within a RRV.

service/policy/db/schema_erd.md
jakedoublev: Yeah, looks like the `models.go` has it as a `pgtype.Text` and IIRC, @c-r33d changed the column value from the proto enum index to the string representation

service/integration/registered_resources_test.go
```diff
@@ -560,6 +565,39 @@ func (s *RegisteredResourcesSuite) Test_CreateRegisteredResourceValue_WithMetada
 	s.Require().Len(created.GetMetadata().GetLabels(), 2)
 }
 
+func (s *RegisteredResourcesSuite) Test_CreateRegisteredResourceValue_With_ActionAttributeValues_Succeeds() {
```
jakedoublev, line 568: Do you think it's worth it to add a multiples scenario or update this one to have two `ActionAttributeValue` combos?

service/integration/registered_resources_test.go
```diff
@@ -587,66 +625,185 @@ func (s *RegisteredResourcesSuite) Test_CreateRegisteredResourceValue_WithNonUni
...
-
-func (s *RegisteredResourcesSuite) Test_GetRegisteredResourceValue_ByInvalidID_Fails() {
-	got, err := s.db.PolicyClient.GetRegisteredResourceValue(s.ctx, &registeredresources.GetRegisteredResourceValueRequest{
-		Identifier: &registeredresources.GetRegisteredResourceValueRequest_Id{
-			Id: invalidID,
+		{
+			name: "By FQN",
+			req: &registeredresources.GetRegisteredResourceValueRequest{
+				Identifier: &registeredresources.GetRegisteredResourceValueRequest_Fqn{
+					Fqn: fmt.Sprintf("https://reg_res/%s/value/%s", existingRes.Name, existingResValue1.Value),
```
jakedoublev, line 736: Would you mind please using the `lib/identifier` functionality here instead of Sprintf?

jakedoublev (commented):
Looks great so far. I know you're going to make a couple more changes, so won't give it a ✅.

service/integration/registered_resources_test.go
ryanulit: `Sprintf` is just so much less verbose than this though...

```
fqn := identifier.FullyQualifiedRegisteredResourceValue{
	Name: existingRes.Name,
	Value: existingResValue1.Value,
}
fqnStr := x.FQN()
```

and it is a test...

service/integration/registered_resources_test.go
ryanulit: Good call, I think updating this test to have 2 is fine and further tests the `copyfrom` sqlc logic. Will add another.

jakedoublev (approved):

strantalis:
👍

PR #2200, opened 2025-05-09
feat(sdk): Use ConnectRPC in the go client

elizabethhealy:
### Proposed Changes

* use connect rpc clients in the go sdk
* introduced sdkconnect wrappers and interfaces to keep the same grpc function signatures (except for grpc call options) -- these are generated via sdk/internal/gencode
* adds a type ConnectRpcConnection which is composed of the http client, endpoint, and connect options
* adds options to support override of new connection types and connect options
* service updates to work over ipc with the memhttp listener
* still supports grpc gateway

### Checklist


elizabethhealy:
passing xtest: https://github.com/opentdf/platform/actions/runs/15117769962/job/42492697748?pr=2200

sdk/sdk.go
```diff
@@ -447,43 +384,33 @@ func IsValidNanoTdf(reader io.ReadSeeker) (bool, error) {
...
-	if err != nil {
-		return errors.Join(ErrGrpcDialFailed, err)
-	}
-	defer conn.Close()
-
-	req := healthpb.HealthCheckRequest{}
-	healthService := healthpb.NewHealthClient(conn)
-	resp, err := healthService.Check(context.Background(), &req)
-	if err != nil || resp.GetStatus() != healthpb.HealthCheckResponse_SERVING {
+func ValidateHealthyPlatformConnection(platformEndpoint string, httpClient *http.Client) error {
```
strantalis, line 388: does this need to be exposed?

sdk/sdk.go
```diff
@@ -306,23 +258,13 @@ func buildIDPTokenSource(c *config) (auth.AccessTokenSource, error) {
 	return ts, err
 }
 
-// Close closes the underlying grpc.ClientConn.
+// TODO: Remove after otdfctl updates
 func (s SDK) Close() error {
```
elizabethhealy, line 262: can remove this after otdfctl changes, dont want to block xtests

dmihalcik-virtru: Don't we need to clean up the collection store?

sdk/internal/codegen/main.go
```diff
@@ -0,0 +1,13 @@
+package main
+
+import (
+	"log"
+
+	"github.com/opentdf/platform/sdk/internal/codegen/runner"
+)
+
+func main() {
```
dmihalcik-virtru, line 9: Aded a comment describing what happens here, and maybe also (re)generate a README.md for the sdkconnect package as well, explaining how to re-generate (or modify, if required)

sdk/internal/codegen/runner/generate.go
```diff
@@ -0,0 +1,245 @@
...
+					}
+					if ts.Name.Name == client.grpcClientInterface {
+						packageName := path.Base(client.grpcPackagePath)
+						if client.packageNameOverride != "" {
+							packageName = client.packageNameOverride
+						}
+						code := generateWrapper(ts.Name.Name, iface, client.grpcPackagePath, packageName, client.suffix)
+						var currentDir string
+						currentDir, err = getCurrentFileDir()
+						outputPath := filepath.Join(currentDir, "..", "..", "..", "sdkconnect", packageName+".go")
```
dmihalcik-virtru, line 123: Maybe add an assertion about the output path to make sure it 'looks right', since this could be called from the 'wrong' folder

sdk/internal/codegen/runner/generate.go
```diff
@@ -0,0 +1,245 @@
...
+						err = os.WriteFile(outputPath, []byte(code), 0o644) //nolint:gosec // ignore G306
```
dmihalcik-virtru, line 124: maybe log this?

sdk/internal/codegen/runner/generate.go
```diff
@@ -0,0 +1,245 @@
...
+			return fmt.Errorf("error writing file: %w", err)
+		}
+	}
+	return nil
+}
+
+func getCurrentFileDir() (string, error) {
+	_, filename, _, ok := runtime.Caller(0)
+	if !ok {
+		return "", errors.New("could not get caller information")
```
dmihalcik-virtru, line 151:
```suggestion
		return "", errors.New("could not get caller current working directory")
```

sdk/internal/codegen/runner/generate.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+// Generate wrapper code for the Connect RPC client interface
+func generateWrapper(interfaceName string, interfaceType *ast.InterfaceType, packagePath string, packageName string, suffix string) string {
+	// Get method names dynamically from the interface
+	methods := getMethodNames(interfaceType)
+	connectPackageName := packageName + "connect"
+
+	// Start generating the wrapper code
+	wrapperCode := fmt.Sprintf(`// Wrapper for %s (generated code) DO NOT EDIT
```
dmihalcik-virtru, line 174: I have a slight pref for https://pkg.go.dev/text/template but this is fine

sdk/internal/codegen/runner/generate.go
elizabethhealy: thats the intention behind the currentfiledir() func, it gets the dir of this generate.go file so it will work no matter where its called from

Makefile
```diff
@@ -1,7 +1,7 @@
 # make
 # To run all lint checks: `LINT_OPTIONS= make lint`
 
-.PHONY: all build clean docker-build fix fmt go-lint license lint proto-generate proto-lint sdk/sdk test tidy toolcheck
+.PHONY: all build clean docker-build fix fmt go-lint license lint proto-generate connect-wrapper-generate proto-lint sdk/sdk test tidy toolcheck
```
dmihalcik-virtru, line 4: alphabetize

```suggestion
.PHONY: all build clean connect-wrapper-generate docker-build fix fmt go-lint license lint proto-generate proto-lint sdk/sdk test tidy toolcheck
```

examples/cmd/benchmark.go
```diff
@@ -106,7 +106,11 @@ func runBenchmark(cmd *cobra.Command, _ []string) error {
 			return err
 		}
 		nanoTDFConfig.EnableECDSAPolicyBinding()
-		err = nanoTDFConfig.SetKasURL(fmt.Sprintf("http://%s/kas", "localhost:8080"))
+		if insecurePlaintextConn || strings.Contains(platformEndpoint, "http://") {
```
dmihalcik-virtru, line 109:
```suggestion
		if insecurePlaintextConn || strings.HasPrefix(platformEndpoint, "http://") {
```

examples/cmd/benchmark.go
```diff
@@ -106,7 +106,11 @@ func runBenchmark(cmd *cobra.Command, _ []string) error {
...
+			err = nanoTDFConfig.SetKasURL(fmt.Sprintf("http://%s/kas", "localhost:8080"))
```
dmihalcik-virtru, line 110: Wait, shouldn't this be based off `platformEndpoint` somehow?

examples/cmd/benchmark.go
```diff
@@ -123,15 +127,26 @@ func runBenchmark(cmd *cobra.Command, _ []string) error {
...
-		tdf, err := client.CreateTDF(
-			out, in,
-			sdk.WithDataAttributes(dataAttributes...),
-			sdk.WithKasInformation(
+		opts := []sdk.TDFOption{sdk.WithDataAttributes(dataAttributes...), sdk.WithAutoconfigure(false)}
+		if insecurePlaintextConn || strings.Contains(platformEndpoint, "http://") {
+			opts = append(opts, sdk.WithKasInformation(
 				sdk.KASInfo{
-					URL:       "http://" + "localhost:8080",
+					URL:       "http://localhost:8080",
```
dmihalcik-virtru, line 134: See note above about platformEndpoint maybe being back in play. Maybe file a follow-up issue about getting examples working with non-localhost servers

sdk/audit/metadata_adding_interceptor.go
```diff
@@ -46,3 +47,33 @@ func MetadataAddingClientInterceptor(
 
 	return err
 }
+
+func MetadataAddingConnectInterceptor() connect.UnaryInterceptorFunc {
```
dmihalcik-virtru, line 51: When/why is this being used?

sdk/audit/metadata_adding_interceptor_test.go
```diff
@@ -3,86 +3,162 @@ package audit
...
+	_, err := clientConnect.PublicKey(ctx, connect.NewRequest(&kas.PublicKeyRequest{}))
 	if err != nil {
 		t.Fatalf("error making call: %v", err)
 	}
-
-	if server.requestID != contextRequestID {
-		t.Fatalf("request ID did not match: %v", server.requestID)
+	_, err = clientGrpc.PublicKey(ctx, &kas.PublicKeyRequest{})
+	if err != nil {
+		t.Fatalf("error making call: %v", err)
```
dmihalcik-virtru, line 65: feel free to upgrade to testify/assert (or require) to save a little bit of line count

sdk/auth/token_adding_interceptor.go
```diff
@@ -82,6 +83,38 @@ func (i TokenAddingInterceptor) AddCredentials(
...
+			if err != nil {
+				slog.ErrorContext(ctx, "error getting access token", "error", err)
+				return nil, connect.NewError(connect.CodeUnauthenticated, err)
+			}
+
+			// Add Authorization header
+			req.Header().Set("Authorization", fmt.Sprintf("DPoP %s", accessToken))
+
+			// Add DPoP header if possible
+			dpopTok, err := i.GetDPoPToken(req.Spec().Procedure, http.MethodPost, string(accessToken))
```
dmihalcik-virtru, line 102: If Dpop doesn't work, shouldn't line 99 be Bearer? I forget :-/

service/pkg/server/services.go
```diff
@@ -162,7 +164,7 @@ func startServices(ctx context.Context, cfg *config.Config, otdf *server.OpenTDF
 				var err error
 				svcDBClient, err = newServiceDBClient(ctx, cfg.Logger, cfg.DB, tracer, ns, svc.DBMigrations())
 				if err != nil {
-					return err
+					return nil, err
```
dmihalcik-virtru, line 167: maybe always return a no-op instead of nil?

service/pkg/server/services.go
```diff
@@ -195,8 +197,18 @@ func startServices(ctx context.Context, cfg *config.Config, otdf *server.OpenTDF
...
 
 			// Register GRPC Gateway Handler using the in-process connect rpc
-			if err := svc.RegisterGRPCGatewayHandler(ctx, otdf.GRPCGatewayMux, otdf.ConnectRPCInProcess.Conn()); err != nil {
+			grpcConn := otdf.ConnectRPCInProcess.GrpcConn()
+			err := svc.RegisterGRPCGatewayHandler(ctx, otdf.GRPCGatewayMux, otdf.ConnectRPCInProcess.GrpcConn())
+			if err != nil {
 				logger.Info("service did not register a grpc gateway handler", slog.String("namespace", ns))
+			} else if gatewayCleanup == nil {
+				gatewayCleanup = func() {
+					slog.Info("executing cleanup")
```
dmihalcik-virtru, line 206: Consider Debug, but also add more detail about what is getting cleaned up

service/internal/server/server.go
```diff
@@ -458,6 +462,10 @@ func (s OpenTDFServer) Stop() {
 		s.logger.Error("failed to shutdown http server", slog.String("error", err.Error()))
 		return
 	}
+	// Close the listener
+	if s.Listener != nil {
+		s.Listener.Close()
```
dmihalcik-virtru, line 467: Should we clear?

service/authorization/authorization_test_structures.go
```diff
@@ -0,0 +1,334 @@
+package authorization
```
dmihalcik-virtru, line 1: Maybe we can move this to authorization_test? I guess we'd have to move authorization_test too.

sdk/sdkconnect/actions.go
```diff
@@ -0,0 +1,70 @@
+// Wrapper for ActionServiceClient (generated code) DO NOT EDIT
```
dmihalcik-virtru, line 1: Add this directory as `linguist-generated` to `./.gitattributes` so they collapse by default in the PR review page

sdk/auth/token_adding_interceptor.go
elizabethhealy: im just mimicking the behavior of the existing grpc interceptor which seems to add "Authorization DPoP ..." https://github.com/opentdf/platform/blob/5ff754e99189d09ec3698128d1bc51b6f7a90994/sdk/auth/token_adding_interceptor.go#L60

sdk/audit/metadata_adding_interceptor.go
elizabethhealy: not sure, i think it just adds request ids to the metadata for tracking. both this and the credentials interceptors already existed for grpc, this is just making an equivalent one for connectrpc

service/internal/server/server.go
elizabethhealy: what do you mean by clear? the listener interface just has accept, close, and addr https://pkg.go.dev/net#Listener

examples/cmd/benchmark.go
elizabethhealy: ya im not sure why were hardcoding it

examples/cmd/benchmark.go
elizabethhealy: can do, ill add it to tech debt

sdk/auth/token_adding_interceptor.go
dmihalcik-virtru: It sounds like it should be Bearer in this case, as the DPoP spec says a server MUST check for the dpop header when the authorization header creds are in the form DPoP ...

https://datatracker.ietf.org/doc/html/rfc9449#section-7.1

That said, if this isn't a regression I'm fine with you filing a follow up ticket that is about improving our DPoP testing infrastructure. Also this is a pretty bad edge case so IDK what the 'right' thing to do here it - it probably is to fail the request, not to try it with a busted dpop

dmihalcik-virtru (approved):

service/internal/server/server.go
dmihalcik-virtru: Should we set `s.Listener = nil`. I'm guessing the servier object is useless after Stop is called but idk

sdk/auth/token_adding_interceptor.go
elizabethhealy: we must be adding the bearer header elsewhere otherwise id assume all the tests would fail at casbin

service/internal/server/server.go
elizabethhealy: ya i dont think weve tested stopping and then re-starting, but for the listener, its instantiated on start so re-calling start should just create another one and set s.Listener to that

strantalis:
👍

strantalis (approved):

elizabethhealy:
👍

pflynn-virtru (approved):

PR #2206, opened 2025-05-12
feat(sdk): Hide connect rpc client side

elizabethhealy:
### Proposed Changes

* generate connect wrappers for the grpc client interfaces, allows the sdk object to still have grpc client interfaces (meaning if someone is running a seperate ERS for example, it doesnt have to be connectrpc, same for the other services)

### Checklist


.github/workflows/checks.yaml
```diff
@@ -346,6 +346,7 @@ jobs:
       focus-sdk: go
       # use commit instead of ref so we can "go get" specific sdk version
       platform-ref: ${{ github.event.pull_request.head.sha || github.sha }} lts
+      otdfctl-ref: c9f4679d5d3205e4ccb9f6e81fdca973f498c8c5
```
elizabethhealy, line 349: TODO: remove before merging, just for tests

PR #2209, opened 2025-05-13
fix: perfsprint lint issues

strantalis:
### Proposed Changes

*

### Checklist

### DIFF
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -59,10 +59,6 @@ jobs:
             kas
             authz
             examples
-            lib/fixtures
-            lib/flattening
-            lib/identifier
-            lib/ocrypto
       # - name: "Issue Accociated with Pull Request"
       #   id: issue-associated
       #   if: contains(fromJSON('["pull_request", "pull_request_target"]'), github.event_name) && !contains(fromJSON('["dependabot[bot]", "opentdf-automation[bot]"]'), github.actor)
--- a/.github/workflows/pr-checks.yaml
+++ b/.github/workflows/pr-checks.yaml
@@ -59,6 +59,10 @@ jobs:
             kas
             authz
             examples
+            lib/fixtures
+            lib/flattening
+            lib/identifier
+            lib/ocrypto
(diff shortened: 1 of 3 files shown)


elizabethhealy (dismissed):

mkleene (dismissed):

jentfoo (dismissed):

.github/workflows/pr-checks.yaml
```diff
@@ -59,6 +59,10 @@ jobs:
             kas
             authz
             examples
```
dmihalcik-virtru, line 61: This is not necessary. Scopes are there to indicate features or domain areas that are helpful for quickly browsing changelogs. These are unhelpful, as release-please already does breaking-things-down by package for us, and this will just clutter the changelogs with redundant information

image

dmihalcik-virtru: path -> which release
type -> changelog section - most useful for the 'breaking changes' section but also good for finding a bug fix, or when a feature was introduced
scope -> feature or purpose, useful for quickly scanning for what you need

Ideally we would 'roll out' features from internal packages - ie if ocrypto says `feat(core): Adds support for secp521`, the dep bump in service or sdk to use that package would also say `*core*: Adds support for secp521` instead of `*deps*: bumps lib/ocrypto from 1.2.3 to 1.3.0` or whatever we should get once dependabot takes over from autobump. 

Right now we can do this by manually inserting the line as needed in the generated commit (e.g. by adding it to the PR and using the PR description squash-and-merge strat)

https://github.com/googleapis/release-please?tab=readme-ov-file#what-if-my-pr-contains-multiple-fixes-or-features

jrschumacher (approved):

.github/workflows/pr-checks.yaml
```diff
@@ -59,6 +59,10 @@ jobs:
...
+            lib/fixtures
+            lib/flattening
+            lib/identifier
+            lib/ocrypto
```
strantalis, line 65:
```suggestion
```

jrschumacher (approved):

PR #2218, opened 2025-05-15
fix(core): Let legacy KAOs use new trust plugins

dmihalcik-virtru:
### Proposed Changes

- The code path for KAOs that do not contain key identifier strings did not use the new trust.KeyIndex plugin, if one is provided
- Now the code will prefer to use the index, if it is available and the 'keyring' config is not set
- Also fixes an issue where defaulting to 'standard' in the config caused the 'in memory' crypto config (standard) to always be used, as `IsEmpty` would always return true

### Checklist

### DIFF
--- a/service/internal/security/crypto_provider.go
+++ b/service/internal/security/crypto_provider.go
@@ -1,34 +1,8 @@
 package security
 
-import (
-	"crypto"
-	"crypto/elliptic"
-)
-
 const (
 	// Key agreement along P-256
 	AlgorithmECP256R1 = "ec:secp256r1"
 	// Used for encryption with RSA of the KAO
 	AlgorithmRSA2048 = "rsa:2048"
 )
-
-type CryptoProvider interface {
-	// Gets some KID associated with a given algorithm.
-	// Returns empty string if none are found.
-	FindKID(alg string) string
-
-	// Gets all KIDs associated with a given algorithm.
-	ListKeysByAlg(alg string) ([]string, error)
-
-	RSAPublicKey(keyID string) (string, error)
-	RSAPublicKeyAsJSON(keyID string) (string, error)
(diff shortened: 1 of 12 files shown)


service/internal/security/crypto_provider.go
```diff
@@ -16,6 +16,10 @@ type CryptoProvider interface {
 	// Gets some KID associated with a given algorithm.
 	// Returns empty string if none are found.
 	FindKID(alg string) string
+
+	// Gets all KIDs associated with a given algorithm.
+	ListKeysByAlg(alg string) ([]string, error)
```
pflynn-virtru, line 21: ListKIDsByAlg ?

pflynn-virtru: we also have `FindKeyByAlgorithm` .  if possible, change to long form

pflynn-virtru (dismissed):

pflynn-virtru (approved):

service/internal/security/crypto_provider.go
```diff
@@ -1,30 +1,8 @@
...
-	"crypto"
-	"crypto/elliptic"
-)
-
 const (
 	// Key agreement along P-256
 	AlgorithmECP256R1 = "ec:secp256r1"
 	// Used for encryption with RSA of the KAO
 	AlgorithmRSA2048 = "rsa:2048"
 )
```
elizabethhealy, line 8: would it make sense to move these to standard_crypto.go and remove this file?

elizabethhealy (approved):
do we have any unit tests for StandardCrypto?

service/internal/security/crypto_provider.go
dmihalcik-virtru: I'll file a follow up ticket that we should export these from the trust package.

PR #2223, opened 2025-05-16
feat(authz): subject mapping plugin support for ABAC with actions

jakedoublev:
### Proposed Changes

* enhance `subjectmappingplugin` with actions in response to subject mapping evaluation

### Checklist

- [X] I have added or updated unit tests

### DIFF
--- a/service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
+++ b/service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
@@ -75,7 +75,6 @@ func EvaluateSubjectMappingsWithActions(
 					m := make(map[string]*policy.Action, len(actions))
 					for _, action := range actions {
 						m[strings.ToLower(action.GetName())] = action
-
 					}
 					entitlementsSet[valueFQN] = append(entitlementsSet[valueFQN], slices.Collect(maps.Values(m))...)
 				}
--- a/service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
+++ b/service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
@@ -2,6 +2,7 @@ package subjectmappingbuiltin
 
 import (
 	"fmt"
+	"maps"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pflynn-virtru (dismissed):

service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
```diff
@@ -0,0 +1,87 @@
...
+
+type AttributeValueFQNsToActions map[string][]*policy.Action
+
+type EntityIDsToEntitlements map[string]AttributeValueFQNsToActions
+
+func EvaluateSubjectMappingMultipleEntitiesWithActions(
+	attributeMappings map[string]*attributes.GetAttributeValuesByFqnsResponse_AttributeAndValue,
+	entityRepresentations []*entityresolutionV2.EntityRepresentation,
+) (EntityIDsToEntitlements, error) {
+	results := make(map[string]AttributeValueFQNsToActions)
```
jentfoo, line 22:
```suggestion
	results := make(map[string]AttributeValueFQNsToActions, len(entityRepresentations))
```

minor optimization if you're interested

service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
```diff
@@ -0,0 +1,87 @@
...
+					if !subjectSetConditionResult {
+						break
+					}
+				}
+
+				// each subject mapping that is true should permit the actions on the mapped value
+				if subjectMappingResult {
+					// add value FQN to the entitlements set
+					if _, ok := entitlementsSet[valueFQN]; !ok {
+						entitlementsSet[valueFQN] = make([]*policy.Action, 0)
```
jentfoo, line 69:
```suggestion
						entitlementsSet[valueFQN] = make([]*policy.Action, 0, len(actions))
```

This might slightly over-allocate, but would avoid reallocations.

Alternatively this would be more performant (technically more temporary memory allocations, but overall faster and less retained memory) and less code by building a map like the following:
```
m := make(map[string]string, len(actions))
for _, action := range actions {
    m[strings.ToUpper(action.GetName())] = action.GetName()
}
entitlementsSet[valueFQN] = slices.Collect(maps.Values(m))
```

Technically last witnessed value wins in the above, where your implementation prefers only the first EqualFold.  But it seems like a minor detail between the two implementations.

jentfoo (approved):
LGTM, feel free to ignore my minor suggestions

service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
jakedoublev: Correct. Order is not of importance, so this is a good optimization. Pushing up.

jentfoo (approved):

pflynn-virtru (approved):

ryanulit:
👍

PR #2254, opened 2025-05-20
feat(policy): Default KAS Keys

c-r33d:
### Proposed Changes

1.) Add a new table to represent default kas keys
2.) Add SetDefaultKey / GetDefaultKeys rpcs 
3.) Allow for keys of the same algorithm to be created even if another is active
4.) Fix an issue with rotate where service level code could not get key

### Checklist


service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -581,6 +581,51 @@ message RotateKeyResponse {
...
+}
+
+message DefaultKasPublicKey {
+  string algorithm = 1;
+  string kid = 2;
+  string pem = 3;
+}
+
+message DefaultKasKey {
+    string tdf_type = 1; // The type of TDF (e.g., ZTDF, Nano)
```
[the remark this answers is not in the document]
c-r33d: Want it to be of type string, since we are using the same structure in the well known config.

service/pkg/db/marshalHelpers.go
```diff
@@ -122,3 +124,65 @@ func KasKeysProtoJSON(keysJSON []byte) ([]*policy.KasKey, error) {
...
+	}
+}
+
+func UnmarshalDefaultKasKey(keysJSON []byte, key *kasregistry.DefaultKasKey) error {
+	if keysJSON != nil {
+		if err := protojson.Unmarshal(keysJSON, key); err != nil {
+			return err
+		}
+
+		alg, err := strconv.Atoi(key.GetPublicKey().GetAlgorithm())
```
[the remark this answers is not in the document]
c-r33d: The algorithm is pulled from the keys table which is stored as an integer then cast to a string to match the expected structure of DefaultKasKey

service/policy/db/key_access_server_registry.go
```diff
@@ -830,3 +828,188 @@ func (c PolicyDBClient) rotatePublicKeyTables(ctx context.Context, oldKeyID, new
...
+	case policy.Algorithm_ALGORITHM_EC_P256, policy.Algorithm_ALGORITHM_EC_P384, policy.Algorithm_ALGORITHM_EC_P521:
+		return true
+	case policy.Algorithm_ALGORITHM_RSA_2048, policy.Algorithm_ALGORITHM_RSA_4096, policy.Algorithm_ALGORITHM_UNSPECIFIED:
+		return false
+	default:
+		return false
+	}
+}
+
+func (c PolicyDBClient) SetDefaultKey(ctx context.Context, r *kasregistry.SetDefaultKeyRequest) (*kasregistry.SetDefaultKeyResponse, error) {
```
strantalis, line 923: Should we add a check once my enum changes are in to not allow public key only keys to be set as default?

service/policy/db/key_access_server_registry.go
```diff
@@ -830,3 +828,188 @@ func (c PolicyDBClient) rotatePublicKeyTables(ctx context.Context, oldKeyID, new
...
+		return nil, err
+	}
+
+	return &kasregistry.SetDefaultKeyResponse{
+		NewDefaultKasKey:      newDefaultKey,
+		PreviousDefaultKasKey: previousDefaultKey,
+	}, nil
+}
+
+func (c PolicyDBClient) SetWellKnownConfig(ctx context.Context) error {
```
strantalis, line 978: We might want to rename this because I could see other areas maybe in policy we want to update the well-known.

service/policy/db/key_access_server_registry.go
c-r33d: Yeah, I agree.

service/policy/kasregistry/key_access_server_registry.proto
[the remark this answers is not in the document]
dmihalcik-virtru: Generally, these will be for the algorithm, not the TDF, right? In the current REST call, the parameter name is 'algorithm', not 'tdf-type', and it takes one of "rsa:2048" or "ec..." as a parameter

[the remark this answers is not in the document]
c-r33d: We have an algorithm parameter that is returned as part of the PublicKey structure defined in the DefaultKasKey structure. The tdf type is to allow us to be able to enforce that nano needs an ECC key.

service/wellknownconfiguration/wellknown_configuration.go
```diff
@@ -34,6 +34,14 @@ func RegisterConfiguration(namespace string, config any) error {
 	return nil
 }
 
+// We should probably have a safe-guard as to what config can be updated
+func UpdateConfiguration(namespace string, config any) error {
```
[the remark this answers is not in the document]
c-r33d: I think this is a fair comment, w.r.t adding an allowed set of namespaces that can be modified. Although, I am not sure if we want to add that now or not.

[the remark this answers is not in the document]
strantalis: We can add it later if needed imo

jentfoo (dismissed):

service/wellknownconfiguration/wellknown_configuration.go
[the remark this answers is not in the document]
c-r33d: Yes, that would work as well.

service/policy/db/key_access_server_registry.go
```diff
@@ -830,3 +831,192 @@ func (c PolicyDBClient) rotatePublicKeyTables(ctx context.Context, oldKeyID, new
...
+		err = db.UnmarshalDefaultKasKey(key, defaultKey)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return defaultKey, nil
+}
+
+func isAlgValidForNano(alg policy.Algorithm) bool {
```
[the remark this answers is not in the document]
c-r33d: I think it's more of inherent knowledge that nano requires an ECC key, at least I haven't seen anything specifically checking anywhere, since we are generating the keys ourselves previously. Ideally we could have a package that can be consumed by this module to validate all this information.

strantalis (approved):

strantalis (approved):

service/integration/kas_registry_key_test.go
```diff
@@ -718,11 +709,226 @@ func (s *KasRegistryKeySuite) Test_RotateKey_NoAttributeKeyMapping_Success() {
...
+		KeyMode:      policy.KeyMode_KEY_MODE_CONFIG_ROOT_KEY,
+		PublicKeyCtx: &policy.KasPublicKeyCtx{Pem: keyCtx},
+		PrivateKeyCtx: &policy.KasPrivateKeyCtx{
+			KeyId:      validKeyID1,
+			WrappedKey: keyCtx,
+		},
+	}
+
+	// Set default key mapping
+	s.Require().NoError(err)
```
dmihalcik-virtru, line 749:
```suggestion
```

service/integration/kas_registry_key_test.go
```diff
@@ -718,11 +709,226 @@ func (s *KasRegistryKeySuite) Test_RotateKey_NoAttributeKeyMapping_Success() {
...
+	s.NotNil(rotatedInKey)
+
+	// Check that the rotated in key is now the ZTDF default key.
+	defaultKasKeys, err = s.db.PolicyClient.GetDefaultKasKeys(s.ctx)
+	s.Require().NoError(err)
+	s.Len(defaultKasKeys, 1)
+	s.Equal(keyMap[nonRotateKey].GetKey().GetKeyId(), defaultKasKeys[0].GetPublicKey().GetKid())
+
+	// Clean up
+	s.cleanupKeys(
```
dmihalcik-virtru, line 774: should this be in an earlier defer block to also trigger if a require hits? probably by appending to a referenced slice instead of hard coding the values

service/integration/kas_registry_key_test.go
```diff
@@ -732,12 +938,589 @@ func (s *KasRegistryKeySuite) Test_RotateKey_NoAttributeKeyMapping_Success() {
...
+		PublicKeyCtx: &policy.KasPublicKeyCtx{Pem: keyCtx},
+		PrivateKeyCtx: &policy.KasPrivateKeyCtx{
+			KeyId:      validKeyID1,
+			WrappedKey: keyCtx,
+		},
+	}
+
+	// Set default key mapping
+	s.Require().NoError(err)
+	_, err = s.db.PolicyClient.SetDefaultKey(s.ctx, &kasregistry.SetDefaultKeyRequest{
```
dmihalcik-virtru, line 964: Why doesn't this return an error?

service/pkg/db/marshalHelpers.go
```diff
@@ -122,3 +124,65 @@ func KasKeysProtoJSON(keysJSON []byte) ([]*policy.KasKey, error) {
...
+		key.PublicKey.Algorithm, err = formatAlg(policy.Algorithm(alg))
+		if err != nil {
+			return err
+		}
+		// Base64 decode the public key
+		pem, err := base64.StdEncoding.DecodeString(key.GetPublicKey().GetPem())
+		if err != nil {
+			return err
+		}
+		key.PublicKey.Pem = string(pem)
```
dmihalcik-virtru, line 166: Maybe try not to partially update the value on error?
```suggestion
		algEnum, err := formatAlg(policy.Algorithm(alg))
		if err != nil {
			return err
		}
		// Base64 decode the public key
		pem, err := base64.StdEncoding.DecodeString(key.GetPublicKey().GetPem())
		if err != nil {
			return err
		}
		key.PublicKey.Algorithm = algEnum
		key.PublicKey.Pem = string(pem)
```

service/policy/kasregistry/key_access_server_registry.go
```diff
@@ -72,6 +72,11 @@ func NewRegistration(ns string, dbRegister serviceregistry.DBRegister) *servicer
 
 				kasrSvc.logger = logger
 				kasrSvc.dbClient = policydb.NewClient(srp.DBClient, logger, int32(cfg.ListRequestLimitMax), int32(cfg.ListRequestLimitDefault))
+				if err = kasrSvc.dbClient.SetDefaultKeyOnWellKnownConfig(context.TODO()); err != nil {
```
dmihalcik-virtru, line 75: File a follow-up to wire through the caller context

service/policy/kasregistry/key_access_server_registry.go
```diff
@@ -423,3 +435,58 @@ func (s KeyAccessServerRegistry) RotateKey(ctx context.Context, r *connect.Reque
...
+		return nil, connect.NewError(connect.CodeInvalidArgument, nil)
+	}
+
+	auditParams := audit.PolicyEventParams{
+		ActionType: audit.ActionTypeUpdate,
+		ObjectType: audit.ObjectTypeKasRegistryKeys,
+		ObjectID:   objectID,
+	}
+
+	err := s.dbClient.RunInTx(ctx, func(txClient *policydb.PolicyDBClient) error {
```
dmihalcik-virtru, line 460: Why run in a TX if there is no way afaict that this will have anything rollback? I mean, is SetDefaultKey nonacid, or is there a plan to panic on audit somehow? (in which case RunInTX needs to be rewritten with a defer block)

Anyway, I filed a followup about both issues (audit has no visible error mode, RunInTx doesn't handle panics)

service/integration/kas_registry_key_test.go
c-r33d: Good point

service/policy/kasregistry/key_access_server_registry.proto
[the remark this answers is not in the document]
strantalis: @dmihalcik-virtru 
With the concept of a default key we don't want the applications or users to decide which key to use as the default. The platform drives what key the sdk's should use. 
If we don't specify tdf type how would the SDK's know whether to use ecc or rsa for ztdf?

service/policy/kasregistry/key_access_server_registry_keys_test.go
```diff
@@ -1393,3 +1394,74 @@ func Test_RotateKeyAccessServer_Keys(t *testing.T) {
...
+				ActiveKey: &kasregistry.SetDefaultKeyRequest_Id{
+					Id: validUUID,
+				},
+				TdfType: kasregistry.TdfType_TDF_TYPE_ZTDF,
+			},
+			expectError: false,
+		},
+	}
+
+	v := getValidator() // Get the validator instance (assuming this is defined elsewhere)
```
strantalis, line 1452:
```suggestion
	v := getValidator()
```

service/integration/kas_registry_key_test.go
c-r33d: Doesn't fail because there is no key within the database that is of tdf type nano rn. Also, the rotateKey is of type ecc. Later in the test, when the rotate is performed, the test is suppose to fail which it does.

service/policy/kasregistry/key_access_server_registry.go
c-r33d: Yeah, I agree. I was under the impression that we were wanting to start putting all update/creates under TXs. This does not necessarily need to be within a TX.

strantalis (dismissed):

strantalis (approved):

PR #2264, opened 2025-05-20
feat(authz): access pdp v2 with actions

jakedoublev:
### Proposed Changes

* PDP which assumes all entitlement policy has been provided to it
    * unblocks embedding a PDP in services
    * unblocks utilization of the PDP library as a playground
    * decouples _how the entity/policy are retrieved_ as totally distinct from _the entity/policy being decisioned/entitled_ with the `JustInTimePDP`
* Tests
* Out of scope (will be handled separately):
	* debug/audit logging parity with auth v1
	* full unification of return types between access PDP and auth service protos
	* named errors

### Checklist


service/internal/access/v2/errors.go
```diff
@@ -0,0 +1,22 @@
+package access
+
+import "errors"
+
+var (
+	ErrMissingRequiredSDK                          = errors.New("access: missing required SDK")
```
dmihalcik-virtru, line 6: Maybe don't group these error values all together in one place abstracted from their use? If you do want to create something like this, only put ones that are actionable by other packages here, and only export ones that need to be exported.

These mostly seem very arbitrary

service/internal/access/v2/pdp.go
```diff
@@ -0,0 +1,314 @@
...
+	entityresolutionV2 "github.com/opentdf/platform/protocol/go/entityresolution/v2"
+	"github.com/opentdf/platform/protocol/go/policy"
+	attrs "github.com/opentdf/platform/protocol/go/policy/attributes"
+	"github.com/opentdf/platform/service/internal/subjectmappingbuiltin"
+	"github.com/opentdf/platform/service/logger"
+)
+
+// Decision represents the overall access decision for an entity.
+type Decision struct {
+	Access  bool               `json:"access" example:"false"`
```
dmihalcik-virtru, line 20: Why are you adding json tags to these?

service/internal/access/v2/just_in_time_pdp.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	p := &JustInTimePDP{
+		sdk:    sdk,
+		logger: l,
+	}
+
+	allAttributes, err := p.fetchAllDefinitions(ctx)
+	if err != nil {
+		l.ErrorContext(ctx, "failed to fetch all attribute definitions", slog.String("error", err.Error()))
```
dmihalcik-virtru, line 56: Again, either do not log these or log at info or below to avoid duplicate error alerts in datadog/whatever

dmihalcik-virtru (commented):
Am i right that the way this works is by downloading the whole attributes table and doing everything in memory?

service/internal/access/v2/errors.go
jakedoublev: This is a good suggestion. I will break them out rather than grouping together. I think I started the file when we only had `pdp.go` and just appended from there, which doesn't make very friendly DX.

service/internal/access/v2/validators.go
```diff
@@ -0,0 +1,140 @@
...
+func validateGetResourceDecision(
+	accessibleAttributeValues map[string]*attrs.GetAttributeValuesByFqnsResponse_AttributeAndValue,
+	entitlements subjectmappingbuiltin.AttributeValueFQNsToActions,
+	action *policy.Action,
+	resource *authzV2.Resource,
+) error {
+	if entitlements == nil {
+		return fmt.Errorf("entitled FQNs to actions are nil: %w", ErrInvalidEntitledFQNsToActions)
+	}
+	if action.GetName() == "" {
```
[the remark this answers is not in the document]
jakedoublev: This is incorrect. See proto definition of method which is nil-safe:
```golang
func (x *Action) GetName() string {
	if x != nil {
		return x.Name
	}
	return ""
}
```

service/internal/access/v2/evaluate.go
```diff
@@ -0,0 +1,323 @@
...
+
+	case *authz.Resource_AttributeValues_:
+		return evaluateResourceAttributeValues(ctx, logger, resource.GetAttributeValues(), resource.GetEphemeralId(), action, entitlements, accessibleAttributeValues)
+
+	default:
+		return nil, fmt.Errorf("unsupported resource type: %w", ErrInvalidResource)
+	}
+
+	// should never reach here
+	return nil, fmt.Errorf("unable to get resource resource decision: %w", ErrInvalidResource)
```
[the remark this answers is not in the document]
jakedoublev: It's dead code but you must return values to compile.

service/internal/access/v2/just_in_time_pdp.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return nil, false, fmt.Errorf("failed to resolve entity identifier: %w", err)
+	}
+
+	var decisions []*Decision
+	allPermitted := true
+	for _, entityRep := range entityRepresentations {
+		d, err := p.pdp.GetDecision(ctx, entityRep, action, resources)
+		if err != nil {
+			// TODO: is it safe to log the entity representation?
```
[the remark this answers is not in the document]
jakedoublev: I think it is safe according to the proto definition, unless ERS puts something sensitive within that should not be there. We can log the ID instead.

service/internal/access/v2/pdp.go
```diff
@@ -0,0 +1,299 @@
...
+	if err := validateGetDecision(entityRepresentation, action, resources); err != nil {
+		return nil, err
+	}
+
+	// Filter all attributes down to only those that relevant to the entitlement decisioning of these specific resources
+	decisionableAttributes := make(map[string]*attrs.GetAttributeValuesByFqnsResponse_AttributeAndValue)
+
+	for idx, resource := range resources {
+		// Assign indexed ephemeral ID for resource if not already set
+		if resource.GetEphemeralId() == "" {
```
jentfoo, line 159: Can you explain the purpose of this ID, it's not immediately obvious to me.

jentfoo (commented):
A lot of remaining TODO notes, but generally looks good to me.  Just a couple things stood out to me.

service/internal/access/v2/evaluate.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+
+	logger.DebugContext(
+		ctx,
+		"getting decision on one resource",
+		slog.Any("resource", resource.GetResource()),
+	)
+
+	switch resource.GetResource().(type) {
+	case *authz.Resource_RegisteredResourceValueFqn:
```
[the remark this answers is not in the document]
jakedoublev: I decided to define and return an explicit error if encountering registered resources while they are not yet supported

service/internal/access/v2/pdp.go
```diff
@@ -0,0 +1,299 @@
...
+			resource.EphemeralId = "resource-" + strconv.Itoa(idx)
+		}
+
+		switch resource.GetResource().(type) {
+		case *authz.Resource_RegisteredResourceValueFqn:
```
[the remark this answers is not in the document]
jakedoublev: I decided to define and return an explicit error if encountering registered resources while they are not yet supported

jakedoublev:
> A lot of remaining TODO notes, but generally looks good to me. Just a couple things stood out to me.

Thanks for the review. All of the remaining TODOs are now callouts to registered resources or a specific error case which needs to be investigated during the follow-on work around errors.

service/internal/access/v2/pdp.go
jentfoo: Thank you for the explanation.  I only saw them used in logging too.  No concerns given that use case, but I did notice that these can be supplied externally as well.  I was concerned there may be some potential for a user to introduce a conflict here.

But if only used in logging, I don't see a risk.

jentfoo (approved):

dmihalcik-virtru (approved):
I have some performance and style worries but nothing blocking. We really should revisit the database design if it isn't meeting our performance goals, but we should formalize what those are, and hopefully get some CI around them before we make any big changes

service/internal/access/v2/evaluate.go
```diff
@@ -0,0 +1,321 @@
...
+	entitlements subjectmappingbuiltin.AttributeValueFQNsToActions,
+	accessibleAttributeValues map[string]*attrs.GetAttributeValuesByFqnsResponse_AttributeAndValue,
+) (*ResourceDecision, error) {
+	// Group value FQNs by parent definition
+	definitionFqnToValueFqns := make(map[string][]string)
+	definitionsLookup := make(map[string]*policy.Attribute)
+
+	for idx, valueFQN := range resourceAttributeValues.GetFqns() {
+		// lowercase the value FQN to ensure case-insensitive matching
+		valueFQN = strings.ToLower(valueFQN)
```
strantalis, line 75: @jakedoublev Why do we have to do ToLower here versus just making sure we use `strings.EqualFold`?

jakedoublev: We can use `strings.EqualFold` when doing fqn and action name comparisons, but when using them as map keys in  lookups, the case sensitivity breaks when keys for the same attribute/value are of different cases. I am pushing up a change to lower case them at the highest level in the `GetDecision` flow rather than here in `evaluate`.

jakedoublev: If we did not do this, clients (like KAS) would need to lower case their own resource attribute value FQNs, on the way in, and it seems better to do this work at the library/auth service level for consistency.

service/internal/access/v2/helpers.go
```diff
@@ -0,0 +1,184 @@
...
+
+	lower := false
+	for _, value := range definition.GetValues() {
+		if lower {
+			alreadyEntitledActions, exists := entitledActionsPerAttributeValueFqn[value.GetFqn()]
+			if !exists {
+				entitledActionsPerAttributeValueFqn[value.GetFqn()] = entitledActions
+			} else {
+				// Ensure the actions are unique
+				mergedActions := mergeDeduplicatedActions(entitledActions.GetActions(), alreadyEntitledActions.GetActions())
```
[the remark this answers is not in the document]
jakedoublev: [`4d33515` (#2264)](https://github.com/opentdf/platform/pull/2264/commits/4d335155d3c97d166934e0ebe69221931973c13b)

service/internal/access/v2/validators_test.go
```diff
@@ -0,0 +1,427 @@
...
+			wantErr:                   ErrInvalidResource,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validateGetResourceDecision(tt.accessibleAttributeValues, tt.entitlements, tt.action, tt.resource)
+			if tt.wantErr != nil {
+				require.Error(t, err)
+				require.ErrorIs(t, err, tt.wantErr)
```
dmihalcik-virtru, line 421:
```suggestion
				require.ErrorIs(t, err, tt.wantErr)
```

service/internal/access/v2/validators_test.go
```diff
@@ -0,0 +1,427 @@
...
+			wantErr:               ErrInvalidEntityChain,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validateEntityRepresentations(tt.entityRepresentations)
+			if tt.wantErr != nil {
+				require.Error(t, err)
+				require.ErrorIs(t, err, tt.wantErr)
```
dmihalcik-virtru, line 309:
```suggestion
				require.ErrorIs(t, err, tt.wantErr)
```

service/internal/access/v2/validators_test.go
```diff
@@ -0,0 +1,427 @@
...
+			wantErr: ErrInvalidAttributeDefinition,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validateAttribute(tt.attribute)
+			if tt.wantErr != nil {
+				require.Error(t, err)
+				require.ErrorIs(t, err, tt.wantErr)
```
dmihalcik-virtru, line 268:
```suggestion
				require.ErrorIs(t, err, tt.wantErr)
```

service/internal/access/v2/validators.go
```diff
@@ -0,0 +1,140 @@
...
+		return fmt.Errorf("attribute FQN is empty: %w", ErrInvalidAttributeDefinition)
+	}
+	if len(attribute.GetValues()) == 0 {
+		return fmt.Errorf("attribute values are empty: %w", ErrInvalidAttributeDefinition)
+	}
+	for _, value := range attribute.GetValues() {
+		if value == nil {
+			return fmt.Errorf("attribute value is nil: %w", ErrInvalidAttributeDefinition)
+		}
+		if value.GetFqn() == "" {
```
dmihalcik-virtru, line 89: Should we be asserting something like:

```suggestion
		if strings.HasPrefix(value.GetFqn(), attribute.GetFqn()) {
```

dmihalcik-virtru (approved):

PR #2272, opened 2025-05-21
fix: handle empty private and public key ctx structs

strantalis:
### Proposed Changes

*

### Checklist


c-r33d (commented):
I have the following in my PR, which would be good to add:

```go
var privateCtx []byte
	if r.GetPrivateKeyCtx() != nil {
		privateCtx, err = json.Marshal(r.GetPrivateKeyCtx())
		if err != nil {
			return nil, db.ErrMarshalValueFailed
		}
	}
```

c-r33d (approved):

PR #2303, opened 2025-05-23
feat: adds basic config root key manager

strantalis:
### Proposed Changes

*

### Checklist

### DIFF
--- a/service/internal/server/server.go
+++ b/service/internal/server/server.go
@@ -28,7 +28,6 @@ import (
 	"github.com/opentdf/platform/service/logger/audit"
 	ctxAuth "github.com/opentdf/platform/service/pkg/auth"
 	"github.com/opentdf/platform/service/tracing"
-	"github.com/opentdf/platform/service/trust"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 	"google.golang.org/grpc"
@@ -58,8 +57,10 @@ type Config struct {
 	CORS                    CORSConfig                               `mapstructure:"cors" json:"cors"`
 	WellKnownConfigRegister func(namespace string, config any) error `mapstructure:"-" json:"-"`
 	// Port to listen on
-	Port int    `mapstructure:"port" json:"port" default:"8080"`
-	Host string `mapstructure:"host,omitempty" json:"host"`
+	Port           int    `mapstructure:"port" json:"port" default:"8080"`
+	Host           string `mapstructure:"host,omitempty" json:"host"`
+	PublicHostname string `mapstructure:"public_hostname,omitempty" json:"publicHostname"`
+
 	// Enable pprof
 	EnablePprof bool `mapstructure:"enable_pprof" json:"enable_pprof" default:"false"`
 	// Trace is for configuring open telemetry based tracing.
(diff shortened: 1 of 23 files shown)


service/kas/access/provider.go
```diff
@@ -103,10 +99,10 @@ func (kasCfg *KASConfig) UpgradeMapToKeyring(c *security.StandardCrypto) {
 	}
 }
 
-func (p *Provider) initSecurityProviderAdapter() {
+func (p *Provider) InitSecurityProviderAdapter() trust.KeyService {
 	// If the CryptoProvider is set, create a SecurityProviderAdapter
-	if p.CryptoProvider == nil || p.KeyManager != nil && p.KeyIndex != nil {
-		return
+	if p.CryptoProvider == nil {
```
c-r33d, line 104: Won't crypto provider be null when we are using the manager/index?

strantalis: I don' think this is needed anymore because I moved the initialization of the delegator to NewRegistration

service/trust/delegating_key_service.go
```diff
@@ -105,37 +117,77 @@ func (d *DelegatingKeyService) Close() {
 	for _, manager := range d.managers {
 		manager.Close()
 	}
+	d.defaultKeyManager = nil
 }
 
+// _defKM retrieves or initializes the default KeyManager.
+// It uses the `defaultMode` field to determine which manager is the default.
 func (d *DelegatingKeyService) _defKM() (KeyManager, error) {
```
c-r33d, line 125: I kinda wish this was just apart of getManager()

c-r33d (dismissed):
Looks good, maybe a few tests around ciphers of different shapes like: RSA (4096) and EC (384/521).

service/internal/server/server.go
```diff
@@ -58,8 +57,10 @@ type Config struct {
 	CORS                    CORSConfig                               `mapstructure:"cors" json:"cors"`
 	WellKnownConfigRegister func(namespace string, config any) error `mapstructure:"-" json:"-"`
 	// Port to listen on
-	Port int    `mapstructure:"port" json:"port" default:"8080"`
-	Host string `mapstructure:"host,omitempty" json:"host"`
+	Port           int    `mapstructure:"port" json:"port" default:"8080"`
+	Host           string `mapstructure:"host,omitempty" json:"host"`
+	PublicHostname string `mapstructure:"public_hostname,omitempty" json:"publicHostname"`
```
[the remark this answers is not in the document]
strantalis: @dmihalcik-virtru Do you not think other services could use a config like this?

dmihalcik-virtru (dismissed):
one optional request to simplify a viper name

service/internal/server/server.go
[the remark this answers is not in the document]
strantalis: We need a way for a KAS to identify itself when the indexer makes a call to GetKey URI is the only column thats unique + required at the moment. 

I don't have any use cases for other services to use it right now but it would be more difficult to move if we had it under the kas configuration.

dmihalcik-virtru (dismissed):

dmihalcik-virtru (approved):

jrschumacher (approved):

service/trust/delegating_key_service.go
```diff
@@ -24,15 +27,18 @@ type DelegatingKeyService struct {
 
 	defaultKeyManager KeyManager
 
+	l *logger.Logger
+
 	// Mutex to protect access to the manager cache
 	mutex sync.Mutex
```
strantalis, line 33: @gemini-code-assist Is a mutex needed in this file ?

PR #2375, opened 2025-06-04
chore(ci): add db migration check

jrschumacher:
fix breaking down migration

### Proposed Changes

*

### Checklist

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -465,3 +465,41 @@ jobs:
           "$(go env GOROOT)"/bin/go run github.com/google/go-licenses@5348b744d0983d85713295ea08a20cca1654a45e
           check --disallowed_types=forbidden --include_tests
           ./examples
+  migrate:
+    name: goose migrate up and down
+    runs-on: ubuntu-22.04
+    services:
+      postgres:
+        image: postgres:15
+        env:
+          POSTGRES_USER: testuser
+          POSTGRES_PASSWORD: testpass
+          POSTGRES_DB: testdb
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd="pg_isready -U testuser" --health-interval=10s --health-timeout=5s --health-retries=5
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install goose
+        run: go install github.com/pressly/goose/v3/cmd/goose@latest
+      - name: Wait for Postgres
+        run: |
+          for i in {1..30}; do
+            pg_isready -h localhost -U testuser && break
+            sleep 1
+          done
+      - name: Run goose up
+        env:
+          PGUSER: testuser
+          PGPASSWORD: testpass
+        run: |
+          export PATH=$PATH:$(go env GOPATH)/bin
+          goose -dir ./service/policy/db/migrations postgres "host=localhost port=5432 user=testuser password=testpass dbname=testdb sslmode=disable" up
+      - name: Run goose down
+        env:
+          PGUSER: testuser
+          PGPASSWORD: testpass
+        run: |
+          export PATH=$PATH:$(go env GOPATH)/bin
+          goose -dir ./service/policy/db/migrations postgres "host=localhost port=5432 user=testuser password=testpass dbname=testdb sslmode=disable" down --all
--- a/service/policy/db/migrations/20241125220354_keys_table.sql
+++ b/service/policy/db/migrations/20241125220354_keys_table.sql
@@ -312,7 +312,7 @@ DROP TRIGGER IF EXISTS maintain_active_key;
 
 DROP FUNCTION IF EXISTS update_active_key;
 
-DROP FUNCTION IF EXISTS update_was_mapped ();
+DROP FUNCTION IF EXISTS update_was_mapped;
 
 DROP TABLE public_keys;
 


service/policy/db/migrations/20240402000000_preserve_value_order.sql
```diff
@@ -36,13 +36,21 @@ EXECUTE FUNCTION update_definition_delete_values_order();
 -- +goose Down
 
 -- +goose StatementBegin
+DROP TRIGGER IF EXISTS trigger_update_definition_add_values_order ON attribute_values;
+-- +goose StatementEnd
```
jakedoublev, line 40: For my learning, do we need to mark these as different statements within the down migration?

service/policy/db/migrations/20240618000000_add_delete_cascade.sql
```diff
@@ -140,110 +140,122 @@ ON DELETE CASCADE;
 
 -- Do not cascade deletion of Attribute Definitions when their parent Namespace is deleted
 ALTER TABLE attribute_definitions
-DROP CONSTRAINT attribute_definitions_namespace_id_fkey_cascades;
-
+DROP CONSTRAINT IF EXISTS attribute_definitions_namespace_id_fkey_cascades;
+ALTER TABLE attribute_definitions
+DROP CONSTRAINT IF EXISTS attribute_definitions_namespace_id_fkey; -- Add this
```
jakedoublev, line 145: Could you please remove the `-- Add this` comments throughout this file for clarity?

service/policy/db/migrations/20240618000000_add_delete_cascade.sql
```diff
@@ -140,110 +140,122 @@ ON DELETE CASCADE;
...
 ADD CONSTRAINT subject_mappings_attribute_value_id_fkey
 FOREIGN KEY (attribute_value_id)
 REFERENCES attribute_values (id);
 
 -- Do not cascade deletion of FQNs when their parent objects are deleted
 ALTER TABLE attribute_fqns
-DROP CONSTRAINT attribute_fqns_namespace_id_fkey_cascades;
-
+DROP CONSTRAINT IF EXISTS attribute_fqns_namespace_id_fkey_cascades;
+ALTER TABLE attribute_fqns
```
jakedoublev, line 184: I could be confused here, but it seems like in this down migration we'd be adding this constraint rather than dropping it.

service/policy/db/migrations/20240618000000_add_delete_cascade.sql
```diff
@@ -140,110 +140,122 @@ ON DELETE CASCADE;
...
 ADD CONSTRAINT attribute_definitions_namespace_id_fkey
 FOREIGN KEY (namespace_id)
 REFERENCES attribute_namespaces (id);
 
 -- Do not cascade deletion of Attribute Values when their parent Definition is deleted
 ALTER TABLE attribute_values
-DROP CONSTRAINT attribute_values_attribute_definition_id_fkey_cascades;
-
+DROP CONSTRAINT IF EXISTS attribute_values_attribute_definition_id_fkey_cascades;
+ALTER TABLE attribute_values
```
jakedoublev, line 154: @gemini-code-assist should these be add foreign key relation constraints since this is the down migration and the up migration dropped these?

.github/workflows/checks.yaml
```diff
@@ -465,3 +465,107 @@ jobs:
...
+        options: >-
+          --health-cmd="pg_isready -U testuser -d testdb" --health-interval=10s --health-timeout=5s --health-retries=5
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.4.2
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Install goose
+        run: go install github.com/pressly/goose/v3/cmd/goose@latest
```
ryanulit, line 489: This version should be pinned to the version in service/go.mod file, not the latest, as any breaking change would break our workflow

jrschumacher:
```suggestion
        run: go install github.com/pressly/goose/v3/cmd/goose@3.24.3
```

jakedoublev:
I think it could be worth a step back before merging this because running the up/down migrations without data in the database could lead to false positives and greater confidence than is justified. I think this is positive progress to run up/down on the schema alone for SQL & migration validity, but without stored data, the test passing could lead to overconfidence.

.github/workflows/checks.yaml
ryanulit, line 1: While this seems to work well enough, I think we should really be using our `migrate up|down` CLI commands to ensure symmetry between CI and actual production code execution. An added benefit of using the CLI commands is that the workflow would provide the integration test coverage.

jrschumacher: This is probably a good move in the future but let's enhance it in another PR.

.github/workflows/checks.yaml
```diff
@@ -465,3 +465,107 @@ jobs:
           "$(go env GOROOT)"/bin/go run github.com/google/go-licenses@5348b744d0983d85713295ea08a20cca1654a45e
           check --disallowed_types=forbidden --include_tests
           ./examples
+  migrate:
+    name: goose migrate up and down
```
jrschumacher, line 469:
```suggestion
    name: goose migrate up and down (⚠️ WARN: this could be a false positive)
```

jakedoublev:
> I think it could be worth a step back before merging this because running the up/down migrations without data in the database could lead to false positives and greater confidence than is justified. I think this is positive progress to run up/down on the schema alone for SQL & migration validity, but without stored data, the test passing could lead to overconfidence.

Revising the above to state that this is great incremental progress and we will choose not to be overconfident and aware of enhancements as capacity allows.

.github/workflows/checks.yaml
```diff
@@ -465,3 +465,107 @@ jobs:
...
+      postgres:
+        image: postgres:15
+        env:
+          POSTGRES_USER: testuser
+          POSTGRES_PASSWORD: testpass
+          POSTGRES_DB: testdb
+        ports:
+          - "5432:5432"
+        options: >
+          --health-cmd="pg_isready -U testuser -d testdb" --health-interval=10s --health-timeout=5s --health-retries=5
```
jrschumacher, line 481:
```suggestion
        options: --health-cmd="pg_isready -U testuser -d testdb" --health-interval=10s --health-timeout=5s --health-retries=5
```

PR #2390, opened 2025-06-05
chore(core): re-organize rego location within service module

jakedoublev:
### Proposed Changes

* Moves the embedded rego policy consumed by auth service v1 into `/service/authorization` as that is the only directory with logic relying on it

### Checklist


jrschumacher:
Looks great!

PR #2392, opened 2025-06-05
feat(authz): DSPX-895 auth svc registered resource GetDecision support

ryanulit:
### Proposed Changes

* add GetEntitlementsRegisteredResources method to pdp access v2 logic
* add test coverage
* policy ListRegisteredResources API behavior update to provide ActionAttributeValues back on RegisteredResourceValues in list

### Checklist


service/internal/access/v2/just_in_time_pdp.go
```diff
@@ -94,10 +100,17 @@ func (p *JustInTimePDP) GetDecision(
 	case *authzV2.EntityIdentifier_Token:
 		entityRepresentations, err = p.resolveEntitiesFromToken(ctx, entityIdentifier.GetToken(), skipEnvironmentEntities)
 
-	// TODO: implement this case
 	case *authzV2.EntityIdentifier_RegisteredResourceValueFqn:
-		p.logger.DebugContext(ctx, "getting decision - resolving registered resource value FQN")
-		return nil, false, errors.New("registered resources not yet implemented")
+		valueFQN := strings.ToLower(entityIdentifier.GetRegisteredResourceValueFqn())
```
jakedoublev, line 104: Maybe we should clarify which types of values we're using when we use `valueFQN` throughout this library (i.e. `attrValueFQN` or `regResValueFQN`).

jakedoublev: What do you think?

service/internal/access/v2/pdp.go
```diff
@@ -227,6 +245,111 @@ func (p *PolicyDecisionPoint) GetDecision(
...
+	registeredResourceValueFQN string,
+	action *policy.Action,
+	resources []*authz.Resource,
+) (*Decision, error) {
+	l := p.logger.With("registeredResourceValueFQN", registeredResourceValueFQN)
+	l = l.With("action", action.GetName())
+	l.DebugContext(ctx, "getting decision", slog.Int("resourcesCount", len(resources)))
+
+	if err := validateGetDecisionRegisteredResource(registeredResourceValueFQN, action, resources); err != nil {
+		return nil, err
```
jakedoublev, line 259: Do you think we should wrap any additional context around these errors?

service/internal/access/v2/pdp.go
```diff
@@ -227,6 +245,111 @@ func (p *PolicyDecisionPoint) GetDecision(
 	return decision, nil
 }
 
+func (p *PolicyDecisionPoint) GetDecisionRegisteredResource(
```
jakedoublev, line 248: Maybe `GetDecisionRegisteredResourceAsEntity`? 🤷

service/internal/access/v2/pdp.go
```diff
@@ -227,6 +245,111 @@ func (p *PolicyDecisionPoint) GetDecision(
...
+			decisionableAttributes[valueFQN] = attributeAndValue
+			err := populateHigherValuesIfHierarchy(ctx, p.logger, valueFQN, attributeAndValue.GetAttribute(), p.allEntitleableAttributesByValueFQN, decisionableAttributes)
+			if err != nil {
+				return nil, fmt.Errorf("error populating higher hierarchy attribute values: %w", err)
+			}
+		}
+	}
+	l.DebugContext(ctx, "filtered to only entitlements relevant to decisioning", slog.Int("decisionableAttributeValuesCount", len(decisionableAttributes)))
+
+	entitledFQNsToActions := make(map[string][]*policy.Action)
```
jakedoublev, line 300: Should we abstract this function so we can separately test it as the logic is also used in GetEntitlements?

jakedoublev (commented):
Have not reviewed tests yet and would prefer to review again after the diff is simplified by merging #2358, but this looks great so far.

service/internal/access/v2/pdp.go
```diff
@@ -227,6 +245,15 @@ func (p *PolicyDecisionPoint) GetDecision(
...
 }
 
+func (p *PolicyDecisionPoint) GetDecisionRegisteredResource(
+	ctx context.Context,
+	registeredResourceValueFQN string,
+	action *policy.Action,
+	resources []*authz.Resource,
+) (*Decision, error) {
+	return nil, fmt.Errorf("GetDecisionRegisteredResource not yet implemented")
+}
```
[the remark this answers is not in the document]
ryanulit: this is obsolete now

service/internal/access/v2/pdp.go
ryanulit: IMO I don't think it needs a different name since we don't call the original one `GetDecisionEntityRepresentationAsEntity`. Seems self-explanatory to me unless I'm missing something? Alternatively we could rename the `registeredResourceValueFQN` param to `entityRegisteredResourceValueFQN` to make it consistent with the `GetDecision` method.

service/internal/access/v2/pdp.go
ryanulit: That method returns explicit errors, so not sure we need any more context.

service/internal/access/v2/just_in_time_pdp.go
ryanulit: Yea, I think that's a good idea. Will refactor at the end.

service/internal/access/v2/pdp.go
jakedoublev: It looks like the current logic including `populateHigherValuesIfHierarchy` aligns with the current attribute value decisioning implementation. I will need to take a deeper look (probably in a sync with you) to answer firmly, but I expect us to have enough unit test scenarios to know clearly if something is missing here by the time test coverage is complete.

jakedoublev: IIRC, we populate the higher values in a hierarchy attribute to be able to resolve subject mappings, which isn't an issue here when we are not applying subject mappings at all to stored action-attribute-values and instead doing a more direct comparison on the entitlements map we build from them.

service/internal/access/v2/pdp.go
ryanulit: param has been renamed

service/internal/access/v2/pdp.go
alkalescent: resolving as stale after talking to @jakedoublev

service/internal/access/v2/pdp.go
```diff
@@ -260,6 +226,83 @@ func (p *PolicyDecisionPoint) GetDecision(
...
+		entitledFQNsToActions[attrValFQN] = actionsList
+	}
+
+	decision := &Decision{
+		Access:  true,
+		Results: make([]ResourceDecision, len(resources)),
+	}
+
+	for idx, resource := range resources {
+		resourceDecision, err := getResourceDecision(ctx, p.logger, decisionableAttributes, p.allRegisteredResourceValuesByFQN, entitledFQNsToActions, action, resource)
```
jakedoublev, line 285:
```suggestion
		resourceDecision, err := getResourceDecision(ctx, l, decisionableAttributes, p.allRegisteredResourceValuesByFQN, entitledFQNsToActions, action, resource)
```

service/internal/access/v2/validators.go
```diff
@@ -41,6 +42,28 @@ func validateGetDecision(entityRepresentation *entityresolutionV2.EntityRepresen
 	return nil
 }
 
+// validateGetDecisionRegisteredResource validates the input parameters for GetDecisionRegisteredResource:
+//   - registeredResourceValueFQN: must be a valid registered resource value FQN
+//   - action: must not be nil
+//   - resources: must not be nil and must contain at least one resource
+func validateGetDecisionRegisteredResource(registeredResourceValueFQN string, action *policy.Action, resources []*authzV2.Resource) error {
```
jakedoublev, line 49: Can we please add a TODO or follow up issue for a unit test of these behaviors so we can detect regressions if something changes logically?

service/internal/access/v2/pdp_test.go
```diff
@@ -336,6 +380,231 @@ func (s *PDPTestSuite) SetupTest() {
...
+			{
+				Value: "secret",
+				ActionAttributeValues: []*policy.RegisteredResourceValue_ActionAttributeValue{
+					{
+						Action: testActionRead,
+						AttributeValue: &policy.Value{
+							Fqn:   testClassSecretFQN,
+							Value: "secret",
+						},
+					},
```
jakedoublev, line 408: If our logic is correct, I don't think we should need this `read` action set on `secret, confidential, or public` attributes because it should propagate down hierarchically from `topsecret`

service/internal/access/v2/pdp.go
alkalescent: fixed

service/internal/access/v2/helpers.go
```diff
@@ -183,3 +185,81 @@ func mergeDeduplicatedActions(actionsSet map[string]*policy.Action, actionsToMer
...
+		switch resource.GetResource().(type) {
+		case *authz.Resource_RegisteredResourceValueFqn:
+			regResValueFQN := strings.ToLower(resource.GetRegisteredResourceValueFqn())
+			regResValue, found := accessibleRegisteredResourceValues[regResValueFQN]
+			if !found {
+				return nil, fmt.Errorf("resource registered resource value FQN not found in memory [%s]: %w", regResValueFQN, ErrInvalidResource)
+			}
+
+			for _, aav := range regResValue.GetActionAttributeValues() {
+				slog.Info("processing action attribute value", slog.Any("aav", aav))
```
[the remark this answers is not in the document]
alkalescent: moved log to commented out TODO block one line below

service/internal/access/v2/pdp.go
```diff
@@ -227,6 +245,111 @@ func (p *PolicyDecisionPoint) GetDecision(
...
+
+		if !slices.ContainsFunc(actionsList, func(a *policy.Action) bool {
+			return a.GetName() == aavAction.GetName()
+		}) {
+			actionsList = append(actionsList, aavAction)
+		}
+
+		entitledFQNsToActions[attrValFQN] = actionsList
+
+		// todo: does hierarchy (low or high) need to be populated here?
```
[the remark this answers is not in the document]
alkalescent: adding review of this comment to DSPX-1295

jakedoublev:
Looks good. I will try to pull the branch and hit the endpoints within the next couple of hours.

jakedoublev (approved):

strantalis (approved):

alkalescent:
prob needs at least one maintainer approval who didn't touch this PR

jrschumacher (approved):

PR #2409, opened 2025-06-06
fix(policy): properly formatted pem in test fixtures

strantalis:
### Proposed Changes

*

### Checklist


opentdf-kas-mode.yaml
```diff
@@ -13,6 +13,10 @@ logger:
   output: stdout
 services:
   kas:
```
c-r33d, line 15: Was there a config value we added for kasUri?

service/internal/fixtures/policy_fixtures.yaml
```diff
@@ -544,22 +543,21 @@ kas_registry_keys:
...
+      provider_config_id: 4ab6b1c4-fa5c-4036-8921-53f8c267b728
```
c-r33d, line 551: Should this be nil since it's mode local

opentdf-kas-mode.yaml
strantalis: yeah `public_hostname`

dmihalcik-virtru (approved):

PR #2425, opened 2025-06-10
feat(sdk): Enable base key support.

c-r33d:
### Proposed Changes

1.) Add base key support in SDK
2.) If the base key is present we attempt to use it, if not we use the default kases passed in.

>[!IMPORTANT]
>Decision points located [here](https://github.com/opentdf/platform/pull/2445/files)

### Checklist


strantalis (approved):

strantalis (dismissed):

sdk/basekey.go
```diff
@@ -0,0 +1,124 @@
...
+	case policy.Algorithm_ALGORITHM_EC_P521:
+		return string(ocrypto.EC521Key), nil
+	case policy.Algorithm_ALGORITHM_UNSPECIFIED:
+		fallthrough
+	default:
+		return "", fmt.Errorf("unsupported algorithm: %s", alg)
+	}
+}
+
+func getBaseKey(ctx context.Context, s SDK) (*policy.SimpleKasKey, error) {
```
jakedoublev, line 62: @gemini-code-assist should SDK be a pointer? 
```golang
func getBaseKey(ctx context.Context, s *SDK) (*policy.SimpleKasKey, error) {
```

strantalis (approved):

sdk/basekey_test.go
```diff
@@ -0,0 +1,334 @@
...
+	s.sdk.wellknownConfiguration = mockService
+
+	// Call getBaseKey
+	baseKey, err := getBaseKey(s.T().Context(), s.sdk)
+
+	// Validate result
+	s.Require().True(mockService.called)
+	s.Require().Error(err)
+	s.Require().Nil(baseKey)
+	s.Require().ErrorContains(err, errBaseKeyInvalidFormat.Error())
```
jakedoublev, line 333: Can we please update all these error assertions to `ErrorIs(err, errBaseKeyInvalidFormat)` instead of `contains`?

sdk/basekeyerrors.go
```diff
@@ -0,0 +1,11 @@
+package sdk
+
+import "errors"
+
+var (
+	errBaseKeyNotFound        = errors.New("base key not found in well-known configuration")
```
jakedoublev, line 6: Do we want to expose all these errors (upper case them) so SDK consumers can assert on them rather than relying on `.Error()` and string matching?

sdk/granter.go
```diff
@@ -264,12 +265,12 @@ func convertAlgEnum2Simple(a policy.KasPublicKeyAlgEnum) policy.Algorithm {
...
 }
 
-type grantTypeMask int
+type grantType int
 
 const (
-	noKeysFound grantTypeMask = 0
-	grantsFound               = 1 << iota
-	mappedFound
+	noKeysFound grantType = iota // No keys found
```
jakedoublev, line 271: @gemini-code-assist what do these bitshifting changes do to the logical flow here?

jakedoublev (approved):

sdk/granter.go
c-r33d: We chose to change this to an enum from a readability perspective.

sdk/basekeyerrors.go
c-r33d: Yes, that's fine; no concern.

mkleene (approved):

PR #2437, opened 2025-06-10
feat(sdk): allow Connect-Protocol-Version RPC header for cors

eugenioenko:
### Proposed Changes
- Adds the  `Connect-Protocol-Version` header to the CORS AllowedHeaders to enable Connect RPC cors calls


opentdf-dev.yaml
```diff
@@ -130,6 +130,7 @@ server:
       - Content-Type
       - X-CSRF-Token
       - X-Request-ID
+      - Connect-Protocol-Version
```
eugenioenko, line 133: Do we need to add this header to the rest of the config examples?

dmihalcik-virtru: yeah, it's a good idea to keep things consistent

strantalis (changes requested):
@eugenioenko We should update the defaults here as well. 

https://github.com/opentdf/platform/blob/43afd48a5338efecc1790f22a533e6d681af510d/service/internal/server/server.go#L114

PR #2447, opened 2025-06-16
chore: improve external contributor check

jrschumacher:
### Proposed Changes

* Get org teams and check if user is in the team

### Checklist

### DIFF
--- a/.github/workflows/label.yaml
+++ b/.github/workflows/label.yaml
@@ -58,6 +58,35 @@ jobs:
 
             const authorAssociation = pr.author_association;
             const isExternal = !internalAssociations.includes(authorAssociation);
+            
+            // Get all teams for the org (paginated)
+            let teams = [];
+            let page = 1;
+            while (true) {
+              const { data } = await github.rest.teams.list({
+                context.repo.owner,
+                per_page: 100,
+                page
+              });
+              if (data.length === 0) break;
+              teams.push(...data);
+              page++;
+            }
+            
+            // Check if the user is a member of any team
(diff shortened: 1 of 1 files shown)


.github/workflows/label.yaml
```diff
@@ -58,6 +58,35 @@ jobs:
...
+                await github.rest.teams.getMembershipForUserInOrg({ context.repo.owner, team_slug: team.slug, username: pr.user.login });
+                isTeamMember = true;
+                break;
+              } catch (error) {
+                if (error.status !== 404) throw error; // only ignore 404s
+              }
+            }
+            
+            // Now use isTeamMember to decide if they're "internal"
+            const isExternal = !isTeamMember;
```
jakedoublev, line 89: I think `isExternal` is already defined above so this will error out?

.github/workflows/label.yaml
```diff
@@ -58,6 +58,35 @@ jobs:
...
+              if (data.length === 0) break;
+              teams.push(...data);
+              page++;
+            }
+            
+            // Check if the user is a member of any team
+            let isTeamMember = false;
+            for (const team of teams) {
+              try {
+                await github.rest.teams.getMembershipForUserInOrg({ context.repo.owner, team_slug: team.slug, username: pr.user.login });
```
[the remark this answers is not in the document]
jrschumacher: Good call

.github/workflows/label.yaml
jrschumacher: Good catch. Surprised the linter didn't catch it.

PR #2457, opened 2025-06-17
feat(authz): cache entitlement policy within authorization service

jakedoublev:
### Proposed Changes

* Consume new platform-wide cache client within authorization service v2 to cache entitlement policy for a configured interval
* Provide cache config within authorization service (with sensible defaults, validation and logging)
* Refactor to share a `EntitlementPolicyStore` interface that a cache satisfies, or else the SDK connection should be utilized to call out to policy services directly
* Refresh the cache is `.IsReady()` is called, because that should only ever happen when the service is fully started and running

### Checklist

### DIFF
--- a/service/authorization/v2/cache_test.go
+++ b/service/authorization/v2/cache_test.go
@@ -13,13 +13,13 @@ import (
 
 var (
 	mockCacheExpiry = 5 * time.Minute
-	mockCache, _    = cache.TestCacheClient(mockCacheExpiry)
 	l               = logger.CreateTestLogger()
 )
 
 func Test_NewEntitlementPolicyCache(t *testing.T) {
 	ctx := t.Context()
 	refreshInterval := 10 * time.Second
+	mockCache, _ := cache.TestCacheClient(mockCacheExpiry)
 
 	c, err := NewEntitlementPolicyCache(ctx, l, nil, mockCache, refreshInterval)
 	require.NoError(t, err)
@@ -31,6 +31,8 @@ func Test_NewEntitlementPolicyCache(t *testing.T) {
 func Test_EntitlementPolicyCache_RefreshInterval(t *testing.T) {
 	var refreshInterval time.Duration
 	ctx := t.Context()
+	mockCache, _ := cache.TestCacheClient(mockCacheExpiry)
+
 	_, err := NewEntitlementPolicyCache(ctx, l, nil, mockCache, refreshInterval)
 	require.ErrorIs(t, err, ErrCacheDisabled)
 
@@ -46,6 +48,7 @@ func Test_EntitlementPolicyCache_Enabled(t *testing.T) {
 		err             error
 		ctx             = t.Context()
 		refreshInterval = 10 * time.Second
+		mockCache, _    = cache.TestCacheClient(mockCacheExpiry)
 	)
 	assert.False(t, c.IsEnabled())
 	assert.False(t, c.IsReady(ctx))
@@ -60,6 +63,7 @@ func Test_EntitlementPolicyCache_Enabled(t *testing.T) {
 
 func Test_EntitlementPolicyCache_CacheMiss(t *testing.T) {
 	ctx := t.Context()
+	mockCache, _ := cache.TestCacheClient(mockCacheExpiry)
 
 	c, err := NewEntitlementPolicyCache(ctx, l, nil, mockCache, 1*time.Hour)
 	require.NoError(t, err)
@@ -80,9 +84,7 @@ func Test_EntitlementPolicyCache_CacheMiss(t *testing.T) {
 
 func Test_EntitlementPolicyCache_CacheHits(t *testing.T) {
 	ctx := t.Context()
-
-	c, err := NewEntitlementPolicyCache(ctx, l, nil, mockCache, 1*time.Hour)
-	require.NoError(t, err)
+	mockCache, _ := cache.TestCacheClient(mockCacheExpiry)
 
 	attrsList := []*policy.Attribute{{Name: "attr1"}}
 	subjMappingsList := []*policy.SubjectMapping{{Id: "id-123"}}
@@ -90,6 +92,12 @@ func Test_EntitlementPolicyCache_CacheHits(t *testing.T) {
 	_ = mockCache.Set(ctx, attributesCacheKey, attrsList, nil)
 	_ = mockCache.Set(ctx, subjectMappingsCacheKey, subjMappingsList, nil)
 	_ = mockCache.Set(ctx, registeredResourcesCacheKey, resourcesList, nil)
+
+	c, err := NewEntitlementPolicyCache(ctx, l, nil, mockCache, 1*time.Hour)
+	require.NoError(t, err)
+
+	// Allow for some concurrency overhead in cache library to prevent flakiness in tests
+	time.Sleep(10 * time.Millisecond)
 
 	attrs, err := c.ListAllAttributes(ctx)
 	require.NoError(t, err)


service/authorization/v2/authorization.go
```diff
@@ -23,27 +27,67 @@ type Service struct {
...
+			Version:           "v2",
+			ServiceDesc:       &authzV2.AuthorizationService_ServiceDesc,
+			OnServicesStarted: startHook,
+			ConnectRPCFunc:    authzV2Connect.NewAuthorizationServiceHandler,
 			RegisterFunc: func(srp serviceregistry.RegistrationParams) (authzV2Connect.AuthorizationServiceHandler, serviceregistry.HandlerServer) {
 				authZCfg := new(Config)
 
+				if err := defaults.Set(authZCfg); err != nil {
+					panic(fmt.Errorf("failed to set defaults for policy service config: %w", err))
+				}
```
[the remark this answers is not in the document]
jakedoublev: callback does not accept an error as a return value

service/entityresolution/v2/entity_resolution.go
```diff
@@ -32,19 +34,40 @@ func NewRegistration() *serviceregistry.Service[entityresolutionv2connect.Entity
...
-				var inputConfig ERSConfig
+				cfg := new(configV2.ERSConfig)
+				var (
+					err error
+					c   *cache.ResponseCache
+				)
+
+				if err := defaults.Set(cfg); err != nil {
+					panic(fmt.Errorf("failed to set default values for ERS config: %w", err))
+				}
```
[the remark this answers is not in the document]
jakedoublev: callback does not accept an error as a return value

service/authorization/v2/cache.go
```diff
@@ -0,0 +1,258 @@
...
+	subjectMappingsCacheKey     = "subject_mappings_cache_key"
+	registeredResourcesCacheKey = "registered_resources_cache_key"
+)
+
+var (
+	// Cache tags for authorization-related data set in the cache
+	authzCacheTags = []string{"authorization", "policy", "entitlements"}
+
+	// stopTimeout is the maximum time to wait for the periodic refresh goroutine to stop
+	stopTimeout = 5 * time.Second
```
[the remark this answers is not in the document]
jakedoublev: This is only used when gracefully exiting, so adding to config does not make sense.

service/authorization/v2/authorization.go
```diff
@@ -123,15 +182,15 @@ func (as *Service) GetDecision(ctx context.Context, req *connect.Request[authzV2
 
 	decisions, permitted, err := pdp.GetDecision(ctx, entityIdentifier, action, []*authzV2.Resource{resource})
 	if err != nil {
-		as.logger.ErrorContext(ctx, "failed to get decision", slog.Any("error", err), slog.Any("request", request))
+		as.logger.ErrorContext(ctx, "failed to get decision", slog.Any("error", err))
```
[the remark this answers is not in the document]
jakedoublev: request contains token information

service/authorization/v2/authorization.go
```diff
@@ -23,41 +27,97 @@ type Service struct {
...
+	cache *EntitlementPolicyCache
+
+	// client managed by platform, provided by registration params
+	platformCacheClient *cache.Cache
+}
+
+type Config struct {
+	// Interval in seconds to refresh the in-memory policy entitlement cache (attributes and subject mappings)
+	// Default: cache is disabled with refresh interval set to 0.
+	CacheRefreshIntervalSeconds int `mapstructure:"cache_refresh_interval_seconds" default:"0"` // Cache disabled by default
```
jakedoublev, line 39: We should have a more sensible default, and separate out an `enabled/disabled` state so that customers only have to worry about "is caching happening" rather than "what is reasonable caching"

jakedoublev: We should also track work to test our CI pipeline against caching with eventual consistency in mind

service/policy/db/subject_mappings.go
```diff
@@ -439,6 +439,34 @@ func (c PolicyDBClient) ListSubjectMappings(ctx context.Context, r *subjectmappi
 	}, nil
 }
 
+// Loads all subject mappings into memory by making iterative db roundtrip requests of defaultObjectListAllLimit size
```
jakedoublev, line 442: Remove this dead code

service/authorization/v2/cache.go
```diff
@@ -0,0 +1,300 @@
...
+			case <-time.After(stopTimeout):
+				// Timeout as a safety mechanism in case the goroutine is stuck
+				c.logger.WarnContext(context.Background(), "Timed out waiting for refresh goroutine to complete")
+			}
+		}
+	}
+}
+
+// Refresh manually refreshes the cache by reaching out to policy services
+func (c *EntitlementPolicyCache) Refresh(ctx context.Context) error {
```
strantalis, line 165: @jakedoublev  @gemini-code-assist 

Would there ever be a case where `ListAllAttributes` fails but `ListAllSubjectMappings` could still pass?

Another question is could there be issues if these caches are not synchronized at the same time?

service/authorization/v2/cache.go
```diff
@@ -0,0 +1,300 @@
...
+var (
+	// Cache tags for authorization-related data set in the cache
+	authzCacheTags = []string{"authorization", "policy", "entitlements"}
+
+	// stopTimeout is the maximum time to wait for the periodic refresh goroutine to stop
+	stopTimeout = 5 * time.Second
+
+	// valid minimum refresh interval for the cache (too frequently may overload policy services)
+	minRefreshInterval = 15 * time.Second
+	maxRefreshInterval = 1 * time.Hour
```
strantalis, line 31: Why do we need to set a max? Shouldn't this be up to the policy admin if they want to only refresh the policy cache once a day.

service/authorization/v2/cache.go
jakedoublev: This is a good note. I'll update the cache refresh behavior to make sure all policy API requests were successful until setting the cache, instead of setting on each successful policy retrieval.

jakedoublev: [`ffc4ffe` (#2457)](https://github.com/opentdf/platform/pull/2457/commits/ffc4ffebe3983366e2739fef3682b0cb1a6620c0)

service/authorization/v2/cache.go
strantalis: Its probably better to keep it more restrictive to start but I could see a higher max with how infrequent we have seen policy actually changing.

jakedoublev: Agreed. We can definitely continue to monitor and adjust over time.

strantalis (dismissed):

service/authorization/v2/cache.go
```diff
@@ -0,0 +1,310 @@
...
+func (c *EntitlementPolicyCache) ListAllAttributes(ctx context.Context) ([]*policy.Attribute, error) {
+	var (
+		attributes []*policy.Attribute
+		ok         bool
+	)
+
+	cached, err := c.cacheClient.Get(ctx, attributesCacheKey)
+	if err != nil {
+		if errors.Is(err, cache.ErrCacheMiss) {
+			return attributes, nil
```
elizabethhealy, line 226: wouldnt a cache miss indicate a larger error here? -- since the cache is populated on startup the key should always be present. would it be best to return the error instead of nil?

elizabethhealy (approved):

strantalis (approved):

service/authorization/v2/cache.go
jakedoublev: I will consider this in a follow up.

PR #2475, opened 2025-06-24
fix(sdk): Prefer KID and Algorithm selection from key maps

dmihalcik-virtru:
When key mappings (grants now with key ids!) are available,
prefer those values when possible, even if that means having multiple
mechanism types in the same TDF (e.g. RSA keys of different lengths).
This was always the intent, but apparently didn't work before.

I've added a couple of small tests,
and now will add more to the integration tests.

Also, updates `start-up-with-containers` to automatically drop unsupported key types on older kases.
This will simplify xtest loading the newly supported key types.

### Proposed Changes

* the `kao template` type is now the preferred input format for `prepareManifest`
* move migration code (from `splitPlan`) into separate method
* modifies tests to correctly associate keys with key ids

### Checklist

### DIFF
--- a/sdk/granter.go
+++ b/sdk/granter.go
@@ -293,16 +293,16 @@ func convertAlgEnum2Simple(a policy.KasPublicKeyAlgEnum) policy.Algorithm {
 
 // convertStringToAlgorithm converts a string algorithm representation to policy.Algorithm
 func convertStringToAlgorithm(alg string) policy.Algorithm {
-	switch strings.ToLower(alg) {
-	case "ec:secp256r1":
+	switch ocrypto.KeyType(strings.ToLower(alg)) {
+	case ocrypto.EC256Key:
 		return policy.Algorithm_ALGORITHM_EC_P256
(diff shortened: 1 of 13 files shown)


sdk/granter.go
```diff
@@ -265,6 +280,24 @@ func convertAlgEnum2Simple(a policy.KasPublicKeyAlgEnum) policy.Algorithm {
 	}
 }
 
+// convertStringToAlgorithm converts a string algorithm representation to policy.Algorithm
+func convertStringToAlgorithm(alg string) policy.Algorithm {
```
c-r33d, line 284: might be good to refactor the same method out of base_key.go and call into this method from base_key. Eventually this should go within ocrypto, I believe.

c-r33d: I think UNSPECIFIED will eventually lead to an error when trying to encrypt, right? Which seems to be how we are using it.

c-r33d (approved):
This looks good.

ttschampel (approved):

PR #2488, opened 2025-06-26
adr(docs): ADR for key deletion.

c-r33d:
### Proposed Changes

1.) ADR on the deletion of keys in platform

### Checklist


adr/decisions/2025-06-26-delete-keys.md
```diff
@@ -0,0 +1,93 @@
...
+
+* 🟩 **Good**, Enhanced Auditability: Retains a historical record of all keys, even those "deleted," which can be invaluable for forensic analysis and compliance.
+* 🟩 **Good**, Reversible (in theory): While the data remains inaccessible, the underlying key might technically be "undeleted" in an emergency (though this would be a complex and highly risky operation itself).
+* 🟨 **Neutral**, Implementation Complexity: Requires a robust, versioned key-value storage backend, which might necessitate significant architectural changes if not already in place.
+* 🟨 **Neutral**, Misleading "Deletion": While marked as deleted, the key data still exists, which might not align with an administrator's expectation of true "deletion" for sensitive information.
+* 🟨 **Neutral**, Storage Overhead: Retaining deleted key versions increases storage requirements.
+* 🟥 **Bad**, Unclear behavior when a key access server is to be deleted and all keys are marked as *deleted*. Should we allow the keys to be deleted? Should we block the deletion of the key access server?
```
jentfoo, line 58: There can be legal needs for data removal.  I have worked in such a system that used key deletion as a method of legally complying to data removal requirements.  I feel like this unclear behavior could be a risk if a customer adopted with a similar use case.

adr/decisions/2025-06-26-delete-keys.md
```diff
@@ -0,0 +1,93 @@
...
+
+>[!NOTE]
+>A derivative of this option could be that we move the key to an archived table
```
jentfoo, line 61: Here is another alternative:

1. Allow an optional public key to be configured.  It would be expected the customer has provided the private key to another entity that they trust, or at least a separation of concerns internally.
2. If no key is configured delete is a actual removal
3. If key is configured, instead the key is encrypted by the configured public key

This would fully allow the customer to control the behavior.  Requiring an additional action to enable key retention, but making it clear, and their responsibility to maintain the private key for recovery separately.

adr/decisions/2025-06-26-delete-keys.md
```diff
@@ -0,0 +1,93 @@
...
+>unclear table relationship issues.
+
+### Option 4: Add exported column / export RPC method
+
+Add a column to the keys table to denote whether or not the key has been exported. If it has, then we allow the unsafe deletion. If the key has not been exported we return a meaningful error instructing the user to call the export rpc.
+
+* 🟩 **Good**, Easy to implement
+* 🟩 **Good**, Provides a way for the platform to tell if the admin has intentionally backed up the key
+* 🟩 **Good**, Allows admins a safe, auditable way for deleting keys
+* 🟨 **Neutral**, Cannot guarantee that the admin has kept the key after calling the export key method.
```
jentfoo, line 72: Why would they want to retain the key?  Our service should be the most secure way to retain keys you need to retain.

adr/decisions/2025-06-26-delete-keys.md
```diff
@@ -0,0 +1,93 @@
...
+* 🟥 **Bad**, No way of recovering your private key if not backed up before deletion.
+
```
jentfoo, line 74:
```suggestion
* 🟥 **Bad**, Requires additional distribution of the key, spreading the potential for compromise even if removed by the service.
```

I dislike this option because of this notable negative, and not a clear understanding of why they should need to retain a key.

adr/decisions/2025-06-26-delete-keys.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Decision Outcome
```
jentfoo, line 76: I would consider my alternative suggestion in option 3.  It seems to give the best of all options, with minimal configuration needed.  Let me know if there are any questions.

jentfoo (approved):

adr/decisions/2025-06-26-delete-keys.md
jentfoo: With the currently encrypted private key, I assume it can be decrypted, meaning the master private key is also present on the system?

The key difference in my solution is that only a public key is configured, so that the encryption key can't be recovered by the key server.

strantalis (dismissed):

jrschumacher:
Looks good!

jrschumacher (dismissed):

jp-ayyappan (dismissed):
see suggestions

jp-ayyappan (approved):
lgtm

PR #2546, opened 2025-07-11
feat(policy): BREAKING CHANGE - Require kas name to be non-null.

c-r33d:
### Proposed Changes

1.) Make KAS Name column non-null.
2.) If KAS name is null, copy the URI into the name.

>[!IMPORTANT]
>If you have registered two KAS's and one of the KAS's name is the 
>same as another KAS's URI, during migration there will be an error
>if the second KAS has no name.

### Checklist


c-r33d:
Closing decided not to pursue.

PR #2579, opened 2025-07-23
feat(policy): add obligation protos

alkalescent:
### Proposed Changes

*

### Checklist


service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+}
+
+message CreateObligationRequest {
+    // Required
+    string namespace_id = 1;
+    string name = 2;
+    // Optional
+    // Common metadata
+    common.MetadataMutable metadata = 100;
+}
```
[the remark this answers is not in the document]
alkalescent: [Issue](https://virtru.atlassian.net/browse/DSPX-1348) says protovalidate rules are handled in a separate issue.

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+
+message CreateObligationResponse {
+    policy.Obligation obligation = 1;
+}
+
+message UpdateObligationRequest {
+    // Required
+    string id = 1;
```
alkalescent, line 38: Should this be oneof `id` or `fqn`?

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+    common.MetadataMutable metadata = 100;
+    common.MetadataUpdateEnum metadata_update_behavior = 101;
+}
+
+message UpdateObligationResponse {
+    policy.Obligation obligation = 1;
+}
+
+message DeleteObligationRequest {
+    string id = 1;
```
alkalescent, line 50: Should this be oneof `id` or `fqn`?

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+}
+
+message GetObligationValueResponse {
+    policy.ObligationValue value = 1;
+}
+
+message CreateObligationValueRequest {
+    // Required
+    string obligation_id = 1;
```
alkalescent, line 83: Should this be oneof `obligation_id` or `obligation_fqn`?

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+    common.MetadataMutable metadata = 100;
+}
+
+message CreateObligationValueResponse {
+    policy.ObligationValue value = 1;
+}
+
+message UpdateObligationValueRequest {
+    // Required
+    string id = 1;
```
alkalescent, line 96: Should this be oneof `id` or `fqn`?

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+    common.MetadataMutable metadata = 100;
+    common.MetadataUpdateEnum metadata_update_behavior = 101;
+}
+
+message UpdateObligationValueResponse {
+    policy.ObligationValue value = 1;
+}
+
+message DeleteObligationValueRequest {
+    string id = 1;
```
alkalescent, line 108: Should this be oneof `id` or `fqn`?

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+}
+
+message DeleteObligationValueResponse {
+    policy.ObligationValue value = 1;
+}
+
+message ListObligationValuesRequest {
+  // Required
+  string obligation_id = 1;
```
alkalescent, line 117: Should this be oneof `obligation_id` or `obligation_fqn`?

Or is this not right and only filter by optional namespace arg?

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+}
+
+message UpdateObligationValueRequest {
+    // Required
+    string id = 1;
+    
+    // Optional
+    common.MetadataMutable metadata = 100;
+    common.MetadataUpdateEnum metadata_update_behavior = 101;
+}
```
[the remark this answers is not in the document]
jakedoublev: I actually prefer the space between related proto fields 🤷

service/policy/obligations/obligations.proto
jakedoublev: I think `id` is preferable because if you change the name of the obligation, its FQN would be different. 

Also, I think changing the name should be allowed here and safe because the obligation FQNs are not stored in TDF manifests and therefore much more dynamic around access and mutations.

service/policy/obligations/obligations.proto
jakedoublev: I think `id` is preferred unless it's unsafe, and deleting an obligation does not retroactively affect access to an existing TDF (just PEP capabilities & behaviors), so it's safe in my opinion.

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+
+message DeleteObligationResponse {
+    policy.Obligation obligation = 1;
+}
+
+message ListObligationsRequest {
+    // Optional
+    // Namespace ID or name
+    string namespace = 1;
```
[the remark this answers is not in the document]
jakedoublev: You could use the identifier approach with the FQN or allow optional filtering based on the id alone. Either is good, though the non-UUID option could be nice because `https://namespace.com` is already human readable at the root of an obligation FQN.

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+
+message ListObligationsResponse {
+    repeated policy.Obligation obligations = 1;
+
+    policy.PageResponse pagination = 10;
+}
+
+// Values
+message GetObligationValueRequest {
+    string id = 1;
```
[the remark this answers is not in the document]
jakedoublev: I think either `id` or `identifier` taking a oneof id/fqn. I'm good with either.

service/policy/obligations/obligations.proto
jakedoublev: If we only provide one, `id` is preferred, but you could make both options with the `identifier` approach: https://github.com/opentdf/platform/blob/cc169d969f0e3380a2341033bc53a1a0eece781a/service/policy/namespaces/namespaces.proto#L60-L75

service/policy/obligations/obligations.proto
jakedoublev: I think we should make the `value` mutable here, then `id` makes sense for `update` because changing the `value` changes its FQN.

service/policy/obligations/obligations.proto
jakedoublev: same as with definitions/obligations: either `id` only or `identifier`

service/policy/obligations/obligations.proto
jakedoublev: This is interesting because we have a `ListAttributeValues` [request](https://github.com/opentdf/platform/blob/cc169d969f0e3380a2341033bc53a1a0eece781a/service/policy/attributes/attributes.proto#L212) but nobody uses it because you consumer users/services rely on `ListAttributes` and `GetAttributeValue` if they need more context about any of the values. I think we should consider skipping this as a result so consumers can:
- ListObligations(optionalNamespace)
- GetObligation(optionalIdentifier)
- GetObligationValue(id)

A get of an obligation and list of its values are functionally the same but we can probably reduce maintenance burden here.

jakedoublev (commented):
Thanks @alkalescent. Responded to your questions and had a couple notes.

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+}
+
+message GetObligationResponse {
+    policy.Obligation obligation = 1;
+}
+
+message CreateObligationRequest {
+    // Required
+    string namespace_id = 1;
```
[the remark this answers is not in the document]
alkalescent: Figured I'd ask since we allow both uuid and name for `ListAttributesRequest` without the identifier approach.
https://github.com/opentdf/platform/blob/cc169d969f0e3380a2341033bc53a1a0eece781a/service/policy/attributes/attributes.proto#L70-L71

service/policy/obligations/obligations.proto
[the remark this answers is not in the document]
alkalescent: took identifier approach

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+import "policy/selectors.proto";
+import "google/protobuf/struct.proto";
+
+///
+/// Obligation
+///
+
+// Definitions
+message GetObligationRequest {
+    string id = 1;
```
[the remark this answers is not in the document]
alkalescent: updated to identifier

service/policy/obligations/obligations.proto
alkalescent: used identifier

service/policy/objects.proto
```diff
@@ -460,6 +460,35 @@ message RegisteredResourceValue {
...
+
+message ObligationTrigger {
+  string id = 1;
+  ObligationValue obligation_value = 2;
+  Action action = 3;
+  Value attribute_value = 4;
+  common.Metadata metadata = 100;
+}
+
+message ObligationFulfiller {
```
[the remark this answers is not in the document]
alkalescent: commented out fulfillers

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,230 @@
...
+    rpc RemoveObligationTrigger(RemoveObligationTriggerRequest) returns (RemoveObligationTriggerResponse) {}
+
+    /*--------------------------------------*
+     * Fulfiller RPCs
+     *--------------------------------------*/
+
+    rpc AddObligationFulfiller(AddObligationFulfillerRequest) returns (AddObligationFulfillerResponse) {}
+
+    rpc RemoveObligationFulfiller(RemoveObligationFulfillerRequest) returns (RemoveObligationFulfillerResponse) {}   
+}
```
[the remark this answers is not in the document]
alkalescent: changed spacing

protocol/go/CHANGELOG.md
jakedoublev, line 1: Should this have been deleted?

docs/openapi/policy/registeredresources/registered_resources.openapi.yaml
jakedoublev, line 1: should these `openapi` docs have been deleted?

service/go.mod
```diff
@@ -179,3 +179,5 @@ require (
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250603155806-513f23925822 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/opentdf/platform/protocol/go => ../protocol/go
```
jakedoublev, line 183: Can you please remove this?

service/policy/objects.proto
```diff
@@ -460,6 +460,48 @@ message RegisteredResourceValue {
   common.Metadata metadata = 100;
 }
 
+message Obligation {
```
jakedoublev, line 463: I think the Obligation requires in the proto its `repeated ObligationValue values` so you can call `ListObligations` or `GetObligation` and see its children.

service/policy/obligations/obligations.go
```diff
@@ -0,0 +1,158 @@
...
+func OnConfigUpdate(s *Service) serviceregistry.OnConfigUpdateHook {
+	return func(_ context.Context, cfg config.ServiceConfig) error {
+		sharedCfg, err := policyconfig.GetSharedPolicyConfig(cfg)
+		if err != nil {
+			return fmt.Errorf("failed to get shared policy config: %w", err)
+		}
+		s.config = sharedCfg
+		s.dbClient = policydb.NewClient(s.dbClient.Client, s.logger, int32(sharedCfg.ListRequestLimitMax), int32(sharedCfg.ListRequestLimitDefault))
+
+		s.logger.Info("obligations service config reloaded")
```
jakedoublev, line 33: Nit: you could use `s.logger.InfoContext(ctx, "obligations service config reloaded")` to use the `context.Context` argument and propagate it into the logs.

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,260 @@
...
+message GetObligationsByFQNsResponse {
+    map<string, policy.Obligation> fqn_obligation_map = 1;
+}
+
+message CreateObligationRequest {
+    // Required
+    oneof identifier {
+        string namespace_id = 1;
+        string namespace_fqn = 2;
+    }
```
jakedoublev, line 39:
```suggestion
    oneof namespace_identifier {
        string id = 1;
        string fqn = 2;
    }
```

service/policy/obligations/obligations.proto
```diff
@@ -0,0 +1,260 @@
...
+message GetObligationValuesByFQNsResponse {
+    map<string, policy.ObligationValue> fqn_value_map = 1;
+}
+
+message CreateObligationValueRequest {
+    // Required
+    oneof identifier {
+        string obligation_id = 1;
+        string obligation_fqn = 2;
+    }
```
jakedoublev, line 118:
```suggestion
    oneof obligation_identifier {
        string id = 1;
        string fqn = 2;
    }
```

jakedoublev (commented):
Thanks for the updates - I think there are a couple remaining fixes.

protocol/go/CHANGELOG.md
alkalescent: We delete this file every time we run `make proto-generate`. For example, here is a commit in which you deleted it: https://github.com/opentdf/platform/commit/b3093cce2ffd1f1cdaec884967dc96a40caa2903

I've restored it now.

service/go.mod
alkalescent: I've removed it, but the "Friendly Reminders" check will fail.

service/policy/obligations/obligations.proto
alkalescent: updated

service/policy/obligations/obligations.go
alkalescent: updated

service/go.mod
jakedoublev: Yes, `protocol/go` and `service` are separate go modules, so to add the RPCs to the service and to add the obligation service client to the SDK, you'll need to:
1. merge the proto gencode
2. release `protocol/go`
3. bump `protocol/go` in the `sdk` module
4. release the `sdk` module
5. bump `protocol/go` and `sdk` in `service`

Since you're only adding RPCs to `service` that depend on unreleased proto types, I think you can skip the SDK-related bumps but would need a release of `protocol/go` for a tidy go mod in CI. The `go.work` resolves all this for you locally.

jakedoublev (approved):

PR #2596, opened 2025-07-29
feat(core): add multi-strategy ERS to support ldap and sql

jrschumacher:
### Proposed Changes

Add a multi-strategy adapter that will support a variety of backends (LDAP, SQL, etc) but support a chaining approach to satisfy the needs for entity chains.

Also added contract focused integration tests to ensure all ERS adapters behave the same given the same input.

### Checklist

### DIFF
--- a/service/entityresolution/entityresolution.go
+++ b/service/entityresolution/entityresolution.go
@@ -64,16 +64,16 @@ func NewRegistration() *serviceregistry.Service[entityresolutionconnect.EntityRe
 				}
 
 				switch inputConfig.Mode {
-				case KeycloakMode:
-					// Keycloak ERS with cache support
-					kcSVC, kcHandler := keycloak.RegisterKeycloakERS(srp.Config, srp.Logger, ersCache)
-					kcSVC.Tracer = srp.Tracer
-					return EntityResolution{EntityResolutionServiceHandler: kcSVC, Tracer: srp.Tracer}, kcHandler
-				default:
-					// Default to claims ERS (works with any IdP)
+				case ClaimsMode:
+					// Claims ERS (works with any IdP)
 					claimsSVC, claimsHandler := claims.RegisterClaimsERS(srp.Config, srp.Logger)
 					claimsSVC.Tracer = srp.Tracer
 					return EntityResolution{EntityResolutionServiceHandler: claimsSVC}, claimsHandler
+				default:
+					// Default to Keycloak ERS with cache support
+					kcSVC, kcHandler := keycloak.RegisterKeycloakERS(srp.Config, srp.Logger, ersCache)
+					kcSVC.Tracer = srp.Tracer
+					return EntityResolution{EntityResolutionServiceHandler: kcSVC, Tracer: srp.Tracer}, kcHandler
 				}
 			},
 		},
--- a/service/entityresolution/v2/entity_resolution.go
+++ b/service/entityresolution/v2/entity_resolution.go
@@ -70,20 +70,15 @@ func NewRegistration() *serviceregistry.Service[entityresolutionv2connect.Entity
 					claimsSVC, claimsHandler := claims.RegisterClaimsERS(srp.Config, srp.Logger)
 					claimsSVC.Tracer = srp.Tracer
 					return EntityResolution{EntityResolutionServiceHandler: claimsSVC}, claimsHandler
-				case KeycloakMode:
-					// Keycloak ERS with cache support
-					kcSVC, kcHandler := keycloak.RegisterKeycloakERS(srp.Config, srp.Logger, ersCache)
-					kcSVC.Tracer = srp.Tracer
-					return EntityResolution{EntityResolutionServiceHandler: kcSVC, Tracer: srp.Tracer}, kcHandler
 				case MultiStrategyMode:
 					multiSVC, multiHandler := multistrategyv2.RegisterERSV2(srp.Config, srp.Logger)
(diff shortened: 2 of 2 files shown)


service/entityresolution/ldap/ldap_entity_resolution.go
```diff
@@ -0,0 +1,544 @@
...
+			return entityresolution.ResolveEntitiesResponse{}, 
+				connect.NewError(connect.CodeInternal, ErrGetRetrievalFailed)
+		}
+
+		var jsonEntities []*structpb.Struct
+		
+		if len(entries) == 0 {
+			// Handle entity not found - check if we should infer the identity
+			if s.shouldInferEntity(ident) {
+				entityStruct, err := s.entityToStructPb(ident)
```
elizabethhealy, line 252: so its expecting the ldap search to return proto formatted entities? how does that work on the admin side? do they just provide a client filter that does conversion from whatever format they have to the json format of the client entity type?

strantalis:
@jrschumacher I don't think we need to have v1 and v2 implementations of the new ers integrations imo.

adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md
```diff
@@ -0,0 +1,751 @@
...
+          jwt_claims:
+            - claim: "email"
+              operator: "exists"
+            - claim: "iss"
+              operator: "equals"
+              values: ["https://keycloak.corp.com"]
+            - claim: "aud"
+              operator: "contains"
+              values: ["opentdf-platform"]
+
```
jrschumacher, line 228:
```suggestion
              entity_type: subject
```

adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md
```diff
@@ -0,0 +1,751 @@
...
+            claim_name: "reporting_manager"
+          - source_column: "groups"
+            claim_name: "group_memberships"       # Generic group concept
+            transformation: "array"
+          - source_column: "authorized_projects"
+            claim_name: "project_assignments"     # Generic project access
+            transformation: "array"
+
+      # Strategy 3: Backup database failover strategy  
+      - name: "corporate_users_backup"
```
jrschumacher, line 278:
```suggestion
      - name: "corporate_service_accounts"
```

adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md
```diff
@@ -0,0 +1,751 @@
...
+        provider: "backup_db"
+        conditions:
+          jwt_claims:
```
jrschumacher, line 281:
```suggestion
          jwt_claims:
          	- claim: "client_id"
          	  operator: "equals"
          	  values: ["service_account_001"]
```

adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md
```diff
@@ -0,0 +1,751 @@
...
+ - multi-strategy
+ - configuration
+ - jwt-mapping
+driver: '@jrschumacher'
+consulted:
+  - '@elizabethhealy'
+  - '@damorris25'
+  - '@biscoe916'
+---
+# Multi-Strategy Entity Resolution Service Implementation
```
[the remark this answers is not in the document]
jrschumacher: I did consider this, but thought it would be best to handle this in the future. I feel that the complexity is already high and adding this kind of feature would make it more challenging.

service/entityresolution/integration/internal/chain_contract_tests.go
```diff
@@ -0,0 +1,255 @@
...
+				},
+			},
+			{
+				Name:        "CreateMultiEntityChainsFromMultipleTokens",
+				Description: "Should create multiple entity chains with consistent multi-entity behavior",
+				Input: ContractInput{
+					Entities: []*entity.Entity{},
+					Tokens: []*entity.Token{
+						CreateTestToken("chain-token-1", "test-client-1", "alice", "alice@opentdf.test"),
+						CreateTestToken("chain-token-2", "test-client-2", "bob", "bob@opentdf.test"),
```
elizabethhealy, line 56: so the tokens created here would be the same i think except for the ephemeral id, but the info inside would be the same?

service/entityresolution/integration/internal/helpers.go
```diff
@@ -0,0 +1,160 @@
...
+
+// CreateResolveEntitiesRequest creates a v2 ResolveEntitiesRequest for testing
+func CreateResolveEntitiesRequest(entities ...*entity.Entity) *entityresolutionV2.ResolveEntitiesRequest {
+	return &entityresolutionV2.ResolveEntitiesRequest{
+		Entities: entities,
+	}
+}
+
+// CreateTestJWT creates a proper JWT token for token-based testing
+func CreateTestJWT(clientID, username, email string) string {
```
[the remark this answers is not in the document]
jrschumacher: Great catch! I'm looking at this again and going to spend more time reviewing

service/entityresolution/integration/internal/chain_contract_tests.go
jrschumacher: This is fixed

jrschumacher:
Bypassed the CodeQL issue since it was green. It was previously an issue with the tests where we were generating a password but didn't have `test` in the function name so it raised a flag.

PR #2606, opened 2025-07-31
chore(ci): remove otdfctl test to improve dx

strantalis:
### Proposed Changes

*

### Checklist


c-r33d:
> > Can we just make the workflow pass even if this fails **continue_on_error**
> > @c-r33d
> > We could but wouldn't you find that confusing seeing the red failure ?

Seems like we are just using this test to make sure you are not breaking anything, might be helpful to keep it. Since 90% of the time it will pass, right?

alkalescent (approved):

PR #2616, opened 2025-08-07
feat(policy): Modify KAS indexer to support legacy keys.

c-r33d:
### Proposed Changes

1.) Create a new function for **key_index** called ListKeysWith allowing developers to pass in common ListKeyOptions
2.) Modify ListKeys/FindKeyByAlgorithm within **key_indexer** to support listing legacy keys.

### Checklist

### DIFF
--- a/service/kas/key_indexer.go
+++ b/service/kas/key_indexer.go
@@ -145,17 +145,21 @@ func (p *KeyIndexer) FindKeyByID(ctx context.Context, id trust.KeyIdentifier) (t
 	}, nil
 }
 
-func (p *KeyIndexer) ListKeys(ctx context.Context, legacyOnly bool) ([]trust.KeyDetails, error) {
-	var legacy *bool
-	if legacyOnly {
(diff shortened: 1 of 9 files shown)


service/kas/key_indexer.go
```diff
@@ -139,11 +145,17 @@ func (p *KeyIndexer) FindKeyByID(ctx context.Context, id trust.KeyIdentifier) (t
 	}, nil
 }
 
-func (p *KeyIndexer) ListKeys(ctx context.Context) ([]trust.KeyDetails, error) {
+func (p *KeyIndexer) ListKeys(ctx context.Context, legacyOnly bool) ([]trust.KeyDetails, error) {
```
dmihalcik-virtru, line 148: Add a new method that is like ListKeyWith(opts...), instead of changing method

dmihalcik-virtru (changes requested):

dmihalcik-virtru (approved):

PR #2643, opened 2025-08-18
feat(core): add support for mode aware services

jrschumacher:
### Proposed Changes

* Add support for mode aware services when extending

### Checklist


jrschumacher:
@strantalis since we can already do `mode: kas,mycustomservice` is this actually needed?

PR #2645, opened 2025-08-19
chore: Add dependency-review workflow that always targets `main`

jentfoo:
### Proposed Changes

The `base-ref` for dependency review is typically the target branch you're merging into. This results in dependency-review alerting on any changed dependencies which may have CVE's. By using `orign/main` as our `base-ref`, we are now also validating that the target branch has adopted all necessary security updates that were adopted within the `main` branch. This will help ensure `release` branches have adopted all necessary security fixes.

### Checklist


jrschumacher (approved):

pflynn-virtru (approved):
I see it's working

PR #2647, opened 2025-08-20
fix(policy)!: Make legacy flag optional

c-r33d:
### Proposed Changes

1.) Make `legacy` flag optional within `AsymmetricKey`.

Doing this because a field of type `bool` with a value of `false` will be omitted from a JSON response, when using `omitempty.`

>[!IMPORTANT]
>Currently not used within any downstream clients.
>Only located within release v0.7 protocol/go
>Plan on backporting.

### Checklist


c-r33d:
Closing because we decided it was fine to have the field omitted when false.

PR #2657, opened 2025-08-26
fix: fixes race conditions in EntitlementPolicyCache 

imdominicreed:
### Proposed Changes

*

### Checklist

### DIFF
--- a/service/authorization/v2/authorization.go
+++ b/service/authorization/v2/authorization.go
@@ -15,7 +15,6 @@ import (
 	otdf "github.com/opentdf/platform/sdk"
 	"github.com/opentdf/platform/service/internal/access/v2"
 	"github.com/opentdf/platform/service/logger"
-	"github.com/opentdf/platform/service/pkg/cache"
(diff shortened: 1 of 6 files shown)


dmihalcik-virtru (approved):

jrschumacher (changes requested):
@imdominicreed this feels pretty aggressive. By shifting the cache to in memory managed by policy it removes the centralized control which will have long term impact on admins from managing cache behavior across the platform.

It is true that ristretto is eventually consistent so the first few hits might be fetching from the DB until the cache is available, but this is expected. If you're encountering some unexpected behavior I'd prefer to address the defect rather than tossing the baby with the bathwater.

I think its important to consider the eviction case, but I don't think its realistic that the policy would be too large at this point unless the admin uses an extremely small cache size.

One thing we're trying to set ourselves up for is support for other cache solutions such as distributive cache via Redis. The cache manager uses https://github.com/eko/gocache so that it will be possible to give admins the option to choose the cache layer they want.

PR #2662, opened 2025-08-28
fix(core): Return new http client for each httputil call, avoid shared state

elizabethhealy:
### Proposed Changes

* return a new instance of the http client for each httputil call
* avoid shared state between callers

### Checklist

### DIFF
--- a/sdk/httputil/http.go
+++ b/sdk/httputil/http.go
@@ -19,23 +19,21 @@ var preventRedirectCheck = func(_ *http.Request, _ []*http.Request) error {
 	return http.ErrUseLastResponse // Prevent following redirects
 }
 
-var safeHTTPClient = &http.Client{
-	Transport:     http.DefaultTransport,
-	Timeout:       defaultTimeout,
-	CheckRedirect: preventRedirectCheck,
-}
-
 // SafeHTTPClient returns a default http client which has sensible timeouts, won't follow redirects, and enables idle
 // connection pooling.
 func SafeHTTPClient() *http.Client {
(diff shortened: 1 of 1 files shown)


dmihalcik-virtru (approved):

elizabethhealy:
> Can you explain the motivation for this change? Reusing the http.Client is the best way to reuse connections, reducing the latency for ssl handshakes.
> 
> It's also considered go best practices to reuse the http.Client.

@jentfoo im worried about different services/peps calling it to get a client, but one service decides they need a longer timeout so they edit the returned client affecting the rest of the services using it
i assume a specific service/pep would reuse their http client within its setup (hopefully), so it would be on them to ensure theyre reducing that overhead

jentfoo:
@elizabethhealy Modifying an already in use http.Client would be a concerning pattern, one we should avoid.  If custom configurations are needed it's expected that instead a specific client will be constructed with one of the helpers (e.g. `SafeHTTPClientWithTransport`).  Because client reuse is important, and a difficult to communicate burden for external implementors I believe it's easiest if it's done by default. Do you think there are possible API changes we can make while still keeping reuse a default behavior?

If the intention that the implementor handles the de-duplication of the http.Client we would need additional changes to support that.  As is this will cause a performance and memory overhead regression.

jentfoo:
@elizabethhealy Let me know if you want me to expand the godocs or have other ideas on this API.  Client reuse is important for both security and performance, I also care a lot about our API ergonomics to ensure this is easy to understand and adopt.

PR #2699, opened 2025-09-06
feat(policy): add protovalidate for obligation defs + vals

alkalescent:
### Proposed Changes

*

### Checklist


service/policy/db/obligations.go
```diff
@@ -16,16 +16,41 @@ import (
 /// Obligation Definitions
 ///
 
+func breakOblFQN(fqn string) (string, string) {
```
jakedoublev, line 19: Could you please move this to the `lib/identifier` module for reusability?

c-r33d (dismissed):

service/policy/obligations/obligations.proto
```diff
@@ -14,10 +13,14 @@ import "buf/validate/validate.proto";
 
 // Definitions
 message GetObligationRequest {
-    oneof identifier {
-        string id = 1;
-        string fqn = 2;
-    }
+    option (buf.validate.message).oneof = { fields: ["id", "fqn"], required: true };
+    string id = 1 [(buf.validate.field).string.uuid = true];
```
jakedoublev, line 17: Should we be using the common identifiers here? https://github.com/opentdf/platform/blob/75d7590ec062f822045027d4eb0b59a48bdea465/service/common/common.proto#L8

jakedoublev (changes requested):

PR #2704, opened 2025-09-08
feat(policy): Include Triggers in GET/LISTable reqs

c-r33d:
### Proposed Changes

Return Obligation Triggers for the following routes:
- GetObligation
- GetObligationByFQNs
- ListObligations
- GetObligationValue
- GetObligationValuesByFQNs

### Checklist


service/policy/db/queries/obligations.sql
```diff
@@ -51,6 +51,29 @@ LEFT JOIN inserted_values iv ON iv.obligation_definition_id = io.id
...
+    SELECT
+        ot.obligation_value_id,
+        JSON_AGG(
+            JSON_BUILD_OBJECT(
+                'id', ot.id,
+                'action', JSON_BUILD_OBJECT(
+                    'id', a.id,
+                    'name', a.name
+                ),
+                'attribute_value', JSON_BUILD_OBJECT(
```
jakedoublev, line 64: We should add a follow-up to do a migration to index all of these relations for faster reads.

PR #2705, opened 2025-09-09
feat(policy): Allow for additional context to be added to obligation triggers

c-r33d:
### Proposed Changes

1.) Add `client_id` column to `obligation_triggers` table.
2.) Modify **ObligationTrigger** protos to take in an optional **RequestContext** where the `client_Id` is stored within the trigger

### Checklist

### DIFF
--- a/service/policy/db/obligations.go
+++ b/service/policy/db/obligations.go
@@ -596,18 +596,13 @@ func (c PolicyDBClient) CreateObligationTrigger(ctx context.Context, r *obligati
 		return nil, fmt.Errorf("failed to get obligation value: %w", err)
 	}
 
-	clientID := ""
-	if r.GetContext() != nil && r.GetContext().GetPep() != nil {
-		clientID = r.GetContext().GetPep().GetClientId()
-	}
-
 	params := createObligationTriggerParams{
 		ObligationValueID: oblVal.GetId(),
 		ActionName:        r.GetAction().GetName(),
 		ActionID:          r.GetAction().GetId(),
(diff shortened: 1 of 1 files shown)


service/policy/db/migrations/20250909000000_add_registered_peps_to_obligations_triggers.sql
```diff
@@ -0,0 +1,16 @@
+-- +goose Up
+-- +goose StatementBegin
+-- Add client_id column to obligation_triggers table
+ALTER TABLE IF EXISTS obligation_triggers
+ADD COLUMN IF NOT EXISTS client_id TEXT NOT NULL;
```
jakedoublev, line 5: Can we please allow this to be null in both the proto and schema so scoping a trigger to a PEP client ID is optional and not required?

service/policy/db/obligations.go
```diff
@@ -596,13 +596,18 @@ func (c PolicyDBClient) CreateObligationTrigger(ctx context.Context, r *obligati
 		return nil, fmt.Errorf("failed to get obligation value: %w", err)
 	}
 
+	clientID := ""
+	if r.GetContext() != nil && r.GetContext().GetPep() != nil {
```
jakedoublev, line 600: This is fine, but gemini actually lied about the nil panic here. The reason the linter pushes `GetField()` methods on these generated proto structs is because they're nil-safe when chained.

jakedoublev (dismissed):

PR #2706, opened 2025-09-09
feat(authz): add obligation policy decision point

jakedoublev:
WIP, current:

1. obligation pdp
2. wire up to JustInTimePDP in the result of a permit decision from access PDP

Future:

1. wire up to auth service with a decision request that also takes in PEP-reported fulfillable obligations


alkalescent (approved):
lgtm

PR #2730, opened 2025-09-16
fix(core): Improves errs with misconfigured keys

dmihalcik-virtru:
### Proposed Changes

* Fixes two existing panics for miscofigured keys:
  * If the private key for a basic wrapped KAS key is empty, error out
  * Error out if key is labeled non-exportable
* For in-process keys, support extended keytypes (larger RSA keys, different curves)

### Checklist


dmihalcik-virtru:
Pulled out ocrypto stuff to #1635

PR #2742, opened 2025-09-18
feat(policy): Return obligations from GetAttributeValue calls

c-r33d:
### Proposed Changes

1.) Update Attribute `Value` structure in  `objects` proto, making **Obligations** part of the object.
2.) Return an array of `Obligations` from the `GetAttributeValue` endpoint

### Checklist

### DIFF
--- a/service/policy/db/attribute_values.go
+++ b/service/policy/db/attribute_values.go
@@ -101,13 +101,10 @@ func (c PolicyDBClient) GetAttributeValue(ctx context.Context, identifier any) (
 		}
 	}
 
-	obligations := make([]*policy.Obligation, 0)
-	if av.Obligations != nil {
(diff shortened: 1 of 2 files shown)


service/policy/db/attribute_values.go
```diff
@@ -101,6 +101,15 @@ func (c PolicyDBClient) GetAttributeValue(ctx context.Context, identifier any) (
 		}
 	}
 
+	obligations := make([]*policy.Obligation, 0)
+	if av.Obligations != nil {
```
jakedoublev, line 105: if your `unmarshalObligations` helper already handles nil bytes passed in, could this be simplified at all?

jakedoublev (approved):

PR #2754, opened 2025-09-19
feat(sdk): sdk should optionally take in a logger

jakedoublev:
If we want to move forward with this, I'll need to split the SDK and service changes out (stacked PRs) to avoid breaking `service` with unreleased `sdk` api changes.

### DIFF
--- a/sdk/idp_cert_exchange.go
+++ b/sdk/idp_cert_exchange.go
@@ -2,6 +2,7 @@ package sdk
 
 import (
 	"context"
+	"log/slog"
 	"net/http"
 	"sync"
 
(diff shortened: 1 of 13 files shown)


sdk/idp_cert_exchange.go
```diff
@@ -21,13 +23,14 @@ type CertExchangeTokenSource struct {
 	key         jwk.Key
 }
 
-func NewCertExchangeTokenSource(info oauth.CertExchangeInfo, credentials oauth.ClientCredentials, idpTokenEndpoint string, dpop *ocrypto.RsaKeyPair) (auth.AccessTokenSource, error) {
+func NewCertExchangeTokenSource(logger *slog.Logger, info oauth.CertExchangeInfo, credentials oauth.ClientCredentials, idpTokenEndpoint string, dpop *ocrypto.RsaKeyPair) (auth.AccessTokenSource, error) {
```
dmihalcik-virtru, line 26: This one is probably safe as no one (at least according to github search) is calling this yet

dmihalcik-virtru (approved):

PR #2759, opened 2025-09-23
feat(sdk): Add obligations support.

c-r33d:
### Proposed Changes

1.) Expose `Obligations` function to TDF/NanoTDF
2.) Expose new `NanoTDFReader` with `LoadNanoTDF`,`Init` functions
3.) Refactor `bulk` to allow consumers to retrieve Obligations before decrypting content
4.) Expose `PrepareBulkDecrypt` to retrieve Obligations before decrypting the content
5.) Expose `ErrRewrapForbidden` for consumption by downstream consumers of the SDK.

Manual E2E tests:
- Ran benchmark and rt tests for bulk
- Nano/ZTDF no access to resource regardless of obligations (returns empty slice of FQNs)
- Nano/ZTDF has access to resource, but does not fulfill obligations (returns required obligations)
- Nano/ZTDF has access to resource and fulfills obligations -> Decrypt successfully.
- Multi-KAS decrypt successful when able to fulfill obligations, failure with returned obligations when not

**Caveats**
>[!NOTE]
>Bulk decrypt is designed in a way that will only allow
>a global set of fulfillable obligations to be applied, not for
>an individual TDF.

### Checklist

### DIFF
--- a/sdk/nanotdf.go
+++ b/sdk/nanotdf.go
@@ -19,7 +19,6 @@ import (
 	"github.com/opentdf/platform/protocol/go/kas"
 	"github.com/opentdf/platform/protocol/go/policy"
 	"github.com/opentdf/platform/sdk/auth"
-	"github.com/opentdf/platform/sdk/sdkconnect"
 
 	"github.com/opentdf/platform/lib/ocrypto"
 )
@@ -882,7 +881,6 @@ type NanoTDFReader struct {
 	httpClient      *http.Client
 	connectOptions  []connect.ClientOption
 	collectionStore *collectionStore
-	authV2Client    sdkconnect.AuthorizationServiceClientV2
 
 	header     NanoTDFHeader
 	headerBuf  []byte
@@ -948,7 +946,6 @@ func (s SDK) LoadNanoTDF(ctx context.Context, reader io.ReadSeeker, opts ...Nano
 		collectionStore: s.collectionStore,
 		header:          header,
 		headerBuf:       headerBuf,
-		authV2Client:    s.AuthorizationV2,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


sdk/bulk.go
```diff
@@ -225,9 +308,36 @@ func (s SDK) BulkDecrypt(ctx context.Context, opts ...BulkDecryptOption) error {
...
+}
+
 func (b *BulkDecryptRequest) appendTDFs(tdfs ...*BulkTDF) {
 	b.TDFs = append(
 		b.TDFs,
 		tdfs...,
 	)
 }
+
+func getFulfillableObligations(decryptor decryptor) []string {
```
jakedoublev: Can we please use the logger on the SDK struct (passing it around if needed) rather than calling `slog` directly?

sdk/kas_client.go
```diff
@@ -272,12 +322,52 @@ func (k *KASClient) processECResponse(response *kas.RewrapResponse, aesGcm ocryp
...
+	if obligationFields == nil {
+		return triggeredFQNs
+	}
+
+	policyOblsValue, ok := obligationFields[policyID]
+	if !ok {
+		return triggeredFQNs
+	}
+
+	listValue := policyOblsValue.GetListValue()
```
jakedoublev: Same here: `policyOblsValue.GetListValue().GetValues()`

sdk/nanotdf.go
```diff
@@ -886,107 +906,87 @@ func createNanoTDFDecryptHandler(reader io.ReadSeeker, writer io.Writer, opts ..
...
-	payloadLength := binary.BigEndian.Uint32(payloadLengthBuf)
-	getLogger().DebugContext(ctx, "decrypt", slog.Uint64("payload_length", uint64(payloadLength)))
+// What do we need:
+// 1. Get obligations before decrypting a file
+// 2. Store obligations triggered during decrypting a file
 
-	cipherData := make([]byte, payloadLength)
-	_, err = n.reader.Read(cipherData)
+// No current way to return obligations for NanoTDF, since there is no reader.
+func (s SDK) LoadNanoTDF(ctx context.Context, reader io.ReadSeeker, opts ...NanoTDFReaderOption) (*NanoTDFReader, error) {
```
jakedoublev: If we're using this function other places to enforce the kas allowlist, can we add a comment please?

jakedoublev: Makes sense to me. Could you please open an issue for this to track that work if it's a follow-up to this PR?

c-r33d: Refactored in [this commit](https://github.com/opentdf/platform/pull/2759/commits/05a41dd0c6127aaf7c1939398e240f6c9304231f)

pflynn-virtru (approved):
The NanoTDF changes are more extensive than I would have thought.  Let's make sure nano has good regression test coverage

sdk/options.go
```diff
@@ -44,6 +44,7 @@ type config struct {
 	entityResolutionConn               *ConnectRPCConnection
 	collectionStore                    *collectionStore
 	shouldValidatePlatformConnectivity bool
+	fulfillableObligationFQNs          []string
```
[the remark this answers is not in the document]
jakedoublev: This comment is a good idea ^^ We allow configurability of total max size of bytes sent as headers serverside: https://github.com/opentdf/platform/blob/8a56a965b934b85ba1a7530a249184e96967cde1/service/internal/server/server.go#L118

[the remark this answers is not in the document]
c-r33d: This is something @jakedoublev and I discussed with @biscoe916. As of now it seems ok to do no validation on the size, since we expose the ability to bump the [MaxHeaderSize](https://github.com/opentdf/platform/blob/main/service/internal/server/server.go#L118).

sujankota (approved):

PR #2771, opened 2025-09-30
feat(policy): namespace root certificates

pflynn-virtru:
This pull request introduces new support for namespace root certificates within the policy service, updates the API to expose these certificates, and makes related improvements to authentication and routing. The most significant changes include extending the `Namespace` object to include root certificates, updating the API and HTTP routing to allow public access to namespace information, and adding a proof-of-concept implementation for injecting a root CA certificate.

`http://localhost:8080/policy.namespaces.NamespaceService/GetNamespace`

```json
{
    "namespace": {
        "id": "8f1d8839-2851-4bf4-8bf4-5243dbfe517d",
        "name": "example.com",
        "fqn": "https://example.com",
        "active": true,
        "metadata": {
            "createdAt": "2025-10-10T17:01:12.477117Z",
            "updatedAt": "2025-10-10T17:01:12.477117Z"
        },
        "rootCerts": [
            {
                "id": "46f1aa69-a458-45fa-9ff3-bc6e2f349773",
                "pem": "-----BEGIN CERTIFICATE-----\nMIIC/TCCAeWgAwIBAgIUSMNrGBq+5ax1wDaMSX8Nnl1m/REwDQYJKoZIhvcNAQEL\nBQAwDjEMMAoGA1UEAwwDa2FzMB4XDTI0MDUyOTE4MzcxOFoXDTI1MDUyOTE4Mzcx\nOFowDjEMMAoGA1UEAwwDa2FzMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKC\nAQEAzbLGCYovxrCy6DccFAdj1DjnDKwmTqjYQoHfOKXbYOF1Yf5mvAyFO9MhtLet\nPRhmCd7mJA3M8Sy4/kqqAZprYHmvagAIwD0zQyQ+DcogrzmirQ8cUYIDkGre1exc\niVKsNYSlM/NcjNO5AqOaw7gQvmUkp2YciSMa7mixPsl3OTArtOtEGMn34RK1NGC+\n+MdUH6TOF2VmhqJ6+F8iSF154ldRoiR01LIzIxvqpHr0RYhU93BrMDcb4Yx92j2M\nlCpwJ8hVf2RPIqhjsnivzXu+BBQh7zynB3H6LnZUGlyWEHgEfM9yhTFWNeKDClKC\nFE2s1DD8ksvMa8ayndIy0+cNXQIDAQABo1MwUTAdBgNVHQ4EFgQUrrmJ35Nvvkpp\nGnQi686Y/DKmsv0wHwYDVR0jBBgwFoAUrrmJ35NvvkppGnQi686Y/DKmsv0wDwYD\nVR0TAQH/BAUwAwEB/zANBgkqhkiG9w0BAQsFAAOCAQEAAXfFJqJYVD+7GqZAWg3v\nw1Sr9jKvwMTtcAhmEJ8KKu9u4C/eGz+P4vU+KwismRfFwQaZVr0AuxRhCtCt3dqh\nKPHZ6sVUw8NG/Fg8v7cQmSpNEuy31vxTMrAWE0/zgScZmIQgS5TnqavsEaxiQGU1\nKTEGOuYQRQZ44KBQJJwUHpwb6znHOwVTKex4OsenkxMLOgsPYa3Qf53kHdZIq8x6\nkxm+METK7jT7RMDNqFWPo4R7XfpvCBR/HS3xjmXe0VVonSc2I4QKdrFPpmx2oOsS\n8Mr8dklyCivjNJ4dM1oKU/YK0hm0bF4OuhoROKpDDBDimDPnFFh/tphl1rzAuuxM\n/A==\n-----END CERTIFICATE-----\n"
            }
        ]
    }
}
```

**Namespace Root Certificate Support:**

* Added a `root_certs` repeated field of type `Certificate` to the `Namespace` message in `policy/objects.proto`, and defined a new `Certificate` message to hold certificate data in PEM format.

### DIFF
--- a/service/policy/db/namespaces.go
+++ b/service/policy/db/namespaces.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"log/slog"
 	"strings"
+	"time"
 
 	"github.com/jackc/pgx/v5/pgtype"
 	"github.com/opentdf/platform/protocol/go/common"
@@ -438,6 +439,20 @@ func validateRootCertificate(pemStr string) error {
 		return errors.Join(db.ErrInvalidCertificate, errors.New("invalid certificate: must be a root certificate (self-signed)"))
 	}
 
+	// Verify the self-signed certificate signature
+	if err := cert.CheckSignatureFrom(cert); err != nil {
+		return errors.Join(db.ErrInvalidCertificate, fmt.Errorf("invalid certificate: signature verification failed: %w", err))
+	}
+
+	// Validate temporal properties (NotBefore and NotAfter)
+	now := time.Now()
+	if now.Before(cert.NotBefore) {
+		return errors.Join(db.ErrInvalidCertificate, fmt.Errorf("invalid certificate: not yet valid (NotBefore: %v, current time: %v)", cert.NotBefore, now))
+	}
+	if now.After(cert.NotAfter) {
+		return errors.Join(db.ErrInvalidCertificate, fmt.Errorf("invalid certificate: expired (NotAfter: %v, current time: %v)", cert.NotAfter, now))
+	}
+
 	return nil
 }
 
@@ -540,20 +555,21 @@ func (c PolicyDBClient) AssignCertificateToNamespace(ctx context.Context, namesp
 
 // CreateAndAssignCertificateToNamespace creates a certificate and assigns it to a namespace in a transaction
 func (c PolicyDBClient) CreateAndAssignCertificateToNamespace(ctx context.Context, namespaceID *common.IdFqnIdentifier, pem string, metadata []byte) (string, error) {
-	// Check if certificate with same PEM already exists
-	existingCert, err := c.queries.getCertificateByPEM(ctx, pem)
-	if err == nil {
-		// Certificate exists, just assign it to namespace
-		err = c.AssignCertificateToNamespace(ctx, namespaceID, existingCert.ID)
-		if err != nil {
-			return "", err
-		}
-		return existingCert.ID, nil
-	}
-
 	var certID string
-	err = c.RunInTx(ctx, func(txClient *PolicyDBClient) error {
-		var err error
+	err := c.RunInTx(ctx, func(txClient *PolicyDBClient) error {
+		// Check if certificate with same PEM already exists (inside transaction to avoid race condition)
+		existingCert, err := txClient.queries.getCertificateByPEM(ctx, pem)
+		if err == nil {
+			// Certificate exists, just assign it to namespace
+			certID = existingCert.ID
+			err = txClient.AssignCertificateToNamespace(ctx, namespaceID, existingCert.ID)
+			if err != nil {
+				return err
+			}
+			return nil
+		}
+
+		// Certificate doesn't exist, create it
 		cert, err := txClient.CreateCertificate(ctx, pem, metadata)
 		if err != nil {
 			return err
--- a/service/policy/db/namespaces_test.go
+++ b/service/policy/db/namespaces_test.go
@@ -1,0 +1,252 @@
+package db
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Helper function to generate a valid self-signed root certificate
+func generateValidRootCert(t *testing.T, notBefore, notAfter time.Time) string {
+	t.Helper()
+
+	// Generate a new ECDSA private key
+	privateKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.NoError(t, err)
+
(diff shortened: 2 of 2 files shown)


service/policy/namespaces/namespaces.go
```diff
@@ -289,3 +291,92 @@ func (ns NamespacesService) RemovePublicKeyFromNamespace(ctx context.Context, r
...
+	if err != nil {
+		ns.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
+		// Clean up the certificate we just created
+		if cleanupErr := ns.dbClient.DeleteCertificate(ctx, certID); cleanupErr != nil {
+			ns.logger.ErrorContext(ctx, "failed to cleanup orphaned certificate after assignment failure",
+				slog.String("cert_id", certID),
+				slog.String("error", cleanupErr.Error()))
+		}
+		return nil, db.StatusifyError(ctx, ns.logger, err, "Failed to assign certificate to namespace")
+	}
```
c-r33d, line 346: Do this in a transaction so if the Assign fails the cert creation will be rolled back.

Ex:
- https://github.com/opentdf/platform/blob/main/service/policy/keymanagement/key_management.go#L123-L144

c-r33d (commented):
Looks good, just a couple small comments. 

Are you wanting to be able to retrieve your created certs? In addition, you will probably want an easy command to list your namespace <-> cert pairs. Key management uses a function called `ListKeyMappings`, which will return all mappings for namespace/definition/value <-> kas asymmetric keys.

dmihalcik-virtru (dismissed):

service/policy/db/namespaces.go
```diff
@@ -386,3 +398,212 @@ func (c PolicyDBClient) RemovePublicKeyFromNamespace(ctx context.Context, k *nam
 		KeyId:       k.GetKeyId(),
 	}, nil
 }
+
+// validateRootCertificate validates that the PEM string is a valid PEM-encoded root certificate
+func validateRootCertificate(pemStr string) error {
```
jentfoo: I don't see anywhere we are validating the certificate signature to be valid.  We need to add a call to `cert.CheckSignatureFrom(cert)` to validate the self signed signature and validate the cert has not been changed.

service/policy/db/namespaces.go
```diff
@@ -386,3 +398,212 @@ func (c PolicyDBClient) RemovePublicKeyFromNamespace(ctx context.Context, k *nam
...
+		return errors.Join(db.ErrInvalidCertificate, fmt.Errorf("invalid PEM type: expected CERTIFICATE, got %s", block.Type))
+	}
+
+	// Parse the certificate
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return errors.Join(db.ErrInvalidCertificate, fmt.Errorf("invalid certificate: not a valid X.509 certificate: %w", err))
+	}
+
+	// Verify it's a root certificate (self-signed)
```
jentfoo: I don't see anywhere we are validating the certificate signature to be valid.  We need to add a call to `cert.CheckSignatureFrom(cert)` to validate the self signed signature and validate the cert has not been changed.

service/policy/db/namespaces.go
```diff
@@ -386,3 +398,212 @@ func (c PolicyDBClient) RemovePublicKeyFromNamespace(ctx context.Context, k *nam
...
+	if err != nil {
+		return db.WrapIfKnownInvalidQueryErr(err)
+	}
+	return nil
+}
+
+// CreateAndAssignCertificateToNamespace creates a certificate and assigns it to a namespace in a transaction
+func (c PolicyDBClient) CreateAndAssignCertificateToNamespace(ctx context.Context, namespaceID *common.IdFqnIdentifier, pem string, metadata []byte) (string, error) {
+	// Check if certificate with same PEM already exists
+	existingCert, err := c.queries.getCertificateByPEM(ctx, pem)
```
jentfoo: There is a check then act race condition here.  The read happening out of the transaction means that multiple concurrent clients could see no certificate and fall to the transaction below, resulting in multiple certificates being generated.

service/policy/db/namespaces.go
jentfoo: I don't see any temporal checks on the certificates here.  Even if we don't currently expire the certs or set Before times, adding the checks proactively could help protect a future change where certificate rotation is introduced.

jentfoo (changes requested):
I really like the idea of a namespace cert, and don't see any significant issues to this design. Marking as requires changes because of a couple small bugs noticed.

jrschumacher (dismissed):
After consultation we've decided to move forward and figure out any additional enhancements in the future.

service/policy/db/namespaces.go
pflynn-virtru: fixed

service/policy/db/namespaces.go
pflynn-virtru: Signature verification via cert.CheckSignatureFrom(cert)

service/policy/db/namespaces.go
pflynn-virtru: Temporal validation checking both NotBefore and NotAfter timestamps

PR #2786, opened 2025-10-06
feat(policy): Add obligation support to KAS

c-r33d:
### Proposed Changes

1.) Add logic to retrieve obligations from request header during rewrap
2.) Add logic to populate required obligations on kas response object

Smoke tests tried manually (no key split or bulk rewrap):
- Obligations are not fulfilled, rewrap fails
- Obligations are fulfilled, rewrap succeeds

### Checklist

### DIFF
--- a/service/kas/access/accessPdp.go
+++ b/service/kas/access/accessPdp.go
@@ -29,7 +29,7 @@ type PDPAccessResult struct {
 	RequiredObligations []string
 }
 
-func (p *Provider) canAccess(ctx context.Context, token *entity.Token, policies []*Policy, obligations []string) ([]PDPAccessResult, error) {
+func (p *Provider) canAccess(ctx context.Context, token *entity.Token, policies []*Policy, fulfillableObligationFQNs []string) ([]PDPAccessResult, error) {
 	var res []PDPAccessResult
 	var resources []*authzV2.Resource
 	idPolicyMap := make(map[string]*Policy)
@@ -68,7 +68,7 @@ func (p *Provider) canAccess(ctx context.Context, token *entity.Token, policies 
 	ctx, span := p.Start(ctx, "checkAttributes")
 	defer span.End()
 
-	resourceDecisions, err := p.checkAttributes(ctx, resources, token, obligations)
+	resourceDecisions, err := p.checkAttributes(ctx, resources, token, fulfillableObligationFQNs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


service/kas/access/rewrap.go
```diff
@@ -766,6 +797,7 @@ func (p *Provider) tdf3Rewrap(ctx context.Context, requests []*kaspb.UnsignedRew
 			}
 
 			if !access {
+				// ? Maybe we should change the error if we get an obligation failure? Not just "forbidden"
```
pflynn-virtru, line 800: definitely a more precise AUDIT event.  the client response IMO is fine

service/kas/access/accessPdp.go
```diff
@@ -23,12 +23,13 @@ var decryptAction = &policy.Action{
...
-	Error  error
-	Policy *Policy
+	Access              bool
+	Error               error
+	Policy              *Policy
+	RequiredObligations []string
 }
 
-func (p *Provider) canAccess(ctx context.Context, token *entity.Token, policies []*Policy) ([]PDPAccessResult, error) {
+func (p *Provider) canAccess(ctx context.Context, token *entity.Token, policies []*Policy, obligations []string) ([]PDPAccessResult, error) {
```
jakedoublev, line 32: can we please name this argument like `fulfillableObligationFQNs` or something to indicate the origin (not policy, but request)?

service/kas/access/rewrap.go
jakedoublev: Audit will be populated with the obligations involved in a decision from the auth service side because every decision result gets logged to audit alongside the rewrap audit logs (important to track key releases).

service/kas/access/rewrap.go
```diff
@@ -996,7 +1042,73 @@ func extractNanoPolicy(symmetricKey ocrypto.ProtectedKey, header sdk.NanoTDFHead
...
+	}
+	fields[policyID] = structpb.NewListValue(&structpb.ListValue{
+		Values: values,
+	})
+	response.Metadata = metadata
+}
+
+// Retrieve additional request context needed for rewrap processing
+// Header is json encoded AdditionalRewrapContext struct
+// Example: `{"obligations": ["https://demo.com/obl/test/value/watermark","https://demo.com/obl/test/value/geofence"]}`
```
jakedoublev, line 1080: This is a good approach, but it can be more extensible if we make `obligations` an object so we can add more keys like `request_flags`, `geospatial`, or general PEP metadata.

PR #2799, opened 2025-10-14
feat(core): introduce typed Secret with inline env/file/literal, BindServiceConfig; integrate in KAS

strantalis:
### Proposed Changes
  - Add Secret type with redaction and safe JSON/log masking
  - Support inline secret directives: env:NAME, file:/path, literal:VALUE
  - Add BindServiceConfig helper with mapstructure decode hooks + validation
  - Implement file-based and env-based secret resolution (lazy/eager)
  - Integrate KAS: use Secret for root_key and BindServiceConfig
  - Improve env handling during reload to surface env-only keys
  - Add unit tests for Secret, BindServiceConfig, and inline decode forms
  - Update docs: “Secrets In Config” with examples and guidance
  - Remove temporary debug logs and experimental nested secret usage in KAS

  DX

  - Plain strings are literals by default; use literal: to disambiguate values starting with env:/file:
  - Example:
      - services.kas.root_key: "env:OPENTDF_SERVICES_KAS_ROOT_KEY"
      - services.kas.preview.key_management: true

### Checklist


dmihalcik-virtru (changes requested):
Can you avoid breaking compatibility with downstream users? Maybe print out a warning if the old `root_key` field is set, and use a new name (suggestion: `envelope_key`) from now on.

service/kas/access/provider.go
```diff
@@ -40,7 +40,7 @@ type KASConfig struct {
 	// Deprecated
 	RSACertID string `mapstructure:"rsacertid" json:"rsacertid"`
 
-	RootKey string `mapstructure:"root_key" json:"root_key"`
+	RootKey config.Secret `mapstructure:"root_key" json:"root_key"`
```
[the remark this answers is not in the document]
strantalis: There shouldn't be downstream impact. This should still support a string literal in yaml without any prefix.

service/pkg/config/config.go
```diff
@@ -190,6 +191,22 @@ func (c *Config) Reload(ctx context.Context) error {
 		assigned = make(map[string]struct{})
 		orderedViper := viper.NewWithOptions(viper.WithLogger(slog.Default()))
 
+		// If an environment loader is present, configure orderedViper to read env directly.
+		for _, loader := range c.loaders {
+			if loader.Name() == LoaderNameEnvironmentValue {
```
ntrevino-virtru: The env loader should already do this. And it should be configured by the consumer of the library.

service/pkg/config/environment_value_loader.go
```diff
@@ -56,7 +59,41 @@ func (l *EnvironmentValueLoader) Get(key string) (any, error) {
...
-	return l.viper.AllKeys(), nil
+	// Start with keys Viper already knows about (from defaults/file loads)
+	keys := make([]string, 0, len(l.viper.AllKeys()))
+	keys = append(keys, l.viper.AllKeys()...)
+
+	// Discover environment variables with the configured prefix and convert them to dotted keys
+	// Example: OPENTDF_SERVICES_KAS_ROOT_KEY -> services.kas.root_key
+	// Note: Viper treats keys case-insensitively; we normalize to lower-case dotted form here
+	prefix := l.envPrefix + "_"
+	for _, env := range os.Environ() {
```
ntrevino-virtru: This bit looks like what Viper should be doing already. If it is adding additional keys we should cache those keys and values so they can be returned by `Get`, otherwise we might hit this: https://github.com/opentdf/platform/blob/main/service/pkg/config/config.go#L219-L229

PR #2803, opened 2025-10-15
feat(policy): Protos List obligation triggers

c-r33d:
### Proposed Changes

1.) Protos for `ListObligationTriggers` rpc

### Checklist


service/policy/obligations/obligations.proto
```diff
@@ -302,6 +302,27 @@ message RemoveObligationTriggerResponse {
     policy.ObligationTrigger trigger = 1;
 }
 
+message ListObligationTriggersRequest {
+    // Required
+    option (buf.validate.message).oneof = { fields: ["obligation_value_id", "obligation_value_fqn"], required: true };
+    string obligation_value_id = 1 [(buf.validate.field).string.uuid = true];
+    string obligation_value_fqn = 2 [
```
jakedoublev: Is the intention of this API more of a global LIST of triggers rather than all associated with one value? I'd assume I could just `GetObligationValue(id or FQN)` to see all the triggers?

c-r33d: Maybe both?? I was following the previous flows with attribute values and listing under a specific attribute. Might be better to just have a List with no filter. Or have the filter be optional. Thoughts?

PR #2823, opened 2025-10-17
feat(policy): List obligation triggers rpc

c-r33d:
### Proposed Changes

1.) Implement the `ListObligationTriggers` RPC, which allows a user to list obligation triggers globally. The user can specify an optional namespace filter.
2.) Add NULLIF check for UUID casts to avoid errors do to early parameter binding optimizations.

### Checklist

### DIFF
--- a/service/policy/db/queries/obligations.sql
+++ b/service/policy/db/queries/obligations.sql
@@ -586,17 +586,6 @@ WHERE id = $1
 RETURNING id;
 
 -- name: listObligationTriggers :many
-WITH counted AS (
-    SELECT COUNT(ot.id) AS total
-    FROM obligation_triggers ot
-    JOIN obligation_values_standard ov ON ot.obligation_value_id = ov.id
-    JOIN obligation_definitions od ON ov.obligation_definition_id = od.id
-    LEFT JOIN attribute_namespaces n ON od.namespace_id = n.id
-    LEFT JOIN attribute_fqns fqns ON fqns.namespace_id = n.id AND fqns.attribute_id IS NULL AND fqns.value_id IS NULL
-    WHERE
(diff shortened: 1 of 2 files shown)


service/policy/db/queries/obligations.sql
```diff
@@ -584,3 +584,66 @@ WHERE obligation_value_id = $1;
 DELETE FROM obligation_triggers
 WHERE id = $1
 RETURNING id;
+
+-- name: listObligationTriggers :many
+SELECT
```
jakedoublev: Doing the counting as a CTE rather than a subquery should be more performant on this read hot path:

```sql
WITH counted AS (
    SELECT COUNT(id) AS total FROM obligation_triggers
)
SELECT ...
```

service/policy/db/queries/obligations.sql
```diff
@@ -584,3 +584,66 @@ WHERE obligation_value_id = $1;
...
+            'obligation_value', JSON_BUILD_OBJECT(
+                'id', ov.id,
+                'value', ov.value,
+                'obligation', JSON_BUILD_OBJECT(
+                    'id', od.id,
+                    'name', od.name,
+                    'namespace', JSON_BUILD_OBJECT(
+                        'id', n.id,
+                        'name', n.name,
+                        'fqn', COALESCE(ns_fqns.fqn, '')
```
jakedoublev: Do we need to COALESCE this? It shouldn't be null?

service/policy/db/queries/obligations.sql
jakedoublev: /gemini can you weigh in on this

service/policy/db/queries/obligations.sql
jakedoublev: Discussed: this is an established pattern we know is working, so we will not use a different approach until we reconsider fqn reindexing/cache consistency state.

jakedoublev (approved):

PR #2873, opened 2025-11-03
chore(ci): Add job to automate backport label creation

elizabethhealy:
### Proposed Changes

Automatically create a backport label when a release branch is created

* only runs when create-release-branch runs+succeeds
* checks for existing label
* logs success/failure

### Checklist


.github/workflows/create-release-branch.yaml
```diff
@@ -16,3 +16,60 @@ jobs:
...
+      - name: Create Backport PR label if it doesn't exist
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const tagName = context.payload.release?.tag_name;
+            const labelPrefix = "backport release/";
+            // Extract the label name by removing patch version
+            let labelName;
+            if (tagName.includes('.')) {
```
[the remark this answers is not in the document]
elizabethhealy: this shouldnt happen -- the job above that this job depends on requires `if: ${{ endsWith(github.event.release.tag_name, '.0') }}` so tag_name should be present

.github/workflows/create-release-branch.yaml
```diff
@@ -16,3 +16,58 @@ jobs:
...
+            const tagName = context.payload.release?.tag_name;
+            const labelPrefix = "backport release/";
+            // Extract the label name by removing patch version
+            let labelName;
+            if (tagName && tagName.includes('.')) {
+              const lastDotIndex = tagName.lastIndexOf('.');
+              labelName = `${labelPrefix}${tagName.substring(0, lastDotIndex)}`;
+            } else {
+              labelName = `${labelPrefix}${tagName}`;
+            }
```
dmihalcik-virtru: I'd recommend being strict here, and only matching known formats, e.g. something like:

```suggestion
            if (!tag_name || !tag_name.match) {
              console.error(`❌ Unable to create label for "${JSON.stringify(context.payload)}"`);
            }
            const m = tag_name.match(/^([\w/]+\/v\d+\.\d+)\.\d+$/);
            if (!m) {
              console.error(`❌ Unable to create label for unrecognized tag format "${tag_name}"`);
            }
            const labelName = m[1];
```

.github/workflows/create-release-branch.yaml
```diff
@@ -16,3 +16,69 @@ jobs:
...
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const tagName = context.payload.release?.tag_name;
+            if (!tagName) {
+              console.error("❌ No tag name found in the release payload");
+              throw new Error("No tag name found");
+            }
+            // check if it matches vX.Y.Z format or something/vX.Y.Z
+            const tagPattern = /^(.*\/)?v\d+\.\d+\.\d+$/;
```
dmihalcik-virtru: I'd suggest being a bit more strict. git tags are already even stricter than the following, with the exception that they will allow more `.` characters. This basically says no`.`s in the prefix

```suggestion
            const tagPattern = /^([\w/-_]*\/)?v\d+\.\d+\.\d+$/;
```

strantalis (approved):

PR #2918, opened 2025-11-18
fix(kas): Ensure root key is not logged.

c-r33d:
### Proposed Changes

1.) Modify the `KasConfig` object to redact the `root_key`
2.) Modify the `ServiceConfig` object to check for the existence of `root_key`, if present redact it.

Logs redacted:

1.) At startup:
Screenshot 2025-11-19 at 8 28 18 AM

2.) Update hook
Screenshot 2025-11-19 at 8 26 13 AM

3.) Startup - Invalid Config
Screenshot 2025-11-19 at 8 24 30 AM

### Checklist

### DIFF
--- a/service/pkg/config/config.go
+++ b/service/pkg/config/config.go
@@ -15,6 +15,8 @@ import (
 	"github.com/spf13/viper"
 )
 
+const rootKeyField = "root_key"
+
 // ChangeHook is a function invoked when the configuration changes.
 type ChangeHook func(configServices ServicesMap) error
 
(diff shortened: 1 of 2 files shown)


service/pkg/config/config.go
```diff
@@ -15,6 +15,8 @@ import (
 	"github.com/spf13/viper"
 )
 
+const rootKeyField = "root_key"
```
jentfoo: There are additional fields we should ensure are not logged:
* `DB` has a `Password` field that should be redacted
* `SDKConfig` has `ClientSecret`

Another related struct that can get tied to config:
* StandardConfig struct in service/internal/security/standard_cryto.go (contains keys)

PR #2941, opened 2025-12-02
feat(core): add additive CORS configuration fields

strantalis:
### Proposed Changes

  Add `additionalheaders`, `additionalmethods`, and `additionalexposedheaders`
  fields to CORS config, allowing operators to append custom values without
  replacing the entire default list.

  - Add mergeStringSlices() for case-sensitive method deduplication
  - Add mergeHeaderSlices() for case-insensitive header deduplication (RFC 7230)
  - Add Effective*() methods to compute merged values at runtime
  - Log effective CORS values at startup for operator visibility
  - Add unit tests for merge functions and Effective* methods
  - Update docs/Configuring.md with CORS configuration section

  Operators can now add custom headers without copying all defaults:

  ```yaml
  server:
    cors: 
      additionalheaders: 
        - X-Custom-Header
```
  Backwards compatible: existing configs work unchanged.

### Checklist

### DIFF
--- a/docs/Configuring.md
+++ b/docs/Configuring.md
@@ -10,12 +10,27 @@ The platform leverages [viper](https://github.com/spf13/viper) to help load conf
   - [SDK Configuration](#sdk-configuration)
   - [Logger Configuration](#logger-configuration)
   - [Server Configuration](#server-configuration)
+    - [CORS Configuration](#cors-configuration)
+      - [Additive Configuration](#additive-configuration)
+      - [Programmatic Configuration](#programmatic-configuration)
     - [Crypto Provider](#crypto-provider)
   - [Database Configuration](#database-configuration)
     - [Tracing Configuration](#tracing-configuration)
+  - [Security Configuration](#security-configuration)
   - [Services Configuration](#services-configuration)
     - [Key Access Server (KAS)](#key-access-server-kas)
     - [Authorization](#authorization)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: service/internal/server/server.go [hunk 1]
```diff
@@ -129,6 +129,89 @@ type CORSConfig struct {
 	AllowCredentials bool     `mapstructure:"allowcredentials" json:"allowcredentials" default:"true"`
 	MaxAge           int      `mapstructure:"maxage" json:"maxage" default:"3600"`
 	Debug            bool     `mapstructure:"debug" json:"debug"`
+
+	// Additive fields - appended to base lists at runtime without replacing defaults
+	AdditionalMethods        []string `mapstructure:"additionalmethods" json:"additionalmethods"`
```
jakedoublev: I'm good with this, but would `AdditionalAllowed*` be a better convention so the associations are even more closely aligned between `AllowedMethods -> AdditionalAllowedMethods` `AllowedHeaders -> AdditionalAllowedHeaders`?

jakedoublev (approved):

service/internal/server/server.go
jakedoublev: I think `AdditionalExposedHeaders` would stay the same since it's different than `AdditionalAllowedHeaders`, but we can just leave this how it is.

PR #2968, opened 2025-12-19
fix(sdk): conditionally set client_id based on auth method

### LINKED ISSUE #3076: SDK client credentials auth fails when using Okta OAuth provider

imdominicreed:
Move client_id setting from generic request setup to auth-method-specific handlers. This ensures client_id is only included when using client assertion authentication, not for all request types.

### Proposed Changes

* fix #3076 

### Checklist

### DIFF
--- a/sdk/auth/oauth/oauth.go
+++ b/sdk/auth/oauth/oauth.go
@@ -290,7 +290,7 @@ func getTokenExchangeRequest(ctx context.Context, tokenEndpoint, dpopNonce strin
 	}
 
 	if len(scopes) > 0 {
-		data.Set("scopes", strings.Join(scopes, " "))
+		data.Set("scope", strings.Join(scopes, " "))
 	}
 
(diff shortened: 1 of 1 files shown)


sdk/auth/oauth/oauth.go
```diff
@@ -293,8 +293,7 @@ func getTokenExchangeRequest(ctx context.Context, tokenEndpoint, dpopNonce strin
 		data.Set("scopes", strings.Join(scopes, " "))
```
jentfoo: This line is incorrectly setting the field `scopes` (verified `scope` as used on line 75 is correct: https://www.rfc-editor.org/rfc/rfc8693.html#name-scope-scopes-claim )

imdominicreed: good catch!

strantalis (approved):

PR #2977, opened 2025-12-23
fix(policy): wrap SQL optional param type casts in null checks

alkalescent:
### Proposed Changes

* PostgreSQL's query optimizer evaluates conditions for WHERE clauses in any order, so we must wrap optional parameter type casts in null checks to ensure expressions are safe.

### Checklist

### DIFF
--- a/service/integration/attributes_test.go
+++ b/service/integration/attributes_test.go
@@ -1468,11 +1468,7 @@ func (s *AttributesSuite) Test_GetAttribute_ByIdAndFqn_ReturnSameResult() {
 		s.Require().NotNil(attrByFQN)
 
 		// Verify both return the same attribute
-		s.Equal(attrByID.GetId(), attrByFQN.GetId(), "ID should match for FQN: %s", fqn)
-		s.Equal(attrByID.GetName(), attrByFQN.GetName(), "Name should match for FQN: %s", fqn)
-		s.Equal(attrByID.GetRule(), attrByFQN.GetRule(), "Rule should match for FQN: %s", fqn)
-		s.Equal(attrByID.GetNamespace().GetId(), attrByFQN.GetNamespace().GetId(), "Namespace ID should match for FQN: %s", fqn)
-		s.Len(attrByID.GetValues(), len(attrByFQN.GetValues()), "Number of values should match for FQN: %s", fqn)
+		s.True(proto.Equal(attrByID, attrByFQN))
 	}
 }
 
--- a/service/integration/kas_registry_test.go
+++ b/service/integration/kas_registry_test.go
@@ -883,12 +883,8 @@ func (s *KasRegistrySuite) Test_GetKeyAccessServer_ByIdNameUri_ReturnSameResult(
 	s.Require().NotNil(kasByURI)
 
 	// Verify all three return the same KAS
-	s.Equal(kasByID.GetId(), kasByName.GetId(), "ID should match between ID and Name lookups")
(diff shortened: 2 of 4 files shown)


service/integration/attributes_test.go
```diff
@@ -1451,6 +1451,31 @@ func (s *AttributesSuite) getAttributeFixtures() map[string]fixtures.FixtureData
...
+		s.Require().NoError(err, "Failed to get attribute by ID for FQN: %s", fqn)
+		s.Require().NotNil(attrByID)
+
+		// Get by FQN
+		attrByFQN, err := s.db.PolicyClient.GetAttribute(s.ctx, &attributes.GetAttributeRequest_Fqn{Fqn: fqn})
+		s.Require().NoError(err, "Failed to get attribute by FQN: %s", fqn)
+		s.Require().NotNil(attrByFQN)
+
+		// Verify both return the same attribute
+		s.Equal(attrByID.GetId(), attrByFQN.GetId(), "ID should match for FQN: %s", fqn)
```
jakedoublev: I think you could optimize all of the matching query results test assertions with the `google.golang.org/protobuf/proto` import as `s.True(proto.Equal(attrByID, attrByQN))`. This would also detect regressions caused by drift when queries start populating new fields.

service/policy/db/queries/attributes.sql
```diff
@@ -3,6 +3,11 @@
 ----------------------------------------------------------------
 
 -- name: listAttributesDetail :many
+WITH params AS (
+    SELECT
+        NULLIF(@namespace_id, '')::uuid as namespace_id,
```
jakedoublev: Is there a reason we cannot unify on `sqlc.narg` or `NULLIF` exclusively in all of these queries rather than using both? It looks like `sqlc.narg` is a little cleaner.

c-r33d (commented):
I think the approach will work. One thing that concerns me is the `CROSS JOIN` on the result set. If the SELECT query returns a lot of results, the cross could take a decent chunk of time to complete. While I think your approach is more readable than doing things inline, maybe it's better to do inline NULLIFs/sqlc narg for LIST commands, and hot path routes? @alkalescent

service/policy/db/queries/attributes.sql
```diff
@@ -342,8 +354,8 @@ LEFT JOIN (
     INNER JOIN key_access_servers kas ON kask.key_access_server_id = kas.id
     GROUP BY k.definition_id
 ) defk ON ad.id = defk.definition_id
-WHERE (sqlc.narg('id')::uuid IS NULL OR ad.id = sqlc.narg('id')::uuid)
-  AND (sqlc.narg('fqn')::text IS NULL OR REGEXP_REPLACE(fqns.fqn, '^https?://', '') = REGEXP_REPLACE(sqlc.narg('fqn')::text, '^https?://', ''))
+WHERE (params.id IS NULL OR ad.id = params.id)
+  AND (params.fqn IS NULL OR REGEXP_REPLACE(fqns.fqn, '^https?://', '') = params.fqn)
```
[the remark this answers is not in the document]
alkalescent: removed "?" from regular expressions

service/policy/db/queries/attributes.sql
alkalescent: `sqlc.narg` and `NULLIF` serve different purposes, and I didn't want to change app code (to prove these SQL changes didn't break anything). However, the app code changes were simple. Removed all `NULLIF` invocations from all SQL queries.

PR #2999, opened 2026-01-10
feat(core): add fine-grain authorization support

alkalescent:
### Proposed Changes

* adds fine-grain authorization support
* adds pluggable authorizer

### Checklist


jrschumacher (dismissed):

service/internal/auth/authz/casbin/casbin.go
```diff
@@ -0,0 +1,573 @@
...
+
+	// Sort keys for canonical ordering
+	keys := make([]string, 0, len(allDims))
+	for k := range allDims {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	// Build canonical string
+	parts := make([]string, 0, len(keys))
```
jentfoo: Are we certain it's impossible for a key to contain `=` or `&`? The point of validation is not obvious to me, and if either were in a key there is the potential for confusion here. We should pick a point to add a defensive check.

jentfoo (changes requested):
A couple defensive recommendations, the rolePrefix check is a critical fix.

service/internal/auth/authn.go
```diff
@@ -277,24 +332,37 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
...
-			http.Error(w, "internal server error", http.StatusInternalServerError)
+
+		// Defensive check: authorizer must be initialized
+		if a.authorizer == nil {
+			log.ErrorContext(ctx, "authorizer not initialized")
+			http.Error(w, "authorization system not configured", http.StatusInternalServerError)
+			return
+		}
+
+		// Build authorization request
```
c-r33d: The gRPC gateway was removed, is this still needed?

c-r33d:
Moved to a cleaner version here:
- https://github.com/opentdf/platform/pull/3614

PR #3053, opened 2026-02-06
fix: Go 1.25

pflynn-virtru:
## Summary
- Upgrades `go` directive to `1.25.0` and `toolchain` to `go1.25.7` in all workspace modules and `go.work`
- Addresses GO-2026-4337 (`crypto/tls` vulnerability, fixed in go1.25.7)
- Updates `golangci-lint` from `v2.1` to `v2.8.0` in CI workflow and Makefile (required for Go 1.25 compatibility)
- Updates `govulncheck` `go-version-input` to `1.25.7` in `checks.yaml` and `vulnerability-check.yaml`
- Updates `sonarcloud.yml` Go version from `1.23` to `1.25.7`

## Test plan
- [x] `go mod tidy` succeeds in all modules
- [x] `go build ./...` succeeds in all modules
- [ ] CI checks pass

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -26,7 +26,7 @@ jobs:
       - name: "Setup Go"
         uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
-          go-version: "1.23"
+          go-version: "1.25.7"
           check-latest: false
           cache-dependency-path: |
             service/go.sum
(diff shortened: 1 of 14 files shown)


.github/workflows/sonarcloud.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           name: code-coverage-report
 
       - name: "SonarCloud Scan"
+        continue-on-error: true
```
elizabethhealy: would this cause the workflow to fail silently?

pflynn-virtru: removed, it is coming from the GitHub app not this workflow

PR #3062, opened 2026-02-09
feat(kas): Add kas uri resolver

c-r33d:
### Proposed Changes

1.) Add a kas uri resolver that can be injected into the server start options
2.) Default is to use the static resolver that reads the `registered_kas_uri` from the kas configuration for previous backwards compatibility.

### Checklist


service/trust/key_index.go
```diff
@@ -80,3 +80,10 @@ type KeyIndex interface {
 	// List keys with options
 	ListKeysWith(ctx context.Context, opts ListKeyOptions) ([]KeyDetails, error)
 }
+
+// RegisteredKasURIResolver resolves the registered KAS URI for KAS when running with key management.
+type RegisteredKasURIResolver interface {
+	fmt.Stringer
+	slog.LogValuer
+	ResolveURI(ctx context.Context) (string, error)
```
jentfoo: What is the contract with the context here? I assume lifecycle only and not intended to be used for communicating request state? Might be worth a godoc here.

PR #3077, opened 2026-02-13
fix(sdk): exp SDK tests and fix

pflynn-virtru:
This pull request adds comprehensive regression and compatibility tests to the experimental TDF SDK, focusing on key management edge cases and ensuring encryption format compatibility with the production SDK. The new tests verify correct handling of legacy attribute grants, segment integrity hashing, and cross-decryption between SDKs. Additionally, new imports were added to support these tests.

See xtest failure: https://github.com/opentdf/tests/pull/414

**Regression and Compatibility Testing Enhancements:**

* Added regression tests to `xor_splitter_test.go` to ensure that when an attribute grant references a KAS URL by URI only (without an embedded public key), the default KAS public key is correctly merged, and that existing keys are not overwritten by the default.
* Added end-to-end regression tests in `writer_test.go` to verify that segment hashes include both nonce and ciphertext (not just ciphertext), and that finalization succeeds for URI-only grants by leveraging the default KAS public key.

**Cross-SDK Compatibility:**

* Introduced a new test in `writer_test.go` to validate that the experimental writer's output is fully compatible with the production SDK's decryption and segment hash verification, including full TDF ZIP assembly and parsing with the production ZIP reader.

**Test Infrastructure Improvements:**

* Registered the new regression and compatibility tests in the main test suite.
* Added necessary imports for cryptographic operations and ZIP parsing to support the new tests. [[1]](diffhunk://#diff-f935252b0fdbd2ed0baff420a925250cdcfef0dd3c877aa984d449b96d93340aR7-R12) [[2]](diffhunk://#diff-f935252b0fdbd2ed0baff420a925250cdcfef0dd3c877aa984d449b96d93340aR21)

### DIFF
--- a/examples/cmd/benchmark_experimental.go
+++ b/examples/cmd/benchmark_experimental.go
@@ -2,9 +2,14 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
+	"crypto/tls"
 	"fmt"
+	"io"
+	"net/http"
+	"os"
 	"sync"
 	"time"
 
@@ -13,7 +18,6 @@ import (
 	kasp "github.com/opentdf/platform/protocol/go/kas"
 	"github.com/opentdf/platform/protocol/go/kas/kasconnect"
 	"github.com/opentdf/platform/protocol/go/policy"
-
 	"github.com/opentdf/platform/sdk/experimental/tdf"
 	"github.com/opentdf/platform/sdk/httputil"
 	"github.com/spf13/cobra"
@@ -35,7 +39,7 @@ func init() {
 	//nolint: mnd // no magic number, this is just default value for payload size
 	benchmarkCmd.Flags().IntVar(&payloadSize, "payload-size", 1024*1024, "Payload size in bytes") // Default 1MB
 	//nolint: mnd  // same as above
-	benchmarkCmd.Flags().IntVar(&segmentChunk, "segment-chunks", 16*1024, "segment chunks ize") // Default 16 segments
+	benchmarkCmd.Flags().IntVar(&segmentChunk, "segment-chunks", 16*1024, "segment chunk size") // Default 16KB
 	ExamplesCmd.AddCommand(benchmarkCmd)
 }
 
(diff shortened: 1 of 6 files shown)


dmihalcik-virtru (dismissed):

dmihalcik-virtru (dismissed):

examples/cmd/benchmark_experimental.go
```diff
@@ -98,12 +109,48 @@ func runExperimentalWriterBenchmark(_ *cobra.Command, _ []string) error {
...
+	}
+	var decrypted bytes.Buffer
+	if _, err = io.Copy(&decrypted, tdfReader); err != nil {
+		return fmt.Errorf("failed to decrypt TDF with production SDK: %w", err)
+	}
+
+	if bytes.Equal(payload, decrypted.Bytes()) {
+		fmt.Println("| Decrypt Verify     | PASS - roundtrip matches |")
+	} else {
+		fmt.Printf("| Decrypt Verify     | FAIL - payload %d bytes, decrypted %d bytes |\n", len(payload), decrypted.Len())
```
dmihalcik-virtru, line 152: Is this accessible? I'd think `Copy` would return an error in most cases I can think of, unless maybe somehow you opened the wrong file?

examples/cmd/benchmark_experimental.go
```diff
@@ -98,12 +109,48 @@ func runExperimentalWriterBenchmark(_ *cobra.Command, _ []string) error {
...
 	totalTime := end.Sub(start)
 
+	// Assemble the complete TDF: segment data (in order) + finalize data
+	var tdfBuf bytes.Buffer
+	for i, sr := range segResults {
+		if _, err := io.Copy(&tdfBuf, sr.TDFData); err != nil {
+			return fmt.Errorf("failed to read segment %d TDF data: %w", i, err)
+		}
+	}
+	tdfBuf.Write(result.Data)
```
dmihalcik-virtru, line 119: Why not just copy to `io.Discard` and record the number of bytes? There isn't much advantage to storing the file twice in memory. See the note below: decrypt is authenticated so mismatched bytes should be unreachable

dmihalcik-virtru (approved):

PR #3142, opened 2026-03-12
feat(sdk): export action name constants

pflynn-virtru:
## Summary
- Export `ActionNameCreate`, `ActionNameRead`, `ActionNameUpdate`, and `ActionNameDelete` constants from the `sdk` package
- Allows downstream consumers (e.g. `secure-enclave`) to reference standard action names without importing `service/policy/actions`, avoiding the heavy `service` module dependency (~300 go.mod lines, testcontainers, DB drivers, etc.)
- Updates internal `readActionName` in `tdf.go` to reference the new exported constant

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./... -short` passes
- [x] `golangci-lint run ./...` — 0 issues
- [x] `gofumpt -l .` — no formatting issues


dmihalcik-virtru (dismissed):
~~Suggestion: Consider making these a typed thing?~~

After discussion, simpler >> compex

marythought (commented):
Thanks for tackling this — avoiding the heavy `service` import from the SDK is a real pain point and worth solving.

A few questions on the approach:

1. **Naming** — The existing constants that all consumers use are `ActionNameCreate`, `ActionNameRead`, etc. (in `service/policy/actions`). Was there a specific reason for choosing `PolicyActionNameCreate` / `DecisionActionNameCreate` over matching the existing convention? Wondering if I'm missing context on a planned naming distinction.

2. **`DecisionActionName*` aliases** — These are identical to the `PolicyActionName*` values today. What's the thinking behind exposing both sets? Is there a near-term plan for decision actions to diverge from policy actions, or is this more of a future-proofing measure?

3. **Duplication vs. shared package** — Was extracting a lightweight leaf package considered (e.g., `service/policy/actionnames` or somewhere in `protocol`) that both `sdk` and `service/policy/actions` could import? Curious whether something made that impractical — the comment-enforced sync between the two files makes me a little nervous long-term.

4. **`tdf.go` change** — What motivated tying `readActionName` to `PolicyActionNameRead` rather than leaving it as the literal `"read"`? My read was that the TDF concept of "read" is independent of the policy service's action naming, but wondering if there's a coupling I'm not seeing.

Happy to discuss — the goal is solid, just want to make sure I understand the design choices before approving.

sdk/actions.go
```diff
@@ -0,0 +1,29 @@
+package sdk
```
dmihalcik-virtru, line 1: Hmm could we put this in a subpackage so the imports are like

```golang
import "github.com/opentdf/platform/sdk/actions"

...
func example() {
	a1 := actions.Read
}
```

something like 
```
package actions
```

or maybe `policyAction` or something? I'm overthinking it...

This also feels like something that could be derived from a proto enum...

PR #3164, opened 2026-03-16
feat(policy): Add actions fqn support across downstream consumers

elizabethhealy:
### Proposed Changes

* maintains name support for backward compatibility, if name provided assumes namespace=null

### Checklist

## Release Notes

* **New Features**
  * Added support for action Fully Qualified Names (FQNs) across policy APIs, allowing actions to be referenced using URIs (e.g., `https://namespace/act/action-name`) in addition to IDs and names.
  * Standard actions (`create`, `read`, `update`, `delete`) are now automatically created when a new namespace is established.

* **Documentation**
  * Updated API schemas and protocol definitions to include FQN fields for actions and identifiers.

### DIFF
--- a/service/policy/adr/0005-action-fqn-support-in-policy-apis.md
+++ b/service/policy/adr/0005-action-fqn-support-in-policy-apis.md
@@ -1,0 +1,37 @@
+# Policy APIs will support action FQNs for downstream references
+
+As policy objects become namespace-aware, referring to actions by `name` alone is ambiguous across namespaces.
+
+This is especially important for downstream APIs that attach actions to other policy objects, including:
+
+1. Obligation Triggers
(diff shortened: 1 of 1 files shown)


service/common/common.proto
```diff
@@ -17,6 +17,10 @@ message IdNameIdentifier {
         expression: "this.matches('^[a-zA-Z0-9](?:[a-zA-Z0-9_-]*[a-zA-Z0-9])?$')"
       }
     ];
+    string fqn = 3 [(buf.validate.field).string = {
```
elizabethhealy: ideally id create a IdFqnNameIdentifier or switch from IdNameIdentifier to IdFqnIdentifier but those would cause breaking changes

PR #3210, opened 2026-03-26
[DRAFT] PQC xwing (codex generated)

dmihalcik-virtru:
- **feat(kas): add ml-kem key wrapping support**
- **feat(kas): add ml-kem-1024 support**
- **Add X-Wing hybrid KAO wrapping support**

PROMPT

> This branch adds support for ML-KEM wrapping to our TDF Key Access Object wrapping. I would like to explicitly support hybrid keys with xwing, as defined in https://www.ietf.org/archive/id/draft-connolly-cfrg-xwing-kem-10.txt. So, like this PR adds the `mlkem-wrapped SchemeType`, it should also offer the `hybrid SchemeType`. 
> This will be supported by a new lib/ocrypto `HybridXWing KeyType = "hpqt:xwing"` in ec_key_pair.go (maybe move those to a new file with a neutral name?). Let KAS and the policy service store and serve a composite public 
 key as a single byte array (or in a PEM). E.g. let the key data be a concatenated set of traditional and post-quantum keys, or an ASN.1 object with additional data that contains both types of keys. Simmilarly, the KEM  objects themselves should be base64 encoded in the same field in the KAO they are now, but will be much longer and more structured with ASN.1 Data so the server can later perform rewrap without additional json fields or data beyond the schema that currently exists (i.e. without needing a new ciphertext field or similar)


dmihalcik-virtru:
Merged in #3276

PR #3226, opened 2026-03-31
feat(authz): Namespaced policy in decisioning

elizabethhealy:
### Proposed Changes

This pull request introduces a namespaced policy evaluation model to the Policy Decision Point (PDP). The primary goal is to resolve ambiguities in access decisioning where action names might overlap across different namespaces. By threading a new `NamespacedPolicy` configuration through the PDP, the system can now enforce strict namespace equality for actions and subject mappings, ensuring a fail-closed and deterministic authorization process suitable for staged migration.

### Highlights

* **Namespaced Policy Enforcement**: Introduced a `NamespacedPolicy` feature flag to enforce strict namespace ownership in access decisioning, preventing cross-namespace action matches.
* **Deterministic Action Resolution**: Implemented namespace-aware action matching logic that resolves actions within specific evaluation contexts, with explicit identity precedence (ID > Name+Namespace > Name).
* **Subject Mapping Filtering**: Added logic to filter out unnamespaced subject mappings when the `NamespacedPolicy` mode is enabled.
* **Conflict Resolution**: Added defensive logic to `EvaluateSubjectMappingsWithActions` to handle and log action name collisions deterministically.

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Add optional strict, namespace-aware authorization (EnforceNamespacedEntitlements) enforcing per-namespace action matching with ID → name+namespace → name precedence and fail-closed behavior.
* **Bug Fixes**
  * Deterministic deduplication of conflicting actions to avoid silent overwrites; direct entitlements are merged rather than replaced.
* **Documentation**
  * Added ADR and deployment template describing namespaced decisioning.
* **Tests**
  * Expanded unit and BDD tests covering namespaced decisioning scenarios and matching semantics.


service/authorization/v2/config.go
```diff
@@ -20,6 +20,9 @@ type Config struct {
 
 	// enable entity direct entitlements that do not require subject mappings
 	AllowDirectEntitlements bool `mapstructure:"allow_direct_entitlements" json:"allow_direct_entitlements" default:"false"`
+
+	// enforce strict namespaced policy evaluation behavior in access decisioning
+	NamespacedPolicy bool `mapstructure:"namespaced_policy" json:"namespaced_policy" default:"false"`
```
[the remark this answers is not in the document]
jakedoublev: Good question. In my mind, they are different settings (require administration of the policy constructs with namespaces, drive entitlement resolution via namespaces), but one feature: namespaced policy. Are there any global/temporary config locations for cross-cutting features like this, or temporary features?

service/internal/access/v2/evaluate.go
```diff
@@ -80,9 +81,28 @@ func getResourceDecision(
 		}
 		for _, aav := range regResValue.GetActionAttributeValues() {
 			aavAttrValueFQN := aav.GetAttributeValue().GetFqn()
+			// First, check whether the request action identity (id or name[/namespace])
+			// could match this AAV action at all. This lightweight pre-check is used to
+			// decide whether strict mode must fail closed when namespace context is
+			// missing for this candidate AAV.
+			matchesRequestIdentity := isRequestedActionMatch(ctx, l, action, "", aav.GetAction(), false)
```
jakedoublev: Nit: please name the empty string and the false boolean arguments for readability.

service/internal/subjectmappingbuiltin/subject_mapping_builtin_actions.go
```diff
@@ -71,10 +72,28 @@ func EvaluateSubjectMappingsWithActions(
...
+					// namespace consistency between subject mappings and referenced actions.
+					// This extra conflict check is defensive for unexpected or mixed legacy
+					// states in-memory; if encountered, keep deterministic behavior and log.
 					m := make(map[string]*policy.Action, len(actions))
 					for _, action := range actions {
-						m[strings.ToLower(action.GetName())] = action
+						key := strings.ToLower(action.GetName())
+						if existing, ok := m[key]; ok {
+							if actionsConflict(existing, action) {
+								slog.Warn(
```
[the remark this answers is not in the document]
elizabethhealy: so theres an import cycle  -- subjectmappingbuiltin -> logger -> logger/audit -> subjectmappingbuiltin -- so instead im just passing the service's slog.Logger

service/internal/access/v2/pdp.go
```diff
@@ -155,11 +171,20 @@ func NewPolicyDecisionPoint(
 				return nil, fmt.Errorf("invalid registered resource value: %w", err)
 			}
 
+			namespaceName := namespaceNameFromPolicyNamespace(rr.GetNamespace())
+
 			fullyQualifiedValue := identifier.FullyQualifiedRegisteredResourceValue{
```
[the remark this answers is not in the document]
elizabethhealy: 1. in that scenario we don’t reject the RR just because it’s unnamespaced.
Flow is: resolve the RR value FQN, evaluate each RR AAV, derive required namespace from the AAV’s attribute value context, in strict mode (EnforceNamespacedEntitlements=true), only allow AAV actions whose namespace matches that required namespace (otherwise fail/deny).
So enforcement is at the AAV/action matching layer, not an upfront “RR must be namespaced” check.
 
1A: EnforceNamespacedEntitlements=true + namespaced RR FQN request, flow is: resolve RR value by the namespaced RR FQN, evaluate RR AAVs for requested action, derive required namespace from each AAV’s attribute-value context, enforce strict namespace match for action evaluation per AAV, return permit/deny from those namespace-scoped results.
And since RR create/update already enforces AAV namespace consistency with the RR namespace, this path stays aligned.

2. assuming they've done the migration/pruning (and the old un-namespaced RRs no longer exist in the db)
Flow: request uses legacy/global RR FQN (https://reg_res/...), all RR data is now namespaced-only (https://<ns>/reg_res/...), lookup for legacy FQN won’t resolve a registered resource value, decision returns deny / not-found path (fail-closed), not permit fallback.

2A: if they havent pruned the un-namespaced RRs, then this behaves like 1.

3. If https://reg_res/example_rr_name/value/value does not exist in the db (ie migration/pruning has taken place) then fail/deny. If it does still exist in the db (no pruning yet) -- then we match the existing behavior -- PDP resolves RR value using legacy/global FQN key (https://reg_res/...) if present. AAV action matching is name-based legacy behavior (no strict namespace requirement). Rules evaluate as before; decision can permit/deny based on entitlement/rule match.
So: legacy RR requests continue to work in non-strict mode, assuming legacy RR entries exist in memory.

cshamrick (dismissed):

service/authorization/v2/config.go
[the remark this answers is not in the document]
jakedoublev: That sounds good to me. Make it easy on the customer and clear for us.

service/internal/access/v2/pdp.go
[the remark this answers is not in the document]
jakedoublev: Does the code get simpler or more complex if `EnforceNamespacedEntitlements=true` enforces namespaces on registered resources in decision requests as well?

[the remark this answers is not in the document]
elizabethhealy: after dicsussion -- we should enforce namespace requirement for RRs if flag on

service/internal/access/v2/evaluate.go
```diff
@@ -80,9 +95,28 @@ func getResourceDecision(
...
+			// This ensures that only attribute values from the correct namespace are considered in the evaluation.
+			if namespacedPolicy {
+				parsed, err := identifier.Parse[*identifier.FullyQualifiedAttribute](aavAttrValueFQN)
+				if err != nil {
+					return nil, fmt.Errorf("invalid attribute value FQN [%s]: %w", aavAttrValueFQN, ErrInvalidResource)
+				}
+				if parsed.Namespace != requiredNamespace {
+					return nil, fmt.Errorf("attribute value FQN [%s] namespace [%s] does not match RR namespace [%s]: %w", aavAttrValueFQN, parsed.Namespace, requiredNamespace, ErrInvalidResource)
+				}
+				// Since we dont have the ns id on the RR value, pull it from the attr val
```
jakedoublev: Did your RR query change in this PR provide it upstream, making this comment and logic outdated?

service/internal/access/v2/evaluate.go
```diff
@@ -414,3 +491,81 @@ func hierarchyRule(
...
+	// that identity constraint regardless of namespacedPolicy mode.
+	if requestNamespace := requestedAction.GetNamespace(); requestNamespace != nil && (requestNamespace.GetId() != "" || requestNamespace.GetFqn() != "") {
+		entitledNamespace := entitledAction.GetNamespace()
+		if entitledNamespace == nil {
+			l.TraceContext(ctx, "action match request namespace mismatch",
+				slog.String("requested_action_namespace_id", requestNamespace.GetId()),
+			)
+			return false
+		}
+		if requestNamespace.GetId() != "" && entitledNamespace.GetId() != requestNamespace.GetId() {
```
jakedoublev, line 494: Can you please describe the conditions that some of these more complex expressions represent?

```suggestion
        specifiedReqNamespaceDoesNotMatchEntitled := requestNamespace.GetId() != "" && entitledNamespace.GetId() != requestNamespace.GetId()
		if  specifiedReqNamespaceDoesNotMatchEntitled {
```

service/internal/access/v2/pdp.go
```diff
@@ -214,10 +272,26 @@ func (p *PolicyDecisionPoint) GetDecision(
 		for _, directEntitlement := range entityRepresentation.GetDirectEntitlements() {
 			fqn := directEntitlement.GetAttributeValueFqn()
 			actionNames := directEntitlement.GetActions()
+			// In strict namespaced-policy mode, direct-entitlement actions must carry
+			// the same namespace context as the entitled attribute value so they can
+			// satisfy namespace-aware action matching during rule evaluation.
+			var actionNamespace *policy.Namespace
```
jakedoublev, line 278: If the direct entitlement FQN is not found (which happens with some UUID scenarios today), we would not have any actions from either lookup map because the FQN wouldn't be a key. Instead of this logic, or perhaps in tandem to it, should we do either of these?

1. check the namespace of the direct entitlement FQN itself and use that for the entitled actions
2. expect that ERS set the direct entitlement action namespaces appropriately

We will always have the namespace name and FQN within the direct entitlement attr value FQN.

jakedoublev: If we rely on the name of the namespace for entitlement checks instead of ID, that can always be pulled off the FQN as self-encapsulated state.

service/policy/db/queries/registered_resources.sql
```diff
@@ -108,7 +108,10 @@ LEFT JOIN LATERAL (
         JSON_BUILD_OBJECT(
             'action', JSON_BUILD_OBJECT(
                 'id', a.id,
-                'name', a.name
+                'name', a.name,
+                'namespace', CASE WHEN a.namespace_id IS NULL THEN NULL
```
jakedoublev, line 112: Could you please add an integration test for this change in `service/integration/registered_resources_test.go`?

service/internal/access/v2/evaluate.go
```diff
@@ -247,11 +284,24 @@ func evaluateDefinition(
...
 func allOfRule(
-	_ context.Context,
-	_ *logger.Logger,
+	ctx context.Context,
+	l *logger.Logger,
+	entitlements subjectmappingbuiltin.AttributeValueFQNsToActions,
+	action *policy.Action,
+	resourceValueFQNs []string,
+) []EntitlementFailure {
+	// Legacy wrapper: evaluate without strict namespace constraints.
```
jakedoublev: Are we still calling `allOfRule` directly outside unit tests, or is this claude wanting to strictly avoid breaking changes?

service/internal/access/v2/evaluate.go
elizabethhealy: kind of, but we can pull it from the RR in the pdp code when we list and that will eliminate a lot of this logic

service/internal/access/v2/evaluate.go
elizabethhealy: theyre being used in some tests, but we can just update those call sites as well

service/internal/access/v2/pdp.go
jakedoublev: @elizabethhealy and I discussed and my concern was incorrect - the FQNs of direct entitlements are added to these lookup maps in preprocessing before we hit this point. However, this may not be needed if we reduce reliance on namespace IDs and use namespace names/FQNs throughout as a namespaced object should always have the namespace name/FQN available self-encapsulated within its own FQN.

service/internal/access/v2/evaluate.go
elizabethhealy: removed

service/internal/access/v2/evaluate.go
elizabethhealy: added comments above each of these explaining whats going on

service/authorization/v2/authorization.go
jakedoublev, line 169: merge conflict?

jakedoublev: Looks like some otel stuff got removed? https://github.com/opentdf/platform/pull/3226/changes/79e3a29312901528468d038e10666a5b24647a91

jakedoublev (approved):

PR #3285, opened 2026-04-08
fix(ci): Prefer go.work for toolchain info

dmihalcik-virtru:
### Proposed Changes

* We have been setting toolchain explicitly to satisfy `govulncheck`
* Setting it in the `go.work` file lets us keep everything up to date while only editing one file
* Updates toolchain to latest on current minor branch: 1.25.9

### Checklist

## Summary by CodeRabbit

* **Chores**
  * Removed explicit per-module toolchain pins in favor of the workspace-specified toolchain.
  * CI workflows now derive the Go toolchain from the workspace file instead of hardcoded versions.
  * Workspace regeneration preserves and conditionally restores the original toolchain when present, and skips unnecessary work for leaf modules to avoid unintended changes.

### DIFF
--- a/.github/scripts/work-init.sh
+++ b/.github/scripts/work-init.sh
@@ -36,10 +36,16 @@ if ! cd "$ROOT_DIR"; then
   exit 1
 fi
 
+# Preserve the toolchain directive from the original go.work so that CI steps
+# reading go-version-file: go.work (e.g. govulncheck) continue to use the
+# correct Go version after the workspace is regenerated.
+ORIG_TOOLCHAIN=$(awk '/^toolchain / {print $2; exit}' go.work 2>/dev/null)
+
 echo "[INFO] Rebuilding partial go.work for [${component}]"
 case $component in
 lib/ocrypto | lib/fixtures | lib/flattening | lib/identifier | protocol/go)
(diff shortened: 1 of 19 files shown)


pflynn-virtru (dismissed):

pflynn-virtru (dismissed):

jentfoo (dismissed):
We can try it, however this will likely make dependabot updates require more manual effort if they need a go version update (not clear if the tool will update the go.work)

dmihalcik-virtru:
> We can try it, however this will likely make dependabot updates require more manual effort if they need a go version update (not clear if the tool will update the go.work)

AFAIK dependabot only updates the toolchain as a side effect of dep updates, so it isn't predictable already. It will continue to add toolchain directives to the go.mod files if a dep adds a toolchain dep greater than the one in the go.work, but it will disappear with mod tidy when the go.work is updated past it.

dmihalcik-virtru:
I've added a [commit](https://github.com/opentdf/platform/pull/3285/changes/71c326edc4a75041aff9545432c8770f1bbe12be) to include the new otdfctl go.mod, as well as consolidate the setup-go actions on go.work

marythought (approved):

strantalis (approved):

PR #3313, opened 2026-04-16
fix(core): preserve backward-compatible Loader.Watch interface

pflynn-virtru:
## Summary
- Reverts `Loader.Watch` to its original 3-arg signature `(ctx, cfg, onChange)` to maintain backward compatibility for downstream consumers that implement `Loader` directly
- Introduces an optional `NamespaceAwareLoader` interface for loaders that want namespace/service information during watch, added via `WatchWithNamespaces(ctx, cfg, onChange, namespaces)`
- `Config.WatchWithNamespaces` type-asserts each loader and routes to `WatchWithNamespaces` when the loader opts in, otherwise falls back to plain `Watch`

## Guidance for downstream implementers

If your custom loader does not need namespace/service info, implement `config.Loader` as before — no change required.

If your custom loader wants namespace/service info (e.g. a remote config source scoped to enabled services), implement `config.NamespaceAwareLoader`:

```go
type MyLoader struct{ /* ... */ }

// Required by Loader
func (l *MyLoader) Watch(ctx context.Context, cfg *config.Config, onChange func(context.Context) error) error {
    // fallback path — used when the caller invokes cfg.Watch(ctx) without namespace info
    return nil
}

// Required by NamespaceAwareLoader
func (l *MyLoader) WatchWithNamespaces(
    ctx context.Context,
    cfg *config.Config,
    onChange func(context.Context) error,
    namespaces []config.NamespaceInfo,
) error {
    // preferred path — invoked when the host calls cfg.WatchWithNamespaces(ctx, ns)
    return nil
}
```

Callers that need namespace-aware behavior must invoke `cfg.WatchWithNamespaces(ctx, namespaces)` — `cfg.Watch(ctx)` passes an empty slice and will not deliver namespace info to `NamespaceAwareLoader` implementers.

## Test plan
- [x] `go build ./service/...` passes
- [x] `go test ./service/pkg/config/...` passes
- [ ] CI checks pass

## Summary by CodeRabbit

* **Refactor**
  * Enhanced configuration loading system architecture by separating standard configuration loading from namespace-aware functionality, improving code organization and enabling more efficient configuration management across loader implementations.

### DIFF
--- a/service/pkg/config/config.go
+++ b/service/pkg/config/config.go
@@ -155,8 +155,8 @@ func (c *Config) AddOnConfigChangeHook(hook ChangeHook) {
 	c.onConfigChangeHooks = append(c.onConfigChangeHooks, hook)
 }
 
-// Watch starts watching the configuration for changes in all config loaders.
-func (c *Config) Watch(ctx context.Context) error {
+// WatchWithNamespaces starts watching the configuration for changes in all config loaders, and provides the namespace info to the loaders.
+func (c *Config) WatchWithNamespaces(ctx context.Context, namespaces []NamespaceInfo) error {
 	if len(c.loaders) == 0 {
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


service/pkg/config/config.go
```diff
@@ -176,7 +176,13 @@ func (c *Config) WatchWithNamespaces(ctx context.Context, namespaces []Namespace
...
 			return c.OnChange(ctx)
 		}
-		if err := loader.Watch(ctx, c, onChangeCallback, namespaces); err != nil {
+		if nsLoader, ok := loader.(NamespaceAwareLoader); ok {
+			if err := nsLoader.WatchWithNamespaces(ctx, c, onChangeCallback, namespaces); err != nil {
+				return err
+			}
+			continue
+		}
+		if err := loader.Watch(ctx, c, onChangeCallback); err != nil {
```
ntrevino-virtru, line 185: I think this should be `else if`

ntrevino-virtru (commented):
Should we consider a standard Go pattern with variadic arguments so we do not keep hitting this?

PR #3379, opened 2026-04-23
feat(sdk): HealthCheck(ctx, service)

### LINKED ISSUE #3370: SDK: Expose public method for post-creation platform health/connectivity check
### Background
The Go SDK currently supports platform connectivity validation only at client creation time via `sdk.WithConnectionValidation()`. This calls the internal (unexported) `validateHealthyPlatformConnection` function, which uses the gRPC health check protocol (grpc.health.v1.Health/Check).

### Problem
Consumers like [Data Harbor](https://github.com/virtru-corp/data-harbor) need to validate platform connectivity after SDK creation — specifically for use in Kubernetes readiness probes. Today the only option is to validate during `sdk.New()`, which means a connectivity failure surfaces as a fatal error in `main.go` rather than a recoverable readiness check failure.

### Proposed Changes Options
1) Expose a public `ValidateConnection()` method on SDK — surfaces the same logic currently used by WithConnectionValidation() so it can be called on-demand after client creation.
2) Expose the health client — make the health client (currently internal, created [here](https://github.com/opentdf/platform/blob/cd931d9a46083151d9cd2a2efea4bd447c43f5cc/sdk/sdk.go#L217)) accessible so consumers can use it directly for custom health check flows.

### Use case
A downstream service (e.g. Data Harbor) creates the SDK instance once at startup, then uses the exposed health check in its K8s readiness probe to report whether the platform is reachable — allowing the pod to go not-ready and stop receiving traffic rather than crashing outright.

pflynn-virtru:
Closes #3370.

## Summary

Adds `SDK.HealthCheck(ctx context.Context, service string) (HealthStatus, error)` so downstream services can run readiness probes through the SDK instead of open-coding `http.Client.Get(healthzEndpoint)`.

## Relationship to #3370

Issue #3370 proposed two options: (1) a public `ValidateConnection()` surfacing the logic behind `WithConnectionValidation()`, or (2) exposing the internal health client.

This PR implements option 1 with two intentional enhancements over the issue description:

- **Named `HealthCheck` rather than `ValidateConnection`.** The gRPC Health v1 RPC it wraps is named `Check`, and "health check" is the canonical term across the ecosystem (k8s probes, gRPC tooling, ALBs, envoy). `ValidateConnection` implies a one-shot connectivity validation; the new method can also be pointed at a specific registered service or all services, so the broader name better matches the capability.
- **Per-service `service string` argument and typed `HealthStatus` return.** The platform's health handler already accepts a `service` field and distinguishes SERVING / NOT_SERVING / UNKNOWN. A `ValidateConnection() error` that only returned nil-or-error would collapse the NOT_SERVING and transport-failure cases into one outcome. The richer return lets a k8s probe distinguish "platform down" from "platform up but KAS not ready" without unwrapping error chains.

If reviewers prefer the narrower `ValidateConnection() error` shape from the issue, I'm happy to add a thin wrapper alongside `HealthCheck` — or swap the name entirely — without changing the implementation.

## Motivation

Downstream consumers today probe with:

```go
resp, err := sdk.Conn().Client.Get(healthzEndpoint)
```

This pattern has four problems:

1. **No context** — `http.Client.Get` hardcodes `context.Background()`.
2. **Timeout is client-wide** via `http.Client.Timeout`, so capping the probe also caps every other RPC sharing the client.
3. **Bypasses interceptors** (auth, audit, caller-injected OTEL) — they live on `Conn().Options`, not on the HTTP client.
4. **Collapses three distinguishable states** (unreachable / NOT_SERVING / unknown service) into one error.

The platform's health handler already supports per-service probes via the gRPC Health v1 `service` field. This method exposes that surface with full ctx / interceptor propagation.

## API

```go
type HealthStatus int

const (
    HealthStatusUnknown HealthStatus = iota
    HealthStatusServing
    HealthStatusNotServing
)

const ErrHealthCheckUnsupported = Error("health check not supported in IPC mode")

func (s SDK) HealthCheck(ctx context.Context, service string) (HealthStatus, error)
```

`service` selects scope:
- `""` — reachability check (handler responds).
- `"all"` — iterate every registered readiness probe.
- other — specific named service (e.g. `"kas"`, `"authorization"`).

Uses `s.conn.Options` so caller-injected `otelconnect.NewInterceptor()` (via `WithExtraClientOptions`) fires automatically. **No new SDK dependency on `go.opentelemetry.io/otel`.**

## Readiness-probe example

```go
func (c *client) IsReady(ctx context.Context) bool {
    status, err := c.sdk.HealthCheck(ctx, "")
    if err != nil {
        if errors.Is(err, sdk.ErrHealthCheckUnsupported) {
            return true // IPC mode — nothing to probe
        }
        slog.ErrorContext(ctx, "platform not ready", slog.Any("error", err))
        return false
    }
    return status == sdk.HealthStatusServing
}
```

## Test plan

- [x] `TestHealthStatus_String` — enum String() method including out-of-range default.
- [x] `TestErrHealthCheckUnsupported_Distinct` — sentinel distinct from `ErrPlatformUnreachable`.
- [x] `TestSDK_HealthCheck_IPCMode_ReturnsErrHealthCheckUnsupported`
- [x] `TestSDK_HealthCheck_Unreachable_ReturnsErrPlatformUnreachable`
- [x] `TestSDK_HealthCheck_ContextCanceled_ReturnsQuickly`
- [x] `TestSDK_HealthCheck_Serving_HappyPath` (4 sub-tests via `httptest` + `grpchealth.NewStaticChecker`)
- [x] `go test ./sdk/... -race -count=1` — green
- [x] `golangci-lint run ./sdk/...` — no new issues
- [x] `TestREADMECodeBlocks` — pass

## Notes

- `connectrpc.com/grpchealth v1.4.0` appears as a direct dep in `sdk/go.mod`. It is **test-only in practice** but Go modules cannot express that distinction. The package is already a transitive dep of the service module.
- Streaming `grpc.health.v1.Health/Watch` is **out of scope** — the platform currently returns `Unimplemented` for Watch.
- **No behavior change to `WithConnectionValidation()`** at construction. Its doc comment now cross-references `HealthCheck` for runtime probes.
- All 4 commits are DCO-signed and GPG/SSH-signed.

### DIFF
--- a/sdk/go.mod
+++ b/sdk/go.mod
@@ -6,6 +6,7 @@ toolchain go1.25.8
 
 require (
 	connectrpc.com/connect v1.19.1
+	connectrpc.com/grpchealth v1.4.0
 	github.com/Masterminds/semver/v3 v3.4.0
 	github.com/google/uuid v1.6.0
 	github.com/gowebpki/jcs v1.0.1
--- a/sdk/sdk.go
+++ b/sdk/sdk.go
@@ -443,21 +443,40 @@ func isValidManifest(manifest string, intensity SchemaValidationIntensity) (bool
 	return true, nil
 }
 
-// Test connectability to the platform and validate a healthy status
-func validateHealthyPlatformConnection(platformEndpoint string, httpClient *http.Client, options []connect.ClientOption) error {
+// checkPlatformHealth issues a single gRPC Health v1 Check against the platform health endpoint.
+// ctx controls deadline, cancellation, and trace-context propagation. service selects scope:
+// "" = reachability, "all" = every registered readiness probe, otherwise a specific service name.
+func checkPlatformHealth(
+	ctx context.Context,
+	endpoint, service string,
+	httpClient *http.Client,
+	options []connect.ClientOption,
+) (healthpb.HealthCheckResponse_ServingStatus, error) {
 	healthClient := connect.NewClient[healthpb.HealthCheckRequest, healthpb.HealthCheckResponse](
 		httpClient,
-		platformEndpoint+"/grpc.health.v1.Health/Check",
+		endpoint+"/grpc.health.v1.Health/Check",
 		options...,
 	)
-	res, err := healthClient.CallUnary(
-		context.Background(),
-		connect.NewRequest(&healthpb.HealthCheckRequest{}),
-	)
-	if err != nil || res.Msg.GetStatus() != healthpb.HealthCheckResponse_SERVING {
+	res, err := healthClient.CallUnary(ctx, connect.NewRequest(&healthpb.HealthCheckRequest{Service: service}))
+	if err != nil {
+		return healthpb.HealthCheckResponse_UNKNOWN, err
+	}
+	return res.Msg.GetStatus(), nil
+}
+
+// validateHealthyPlatformConnection is the construction-time reachability gate used by New when
+// WithConnectionValidation is set. It uses context.Background() to preserve today's behavior.
+// Callers pass cfg.extraClientOptions (pre-auth) because the auth and audit interceptors are
+// assembled after this gate fires; the runtime SDK.HealthCheck method uses the post-interceptor
+// s.conn.Options instead.
+func validateHealthyPlatformConnection(platformEndpoint string, httpClient *http.Client, options []connect.ClientOption) error {
+	status, err := checkPlatformHealth(context.Background(), platformEndpoint, "", httpClient, options)
+	if err != nil {
 		return errors.Join(ErrPlatformUnreachable, err)
 	}
-
+	if status != healthpb.HealthCheckResponse_SERVING {
+		return errors.Join(ErrPlatformUnreachable, fmt.Errorf("platform health status %q", status))
+	}
 	return nil
 }
 
--- a/sdk/sdk.go
+++ b/sdk/sdk.go
@@ -503,9 +503,13 @@ func checkPlatformHealth(
 	httpClient *http.Client,
 	options []connect.ClientOption,
 ) (healthpb.HealthCheckResponse_ServingStatus, error) {
+	checkURL, err := url.JoinPath(endpoint, "grpc.health.v1.Health", "Check")
+	if err != nil {
+		return healthpb.HealthCheckResponse_UNKNOWN, err
+	}
 	healthClient := connect.NewClient[healthpb.HealthCheckRequest, healthpb.HealthCheckResponse](
 		httpClient,
-		endpoint+"/grpc.health.v1.Health/Check",
+		checkURL,
 		options...,
 	)
 	res, err := healthClient.CallUnary(ctx, connect.NewRequest(&healthpb.HealthCheckRequest{Service: service}))
--- a/sdk/sdk.go
+++ b/sdk/sdk.go
@@ -32,9 +32,12 @@ import (
 const (
 	// Failure while connecting to a service.
 	// Check your configuration and/or retry.
-	ErrGrpcDialFailed                = Error("failed to dial grpc endpoint")
-	ErrShutdownFailed                = Error("failed to shutdown sdk")
-	ErrPlatformUnreachable           = Error("platform unreachable or not responding")
+	ErrGrpcDialFailed      = Error("failed to dial grpc endpoint")
+	ErrShutdownFailed      = Error("failed to shutdown sdk")
+	ErrPlatformUnreachable = Error("platform unreachable or not responding")
+	// ErrHealthCheckUnsupported is returned by SDK.HealthCheck when the SDK is configured
+	// in IPC mode, which does not support the gRPC Health protocol.
+	ErrHealthCheckUnsupported        = Error("health check not supported in IPC mode")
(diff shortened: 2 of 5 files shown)


pflynn-virtru:
Thanks for the bot reviews. Responses to each finding:

### ✅ Gemini — `url.JoinPath` (adopted)

Fixed in `dcbf1c7e`. Also adds `TestSDK_HealthCheck_TrailingSlashEndpoint` as a regression guard — constructs an SDK with a trailing-slash endpoint and verifies the health check still resolves SERVING against the `httptest` server. Side benefit: the construction-time `validateHealthyPlatformConnection` path inherits the same fix since both callers share the `checkPlatformHealth` helper from the Task 2 refactor.

### ❌ CodeRabbit — fixed-port flakiness (declining)

`badPlatformEndpoint = "http://localhost:9999"` is a pre-existing package-level constant already used by `Test_New_ShouldFailWithDisconnectedPlatform` (`sdk/sdk_test.go:205`). Fixing only the new tests would leave the file with two conventions for the same concept.

The proposed `httptest.NewServer(...); Close()` pattern has its own TOCTOU: between `Close()` releasing the ephemeral port and the test's `HealthCheck` call, another process can bind to that port. In practice the window is microseconds, but it's not strictly more reliable than a stable `:9999` that nothing's bound to on CI.

If the maintainers want to eliminate the pattern entirely, happy to do it as a separate `test(sdk):` cleanup PR that updates every call site — not in this feature PR.

### ❌ CodeRabbit — docstring coverage 42.86% (false positive for this PR)

Every symbol this PR adds has a doc comment:

- `HealthStatus`, `HealthStatusUnknown`, `HealthStatusServing`, `HealthStatusNotServing`, `String()` ✓
- `HealthCheck`, `ErrHealthCheckUnsupported` ✓
- `checkPlatformHealth` (unexported), `validateHealthyPlatformConnection` ✓

The coverage deficit is pre-existing in `sdk/sdk.go` (e.g., the `Err*` constant block, `GetTdfType`, `IsValidTdf`, etc.). Addressing it is a repo-wide docs pass that shouldn't ride on a feature PR.

sdk/sdk.go
```diff
@@ -310,6 +340,57 @@ func (s SDK) Conn() *ConnectRPCConnection {
...
+// and returns the reported serving status.
+//
+// The service argument selects the check scope:
+//
+//   - ""     — reachability check. Returns SERVING if the health handler responds. Does
+//     not exercise any registered readiness probe.
+//   - "all"  — runs every readiness check the platform has registered. Returns
+//     NOT_SERVING if any registered service reports an error.
+//   - other  — runs only the named service's readiness check (e.g. "kas", "authorization").
+//     Returns HealthStatusUnknown if the service is not registered on this deployment.
```
marythought, line 353: The valid values for service ("", "all", "kas", "authorization", etc.) are documented here but there are no constants for them. Consider exported constants like HealthServiceAll = "all" (even just for "all" and "") for discoverability since this is a public SDK API.

(Or at minimum, a note in the README/docs listing the known service names)

sdk/sdk.go
```diff
@@ -32,9 +32,12 @@ import (
 const (
 	// Failure while connecting to a service.
 	// Check your configuration and/or retry.
-	ErrGrpcDialFailed                = Error("failed to dial grpc endpoint")
-	ErrShutdownFailed                = Error("failed to shutdown sdk")
-	ErrPlatformUnreachable           = Error("platform unreachable or not responding")
+	ErrGrpcDialFailed      = Error("failed to dial grpc endpoint")
```
marythought, line 35: nit: alignment inconsistency

sdk/go.mod
```diff
@@ -6,6 +6,7 @@ toolchain go1.25.8
 
 require (
 	connectrpc.com/connect v1.19.1
+	connectrpc.com/grpchealth v1.4.0
```
marythought, line 9: nit (non-blocking): You already flagged this in the PR description — `grpchealth` is test-only but Go modules can't express that, so this promotes it to a direct dep in sdk/go.mod for all downstream consumers.

Can we move newHealthTestServer into `sdk/internal/healthtest/` with its own go.mod? That keeps grpchealth out of the SDK's public dependency surface. Fine as a follow-up commit on this PR or a separate PR — no need to block on it.

marythought (approved):
This is a great, well-executed and timely feature! Some minor suggestions.

sdk/sdk.go
strantalis: @pflynn-virtru Any reason not to just simplify this to `IsHealthy(ctx context.Context)` or something like `IsReachable(ctx context.Context)`. If we don't want to simplify this we need expose const like @marythought proposed above imo.

marythought:
Hey @pflynn-virtru — taking over this one to incorporate @strantalis' review feedback in #3412. Your 5 commits are preserved on the new branch (`DSPX-3008-sdk-healthcheck`) with attribution; my refactor commit on top simplifies `HealthCheck(ctx, service)` → `IsHealthy(ctx) (bool, error)` per Sean's nudge for an idiomatic Go predicate.

I'll close this PR once #3412 is approved.

Thanks for the solid groundwork — `url.JoinPath` fix and the test scaffolding came along for the ride.

PR #3389, opened 2026-04-23
chore(ci): document why protocol/go is excluded from partial go.work

marythought:
## Summary

- Adds comments to the `sdk`, `service`, and `otdfctl` cases in `.github/scripts/work-init.sh` explaining why `protocol/go` is intentionally excluded from the partial workspace

## Context

The partial `go.work` generated by `work-init.sh` deliberately omits upstream deps like `protocol/go`. This ensures they must be released and tagged before downstream components (sdk, service, otdfctl) can be released — using local source would mask unreleased `protocol/go` changes and allow tagging a component against an unpublished version.

This wasn't documented in the script, which led to confusion (including the original version of this PR, which tried to add `protocol/go` to the workspace). Adding comments so future contributors and AI tools don't repeat the mistake.

## Test plan

- [x] `shellcheck .github/scripts/work-init.sh` passes
- [ ] CI passes on this PR

## Summary by CodeRabbit

* **Chores**
  * Added clarifying documentation to internal initialization scripts regarding dependency handling.

### DIFF
--- a/.github/scripts/work-init.sh
+++ b/.github/scripts/work-init.sh
@@ -45,12 +45,14 @@ sdk)
   rm -f go.work go.work.sum &&
     go work init &&
     go work use ./sdk &&
+    go work use ./protocol/go &&
     go work use ./service &&
     go work use ./examples
   ;;
 service)
   rm -f go.work go.work.sum &&
     go work init &&
+    go work use ./protocol/go &&
     go work use ./service &&
     go work use ./examples
   ;;
@@ -63,6 +65,7 @@ otdfctl)
   rm -f go.work go.work.sum &&
     go work init &&
     go work use ./otdfctl &&
+    go work use ./protocol/go &&
     # service and examples are needed for release branch checks
     go work use ./service &&
     go work use ./examples
--- a/.github/scripts/work-init.sh
+++ b/.github/scripts/work-init.sh
@@ -42,17 +42,20 @@ lib/ocrypto | lib/fixtures | lib/flattening | lib/identifier | protocol/go)
   echo "[INFO] skipping for leaf package"
   ;;
 sdk)
+  # NOTE: protocol/go is intentionally excluded. Upstream deps like protocol/go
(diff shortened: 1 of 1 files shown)


.github/scripts/work-init.sh
```diff
@@ -45,12 +45,14 @@ sdk)
   rm -f go.work go.work.sum &&
     go work init &&
     go work use ./sdk &&
+    go work use ./protocol/go &&
```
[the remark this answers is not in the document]
marythought: @dmihalcik-virtru @elizabethhealy should I add these others? Seems harmless to add them but just checking in case they are excluded for a reason

.github/scripts/work-init.sh
```diff
@@ -45,12 +45,14 @@ sdk)
...
     go work use ./service &&
```
c-r33d, line 52: Is service still needed here? As I understand the use of this script, it is to ensure that a release is not depending on unreleased changes, but SDK does not depend on service.

c-r33d: Believe it is since the calling workflows will start the service.

c-r33d (dismissed):

.github/scripts/work-init.sh
marythought: I'll merge this and look into that as followup (as well as the other two open questions above)
edit: jk looks like I need 2!

dmihalcik-virtru (changes requested):
No, the point of this code is to explicitly require all 'upstream' dependencies to be released first. With this change, you will be able to tag `sdk` with a version of `protocol/go` that does not contain required features

marythought:
> No, the point of this code is to explicitly require all 'upstream' dependencies to be released first. With this change, you will be able to tag `sdk` with a version of `protocol/go` that does not contain required features

Gotcha -- ok, I'll update this PR to add a note that this is purposeful so we/AI doesn't get tripped up about it in the future!

dmihalcik-virtru (approved):

PR #3398, opened 2026-04-24
feat(policy): hydrate resource_mappings on ListAttributes values

KevLehman:
## Summary

- `policy.attributes.AttributesService/ListAttributes` now returns `resource_mappings` on every `Value`, matching the shape already produced by `GetAttributeValuesByFqns` and already advertised in the generated OpenAPI spec.
- The SQL change extends `listAttributesDetail` with a pre-aggregated `value_resource_mappings` CTE (same pattern used by `listAttributesByDefOrValueFqns`), so mappings are hydrated without introducing row-explosion through the main `GROUP BY`.
- No proto change. No new request toggle. No behavior change for callers that ignore the new field.

### Before / after

Before (trimmed):

```json
{
  "id": "…",
  "value": "topsecret",
  "fqn": "https://demo.com/attr/classification/value/topsecret",
  "active": true
}
```

After:

```json
{
  "id": "…",
  "value": "topsecret",
  "fqn": "https://demo.com/attr/classification/value/topsecret",
  "active": true,
  "resourceMappings": [
    { "id": "…", "terms": ["ts", "top-secret", "top_secret"] }
  ]
}
```

Reproduction:

```bash
grpcurl -plaintext -d '{"pagination":{"limit":500}}' \
  localhost:8080 policy.attributes.AttributesService/ListAttributes
```

### Why

Clients that need label/term data (UI pickers, classifiers, policy editors) currently have to enumerate FQNs via `ListAttributes` and then batch‑hydrate them through `GetAttributeValuesByFqns` (250 FQN cap per call). The OpenAPI spec already documents `resourceMappings` on `ListAttributes`, so the current server response is inconsistent with the schema clients read. Hydrating directly collapses the two round trips into one.

### Implementation notes

- `listAttributesDetail` gets a new CTE (`value_resource_mappings`) keyed by `attribute_value_id`, producing a `JSON_AGG` of `{id, terms, group}` per value — exactly the shape `listAttributesByDefOrValueFqns` emits.
- The CTE is joined once against the flattened `avt` subquery by `av.id`. Row count per attribute definition is unchanged (one row per value), so `JSON_AGG … FILTER (WHERE avt.id IS NOT NULL)` and the `GROUP BY ad.id, n.name, fqns.fqn` + `COUNT(*) OVER()` pagination continue to behave the same.
- `Value` already declares `repeated ResourceMapping resource_mappings = 10;`, so the Go DAO (`attributesValuesProtojson` → `protojson.Unmarshal`) picks up the new field automatically. No scanning code changed.
- `GetAttribute` is intentionally out of scope — it uses the separate `getAttribute` query and was left unchanged.

### Out of scope

- `subject_mappings` / `kas_keys` / `grants` / `obligations` / value-level `metadata` hydration on `ListAttributes`. Each is its own JOIN and its own tradeoff, tracked as follow-ups.

## Test plan

- [x] `go build ./service/...`
- [x] `go vet ./service/policy/... ./service/integration/...`
- [x] `golangci-lint run -c .golangci.yaml ./policy/db/...` — 0 issues
- [x] `go test ./service/policy/...` — 761 passed
- [x] `go test -run '^TestAttributesSuite$' ./service/integration/...` — 92 passed (includes new `Test_ListAttributes_HydratesResourceMappings`)
- [x] `go test -run '^TestAttributeFqnSuite$' ./service/integration/...` — 38 passed
- [x] `go test -run '^TestResourceMappingsSuite$|^TestAttributeValuesSuite$|^TestNamespacesSuite$' ./service/integration/...` — 161 passed

## Summary by CodeRabbit

* **New Features**
  * Attribute listings now include resource mapping information for each attribute value.

* **Tests**
  * Added integration test verifying resource mappings are properly populated for attribute values, including both grouped and ungrouped mappings.

### DIFF
--- a/service/integration/attributes_test.go
+++ b/service/integration/attributes_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/opentdf/platform/protocol/go/policy/attributes"
 	"github.com/opentdf/platform/protocol/go/policy/kasregistry"
 	"github.com/opentdf/platform/protocol/go/policy/namespaces"
+	"github.com/opentdf/platform/protocol/go/policy/resourcemapping"
 	"github.com/opentdf/platform/protocol/go/policy/unsafe"
 	"github.com/opentdf/platform/service/internal/fixtures"
 	"github.com/opentdf/platform/service/pkg/db"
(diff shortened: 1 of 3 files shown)


jakedoublev:
@KevLehman Thanks so much for contributing! We discussed as a team and, while this may be a useful thing to have back from `ListAttributes`, we want to avoid additional response overhead since this RPC is on a lot of hot paths. We are also considering either more granular APIs for specific use cases or possible selectors/expanders, which both have their own tradeoffs. All of that said, we can't quite move forward with this at the moment, but we appreciate the idea and will definitely keep it in mind.

KevLehman:
Hey, thanks for checking.

Would you be open to do this change on the API only? when calling v2.ListAttributes?

That way, we would have the information needed without having to do a 2nd (+n) paginated call to ListResourceMappings and map locally

marythought:
Hi @KevLehman — thanks again for the PR!

Of the use cases you list (UI pickers, classifiers, policy editors), the classifier path is exactly what the Virtru Data Security Platform Tagging Service is built for. Instead of clients pulling label/term data and resolving locally, you send content (or already-extracted tags) to the Tagging Service and get back normalized attribute FQNs — synonym dictionary and resource mappings stay server-side, no client enumeration needed.

Hand-rolled in client:
1. Enumerate via `ListAttributes` / `GetAttributeValuesByFqns` (and pay the 250-FQN cap)
2. Resolve label → FQN locally
3. Call `GetDecisions` with FQNs

With the Tagging Service:
1. Send content (or tags) → Tagging Service returns FQNs
2. Call `GetDecisions` with FQNs

One thing worth calling out: keeping that logic server-side means changes to the synonym dictionary, resource mappings, or extraction rules land as Tagging Service config inside DSP — no client redeploy, no SDK upgrade, and no version skew between what your client thinks the labels are and what the policy says they are.

We're also actively working on an SDK helper that wraps `(identity + content) → allowed/denied` into a single call — tag normalization and decision in one step. Happy to share more once it's a bit further along!

Glad to walk through the Tagging Service fit if useful.

 -- Mary

PR #3402, opened 2026-04-27
refactor(policy): decouple sort field/direction mapping and add SQL-level sort defaults

dsm20:
refactor(policy): decouple sort field/direction mapping and add SQL-level sort defaults

## Proposed Changes

Fixes a silent bug where providing a sort direction with an UNSPECIFIED field caused the direction to be discarded. Refactors the Go sort mappers to be pure enum-to-string translators, moves all default logic into SQL via CTEs, and documents all default behavior at the proto level.

## Bug Fixed

Get*SortParams previously coupled field and direction mapping. When the sort field was UNSPECIFIED, both field AND direction were returned as empty strings (return "", ""), silently discarding any user-provided direction. A user sending field=UNSPECIFIED, direction=ASC would get created_at DESC instead of created_at ASC.

## Sort Defaults

All defaults are now resolved in SQL via a CTE:
- **field UNSPECIFIED** defaults to created_at
- **direction UNSPECIFIED** defaults to DESC
- **both UNSPECIFIED or sort omitted** defaults to created_at DESC
- **id ASC** tiebreaker on all queries for deterministic ordering

### Go Sort Mappers -- service/policy/db/utils.go

- getSortDirection: UNSPECIFIED now returns "" (was "ASC"), with explicit fallthrough for readability
- Added 8 get*SortField helpers: pure enum-to-string translators, one per resource, UNSPECIFIED returns ""
- Refactored 8 Get*SortParams: delegate to get*SortField and getSortDirection independently. Removed sort[0] == nil guard (proto never produces nil in repeated fields). Signature unchanged, call sites unchanged.

### SQL Queries -- service/policy/db/queries/*.sql

Added CTE-based sort default resolution to all 9 list queries (8 RPCs):

```sql
WITH params AS (
    SELECT
        COALESCE(NULLIF(@sort_field::text, ''), 'created_at') AS resolved_field,
        COALESCE(NULLIF(@sort_direction::text, ''), 'DESC') AS resolved_direction
)
```

Also added id ASC tiebreaker and p.resolved_field, p.resolved_direction to GROUP BY where needed.

Affected queries: listNamespaces, listAttributesDetail, listAttributesSummary, listObligations, listRegisteredResources, listKeyAccessServers, listKeys, listSubjectConditionSets, listSubjectMappings

### Proto Docs -- all List*Request sort field comments

Replaced vague "the service applies its default ordering" with explicit defaults:
```
// Sort defaults:
//   - direction UNSPECIFIED defaults to DESC for the specified field
//   - field UNSPECIFIED defaults to created_at with the specified direction
//   - both UNSPECIFIED or sort omitted defaults to created_at DESC
```

### Unit Tests -- service/policy/db/utils_test.go

Updated all 8 test functions:
- Removed nil element test cases (proto handles this)
- Added "UNSPECIFIED field with ASC/DESC preserves direction" test cases
- Added "both UNSPECIFIED returns empty strings" test cases
- Updated "unspecified direction" expectations from "ASC" to ""

### Integration Tests -- service/integration/*_test.go

Replaced 8 FallsBackToDefault tests with 24 targeted default tests (3 per RPC):
- SortByUnspecifiedField_DefaultsToCreatedAt: field UNSPECIFIED + explicit ASC, expects created_at ASC
- SortByUnspecifiedDirection_DefaultsToDESC: explicit created_at + direction UNSPECIFIED, expects created_at DESC
- SortByBothUnspecified_DefaultsToCreatedAtDESC: both UNSPECIFIED, expects created_at DESC

### Checklist

### Testing Instructions

Unit tests: go test ./service/policy/db/... -run "Sort"

## Release Notes

* **Improvements**
  * Standardized list operation sorting behavior across all endpoints: sort defaults to `created_at DESC` when omitted, `created_at` with specified direction when field is unspecified, and `DESC` with specified field when direction is unspecified.
  * Enhanced validation: sort field and direction parameters now strictly enforce defined enum values.

* **Documentation**
  * Updated API documentation to clarify consistent sort defaults and behavior across all list operations.


service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -52,16 +52,16 @@ enum SortKeyAccessServersType {
...
 }
 
 message ListKeyAccessServersRequest {
   // Optional
   policy.PageRequest pagination = 10;
   // Optional - CONSTRAINT: max 1 item
-  // Default ordering when omitted or when the first entry's field is UNSPECIFIED:
-  // created_at DESC, then id ASC (tie-breaker).
+  // When sort is provided and sort.direction is UNSPECIFIED, direction defaults to ASC.
+  // When sort is omitted or sort.field is UNSPECIFIED, the service applies its default ordering.
```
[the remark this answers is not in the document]
dsm20: @gemini-code-assist Query behavior should not be fully documented at the proto level, as it could change and the proto docs should not be tethered to that degree. Here we document the unspecified sort call, and then defer the actual query level behavior for no sort + unspecified to the service level. This is a happy middle ground without over-documenting at the proto level.

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -495,8 +495,8 @@ message ListKeysRequest {
   policy.PageRequest pagination = 10; // Pagination request for the list of keys
 
   // Optional - CONSTRAINT: max 1 item
-  // Default ordering when omitted or when the first entry's field is UNSPECIFIED:
-  // created_at DESC.
+  // When sort is provided and sort.direction is UNSPECIFIED, direction defaults to ASC.
+  // When sort is omitted or sort.field is UNSPECIFIED, the service applies its default ordering.
```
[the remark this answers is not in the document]
dsm20: Query behavior should not be fully documented at the proto level, as it could change and the proto docs should not be tethered to that degree. Here we document the unspecified sort call, and then defer the actual query level behavior for no sort + unspecified to the service level. This is a happy middle ground without over-documenting at the proto level.

service/policy/obligations/obligations.proto
```diff
@@ -166,8 +166,8 @@ message ListObligationsRequest {
     policy.PageRequest pagination = 10;
 
     // Optional - CONSTRAINT: max 1 item
-    // Default ordering when omitted or when the first entry's field is UNSPECIFIED:
-    // created_at DESC, then id ASC (tie-breaker).
+    // When sort is provided and sort.direction is UNSPECIFIED, direction defaults to ASC.
+    // When sort is omitted or sort.field is UNSPECIFIED, the service applies its default ordering.
```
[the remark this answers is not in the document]
dsm20: Query behavior should not be fully documented at the proto level, as it could change and the proto docs should not be tethered to that degree. Here we document the unspecified sort call, and then defer the actual query level behavior for no sort + unspecified to the service level. This is a happy middle ground without over-documenting at the proto level.

service/policy/registeredresources/registered_resources.proto
```diff
@@ -120,8 +120,8 @@ message ListRegisteredResourcesRequest {
   policy.PageRequest pagination = 10;
 
   // Optional - CONSTRAINT: max 1 item
-  // Default ordering when omitted or when the first entry's field is UNSPECIFIED:
-  // created_at DESC.
+  // When sort is provided and sort.direction is UNSPECIFIED, direction defaults to ASC.
+  // When sort is omitted or sort.field is UNSPECIFIED, the service applies its default ordering.
```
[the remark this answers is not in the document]
dsm20: Query behavior should not be fully documented at the proto level, as it could change and the proto docs should not be tethered to that degree. Here we document the unspecified sort call, and then defer the actual query level behavior for no sort + unspecified to the service level. This is a happy middle ground without over-documenting at the proto level.

service/policy/subjectmapping/subject_mapping.proto
```diff
@@ -186,8 +188,8 @@ message ListSubjectConditionSetsRequest {
   // Optional
   policy.PageRequest pagination = 10;
   // Optional - CONSTRAINT: max 1 item
-  // Default ordering when omitted or when the first entry's field is UNSPECIFIED:
-  // created_at DESC, then id ASC (tie-breaker).
+  // When sort is provided and sort.direction is UNSPECIFIED, direction defaults to ASC.
+  // When sort is omitted or sort.field is UNSPECIFIED, the service applies its default ordering.
```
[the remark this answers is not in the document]
dsm20: Query behavior should not be fully documented at the proto level, as it could change and the proto docs should not be tethered to that degree. Here we document the unspecified sort call, and then defer the actual query level behavior for no sort + unspecified to the service level. This is a happy middle ground without over-documenting at the proto level.

docs/grpc/index.html
```diff
@@ -7503,7 +7503,7 @@ <h3 id="policy.PageResponse">PageResponse</h3>
 
       
         <h3 id="policy.SortDirection">SortDirection</h3>
-        <p>Sorting direction shared across list APIs.</p><p>Notes:</p><p>- When a sort field is provided, UNSPECIFIED is treated as ASC.</p><p>- When the 'sort' field is omitted or the chosen sort 'field' is UNSPECIFIED,</p><p>the endpoint's request message defines the default ordering; see the</p><p>specific List* request docs.</p>
+        <p>Sorting direction shared across list APIs.</p><p>When the 'sort' field is omitted or the chosen sort 'field' is UNSPECIFIED,</p><p>the endpoint's request message defines the default ordering; see the</p><p>specific List* request docs.</p>
```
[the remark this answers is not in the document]
dsm20: heres a carrot 🥕

service/policy/attributes/attributes.proto
```diff
@@ -86,6 +86,8 @@ message ListAttributesRequest {
   // Optional
   policy.PageRequest pagination = 10;
   // Optional - CONSTRAINT: max 1 item
+  // When sort is provided and sort.direction is UNSPECIFIED, direction defaults to ASC.
```
c-r33d: These are fine, but I would just leave it as the default sort order is DESC `created_at`. Looking at the SQL these should default to the former. Is that true?

service/policy/attributes/attributes.proto
```diff
@@ -86,6 +86,10 @@ message ListAttributesRequest {
   // Optional
   policy.PageRequest pagination = 10;
   // Optional - CONSTRAINT: max 1 item
+  // Sort defaults:
+  //   - direction UNSPECIFIED defaults to ASC for the specified field
```
c-r33d: Why ASC? I think you could loose the both case if you just always default to DESC.

service/policy/db/queries/attributes.sql
```diff
@@ -3,6 +3,14 @@
 ----------------------------------------------------------------
 
 -- name: listAttributesDetail :many
+WITH params AS (
+    SELECT
+        COALESCE(NULLIF(@sort_field::text, ''), 'created_at') AS resolved_field,
+        CASE
```
c-r33d: I would just default this to `DESC` always

service/policy/attributes/attributes.proto
dsm20: [ffda3f8](https://github.com/opentdf/platform/pull/3402/commits/ffda3f8bc426694f9ce0f72bcbc7942717add701) updated

service/policy/db/queries/attributes.sql
dsm20: [ffda3f8](https://github.com/opentdf/platform/pull/3402/commits/ffda3f8bc426694f9ce0f72bcbc7942717add701) updated

docs/openapi/policy/attributes/attributes.openapi.yaml
```diff
@@ -759,11 +759,9 @@ components:
...
       description: |-
         Sorting direction shared across list APIs.
-         Notes:
-         - When a sort field is provided, UNSPECIFIED is treated as ASC.
-         - When the 'sort' field is omitted or the chosen sort 'field' is UNSPECIFIED,
-           the endpoint's request message defines the default ordering; see the
-           specific List* request docs.
+         When the 'sort' field is omitted or the chosen sort 'field' is UNSPECIFIED,
+         the endpoint's request message defines the default ordering; see the
+         specific List* request docs.
```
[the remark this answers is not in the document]
dsm20: the default ordering is placed in protos on purpose, end to end behavior is documented here now

service/integration/attributes_test.go
```diff
@@ -645,7 +645,41 @@ func (s *AttributesSuite) Test_ListAttributes_SortByUnspecifiedField_FallsBackTo
...
+}
+
+func (s *AttributesSuite) Test_ListAttributes_SortByBothUnspecified_DefaultsToCreatedAtDESC() {
+	nsID := s.createSortTestNamespace("sort-both-unspecified")
+	ids := s.createSortTestAttributes(nsID, "both-unspecified-attr", 3)
+
+	listRsp, err := s.db.PolicyClient.ListAttributes(s.ctx, &attributes.ListAttributesRequest{
+		Namespace: nsID,
+		Sort: []*attributes.AttributesSort{
+			{Field: attributes.SortAttributesType_SORT_ATTRIBUTES_TYPE_UNSPECIFIED, Direction: policy.SortDirection_SORT_DIRECTION_UNSPECIFIED},
```
c-r33d, line 676: I would modify this test or add another where no sort field is present at all, just to make sure we aren't directly accessing any Sort field parameters, which could lead to a panic.

c-r33d (commented):
Looks good overall, I would add a couple nil checks around db utils methods when calling `sort[0].Get...`. While the Get methods are nil safe, it's probably good to have a small test where the Sort[] has one nil object.

service/integration/attributes_test.go
dsm20: added new tests to cover this

service/policy/db/utils.go
```diff
@@ -41,125 +41,125 @@ func getListLimit(limit int32, fallback int32) int32 {
...
-func GetSubjectConditionSetsSortParams(sort []*subjectmapping.SubjectConditionSetsSort) (string, string) {
-	if len(sort) == 0 || sort[0] == nil {
+// GetNamespacesSortParams resolves sort field and direction independently,
+// returning SQL-compatible strings. Empty strings delegate defaults to SQL.
+func GetNamespacesSortParams(sort []*namespaces.NamespacesSort) (string, string) {
+	if len(sort) == 0 {
 		return "", ""
 	}
-	s := sort[0]
+	return getNamespacesSortField(sort[0].GetField()), getSortDirection(sort[0].GetDirection())
```
elizabethhealy, line 84: related to chris's comment, are we guaranteed that sort[0] wont be a nil pointer? if not may be good to add a nil check here

service/policy/db/utils.go
```diff
@@ -41,125 +41,125 @@ func getListLimit(limit int32, fallback int32) int32 {
...
-func GetObligationsSortParams(sort []*obligations.ObligationsSort) (string, string) {
-	if len(sort) == 0 || sort[0] == nil {
+// GetSubjectConditionSetsSortParams resolves sort field and direction independently,
+// returning SQL-compatible strings. Empty strings delegate defaults to SQL.
+func GetSubjectConditionSetsSortParams(sort []*subjectmapping.SubjectConditionSetsSort) (string, string) {
+	if len(sort) == 0 {
 		return "", ""
 	}
-	s := sort[0]
+	return getSubjectConditionSetsSortField(sort[0].GetField()), getSortDirection(sort[0].GetDirection())
```
elizabethhealy, line 108: same here

elizabethhealy (approved):

service/policy/db/utils.go
elizabethhealy: edit: theres a test for GetNamespaceSort that passes in a sort array with a nil item, it does not panic

PR #3411, opened 2026-04-29
feat(cli): Add prune planner.

c-r33d:
## Summary

  Add a new prune planner for namespaced policy migration cleanup.

  This branch introduces prune planning for:

  - actions
  - subject condition sets
  - subject mappings
  - registered resources
  - obligation triggers

  It also exposes resolved migration state from the existing planner so prune logic can reuse it where appropriate.

  ## Details

  - Added new prune plan models and structured prune status/reason output.
  - Added a new PrunePlanner with scope-specific behavior:
      - actions / subject condition sets use direct prune planning based on:
          - current legacy usage
          - canonical migrated target lookup
      - subject mappings / registered resources / obligation triggers use resolved migration state from the planner
  - Added RR-specific prune verification using both:
      - resolved source view
      - authoritative full source reloaded from the global namespace
  - Added comprehensive prune planner test coverage across statuses:
      - delete
      - blocked
      - unresolved

  ## Supporting refactors

  - Split Planner.Plan() so resolved targets can be reused via resolve()
  - Moved shared helper logic into plan_utils.go
  - Deduplicated isStandardAction
  - Cleaned up prune naming around:
      - PruneStatusReason
      - PruneStatusReasonType
      - TargetRef

## Summary by CodeRabbit

* **New Features**
  * Added prune planning with detailed status classification (Delete/Blocked/Unresolved), reasons, and options for interactive review and page sizing to support policy migration workflows.

* **Refactor**
  * Reorganized migration planning flow to separate resolution and finalization for clearer, more robust planning.

* **Tests**
  * Added a comprehensive test suite covering prune classification, blocking/unresolved logic, and edge cases.

### DIFF
--- a/otdfctl/migrations/namespacedpolicy/prune_planner.go
+++ b/otdfctl/migrations/namespacedpolicy/prune_planner.go
@@ -619,6 +619,13 @@ func matchedSubjectConditionSetTargets(source *policy.SubjectConditionSet, targe
 	return targets, foundCanonical, labelsMatch, nil
 }
 
+// For actions and subject condition sets, prune runs after legacy policy graph
+// objects are expected to be gone, so the planner can no longer reliably infer
+// which target namespace a source object was intended to migrate into. In that
+// state, a canonical match is only operator context. Delete requires that no
+// legacy object is still using the source and that at least one canonical match
+// carries the expected migrated_from label; additional canonical matches may
+// still appear in the returned targets without that label.
 func pruneStatusForCanonicalTargets(used, foundCanonical, labelsMatch bool, targets []TargetRef) (PruneStatus, PruneStatusReason, []TargetRef) {
 	if used {
 		return PruneStatusBlocked, newPruneReason(PruneStatusReasonTypeInUse, pruneStatusReasonMessageInUse), targets
--- a/otdfctl/migrations/namespacedpolicy/prune_planner.go
+++ b/otdfctl/migrations/namespacedpolicy/prune_planner.go
@@ -13,7 +13,9 @@ import (
 const (
 	pruneStatusReasonMessageMigratedTargetNotFound              = "migrated target not found"
 	pruneStatusReasonMessageInUse                               = "in-use"
-	pruneStatusReasonMessageMigrationLabelNotFound              = "migrated target missing migration label"
+	pruneStatusReasonMessageNoMatchingLabelsFound               = "no canonical migrated target had a matching migration label"
+	pruneStatusReasonMessageMismatchedMigrationLabel            = "migrated target has mismatched migration label"
+	pruneStatusReasonMessageMissingMigrationLabel               = "migrated target missing migration label"
 	pruneStatusReasonMessageNeedsMigration                      = "needs-migration"
 	pruneStatusReasonMessageRegisteredResourceSourceMismatchFmt = "source registered resource contains values outside resolved migration view for target namespace %q; manual review required before source deletion"
 )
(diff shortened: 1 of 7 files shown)


otdfctl/migrations/namespacedpolicy/prune_planner.go
```diff
@@ -0,0 +1,700 @@
...
+
+func singleMigratedTarget(existingID string, namespace *policy.Namespace) TargetRef {
+	if existingID == "" {
+		return TargetRef{}
+	}
+	return TargetRef{
+		ID:           existingID,
+		NamespaceID:  namespace.GetId(),
+		NamespaceFQN: namespace.GetFqn(),
+	}
```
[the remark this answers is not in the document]
c-r33d: protobuf generated methods are nil safe.

otdfctl/migrations/namespacedpolicy/prune_planner.go
```diff
@@ -0,0 +1,700 @@
...
+			targets = append(targets, singleMigratedTarget(target.GetId(), namespace))
+			if migratedFromID(target) == source.GetId() {
+				labelsMatch = true
+			}
+			break
+		}
+	}
+
+	return targets, foundCanonical, labelsMatch, nil
+}
```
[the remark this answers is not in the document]
c-r33d: I addressed this in a comment. Basically, the issue with actions / scs is that they are expected to be deleted last. Because of that we can't derive/infer a plan on where the action/scs should have gone. So we use the fact that at least 1 action canonically matches, has a migrated_from=source, and is not being used by any other legacy object as enough proof to prune.

otdfctl/migrations/namespacedpolicy/prune_planner.go
```diff
@@ -0,0 +1,707 @@
...
+	"strings"
+
+	"github.com/opentdf/platform/protocol/go/common"
+	"github.com/opentdf/platform/protocol/go/policy"
+)
+
+const (
+	pruneStatusReasonMessageMigratedTargetNotFound              = "migrated target not found"
+	pruneStatusReasonMessageInUse                               = "in-use"
+	pruneStatusReasonMessageMigrationLabelNotFound              = "migrated target missing migration label"
```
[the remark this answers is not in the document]
c-r33d: Fixed this where actions / scs return a NoMatchingLabels found and OT/SM/RRs return either MissingLabel or MismatchLabel

otdfctl/migrations/namespacedpolicy/prune_planner.go
```diff
@@ -0,0 +1,721 @@
...
+	if config.pageSize <= 0 {
+		config.pageSize = defaultPlannerPageSize
+	}
+
+	plannerOpts := []Option{WithPageSize(config.pageSize)}
+	if config.reviewer != nil {
+		plannerOpts = append(plannerOpts, WithInteractiveReviewer(config.reviewer))
+	}
+
+	planner, err := NewPlanner(handler, scopeCSV, plannerOpts...)
```
elizabethhealy, line 102: we already parse scopecsv above in this func, then it gets parsed again in NewPlanner, is it possible to avoid the double parse? or is it low enough comp that doing it twice doesn't matter too much

otdfctl/migrations/namespacedpolicy/prune_planner.go
```diff
@@ -0,0 +1,721 @@
...
+
+	resolved, err := p.planner.resolve(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if resolved == nil {
+		return nil, ErrNilResolvedTargets
+	}
+
+	sourceRegisteredResources, err := p.sourceRegisteredResources(ctx)
```
elizabethhealy, line 154: nit: this ends retrieving the registered resources twice between resolve and sourceRegisteredResources

elizabethhealy (approved):

otdfctl/migrations/namespacedpolicy/prune_planner.go
c-r33d: We could probably design around this, but the `resolve` method returns the state of the source object's -- whether or not they should be migrated. Those source objects, for RRs, can be modified during an interactive review (i.e. A namespace collision that a user chooses to resolve). So we fetch the full candidates again to ensure to catch an edge case where the user could have partially migrated an RR. I could modify the RR to include a fullSource and filteredSource like I do with the prune, or return a copy of the slice before modification, but I don't think efficiency is a huge deal here. I'd rather stick with readability / caution here.

otdfctl/migrations/namespacedpolicy/prune_planner.go
c-r33d: Yeah we could pass it in as an option, but for pruning we only allow one scope at a time, so the comp is negligible. I will look at fixing this in a followup PR.

PR #3460, opened 2026-05-11
feat(policy): add GetRegisteredResources batch RPC proto

ronelliott:
## Summary
- Add `GetRegisteredResourcesRequest` and `GetRegisteredResourcesResponse` messages to the registered resources proto
- Reuses `GetRegisteredResourceRequest` as the identifier element type (supports `oneof` id or name with optional namespace context) to avoid duplicating field definitions; max 250 identifiers per request
- Add `GetRegisteredResources` RPC to `RegisteredResourcesService` for batch lookup intended to support ABAC authorization checks before bulk operations
- Regenerate `protocol/go` types, connect handlers, SDK wrappers, and API docs

## Notes
- Service-layer implementation is intentionally excluded; `protocol/go` must be released first before bumping in `service/go.mod` and adding the implementation

## Test plan
- [ ] `buf lint service` passes clean
- [ ] Generated types present in `protocol/go/policy/registeredresources/`
- [ ] CI checks pass

## Summary by CodeRabbit

* **New Features**
  * Added a batch GetRegisteredResources endpoint to request multiple registered resources in one call.

* **Documentation**
  * Updated gRPC and OpenAPI docs to include the new endpoint and request/response schemas; also adjusted example formatting and some description whitespace for clarity.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opentdf/platform/pull/3460)


docs/grpc/index.html
```diff
@@ -14925,6 +14933,55 @@ <h3 id="policy.registeredresources.GetRegisteredResourceValuesByFQNsResponse.Fqn
...
+            </thead>
+            <tbody>
+              
+                <tr>
+                  <td>identifiers</td>
+                  <td><a href="#policy.registeredresources.GetRegisteredResourceRequest">GetRegisteredResourceRequest</a></td>
+                  <td>repeated</td>
+                  <td><p>Required
+List of identifiers to look up; each supports the same oneof (id or name) and optional namespace context as GetRegisteredResourceRequest </p></td>
+                </tr>
```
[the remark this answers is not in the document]
ronelliott: fixed.

service/policy/registeredresources/registered_resources.proto
```diff
@@ -92,6 +92,20 @@ message GetRegisteredResourceResponse {
...
+  repeated GetRegisteredResourceRequest identifiers = 1 [
+    (buf.validate.field).repeated = {
+      min_items: 1,
+      max_items: 250
+    }
+  ];
+}
+message GetRegisteredResourcesResponse {
+  repeated policy.RegisteredResource resources = 1;
+}
```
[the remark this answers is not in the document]
ronelliott: fixed.

protocol/go/policy/registeredresources/registeredresourcesconnect/registered_resources.connect.go
```diff
@@ -252,6 +269,7 @@ func (c *registeredResourcesServiceClient) DeleteRegisteredResourceValue(ctx con
 type RegisteredResourcesServiceHandler interface {
 	CreateRegisteredResource(context.Context, *connect.Request[registeredresources.CreateRegisteredResourceRequest]) (*connect.Response[registeredresources.CreateRegisteredResourceResponse], error)
 	GetRegisteredResource(context.Context, *connect.Request[registeredresources.GetRegisteredResourceRequest]) (*connect.Response[registeredresources.GetRegisteredResourceResponse], error)
+	GetRegisteredResources(context.Context, *connect.Request[registeredresources.GetRegisteredResourcesRequest]) (*connect.Response[registeredresources.GetRegisteredResourcesResponse], error)
 	ListRegisteredResources(context.Context, *connect.Request[registeredresources.ListRegisteredResourcesRequest]) (*connect.Response[registeredresources.ListRegisteredResourcesResponse], error)
```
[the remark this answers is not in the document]
ronelliott: fixed

c-r33d (approved):

c-r33d (approved):
I would make sure to add tests to:
- https://github.com/opentdf/platform/blob/main/service/policy/registeredresources/registered_resources_test.go

If you haven't already. Just to make sure your request structure works as expected. You can comment them out to avoid any dep issues with not having released the protos first. Once the protos are released, uncomment.

service/policy/registeredresources/registered_resources.proto
```diff
@@ -92,6 +92,20 @@ message GetRegisteredResourceResponse {
   policy.RegisteredResource resource = 1;
 }
 
+message GetRegisteredResourcesRequest {
+  // Required
+  // List of identifiers to look up; each supports the same oneof (id or name) and optional namespace context as GetRegisteredResourceRequest
+  repeated GetRegisteredResourceRequest identifiers = 1 [
```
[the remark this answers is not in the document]
ronelliott: this was updated to use a separate shape.

service/policy/registeredresources/registered_resources.proto
```diff
@@ -92,6 +92,20 @@ message GetRegisteredResourceResponse {
...
+    (buf.validate.field).repeated = {
+      min_items: 1,
+      max_items: 250
+    }
+  ];
+}
+message GetRegisteredResourcesResponse {
+  repeated policy.RegisteredResource resources = 1;
```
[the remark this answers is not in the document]
ronelliott: end users can correlate using names or ids

service/policy/registeredresources/registered_resources.proto
```diff
@@ -92,6 +92,51 @@ message GetRegisteredResourceResponse {
...
+      min_len : 1
+      uri : true
+    }
+  ];
+  string namespace_id = 4 [(buf.validate.field).string.uuid = true];
+}
+
+message GetRegisteredResourcesRequest {
+  // Required - maximum 250 identifiers per request
+  repeated RegisteredResourceIdentifier identifiers = 1 [
```
c-r33d, line 129: Is there any concern with uniqueness? As mentioned in the gemini comment?

c-r33d (approved):
Approved. Spoke with Ron about using the `GetRegisteredResourceValuesByFQNsRequest`, but in their use case they only have the name of the RR, not FQN of the value.

marythought (commented):
Thanks for splitting this out so `protocol/go` can be released ahead of the service-layer implementation — that's the right sequencing.

A couple of consistency issues caught my eye that I think want to be resolved before this merges, plus some ergonomics follow-ups (one of which c-r33d already raised):

**Consistency**

1. The `GetRegisteredResources` RPC isn't actually declared in the `service RegisteredResourcesService { ... }` block in `service/policy/registeredresources/registered_resources.proto`. The messages exist and `.pb.go` is regenerated for them, but there's no `rpc GetRegisteredResources(...)` entry, so the PR title ("add … batch RPC") and body ("Add `GetRegisteredResources` RPC to `RegisteredResourcesService`") don't match what lands. If the intent is messages-only-now / RPC-later, I'd suggest adjusting the title + description to reflect that ("add proto types for the batch lookup RPC") so future readers and release notes aren't confused. If messages-only is the intent: would also recommend dropping the OpenAPI path additions in this PR so the docs don't promise an endpoint that the proto hasn't registered.
2. `docs/openapi/policy/registeredresources/registered_resources.openapi.yaml` is out of sync with the proto — it still advertises a `POST .../GetRegisteredResources` endpoint and its `identifiers.items` still `$ref`s `GetRegisteredResourceRequest` instead of the new `RegisteredResourceIdentifier`. The gencode CI check only validates `.pb.go`, which is why this slipped past. Worth regenerating, or dropping the OpenAPI changes from this PR if the RPC is intentionally deferred.
3. The PR body says "Reuses `GetRegisteredResourceRequest` as the identifier element type" — the latest revision uses a dedicated `RegisteredResourceIdentifier` instead, so that line is stale.

**Ergonomics** — inline comments below on:

- Uniqueness on `identifiers` (following c-r33d's open question)
- Response shape doesn't help callers correlate to inputs or detect partial misses — matters specifically for the ABAC pre-check use case this PR justifies
- `RegisteredResourceIdentifier` duplicates `GetRegisteredResourceRequest`'s validation byte-for-byte; the two will drift

Happy to chat through any of these if it's easier than going back and forth in comments.

service/policy/registeredresources/registered_resources.proto
marythought, line 129: +1 to c-r33d's question on uniqueness. The sibling `GetRegisteredResourceValuesByFQNsRequest.fqns` already enforces `unique: true`, and without it here, callers can submit duplicates (or the same resource referenced once by `id` and once by `name+namespace`) — wasted work and ambiguous response interpretation. Recommend `unique: true` unless duplicates are intentionally allowed, in which case a comment explaining why would help.

service/policy/registeredresources/registered_resources.proto
```diff
@@ -92,6 +92,51 @@ message GetRegisteredResourceResponse {
...
+    (buf.validate.field).repeated = {
+      min_items: 1,
+      max_items: 250
+    }
+  ];
+}
+message GetRegisteredResourcesResponse {
+  repeated policy.RegisteredResource resources = 1;
```
marythought, line 137: Following up on the earlier suggestion about correlation — "callers can correlate using names or ids" works for the happy path, but the use case this PR calls out is ABAC pre-checks before bulk operations, and for that the important question is *which of my N inputs were misses*. With `repeated policy.RegisteredResource resources` and no preserved order or "not found" markers, the caller has to diff the request against the response and re-map (which gets messier when a batch mixes `id`-keyed and `name+namespace`-keyed identifiers).

Two paths that preserve ergonomics without a big design change:

1. **Document the contract** in the proto comment: "order preserved; missing identifiers are omitted" (or whatever you actually want). Right now it's undefined, which is the part I'd most want to fix before this ships.
2. **Return paired results** — e.g. `repeated GetRegisteredResourceResult` where each entry carries the input identifier alongside an optional resource. Cheap now, much harder to add later without breaking callers.

Even (1) alone would be a meaningful improvement.

service/policy/registeredresources/registered_resources.proto
```diff
@@ -92,6 +92,51 @@ message GetRegisteredResourceResponse {
   policy.RegisteredResource resource = 1;
 }
 
+// Identifier for a single registered resource lookup within a batch request.
+// Mirrors the identifier fields of GetRegisteredResourceRequest.
+message RegisteredResourceIdentifier {
```
marythought, line 97: Now that this is a dedicated message (which I think is the right call), it's a byte-for-byte duplicate of the identifier fields on `GetRegisteredResourceRequest` — same `oneof`, same CEL rule, same namespace context. These will drift the next time someone updates one and forgets the other.

Two options:

- **Extract a shared inner message** that both wrap (e.g., `GetRegisteredResourceRequest` becomes `{ RegisteredResourceIdentifier identifier = 1; }`). One source of truth for validation.
- **Accept the duplication** and add a `// Keep in sync with GetRegisteredResourceRequest` comment so the next person knows.

Either is fine — leaving it implicit is the risky path.

docs/openapi/policy/registeredresources/registered_resources.openapi.yaml
```diff
@@ -72,6 +72,41 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/policy.registeredresources.GetRegisteredResourceResponse'
+  /policy.registeredresources.RegisteredResourcesService/GetRegisteredResources:
```
marythought, line 75: This yaml is out of sync with the proto on the current branch:
- It registers this `POST .../GetRegisteredResources` endpoint, but the proto's service block has no such RPC.
- Down in the schema definition, `identifiers.items` `$ref`s `GetRegisteredResourceRequest`, but the proto now uses `RegisteredResourceIdentifier`.

CI's gencode check only validates `.pb.go`, so the OpenAPI drift slipped through. Re-generate, or drop these OpenAPI changes from this PR if the RPC is being deferred.

jakedoublev:
> * Response shape doesn't help callers correlate to inputs or detect partial misses — matters specifically for the ABAC pre-check use case this PR justifies

This is a big one. Is it better to correlate with guaranteed order from input to output or respond with a map from input name to output `name: resource`? Also, if any are not found, does the entire request fail? 

We also have a pretty solidified `Get, Create, Update, Delete, List` and maybe `GetByFQN` or `UnsafeXYZ` per policy object. I'm pretty sure this will be the only `Get` RPC that is a bulk lookup not relying on FQNs. Why is it not possible to use FQNs? If it is indeed not possible, can we please name it `GetRegisteredResourcesBulk` to differentiate?

c-r33d:
> > * Response shape doesn't help callers correlate to inputs or detect partial misses — matters specifically for the ABAC pre-check use case this PR justifies
> 
> This is a big one. Is it better to correlate with guaranteed order from input to output or respond with a map from input name to output `name: resource`? Also, if any are not found, does the entire request fail?
> 
> We also have a pretty solidified `Get, Create, Update, Delete, List` and maybe `GetByFQN` or `UnsafeXYZ` per policy object. I'm pretty sure this will be the only `Get` RPC that is a bulk lookup not relying on FQNs. Why is it not possible to use FQNs? If it is indeed not possible, can we please name it `GetRegisteredResourcesBulk` to differentiate?

I agree with the renaming of the RPC. After speaking with Ron, he mentioned that S4 only has the `name` of the RR not the namespace. But now that I think about it that might break when namespacing is released. So S4 might need to update their code to start storing the FQN.

service/policy/registeredresources/registered_resources.proto
ronelliott: for my use case no, but others perhaps. should this be updated to a map structure?

ronelliott:
@jakedoublev @c-r33d should we maybe change it to GetRegisteredResourcesByName, returning the map structure similar to the ByFQN methods and drop the by ID lookup? this more closely matches our specific use case and broadens the API surface a bit less.

ronelliott:
per the discussion we will go a different route so I'm closing this PR

PR #3472, opened 2026-05-13
fix(policy): include action_attribute_values in GetRegisteredResource response

### LINKED ISSUE #3471: GetRegisteredResource omits action_attribute_values from nested values
## Summary

`policy.RegisteredResourcesService/GetRegisteredResource` returns each `RegisteredResourceValue` with only `id`, `value`, and `fqn` populated. The `action_attribute_values` repeated field is always empty in the response, even when mappings exist (and are returned correctly by `GetRegisteredResourceValue` for the same value id).

This makes `GetRegisteredResource` insufficient for any caller that needs to know what attributes a resource's action is gated by — they have to do N+1 follow-up `GetRegisteredResourceValue` calls per value.

## Repro

1. Create a registered resource with one value bound to an action and an attribute:
   ```
   dsp tructl policy registered-resources create -s https://demo.com -n s4-cache-demo -v initial
   dsp tructl policy registered-resources values create -r s4-cache-demo -s https://demo.com -v mapped \
       -a "create;https://demo.com/attr/classification/value/secret"
   ```
2. Fetch the value directly — action mapping is present:
   ```
   dsp tructl policy registered-resources values get --id <value-id> --json
   # action_attribute_values: [{action: {name: "create"}, attribute_value: {fqn: ".../secret"}}]
   ```
3. Fetch the resource — same value comes back without action mappings:
   ```
   dsp tructl policy registered-resources get --id <rr-id> --json
   ```
   ```json
   {
     "id": "a2696888-...",
     "name": "s4-cache-demo",
     "values": [
       {"id": "722de795-...", "value": "initial"},
       {"id": "d73e383a-...", "value": "mapped"}
     ]
   }
   ```

## Expected

`values[].action_attribute_values` is populated for each value, matching what `GetRegisteredResourceValue` returns for that value's id. The proto field exists (`policy.RegisteredResourceValue.action_attribute_values`, field 4 in `service/policy/objects.proto`) and consumers reasonably expect it filled.

## Root cause

`service/policy/db/queries/registered_resources.sql`, the `getRegisteredResource` query:

```sql
JSON_AGG(
    JSON_BUILD_OBJECT(
        'id', v.id,
        'value', v.value
    )
) FILTER (WHERE v.id IS NOT NULL) as values
```

The JSON object built for each value omits `action_attribute_values`. The same query for `getRegisteredResourceValue` joins `registered_resource_action_attribute_values` and populates it; the resource-level query does not.

## Impact

Any client that authorizes on `(resource, action) → attributes` using `GetRegisteredResource` silently fails:

- service's authz wrapper (`pkg/backend/authz/registered_resource.go`) reads `value.GetActionAttributeValues()` from the `GetRegisteredResource` response. Against the live platform every request is denied with `"no attribute mappings found for resource"`. The wrapper's BDD coverage passes because the in-process fake (`pkg/fake/dsp.go`) populates `action_attribute_values` on the mock response, masking the divergence.

Confirmed empirically by patching S4 to log the response from a live DSP:

```
{"msg":"DEBUG-PROBE rr value","value_id":"722de795-...","value":"initial","aav_count":0}
{"msg":"DEBUG-PROBE rr value","value_id":"d73e383a-...","value":"mapped",  "aav_count":0}
{"msg":"no attribute mappings found for resource, denying","resource":"s4-cache-demo","action":"create"}
```

## Suggested fix

Update `getRegisteredResource` in `queries/registered_resources.sql` to include `action_attribute_values` in the per-value JSON object, mirroring the logic used by `getRegisteredResourceValue`. The fields are the same shape; it should be a JOIN against `registered_resource_action_attribute_values` (and `actions`, `attribute_values`) aggregated per value id.

Alternatively, if the omission is intentional for payload-size reasons, the proto field should be documented as not populated on `GetRegisteredResource` and callers should be required to call `GetRegisteredResourceValue` per value — but this contradicts what consumers (including the platform's own SDK examples) appear to assume.

## Environment

- platform: `main` at `6a0b3c63 feat(policy): Add resource mapping group FQNs (#3447)` (local clone)

pflynn-virtru:
## Summary

`GetRegisteredResource` returned each value with only `id`, `value`, and `fqn` populated — the `action_attribute_values` field was always empty even when mappings existed. Callers that authorize on `(resource, action) → attributes` via this RPC were silently denied; `GetRegisteredResourceValue` returned the same value with mappings populated, exposing the divergence.

Root cause: the `getRegisteredResource` SQL in `service/policy/db/queries/registered_resources.sql` built each value's JSON object with only `id` and `value`. `listRegisteredResources` already used the correct `LEFT JOIN LATERAL` pattern; this PR mirrors it into the single-resource query and regenerates the sqlc binding. The follow-up commit also wraps the inner `JSON_AGG` in both queries with `COALESCE(..., '[]'::json)` so a value with no mappings produces an empty array rather than SQL NULL.

Fixes #3471

## Why this is different from #3398

PR #3398 (also a "hydrate a nested field that some queries populate and others don't" change, on `ListAttributes`) was closed with the explanation that `ListAttributes` is on a lot of hot paths and the team would rather pursue selectors/expanders than expand the default response payload. That concern is real for paginated list RPCs but does not apply to this change:

- **Bug fix, not feature expansion.** `GetRegisteredResource` is the only sibling that fails to populate `action_attribute_values`. `ListRegisteredResources` and `GetRegisteredResourceValue` both already hydrate it. The proto field `RegisteredResourceValue.action_attribute_values` is declared and documented; leaving it empty here is divergence, not an opt-in payload addition.
- **Silent authorization failures in production.** Downstream callers that read `value.GetActionAttributeValues()` off the `GetRegisteredResource` response see an empty slice and deny the request with `"no attribute mappings found for resource"`, even though the mappings exist in the database and are returned by every sibling RPC. This is the bug described in #3471 — not a payload-completeness preference.
- **Not a hot-path list endpoint.** `GetRegisteredResource` is a single-item lookup by id or name. The cardinality argument that motivated closing #3398 (thousands of attribute values × N mappings each, on a frequently-called paginated endpoint) does not transfer.
- **No proto change, no new field, no new opt-in toggle.** The shape this PR returns is already what `listRegisteredResources` returns today.

If the team would prefer to address this through a `GetRegisteredResourceWithMappings` selector or an `expand=action_attribute_values` parameter instead of fixing the divergence inline, happy to redirect — but the silent-denial bug is real and the rest of the policy DB queries already treat populating this field as the default.

## Test plan

- [x] Wrote failing integration test first; confirmed it fails against unfixed SQL with `Should NOT be empty, but was []` (red-green verified).
- [x] Confirmed it passes after the SQL fix.
- [x] Full `TestRegisteredResourcesSuite` integration suite passes (testcontainers/postgres) — both the new test and the existing fixture-based tests against values with no AAVs.
- [x] `go test ./service/policy/...` unit tests pass.
- [x] `golangci-lint run` on changed files: 0 new issues.
- [x] `gofumpt` clean on changed Go files.
- [ ] Reviewer: re-verify with `cd service && go test ./integration/ -run TestRegisteredResourcesSuite -count=1`.

## Summary by CodeRabbit

* **Tests**
  * Added integration test for retrieving registered resources with action attribute values.
  * Enhanced existing test with stricter validation of namespaced action information and attribute identifiers.

* **Bug Fixes**
  * Registered resources now properly return action attribute values with complete namespace details when fetched.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opentdf/platform/pull/3472)

### DIFF
--- a/service/integration/registered_resources_test.go
+++ b/service/integration/registered_resources_test.go
@@ -382,6 +382,93 @@ func (s *RegisteredResourcesSuite) Test_ListRegisteredResources_RegResValuesCont
 	s.True(foundVal2, "Value 2 not found in registered resource values")
 }
 
+func (s *RegisteredResourcesSuite) Test_GetRegisteredResource_RegResValuesContainActionAttributeValues() {
+	// Create a registered resource with values that have action attribute values
+	newRegRes, err := s.db.PolicyClient.CreateRegisteredResource(s.ctx, &registeredresources.CreateRegisteredResourceRequest{
+		NamespaceId: s.getNamespaceID("example.com"),
+		Name:        "test_get_reg_res_with_action_attr_values",
+	})
+	s.Require().NoError(err)
+	s.NotNil(newRegRes)
+	regResID := newRegRes.GetId()
+
+	val1, err := s.db.PolicyClient.CreateRegisteredResourceValue(s.ctx, &registeredresources.CreateRegisteredResourceValueRequest{
+		ResourceId: regResID,
+		Value:      "test_value_1",
+		ActionAttributeValues: []*registeredresources.ActionAttributeValue{
+			{
+				ActionIdentifier: &registeredresources.ActionAttributeValue_ActionName{
+					ActionName: actions.ActionNameCreate,
+				},
+				AttributeValueIdentifier: &registeredresources.ActionAttributeValue_AttributeValueFqn{
+					AttributeValueFqn: "https://example.com/attr/attr1/value/value1",
+				},
+			},
+		},
+	})
+	s.Require().NoError(err)
+	s.NotNil(val1)
+
+	val2, err := s.db.PolicyClient.CreateRegisteredResourceValue(s.ctx, &registeredresources.CreateRegisteredResourceValueRequest{
+		ResourceId: regResID,
+		Value:      "test_value_2",
+		ActionAttributeValues: []*registeredresources.ActionAttributeValue{
+			{
+				ActionIdentifier: &registeredresources.ActionAttributeValue_ActionName{
+					ActionName: actions.ActionNameUpdate,
+				},
+				AttributeValueIdentifier: &registeredresources.ActionAttributeValue_AttributeValueFqn{
+					AttributeValueFqn: "https://example.com/attr/attr2/value/value2",
+				},
+			},
+		},
+	})
+	s.Require().NoError(err)
+	s.NotNil(val2)
+
+	// Get the registered resource and check if values contain action attribute values
+	got, err := s.db.PolicyClient.GetRegisteredResource(s.ctx, &registeredresources.GetRegisteredResourceRequest{
+		Identifier: &registeredresources.GetRegisteredResourceRequest_Id{
+			Id: regResID,
+		},
+	})
+	s.Require().NoError(err)
+	s.NotNil(got)
+	s.Equal("test_get_reg_res_with_action_attr_values", got.GetName())
+
+	values := got.GetValues()
+	s.Require().Len(values, 2)
+
+	foundVal1 := false
+	foundVal2 := false
+	for _, v := range values {
+		if v.GetId() == val1.GetId() {
+			foundVal1 = true
+			actionAttrValues := v.GetActionAttributeValues()
+			s.Require().NotEmpty(actionAttrValues)
+			for _, aav := range actionAttrValues {
+				s.NotNil(aav.GetAction())
+				s.NotNil(aav.GetAttributeValue())
+				s.NotNil(aav.GetAction().GetNamespace(), "action namespace should be populated for namespaced RR")
+				s.Equal("example.com", aav.GetAction().GetNamespace().GetName())
+			}
+		}
+		if v.GetId() == val2.GetId() {
+			foundVal2 = true
+			actionAttrValues := v.GetActionAttributeValues()
+			s.Require().NotEmpty(actionAttrValues)
+			for _, aav := range actionAttrValues {
+				s.NotNil(aav.GetAction())
+				s.NotNil(aav.GetAttributeValue())
+				s.NotNil(aav.GetAction().GetNamespace(), "action namespace should be populated for namespaced RR")
+				s.Equal("example.com", aav.GetAction().GetNamespace().GetName())
+			}
+		}
+	}
+	s.True(foundVal1, "Value 1 not found in registered resource values")
+	s.True(foundVal2, "Value 2 not found in registered resource values")
+}
+
 func (s *RegisteredResourcesSuite) Test_ListRegisteredResources_Limit_Succeeds() {
 	var limit int32 = 1
 	list, err := s.db.PolicyClient.ListRegisteredResources(s.ctx, &registeredresources.ListRegisteredResourcesRequest{
--- a/service/integration/registered_resources_test.go
+++ b/service/integration/registered_resources_test.go
@@ -352,22 +352,26 @@ func (s *RegisteredResourcesSuite) Test_ListRegisteredResources_RegResValuesCont
 		foundRegRes = true
 
 		for _, v := range values {
+			actionAttrValues := v.GetActionAttributeValues()
+			s.Require().Len(actionAttrValues, 1)
+			aav := actionAttrValues[0]
+			s.Require().NotNil(aav.GetAction())
+			s.Require().NotNil(aav.GetAttributeValue())
+			s.NotNil(aav.GetAction().GetNamespace(), "action namespace should be populated for namespaced RR")
+			s.Equal("example.com", aav.GetAction().GetNamespace().GetName())
+			s.Equal("https://example.com", aav.GetAction().GetNamespace().GetFqn())
+
 			switch v.GetId() {
 			case val1.GetId():
 				foundVal1 = true
+				s.Equal(actions.ActionNameCreate, aav.GetAction().GetName())
+				s.Equal("https://example.com/attr/attr1/value/value1", aav.GetAttributeValue().GetFqn())
 			case val2.GetId():
 				foundVal2 = true
+				s.Equal(actions.ActionNameUpdate, aav.GetAction().GetName())
+				s.Equal("https://example.com/attr/attr2/value/value2", aav.GetAttributeValue().GetFqn())
 			default:
 				s.FailNow("unexpected value found", "value id: %s", v.GetId())
-			}
-
-			actionAttrValues := v.GetActionAttributeValues()
-			s.Require().NotEmpty(actionAttrValues)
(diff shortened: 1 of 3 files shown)


pflynn-virtru:
@coderabbitai help

pflynn-virtru:
@coderabbitai full review

service/policy/db/queries/registered_resources.sql
```diff
@@ -50,13 +50,42 @@ SELECT
     JSON_AGG(
         JSON_BUILD_OBJECT(
             'id', v.id,
-            'value', v.value
+            'value', v.value,
+            'action_attribute_values', action_attrs.values
```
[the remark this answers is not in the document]
pflynn-virtru: The risk is overstated for this codebase:

1. The existing `listRegisteredResources` query (the block this PR mirrors, lines 97–138 in `registered_resources.sql`) uses the identical `LATERAL` + `JSON_AGG` shape without `COALESCE`, and has been in production. If `null` here were a real `protojson.Unmarshal` problem, the list path would already fail.
2. Fixtures already exercise the null path: `res_with_values__value2` has no AAVs (`service/internal/fixtures/policy_fixtures.yaml:630`). `Test_ListRegisteredResources_NoPagination_Succeeds` and `Test_GetRegisteredResource_WithValues_Succeeds` both pass against that fixture with my fix in place.
3. `protojson.Unmarshal` accepts JSON `null` for repeated fields per spec — it's treated as the default (empty) value, not an error.

`COALESCE(action_attrs.values, '[]'::json)` would make the JSON shape more uniform, which is a nice-to-have. But applying it only to `getRegisteredResource` would diverge from `listRegisteredResources` — worse than the symmetry I have now. Better as a follow-up that touches both queries together.

service/integration/registered_resources_test.go
```diff
@@ -382,6 +382,93 @@ func (s *RegisteredResourcesSuite) Test_ListRegisteredResources_RegResValuesCont
...
+			for _, aav := range actionAttrValues {
+				s.NotNil(aav.GetAction())
+				s.NotNil(aav.GetAttributeValue())
+				s.NotNil(aav.GetAction().GetNamespace(), "action namespace should be populated for namespaced RR")
+				s.Equal("example.com", aav.GetAction().GetNamespace().GetName())
+			}
+		}
+	}
+	s.True(foundVal1, "Value 1 not found in registered resource values")
+	s.True(foundVal2, "Value 2 not found in registered resource values")
```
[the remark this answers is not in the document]
pflynn-virtru: The duplication is intentional: this test was deliberately mirrored from `Test_ListRegisteredResources_RegResValuesContainActionAttributeValues` (lines 354–377 in the same file) so reviewers can compare the two side-by-side as evidence that the `get` and `list` paths now behave identically.

That said, the `default: s.FailNow(...)` clause IS a real correctness improvement — the current `if v.GetId() == ...` arms would silently skip an unexpected value rather than failing the test. Worth doing, but the right fix is to refactor both the list-side and get-side tests together in a follow-up, not just this one. I'd rather leave the symmetry intact in this PR than introduce drift between two near-identical tests.

service/policy/db/queries/registered_resources.sql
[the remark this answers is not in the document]
pflynn-virtru: Reconsidered — applied to both queries in 756138bd. `COALESCE(JSON_AGG(...), '[]'::json)` in both `getRegisteredResource` and `listRegisteredResources` LATERAL subqueries so the per-value JSON always has an array for `action_attribute_values`, no SQL NULL.

service/integration/registered_resources_test.go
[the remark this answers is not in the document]
pflynn-virtru: Reconsidered — applied to both tests in 756138bd. Switch with `default: s.FailNow(...)` in `Test_GetRegisteredResource_...` and `Test_ListRegisteredResources_...` so an unexpected value id fails the test rather than being silently skipped.

service/integration/registered_resources_test.go
```diff
@@ -341,39 +341,33 @@ func (s *RegisteredResourcesSuite) Test_ListRegisteredResources_RegResValuesCont
...
+			case val2.GetId():
+				foundVal2 = true
+			default:
+				s.FailNow("unexpected value found", "value id: %s", v.GetId())
+			}
+
+			actionAttrValues := v.GetActionAttributeValues()
+			s.Require().NotEmpty(actionAttrValues)
+			for _, aav := range actionAttrValues {
+				s.NotNil(aav.GetAction())
```
jakedoublev: Can you please assert that the correct action was found, not that any non-nil action was found, and the same with the attribute value?

jakedoublev: Same with the test below. This test looks like it was lacking originally.

service/policy/db/queries/registered_resources.sql
```diff
@@ -50,13 +50,44 @@ SELECT
...
+LEFT JOIN LATERAL (
+    -- COALESCE so a value with no mappings yields '[]' rather than SQL NULL,
+    -- giving consumers a consistent JSON array shape for action_attribute_values.
+    SELECT COALESCE(JSON_AGG(
+        JSON_BUILD_OBJECT(
+            'action', JSON_BUILD_OBJECT(
+                'id', a.id,
+                'name', a.name,
+                'namespace', CASE WHEN a.namespace_id IS NULL THEN NULL
+                    ELSE JSON_BUILD_OBJECT('id', ans.id, 'name', ans.name, 'fqn', ans_fqns.fqn)
```
jakedoublev: For the `fqn`, optimize with a CONCAT here instead of another join to the FQNs table: https://github.com/opentdf/platform/blob/main/service/policy/db/queries/obligations.sql#L157-L161

service/integration/registered_resources_test.go
pflynn-virtru: Applied in a2081683 to both `Test_GetRegisteredResource_RegResValuesContainActionAttributeValues` and `Test_ListRegisteredResources_RegResValuesContainActionAttributeValues`. Each value now asserts the specific action name (`ActionNameCreate` for val1, `ActionNameUpdate` for val2) and the specific attribute value FQN (`.../attr1/value/value1` and `.../attr2/value/value2`) that was created, instead of just non-nil.

service/policy/db/queries/registered_resources.sql
pflynn-virtru: Applied in a2081683 to both queries. Dropped the `LEFT JOIN attribute_fqns ans_fqns` for the action's namespace FQN and inlined it as `CONCAT('https://', ans.name)` in the `JSON_BUILD_OBJECT`, matching the pattern in `obligations.sql`. The attribute value FQN join still stays because value FQNs aren't a simple concatenation.

jakedoublev (approved):

PR #3488, opened 2026-05-19
fix(cli): Prune was not classifying multi-namespaced RRs properly.

c-r33d:
1.) RRs with multi-namespaces should be classified as `Blocked` and require manual deletion.

>[!NOTE]
>While we could attempt to locate all the `pieces` of a split RR across different namespaces. I think it's better to just 
>have the customer manually delete / handle multi-namespace pruning since trying to do this for customers could
>result in a brittle workflow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Registered resources spanning multiple namespaces are now properly marked for manual deletion instead of reporting source mismatches.

* **Chores**
  * Simplified prune planning logic and removed the interactive reviewer option.
  * Streamlined prune plan details and source state verification.

[Review Change Stack](https://app.coderabbit.ai/change-stack/opentdf/platform/pull/3488?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


c-r33d:
@coderabbitai review

elizabethhealy (approved):

PR #3543, opened 2026-05-28
fix(ers): use pgx database/sql driver in SQL ERS provider

### LINKED ISSUE #3539: multi-strategy ERS SQL provider should register its own database/sql driver
## Summary

The multi-strategy ERS SQL provider (`service/entityresolution/multi-strategy/providers/sql/sql_provider.go`) calls `sql.Open(config.Driver, connStr)` which requires the named driver to be pre-registered via a blank import. The provider source acknowledges this with a comment:

\`\`\`go
// Database drivers would be imported here:
// _ "github.com/lib/pq"           // PostgreSQL driver
// _ "github.com/go-sql-driver/mysql" // MySQL driver
// _ "github.com/mattn/go-sqlite3" // SQLite driver
\`\`\`

This design delegates driver registration to every consumer of the package. For the DSP platform, this means adding \`_ "github.com/lib/pq"\` in \`main.go\` — a side-effect that affects the entire binary even for users who never use the SQL ERS tier.

## Suggested fix

Register the driver inside the SQL provider's own \`init()\` or \`NewProvider()\` function, or document a supported pattern for lazy/conditional registration. The LDAP and claims providers have no equivalent external dependency, so this inconsistency is surprising for consumers.

## Context
Discovered while implementing a multi-strategy ERS reference implementation in the Virtru Data Security Platform.

jp-ayyappan:
## Summary

Fixes #3539.

The multi-strategy ERS SQL provider called `sql.Open(config.Driver, ...)` but did not ensure the PostgreSQL `database/sql` driver was available in the platform service binary. This PR makes PostgreSQL SQL provider configs work without requiring consumers to add their own blank import.

## Changes

1. Imports `github.com/jackc/pgx/v5/stdlib` in the SQL provider package so the pgx `database/sql` drivers are registered when the provider package is loaded.
2. Defaults SQL provider configs to the versioned `pgx/v5` driver name.
3. Normalizes PostgreSQL aliases before opening the database connection:
   - `pgx` -> `pgx/v5`
   - `postgres` -> `pgx/v5`
   - `postgresql` -> `pgx/v5`
   - mixed-case and whitespace-padded values are trimmed/lowercased first
4. Updates README and example configs to recommend `driver: pgx/v5`.
5. Adds tests for the default driver, alias normalization, and connection-string construction for supported PostgreSQL aliases.

## Behavior

| Scenario | Result |
|---|---|
| SQL provider uses default config | Opens with `pgx/v5` |
| Consumer passes `driver: pgx/v5` | Opens with `pgx/v5` |
| Consumer passes `driver: pgx` | Normalized to `pgx/v5` |
| Consumer passes `driver: postgres` or `driver: postgresql` | Normalized to `pgx/v5` |
| Consumer passes `driver: Postgres` | Normalized to `pgx/v5` |
| Consumer already blank-imports `github.com/lib/pq` | SQL ERS provider still uses `pgx/v5`; the existing `postgres` registration is ignored because the provider normalizes to the versioned pgx driver name |
| Consumer passes `driver: mysql` | Not auto-registered; consumer must import/register the driver |
| Consumer passes `driver: sqlite` or `driver: sqlite3` | Not auto-registered; consumer must import/register the driver |

## Testing

- `go test ./entityresolution/multi-strategy/providers/sql -race`
- `go test ./entityresolution/multi-strategy/...`
- `GOCACHE=/private/tmp/go-build-cache GOLANGCI_LINT_CACHE=/private/tmp/golangci-lint-cache golangci-lint run --config ../.golangci.yaml ./entityresolution/multi-strategy/providers/sql --new-from-rev=origin/main`

## Notes

- Existing configs using `driver: postgres`, `driver: postgresql`, or `driver: pgx` continue to work via alias normalization.
- The documented recommended value is now `driver: pgx/v5` to avoid ambiguity with the process-global unversioned `pgx` driver name.

## Summary by CodeRabbit

* **New Features**
  * PostgreSQL driver is now available to the SQL ERS provider without consumer-side blank imports.

* **Improvements**
  * PostgreSQL driver aliases are normalized to the versioned `pgx/v5` driver name for consistent resolution.
  * The default PostgreSQL driver value is standardized to a single constant.

* **Notes**
  * MySQL and SQLite still require consumer-provided driver imports.


service/entityresolution/multi-strategy/providers/sql/sql_config.go
```diff
@@ -6,6 +6,10 @@ import (
 
 const (
 	// Default SQL configuration values
+	defaultPostgreSQLDriver   = "pgx/v5"
```
c-r33d: I'd recommend just keeping it `postgres`, less to test and probably provides the same value.

service/entityresolution/multi-strategy/providers/sql/sql_config_test.go
```diff
@@ -0,0 +1,83 @@
...
+					Username: "ers_user",
+					Password: "ers_password",
+					SSLMode:  "require",
+				},
+			}
+
+			connStr, err := provider.buildConnectionString()
+			if err != nil {
+				t.Fatalf("expected postgres alias to build connection string: %v", err)
+			}
```
c-r33d:
```suggestion
			t.Require().NoError(err)
```

Use the testify module with require.

service/entityresolution/multi-strategy/providers/sql/sql_config_test.go
```diff
@@ -0,0 +1,83 @@
...
+			if !strings.Contains(connStr, "dbname=identity_db") {
+				t.Fatalf("expected database name in connection string, got %q", connStr)
+			}
```
c-r33d:
```suggestion
       t.Contains(connStr, "dbname=identity_db")
```

service/entityresolution/multi-strategy/providers/sql/sql_provider.go
```diff
@@ -6,14 +6,21 @@ import (
 	"fmt"
 	"strings"
 
-	// Database drivers would be imported here:
-	// _ "github.com/lib/pq"           // PostgreSQL driver
-	// _ "github.com/go-sql-driver/mysql" // MySQL driver
-	// _ "github.com/mattn/go-sqlite3" // SQLite driver
-
+	// Register the pgx/v5 database/sql driver for SQL providers.
+	_ "github.com/jackc/pgx/v5/stdlib"
```
c-r33d, line 10:
```suggestion
	"github.com/jackc/pgx/v5/stdlib"
```

strantalis (approved):

service/entityresolution/multi-strategy/providers/sql/sql_provider.go
ryanulit: The code won't compile if you apply this suggestion, so resolving.

ryanulit (approved):

PR #3555, opened 2026-06-01
feat(policy): add search support to ListSubjectMappings

c-r33d:
## Summary
- Adds ListSubjectMappings RPC search support by wiring request search into the policy DB list query.
- Applies escaped, case-insensitive matching in the subject mappings SQL path and adds integration coverage for search behavior, wildcard literals, empty search, and pagination after filtering.

## Summary by CodeRabbit

* **New Features**
  * Subject mappings now support search filtering by attribute values and metadata labels
  * Search results can be combined with namespace filters for refined queries
  * Automatic whitespace trimming and special character handling ensure accurate searches
  * Result pagination works correctly with all filters applied

### DIFF
--- a/service/integration/subject_mappings_test.go
+++ b/service/integration/subject_mappings_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/opentdf/platform/protocol/go/common"
 	"github.com/opentdf/platform/protocol/go/policy"
 	"github.com/opentdf/platform/protocol/go/policy/actions"
+	"github.com/opentdf/platform/protocol/go/policy/attributes"
 	"github.com/opentdf/platform/protocol/go/policy/namespaces"
 	"github.com/opentdf/platform/protocol/go/policy/subjectmapping"
 	"github.com/opentdf/platform/service/internal/fixtures"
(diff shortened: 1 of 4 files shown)


c-r33d:
@coderabbitai review

service/policy/db/queries/subject_mappings.sql
```diff
@@ -116,15 +116,33 @@ subject_actions AS (
...
+            OR sm_ns_fqns.fqn = sqlc.narg('namespace_fqn')::text
+        )
+        AND CASE
+            WHEN sqlc.narg('search')::TEXT IS NULL THEN TRUE
+            ELSE (
+                fqns.fqn LIKE sqlc.narg('search')::TEXT ESCAPE '\'
+                OR EXISTS (
+                    SELECT 1
+                    FROM JSONB_EACH_TEXT(COALESCE(sm.metadata -> 'labels', '{}'::JSONB)) AS label(key, value)
+                    WHERE label.value ILIKE sqlc.narg('search')::TEXT ESCAPE '\'
```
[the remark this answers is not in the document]
c-r33d: The fqns are always lowercased, so using `LIKE` is ok to use here since we lower case the query coming in. In addition, `LIKE` performs better than `ILIKE` do to case-sensitivity.

service/integration/subject_mappings_test.go
```diff
@@ -1040,6 +1041,220 @@ func (s *SubjectMappingsSuite) Test_ListSubjectMappings_NoNamespaceFilter_Return
...
+	attrName := fmt.Sprintf("list-sm-search-attr-%d", suffix)
+	valueToken := fmt.Sprintf("fqn-only-%d", suffix)
+	ns, values := s.createSearchSubjectMappingNamespace(suffix, []string{
+		valueToken,
+		fmt.Sprintf("label-holder-%d", suffix),
+		fmt.Sprintf("label-key-holder-%d", suffix),
+		fmt.Sprintf("condition-holder-%d", suffix),
+		fmt.Sprintf("unmatched-%d", suffix),
+	})
+	defer s.deleteSearchSubjectMappingNamespace(ns)
```
jakedoublev: It looks like might be better to use `t.Cleanup` instead of `defer` for cleanups within go unit tests: https://stackoverflow.com/questions/61609085/what-is-useful-for-t-cleanup I just learned that myself.

service/integration/subject_mappings_test.go
```diff
@@ -1040,6 +1041,220 @@ func (s *SubjectMappingsSuite) Test_ListSubjectMappings_NoNamespaceFilter_Return
...
+			gotIDs := make([]string, 0, len(list.GetSubjectMappings()))
+			for _, sm := range list.GetSubjectMappings() {
+				gotIDs = append(gotIDs, sm.GetId())
+			}
+			s.ElementsMatch(tc.ids, gotIDs)
+			s.Equal(int32(len(tc.ids)), list.GetPagination().GetTotal())
+		})
+	}
+
+	for _, term := range []string{
```
jakedoublev, line 1104: Is it possible to make it clearer in the test, even with just a comment or something, why these are not expected to match filtered List results?

service/integration/subject_mappings_test.go
```diff
@@ -1040,6 +1041,220 @@ func (s *SubjectMappingsSuite) Test_ListSubjectMappings_NoNamespaceFilter_Return
...
+	emptySearch, err := s.db.PolicyClient.ListSubjectMappings(s.ctx, &subjectmapping.ListSubjectMappingsRequest{
+		NamespaceId: ns.GetId(),
+		Search:      &policy.Search{Term: ""},
+	})
+	s.Require().NoError(err)
+	s.Equal(noSearch.GetPagination().GetTotal(), emptySearch.GetPagination().GetTotal())
+	s.Len(emptySearch.GetSubjectMappings(), len(noSearch.GetSubjectMappings()))
+}
+
+func (s *SubjectMappingsSuite) Test_ListSubjectMappings_SearchPaginationAppliesAfterFiltering_Succeeds() {
```
jakedoublev, line 1225: 🔥 great thing to test

jakedoublev (dismissed):

service/integration/subject_mappings_test.go
c-r33d: Yeah t.Cleanup is the way to go

service/integration/subject_mappings_test.go
c-r33d: Yeah, I can add a comment.

jakedoublev (approved):

PR #3557, opened 2026-06-01
feat(policy): add search support to ListKeyAccessServers

c-r33d:
## Summary
- Adds ListKeyAccessServers RPC search support by wiring request search into the policy DB list query.
- Applies escaped, case-insensitive matching in the KAS registry SQL path and adds integration coverage for search behavior, wildcard literals, empty search, and pagination after filtering.


c-r33d:
@coderabbitai review

alkalescent (approved):

PR #3564, opened 2026-06-04
feat(kas): emit INFO log of supported mechanisms at startup (DSPX-3456)

dmihalcik-virtru:
## Summary
- Adds `INFO msg="kas trust mechanisms initialized" mechanisms=[...]` log line at KAS startup and on hot config reload
- Algorithm list sourced from static metadata declared at manager registration time
- Filtered by preview-feature gating to match `rewrap.o\` acceptance logic (ec/hybrid types suppressed unless the corresponding flag is enabled)
- Gives operators and integration tests an authoritative, low-noise view of which cryptographic mechanisms this KAS instance will serve

## Implementation

Extended `NamedKeyManagerCtxFactory` with a `SupportedAlgorithms []ocrypto.KeyType` field (optional; empty means the manager contributes nothing to capability listings)


sujankota (dismissed):

service/internal/security/in_process_provider.go
```diff
@@ -17,6 +17,20 @@ import (
...
 
+// inProcessSupportedAlgorithms is the canonical set of algorithms the
+// InProcessProvider knows how to serve when a corresponding key is loaded.
+// Keep in sync with the switch in Decrypt.
+var inProcessSupportedAlgorithms = []ocrypto.KeyType{
+	ocrypto.RSA2048Key,
+	ocrypto.RSA4096Key,
+	ocrypto.EC256Key,
+	ocrypto.EC384Key,
+	ocrypto.EC521Key,
```
pflynn-virtru: `InProcessProvider` over-advertises EC384/EC521 here. The comment above says "Keep in sync with the switch in Decrypt," but `Decrypt` (in_process_provider.go:274) only handles `case AlgorithmECP256R1:` — an EC384/EC521 key falls through to `default` → `"unsupported key algorithm"`. So in a legacy in-process deployment this log would advertise `ec:secp384r1`/`ec:secp521r1` as supported when `rewrap` would actually reject them.

Since the goal is authoritative discovery, either:
- trim this list to `EC256Key` only (matching the real `Decrypt`), or
- widen the `Decrypt` switch to `AlgorithmECP256R1, AlgorithmECP384R1, AlgorithmECP521R1` if it is genuinely capable.

(`BasicManager` handles all five, so its list is correct — this only affects the in-process adapter.)

service/internal/security/basic_manager.go
```diff
@@ -26,6 +26,20 @@ const (
 	ristrettoCacheTTL    = 30
 )
 
+// basicManagerSupportedAlgorithms is the canonical set of algorithms the
+// BasicManager knows how to serve when a key has been provisioned. Keep in
+// sync with the switch in Decrypt.
```
elizabethhealy, line 31: can we add a comment in decrypt noting it must be kept in sync with this list so someone updating that doesnt miss it

pflynn-virtru (dismissed):

service/internal/security/in_process_provider.go
pflynn-virtru: legacy won't hit this

service/trust/delegating_key_service.go
```diff
@@ -107,6 +107,42 @@ func (d *DelegatingKeyService) ListKeysWith(ctx context.Context, opts ListKeyOpt
...
+// A factory that fails to instantiate is logged at WARN and skipped — listing
+// capabilities never blocks startup.
+func (d *DelegatingKeyService) SupportedAlgorithms(ctx context.Context) []ocrypto.KeyType {
+	d.mutex.Lock()
+	names := slices.Collect(maps.Keys(d.managerFactories))
+	d.mutex.Unlock()
+
+	seen := make(map[ocrypto.KeyType]struct{})
+	for _, name := range names {
+		mgr, err := d.getKeyManager(ctx, &policy.KeyProviderConfig{Manager: name})
```
elizabethhealy: getKeyManager stores every manager it creates in d.managers, keyed by {Manager, Name}. Here Name is always "" since we're not setting it, but real rewrap calls come in with Name: "static" so they always miss the cache and create a second instance. The startup one just sits there leaking forever.

Better to bypass getKeyManager entirely here; copy the factories map under the lock, call each factory directly, grab the algorithms, then Close() and discard without ever touching d.managers.

service/internal/security/in_process_provider.go
```diff
@@ -214,9 +236,10 @@ func (a *InProcessProvider) ListKeysWith(ctx context.Context, opts trust.ListKey
 			}
 		} else if err != nil {
 			if a.logger != nil {
-				a.logger.WarnContext(ctx,
+				a.logger.WarnContext(
+					ctx,
 					"failed to list keys by algorithm",
-					slog.String("algorithm", alg),
+					slog.Any("algorithm", alg),
```
elizabethhealy, line 232: nit: may be cheaper to keep this as slog.String then just have string(alg) to avoid reflection

service/trust/delegating_key_service_test.go
```diff
@@ -321,3 +323,127 @@ func (suite *DelegatingKeyServiceTestSuite) TestGenerateECSessionKey() {
...
+			t.Parallel()
+			d := NewDelegatingKeyService(&MockKeyIndex{}, logger.CreateTestLogger(), nil)
+			tc.register(d)
+			got := d.SupportedAlgorithms(context.Background())
+			if got == nil {
+				got = []ocrypto.KeyType{}
+			}
+			if !slices.Equal(got, tc.want) {
+				t.Fatalf("SupportedAlgorithms = %v, want %v", got, tc.want)
+			}
```
elizabethhealy: nit: use assert/require

elizabethhealy (dismissed):
left a few comments but i dont think any of them are necessarily blocking

service/internal/security/in_process_provider.go
dmihalcik-virtru: Generally I like to use Any for non-string types, since this defers stringification to the logline serializer

service/internal/security/basic_manager.go
dmihalcik-virtru: I'll do it in a followup

service/trust/delegating_key_service.go
dmihalcik-virtru: ugh, yeah this is bad, since it means the factories are initialized too early in the application life cycle - and every time config is reloaded. The intention of the key provider factory is they are lazily created on first key access. I'll change the pattern here and resubmit

dmihalcik-virtru: okay, now the kas starter has to declare the supported algorithms on registration

service/kas/kas.go
```diff
@@ -28,7 +29,8 @@ func OnConfigUpdate(p *access.Provider) serviceregistry.OnConfigUpdateHook {
 		}
 
 		p.ApplyConfig(kasCfg, p.SecurityConfig())
-		p.Logger.TraceContext(ctx, "kas config reloaded")
+		p.Logger.TraceContext(ctx, "kas config reloaded", slog.Any("config", kasCfg))
```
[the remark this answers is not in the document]
dmihalcik-virtru: `kasCfg` implements `LogValuer`

service/kas/kas.go
```diff
@@ -28,7 +29,8 @@ func OnConfigUpdate(p *access.Provider) serviceregistry.OnConfigUpdateHook {
...
+		logSupportedMechanisms(ctx, p.Logger, p.KeyDelegator, &kasCfg)
```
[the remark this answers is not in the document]
dmihalcik-virtru: access.KASConfig implements LogValuer which should skip all secrets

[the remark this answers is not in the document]
dmihalcik-virtru: yeah, I've double checked that it does redact rootkey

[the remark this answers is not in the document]
dmihalcik-virtru: https://github.com/opentdf/platform/blob/79ab34fe4dd1e42af52acd6416dc5e02a3c2ee1b/service/kas/access/provider.go#L161

elizabethhealy (approved):

PR #3568, opened 2026-06-04
feat(policy): dynamic attribute value entitlement mappings

alkalescent:
### Proposed Changes

Implements the `DynamicValueMapping` policy primitive end to end (the model chosen by the spike / ADR 0005). It raises entitlement authority from a concrete `AttributeValue` to the `AttributeDefinition`: a single mapping entitles dynamically-requested values by comparing the requested resource value segment against the entity representation at decision time, instead of pre-provisioning a value + subject mapping per discrete ID (MRNs, account IDs, etc.).

Design highlights:

* **Operator model**: `DynamicValueResolver` reuses `SubjectMappingOperatorEnum` (`IN` = the resource segment equals any resolved entity value; `IN_CONTAINS` = substring). `NOT_IN` is rejected because dynamic resolution is existential over the resolved values. This keeps the resolver and subject-mapping vocabularies aligned and avoids a parallel operator enum. Follows the operator revert in #3685; the earlier decomposed `comparison`/`quantifier`/`case_insensitive` model was undone.
* **Optional static `SubjectConditionSet` pre-gate** (normal static semantics, no field overload) to support compound "entity attribute AND resource value" conditions.
* **No-coexistence enforced both directions** between value-level subject mappings and dynamic mappings on the same definition (attribute *values* may still exist, e.g. for obligation triggers).
* **`HIERARCHY` rejected**; matching is exact and case-sensitive.
* **Experimental, off by default** behind the `allow_dynamic_value_mappings` config; dynamic mappings are loaded (including in the authz cache), evaluated in decisioning, and gated in the cache refresh only when enabled.
* Dynamic values are first-class in decisioning: they flow into the same entitlements map and `ANY_OF` / `ALL_OF` rule layer as static values (manifest stays a list of attribute FQNs).

Wiring: proto + generated code + SDK client; DB migration + sqlc + CRUD; dedicated `DynamicValueMappingService`; decision-time evaluator; PDP load/merge with synthetic-value support; and authz cache. Replaces the throwaway spike package and ports its tests. `GetEntitlements` intentionally does not enumerate dynamic values (high-cardinality, resource-scoped).

### Checklist

### Testing Instructions

- `cd service && go build ./...`
- `cd service && go test ./internal/subjectmappingbuiltin/... ./internal/access/v2/... ./authorization/v2/...`
- `cd service && golangci-lint run ./internal/subjectmappingbuiltin/... ./internal/access/v2/ ./authorization/v2/ ./policy/db/ ./policy/dynamicvaluemapping/`
- DB-layer CRUD and end-to-end GetDecision verification require a live Postgres/platform.

### Related

- Depends on #3685 (operator revert — the current direction)
- ADR findings: `service/policy/adr/0005-dynamic-attribute-value-entitlements-spike.md`
- Protocol PR: #3580 (merged); SDK PR: #3635 (merged)

## Summary by CodeRabbit

* **New Features**
  * Added definition-scoped dynamic value mappings with full CRUD and end-to-end authorization support (create/list/get/update/delete).
  * Introduced `allow_dynamic_value_mappings` flag to enable/disable dynamic mapping evaluation and related direct entitlement synthesis.
  * Extended entitlement cache to expose listing of all cached dynamic value mappings.
* **Bug Fixes**
  * Improved entitlement cache refresh correctness to avoid marking the cache “ready” after partial failures.
  * Added validation and enforcement for dynamic mappings, including rejecting unsupported resolver operators and preventing conflicting coexistence/rule updates.
* **Tests**
  * Added integration and unit test coverage for dynamic mapping evaluation, pagination, updates/deletes, and safety restrictions.
* **Documentation**
  * Added ADR/spike and schema/migration docs for dynamic value mappings.


service/policy/db/definition_value_entitlement_mappings.go
```diff
@@ -0,0 +1,397 @@
...
+	actions                      interface{}
+	metadata                     []byte
+	namespace                    interface{}
+}
+
+func (c PolicyDBClient) CreateDefinitionValueEntitlementMapping(ctx context.Context, r *definitionvalueentitlement.CreateDefinitionValueEntitlementMappingRequest) (*policy.DefinitionValueEntitlementMapping, error) {
+	resolver := r.GetValueResolver()
+	if resolver.GetOperator() == policy.DynamicValueOperatorEnum_DYNAMIC_VALUE_OPERATOR_ENUM_UNSPECIFIED {
+		return nil, errors.Join(db.ErrEnumValueInvalid, errors.New("value_resolver.operator must be specified"))
+	}
```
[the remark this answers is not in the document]
jakedoublev: accessors are nil-safe

service/policy/db/migrations/20260604000000_add_definition_value_entitlement_mappings.sql
```diff
@@ -0,0 +1,62 @@
...
+);
+
+CREATE INDEX idx_definition_value_entitlement_mappings_definition_id
+  ON definition_value_entitlement_mappings(attribute_definition_id);
+CREATE INDEX idx_definition_value_entitlement_mappings_scs_id
+  ON definition_value_entitlement_mappings(subject_condition_set_id);
+CREATE INDEX idx_definition_value_entitlement_mappings_namespace_id
+  ON definition_value_entitlement_mappings(namespace_id);
+CREATE INDEX idx_definition_value_entitlement_mapping_actions_mapping_action
+  ON definition_value_entitlement_mapping_actions(definition_value_entitlement_mapping_id, action_id);
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/db/migrations/20260604000000_add_definition_value_entitlement_mappings.sql
```diff
@@ -0,0 +1,62 @@
...
+
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+
+DROP INDEX IF EXISTS idx_definition_value_entitlement_mapping_actions_mapping_action;
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/definitionvalueentitlement/definition_value_entitlement.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/opentdf/platform/service/pkg/serviceregistry"
+	policyconfig "github.com/opentdf/platform/service/policy/config"
+	policydb "github.com/opentdf/platform/service/policy/db"
+)
+
+type DefinitionValueEntitlementMappingService struct { //nolint:revive // descriptive name mirrors the policy object
+	dbClient policydb.PolicyDBClient
+	logger   *logger.Logger
+	config   *policyconfig.Config
+}
```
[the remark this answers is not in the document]
alkalescent: deferred for consistency

service/policy/db/definition_value_entitlement_mappings.go
```diff
@@ -0,0 +1,397 @@
...
+		},
+	}, nil
+}
+
+func (c PolicyDBClient) UpdateDefinitionValueEntitlementMapping(ctx context.Context, r *definitionvalueentitlement.UpdateDefinitionValueEntitlementMappingRequest) (*policy.DefinitionValueEntitlementMapping, error) {
+	id := r.GetId()
+	before, err := c.GetDefinitionValueEntitlementMapping(ctx, id)
+	if err != nil || before == nil {
+		return nil, db.WrapIfKnownInvalidQueryErr(err)
+	}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
```diff
@@ -0,0 +1,148 @@
...
+func EvaluateDefinitionValueEntitlementMappingsWithActions(
+	mappingsByDefinitionFQN DefinitionValueEntitlementMappingsByDefinitionFQN,
+	decisionableAttributes map[string]*attributes.GetAttributeValuesByFqnsResponse_AttributeAndValue,
+	entityRepresentation *entityresolutionV2.EntityRepresentation,
+	l *slog.Logger,
+) (AttributeValueFQNsToActions, error) {
+	entitlementsSet := make(AttributeValueFQNsToActions)
+	if len(mappingsByDefinitionFQN) == 0 {
+		return entitlementsSet, nil
+	}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+		}
+		return false, nil
+	case policy.DynamicValueOperatorEnum_DYNAMIC_VALUE_OPERATOR_ENUM_RESOURCE_VALUE_IN_CONTAINS:
+		for _, ev := range entityValues {
+			if strings.Contains(canonicalizeValueSegment(fmt.Sprintf("%v", ev)), target) {
+				return true, nil
+			}
+		}
+		return false, nil
```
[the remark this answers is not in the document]
alkalescent: fixed

service/authorization/v2/cache.go
```diff
@@ -270,6 +282,28 @@ func (c *EntitlementPolicyCache) ListAllSubjectMappings(ctx context.Context) ([]
...
+			return mappings, nil
+		}
+		return nil, fmt.Errorf("%w, definition value entitlement mappings: %w", ErrFailedToGet, err)
+	}
+
+	mappings, ok = cached.([]*policy.DefinitionValueEntitlementMapping)
+	if !ok {
+		return nil, fmt.Errorf("%w: %T", ErrCachedTypeNotExpected, mappings)
+	}
+	return mappings, nil
```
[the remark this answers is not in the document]
alkalescent: fixed

service/internal/access/v2/helpers.go
```diff
@@ -251,24 +255,32 @@ func getResourceDecisionableAttributes(
...
-			logger.DebugContext(ctx, "processing direct entitlement for resource decisionable attribute value", slog.String("attribute_value_fqn", attrValueFQN))
-
-			// try to find the definition by extracting partial FQN from direct entitlement synthetic value FQN
-			parentDefinition, err := getDefinition(attrValueFQN, entitleableAttributesByDefinitionFQN)
 			if err != nil {
-				// if definition not found, add to not found list and skip to next attribute value FQN
+				// definition not found: add to not found list and skip
 				notFoundFQNs = append(notFoundFQNs, attrValueFQN)
 				continue
 			}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/internal/access/v2/pdp.go
```diff
@@ -160,6 +186,43 @@ func NewPolicyDecisionPoint(
...
+			l.WarnContext(ctx,
+				"dynamic value mapping references unknown attribute definition - skipping",
+				slog.String("dynamic_value_mapping_id", mapping.GetId()),
+				slog.String("attribute_definition_fqn", definitionFQN),
+			)
+			continue
+		}
+
+		dynamicMappingsByDefinitionFQN[definitionFQN] = append(dynamicMappingsByDefinitionFQN[definitionFQN], mapping)
+	}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/db/dynamic_value_mappings.go
```diff
@@ -0,0 +1,405 @@
...
+	}
+
+	targetNamespaceID := before.GetNamespace().GetId()
+	if actions := r.GetActions(); actions != nil {
+		actionIDs, err := c.resolveSubjectMappingActions(ctx, actions, pgtypeUUID(targetNamespaceID))
+		if err != nil {
+			return nil, err
+		}
+		updateParams.ActionIds = actionIDs
+	}
```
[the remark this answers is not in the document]
alkalescent: keeping this consistent with `UpdateSubjectMapping` in which we also do not re-run the validate function (only on create)

service/policy/dynamicvaluemapping/dynamic_value_mapping.go
```diff
@@ -0,0 +1,189 @@
...
+		auditParams.ObjectID = mapping.GetId()
+		auditParams.Original = mapping
+		s.logger.Audit.PolicyCRUDSuccess(ctx, auditParams)
+
+		rsp.DynamicValueMapping = mapping
+		return nil
+	})
+	if err != nil {
+		return nil, db.StatusifyError(ctx, s.logger, err, db.ErrTextCreationFailed, slog.String("dynamicValueMapping", req.Msg.String()))
+	}
```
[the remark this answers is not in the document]
alkalescent: This is consistent with our other services (SM).

service/policy/dynamicvaluemapping/dynamic_value_mapping.go
```diff
@@ -0,0 +1,189 @@
...
+	return func(_ context.Context, cfg config.ServiceConfig) error {
+		sharedCfg, err := policyconfig.GetSharedPolicyConfig(cfg)
+		if err != nil {
+			return fmt.Errorf("failed to get shared policy config: %w", err)
+		}
+		svc.config = sharedCfg
+		svc.dbClient = policydb.NewClient(svc.dbClient.Client, svc.logger, int32(sharedCfg.ListRequestLimitMax), int32(sharedCfg.ListRequestLimitDefault))
+		svc.logger.Info("dynamic value mapping service config reloaded")
+		return nil
+	}
```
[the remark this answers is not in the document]
alkalescent: This is consistent with our other services (SM).

service/policy/db/subject_mappings.go
```diff
@@ -266,6 +266,13 @@ func (c PolicyDBClient) DeleteAllUnmappedSubjectConditionSets(ctx context.Contex
 // If a new subject condition set is provided, it will be created. The existing subject condition set id takes precedence.
 func (c PolicyDBClient) CreateSubjectMapping(ctx context.Context, s *subjectmapping.CreateSubjectMappingRequest) (*policy.SubjectMapping, error) {
 	attributeValueID := s.GetAttributeValueId()
+
+	// Enforce no-coexistence: a value-level subject mapping cannot be created on a
+	// definition that already has a dynamic value entitlement mapping (DSPX-2754 / ADR 0005).
+	if err := c.ensureNoDynamicValueMappingCoexistence(ctx, attributeValueID); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
alkalescent: This is consistent with our other policy objects (SM).

service/policy/db/queries/dynamic_value_mappings.sql
```diff
@@ -0,0 +1,220 @@
...
+    ),
+    update_count AS (
+        SELECT COUNT(*) AS cnt
+        FROM mapping_update
+    )
+SELECT cnt
+FROM update_count;
+
+-- name: deleteDynamicValueMapping :execrows
+DELETE FROM dynamic_value_mappings WHERE id = $1;
```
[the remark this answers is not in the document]
alkalescent: This is consistent with our other policy objects (SM).

service/policy/db/queries/dynamic_value_mappings.sql
```diff
@@ -0,0 +1,220 @@
...
+            sqlc.arg('id'),
+            a
+        FROM unnest(sqlc.narg('action_ids')::UUID[]) AS a
+        WHERE
+            sqlc.narg('action_ids')::UUID[] IS NOT NULL
+            AND NOT EXISTS (
+                SELECT 1
+                FROM dynamic_value_mapping_actions
+                WHERE dynamic_value_mapping_id = sqlc.arg('id') AND action_id = a
+            )
```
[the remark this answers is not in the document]
alkalescent: keeping this consistent with `updateSubjectMapping`

service/integration/dynamic_value_mappings_test.go
```diff
@@ -0,0 +1,236 @@
...
+	})
+	s.Require().NoError(err)
+
+	resp, err := s.db.PolicyClient.ListDynamicValueMappings(s.ctx, &dynamicvaluemapping.ListDynamicValueMappingsRequest{
+		AttributeDefinitionId: attr.GetId(),
+	})
+	s.Require().NoError(err)
+	s.Require().Len(resp.GetDynamicValueMappings(), 1)
+	s.Equal(attr.GetId(), resp.GetDynamicValueMappings()[0].GetAttributeDefinition().GetId())
+}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/db/queries/dynamic_value_mappings.sql
```diff
@@ -0,0 +1,220 @@
...
+    ),
+    update_count AS (
+        SELECT COUNT(*) AS cnt
+        FROM mapping_update
+    )
+SELECT cnt
+FROM update_count;
```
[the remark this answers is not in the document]
alkalescent: keeping SM convention

service/integration/dynamic_value_mappings_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// second page: remaining item, no further pages
+	second, err := s.db.PolicyClient.ListDynamicValueMappings(s.ctx, &dynamicvaluemapping.ListDynamicValueMappingsRequest{
+		AttributeDefinitionId: attr.GetId(),
+		Pagination:            &policy.PageRequest{Limit: 2, Offset: 2},
+	})
+	s.Require().NoError(err)
+	s.Len(second.GetDynamicValueMappings(), 1)
+	s.Equal(int32(3), second.GetPagination().GetTotal())
+	s.Equal(int32(0), second.GetPagination().GetNextOffset())
```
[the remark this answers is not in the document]
alkalescent: fixed

service/internal/access/v2/helpers.go
```diff
@@ -251,24 +255,29 @@ func getResourceDecisionableAttributes(
 		attributeAndValue, ok := entitleableAttributesByValueFQN[attrValueFQN]
 
 		if !ok {
-			// if the attribute value FQN is not found, then check if direct entitlements with synthetic values are enabled (experimental)
-			if !allowDirectEntitlements {
-				// if disabled, add to not found list and skip to next attribute value FQN
+			// The value FQN is not a concrete policy value. A synthetic value is created
+			// when either direct entitlements are enabled (experimental) OR the parent
+			// definition carries a dynamic value entitlement mapping (DSPX-2754), since
```
jakedoublev: Nit: avoid ticket numbers in comments

service/internal/access/v2/just_in_time_pdp.go
```diff
@@ -91,8 +91,12 @@ func NewJustInTimePDP(
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch all obligations: %w", err)
 	}
+	allDynamicValueMappings, err := store.ListAllDynamicValueMappings(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch all dynamic value mappings: %w", err)
+	}
 
-	pdp, err := NewPolicyDecisionPoint(ctx, log, allAttributes, allSubjectMappings, allRegisteredResources, allowDirectEntitlements, namespacedPolicy)
+	pdp, err := NewPolicyDecisionPointWithDynamicValueMappings(ctx, log, allAttributes, allSubjectMappings, allDynamicValueMappings, allRegisteredResources, allowDirectEntitlements, namespacedPolicy)
```
jakedoublev: For this feature to be initially experimental, I think we need to be able to disable it like we can direct entitlements with `allowDirectEntitlements` as an argument here.

service/internal/access/v2/pdp.go
```diff
@@ -85,6 +86,31 @@ func NewPolicyDecisionPoint(
 	allRegisteredResources []*policy.RegisteredResource,
 	allowDirectEntitlements bool,
 	namespacedPolicy bool,
+) (*PolicyDecisionPoint, error) {
+	return NewPolicyDecisionPointWithDynamicValueMappings(
```
jakedoublev: I've noticed an LLM preference to keep old function signatures around and hide them with new signatures beneath, just like this. Was this intentional or agentic? I think we should carefully consider whether or not two NewPolicyDecisionPoint functions makes sense. The eventual goal is to move the PDP out of the `internal` package and expose it as a library to PEPs or playgrounds.

service/internal/access/v2/pdp.go
```diff
@@ -160,6 +186,55 @@ func NewPolicyDecisionPoint(
...
+		if err := validateDynamicValueMapping(mapping); err != nil {
+			l.WarnContext(ctx,
+				"invalid dynamic value mapping - skipping",
+				slog.Any("dynamic_value_mapping", mapping),
+				slog.Any("error", err),
+			)
+			continue
+		}
+
+		if namespacedPolicy {
```
jakedoublev: We should never encounter dynamic value mappings that are not namespaced since they're a new object and we are enforcing namespacing completely in the policy CRUD, right?

service/internal/access/v2/pdp.go
```diff
@@ -160,6 +186,55 @@ func NewPolicyDecisionPoint(
...
+					slog.String("attribute_definition_fqn", mapping.GetAttributeDefinition().GetFqn()),
+				)
+				continue
+			}
+		}
+
+		definitionFQN := mapping.GetAttributeDefinition().GetFqn()
+		canonicalDef, ok := allAttributesByDefinitionFQN[definitionFQN]
+		if !ok {
+			l.WarnContext(ctx,
```
jakedoublev: Is there a scenario that makes this likely, or was this purely defensive?

service/internal/access/v2/pdp.go
```diff
@@ -160,6 +186,55 @@ func NewPolicyDecisionPoint(
...
+				slog.String("dynamic_value_mapping_id", mapping.GetId()),
+				slog.String("attribute_definition_fqn", definitionFQN),
+			)
+			continue
+		}
+
+		// Defense in depth alongside validateDynamicValueMapping: the mapping's own definition may
+		// carry an unset rule, so reject HIERARCHY using the canonical definition.
+		if canonicalDef.GetRule() == policy.AttributeRuleTypeEnum_ATTRIBUTE_RULE_TYPE_ENUM_HIERARCHY {
+			l.WarnContext(ctx,
```
jakedoublev: This seems like an error case. We probably still want to make the decision but definitely have something messed up in the db/policy to have a dynamic value mapping to a hierarchical attribute definition. IMO, we should log at the error level since there is corrective action involved to rectify it.

service/internal/access/v2/pdp.go
```diff
@@ -301,6 +378,24 @@ func (p *PolicyDecisionPoint) GetDecision(
 		}
 	}
 
+	// Evaluate dynamic, definition-level value entitlement mappings (DSPX-2754) and merge
```
jakedoublev: Nit: avoid ticket numbers in codebase unless they're TODOs

service/internal/access/v2/pdp.go
```diff
@@ -301,6 +378,24 @@ func (p *PolicyDecisionPoint) GetDecision(
...
+	// their results into the entitled FQNs before rule evaluation.
+	if len(p.dynamicMappingsByDefinitionFQN) > 0 {
+		dynamicEntitledFQNsToActions, err := subjectmappingbuiltin.EvaluateDynamicValueMappingsWithActions(
+			p.dynamicMappingsByDefinitionFQN,
+			decisionableAttributes,
+			entityRepresentation,
+			l.Logger,
+		)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error evaluating dynamic value mappings: %w", err)
```
jakedoublev: We should use a named error here. IIRC, there's an open TODO to refactor to named errors throughout, but we should avoid extending the size of the refactor.

service/internal/access/v2/pdp.go
```diff
@@ -301,6 +378,24 @@ func (p *PolicyDecisionPoint) GetDecision(
...
+		}
+		for fqn, actions := range dynamicEntitledFQNsToActions {
+			entitledFQNsToActions[fqn] = append(entitledFQNsToActions[fqn], actions...)
```
jakedoublev, line 378: Should we be deduplicating actions? Should we error if we've found both static entitlement and dynamic entitlement to the same attribute value?

service/internal/access/v2/validators.go
```diff
@@ -127,6 +127,40 @@ func validateAttribute(attribute *policy.Attribute) error {
...
+//   - must reference an attribute definition with a non-empty FQN
+//   - the definition must not be HIERARCHY (ordered static values are incompatible)
+//   - must have a value resolver with a selector and a specified operator
+//   - must have at least one action
+func validateDynamicValueMapping(mapping *policy.DynamicValueMapping) error {
+	if mapping == nil {
+		return fmt.Errorf("dynamic value mapping is nil: %w", ErrInvalidDynamicValueMapping)
+	}
+	def := mapping.GetAttributeDefinition()
+	if def == nil || def.GetFqn() == "" {
```
jakedoublev: nil check is not necessary due to proto getters

service/internal/access/v2/validators.go
```diff
@@ -127,6 +127,40 @@ func validateAttribute(attribute *policy.Attribute) error {
...
+		return fmt.Errorf("mapping's attribute definition is missing: %w", ErrInvalidDynamicValueMapping)
+	}
+	if def.GetRule() == policy.AttributeRuleTypeEnum_ATTRIBUTE_RULE_TYPE_ENUM_HIERARCHY {
+		return fmt.Errorf("HIERARCHY definitions are not supported for dynamic value entitlement: %w", ErrInvalidDynamicValueMapping)
+	}
+	resolver := mapping.GetValueResolver()
+	if resolver == nil || resolver.GetSubjectExternalSelectorValue() == "" {
```
jakedoublev: nil check is not necessary due to proto getters

service/policy/db/subject_mappings.go
```diff
@@ -266,6 +266,13 @@ func (c PolicyDBClient) DeleteAllUnmappedSubjectConditionSets(ctx context.Contex
 // If a new subject condition set is provided, it will be created. The existing subject condition set id takes precedence.
 func (c PolicyDBClient) CreateSubjectMapping(ctx context.Context, s *subjectmapping.CreateSubjectMappingRequest) (*policy.SubjectMapping, error) {
 	attributeValueID := s.GetAttributeValueId()
+
+	// Enforce no-coexistence: a value-level subject mapping cannot be created on a
+	// definition that already has a dynamic value entitlement mapping (DSPX-2754 / ADR 0005).
```
jakedoublev: Nit: avoid ticket references in codebase

service/policy/dynamicvaluemapping/dynamic_value_mapping.go
```diff
@@ -0,0 +1,189 @@
...
+	id := req.Msg.GetId()
+	s.logger.DebugContext(ctx, "updating dynamic value mapping", slog.String("id", id))
+
+	auditParams := audit.PolicyEventParams{
+		ActionType: audit.ActionTypeUpdate,
+		ObjectType: audit.ObjectTypeDynamicValueMapping,
+		ObjectID:   id,
+	}
+
+	original, err := s.dbClient.GetDynamicValueMapping(ctx, id)
```
jakedoublev: Should we use a transaction here?

service/policy/adr/0005-dynamic-attribute-value-entitlements-spike.md
```diff
@@ -0,0 +1,140 @@
+# Dynamic Attribute Value Entitlement
+
+Entitling highly dynamic, high-cardinality attribute values (medical record numbers, account IDs,
+email-like identifiers) is impractical today: each value must be duplicated as an `AttributeValue` and
+paired with its own `SubjectMapping` + `SubjectConditionSet`, then kept constantly in sync with an
+external system of record. The cross-repo ADR [virtru-corp/adr#266](https://github.com/virtru-corp/adr/pull/266)
+chose a definition-level dynamic entitlement model (its Option 3) but **explicitly deferred to an
+implementation spike** the question of *how* to model it. This document records what that spike
+([DSPX-2754](https://virtru.atlassian.net/browse/DSPX-2754)) found.
```
jakedoublev: Please remove references to Virtru repos and atlassian instances

service/policy/adr/0005-dynamic-attribute-value-entitlements-spike.md
```diff
@@ -0,0 +1,140 @@
...
+- **Canonicalization** (@biscoe916): external systems disagree with policy on case and whitespace. Without a
+  normalization step, `MRN-123` from the IdP fails to match `mrn-123` in the FQN. The spike applies a
+  pluggable `Canonicalizer` (default: lowercase + trim) to both sides. `TestCanonicalization` shows the
+  match succeed with it and fail without it. A real implementation must decide where canonicalization is
+  authoritative and whether it is configurable per definition.
+- **Cross-Definition / Namespace Collisions** (@jakedoublev): because entitlement is keyed to the value's
+  *parent definition FQN*, the same pass-through segment under a different definition is **not** granted
+  (`TestCrossDefinitionNoLeak`). This is the key advantage of entitling concrete value FQNs over entitling
+  bare pass-through values.
+- **Multi-Value Resources** (ADR decision-flow step 6): a single resource carrying several values under
```
jakedoublev, line 126: I wasn't sure where to put this question, but is the intention that you can or cannot have values on a definition and use a DynamicValueMapping? If the entity attribute value is always mapped 1:1 with the attribute value (the essence of 'dynamic'), should we enforce that those values don't exist? It seems like there are cases where they need to for things like obligation triggers, right?

service/policy/db/dynamic_value_mappings.go
```diff
@@ -0,0 +1,412 @@
...
+	actions                      interface{}
+	metadata                     []byte
+	namespace                    interface{}
+}
+
+func (c PolicyDBClient) CreateDynamicValueMapping(ctx context.Context, r *dynamicvaluemapping.CreateDynamicValueMappingRequest) (*policy.DynamicValueMapping, error) {
+	resolver := r.GetValueResolver()
+	if resolver.GetComparison() == policy.ConditionComparisonOperatorEnum_CONDITION_COMPARISON_OPERATOR_ENUM_UNSPECIFIED {
+		return nil, errors.Join(db.ErrEnumValueInvalid, errors.New("value_resolver.comparison must be specified"))
+	}
```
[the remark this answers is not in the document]
alkalescent: irrelevant now

service/authorization/v2/cache.go
```diff
@@ -178,6 +180,10 @@ func (c *EntitlementPolicyCache) Refresh(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	dynamicValueMappings, err := c.retriever.ListAllDynamicValueMappings(ctx)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/dynamicvaluemapping/dynamic_value_mapping.go
alkalescent: We could. I was trying to follow convention in the repo. Would you rather do it piecemeal or all at once?
Edit: made the transaction change

service/policy/adr/0005-dynamic-attribute-value-entitlements-spike.md
alkalescent: done

service/policy/db/subject_mappings.go
alkalescent: done

service/internal/access/v2/validators.go
alkalescent: removed unnecessary nil check

service/internal/access/v2/helpers.go
alkalescent: done

service/internal/access/v2/pdp.go
alkalescent: done

service/integration/dynamic_value_mappings_test.go
```diff
@@ -0,0 +1,282 @@
...
+func (s *DynamicValueMappingsSuite) TestRejectsHierarchyDefinition() {
+	attr := s.createDefinition("dvem_hierarchy", policy.AttributeRuleTypeEnum_ATTRIBUTE_RULE_TYPE_ENUM_HIERARCHY)
+
+	_, err := s.db.PolicyClient.CreateDynamicValueMapping(s.ctx, &dynamicvaluemapping.CreateDynamicValueMappingRequest{
+		AttributeDefinitionId: attr.GetId(),
+		ValueResolver:         s.resolver(".x[]", policy.ConditionComparisonOperatorEnum_CONDITION_COMPARISON_OPERATOR_ENUM_EQUALS),
+		Actions:               []*policy.Action{s.readAction()},
+	})
+	s.Require().Error(err, "HIERARCHY definitions must be rejected")
+}
```
alkalescent, line 96: asserted `ErrEnumValueInvalid`

service/internal/access/v2/validators.go
alkalescent: removed unnecessary nil check

service/internal/access/v2/just_in_time_pdp.go
alkalescent: changed to `NewJustInTimePDP(..., allowDynamicValueMappings)`

service/internal/access/v2/pdp.go
alkalescent: agentic/not intentional
consolidated to a single constructor with a `WithDynamicValueMappings` option

service/policy/adr/0005-dynamic-attribute-value-entitlements-spike.md
alkalescent: Values should be able to exist on a definition like for obligation triggers. However, static subject mappings being associated with a value should be mutually exclusive with a dynamic mapping. Updated the ADR a few lines down.

service/internal/access/v2/pdp.go
alkalescent: `ensureNoValueSubjectMappingCoexistence` ensures a value cannot carry both static and dynamic entitlements and actions are deduped with `dedupeSubjectMappingActions`

service/internal/access/v2/pdp.go
alkalescent: uses `ErrDynamicValueMappingEvaluation` now

service/internal/access/v2/pdp.go
alkalescent: Okay, we now reject HIERARCHY, log at `ErrorContext`, and skip that mapping while the decision proceeds.

service/internal/access/v2/pdp.go
alkalescent: True, removed the namespace skip.

service/internal/access/v2/pdp.go
alkalescent: Purely defensive, but prob overkill so removed.

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
```diff
@@ -0,0 +1,145 @@
...
+				entitlementsSet[valueFQN] = append(
+					entitlementsSet[valueFQN],
+					dedupeSubjectMappingActions(mapping.GetActions(), l)...,
+				)
+			}
+		}
+	}
+
+	return entitlementsSet, nil
+}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/integration/dynamic_value_mappings_test.go
```diff
@@ -0,0 +1,283 @@
...
+	s.Require().NoError(err)
+	s.Equal(".accounts[]", updated.GetValueResolver().GetSubjectExternalSelectorValue())
+	s.Equal(policy.SubjectMappingOperatorEnum_SUBJECT_MAPPING_OPERATOR_ENUM_IN_CONTAINS, updated.GetValueResolver().GetOperator())
+
+	_, err = s.db.PolicyClient.DeleteDynamicValueMapping(s.ctx, created.GetId())
+	s.Require().NoError(err)
+
+	_, err = s.db.PolicyClient.GetDynamicValueMapping(s.ctx, created.GetId())
+	s.Require().Error(err, "mapping should be gone after delete")
+}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/integration/dynamic_value_mappings_test.go
```diff
@@ -0,0 +1,283 @@
...
+func (s *DynamicValueMappingsSuite) TestRejectsHierarchyDefinition() {
+	attr := s.createDefinition("dvem_hierarchy", policy.AttributeRuleTypeEnum_ATTRIBUTE_RULE_TYPE_ENUM_HIERARCHY)
+
+	_, err := s.db.PolicyClient.CreateDynamicValueMapping(s.ctx, &dynamicvaluemapping.CreateDynamicValueMappingRequest{
+		AttributeDefinitionId: attr.GetId(),
+		ValueResolver:         s.resolver(".x[]", policy.SubjectMappingOperatorEnum_SUBJECT_MAPPING_OPERATOR_ENUM_IN),
+		Actions:               []*policy.Action{s.readAction()},
+	})
+	s.Require().ErrorIs(err, db.ErrEnumValueInvalid, "HIERARCHY definitions must be rejected")
+}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/authorization/v2/cache_test.go
```diff
@@ -21,7 +21,7 @@ func Test_NewEntitlementPolicyCache(t *testing.T) {
 	refreshInterval := 10 * time.Second
 	mockCache, _ := cache.TestCacheClient(mockCacheExpiry)
 
-	c, err := NewEntitlementPolicyCache(ctx, l, nil, mockCache, refreshInterval)
+	c, err := NewEntitlementPolicyCache(ctx, l, nil, mockCache, refreshInterval, false)
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/db/dynamic_value_mappings.go
```diff
@@ -0,0 +1,421 @@
...
+
+	return c.GetDynamicValueMapping(ctx, createdID)
+}
+
+func (c PolicyDBClient) GetDynamicValueMapping(ctx context.Context, id string) (*policy.DynamicValueMapping, error) {
+	row, err := c.queries.getDynamicValueMapping(ctx, id)
+	if err != nil {
+		return nil, db.WrapIfKnownInvalidQueryErr(err)
+	}
+	if row.ID == "" {
```
jakedoublev: I think we would have gotten `ErrNoRows` from line 90 if the row wasn't there, and every row has to be an ID, so this seems like dead code?

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
```diff
@@ -0,0 +1,158 @@
...
+
+// evaluateDynamicValueMapping returns true when the optional static gate
+// passes (if present) AND the dynamic resolver matches the resource value segment.
+func evaluateDynamicValueMapping(
+	mapping *policy.DynamicValueMapping,
+	entity flattening.Flattened,
+	segment string,
+) (bool, error) {
+	// optional static pre-gate: all subject sets AND together with normal semantics
```
jakedoublev: Can you please double check and add a comment around expected behavior of multiple subject sets being AND or OR? This logic looks like AND (if any subject set is false, the SCS is rejected), but we should double-check that behavior within Subject Mappings for consistency.

jakedoublev: We should also ensure test coverage for whichever is the valid relationship of multiple Subject Sets.

service/policy/dynamicvaluemapping/dynamic_value_mapping.go
```diff
@@ -0,0 +1,198 @@
...
+			Namespace:      ns,
+			DB:             dbRegister,
+			ServiceDesc:    &dvm.DynamicValueMappingService_ServiceDesc,
+			ConnectRPCFunc: dynamicvaluemappingconnect.NewDynamicValueMappingServiceHandler,
+			OnConfigUpdate: onUpdateConfigHook,
+			RegisterFunc: func(srp serviceregistry.RegistrationParams) (dynamicvaluemappingconnect.DynamicValueMappingServiceHandler, serviceregistry.HandlerServer) {
+				logger := srp.Logger
+				cfg, err := policyconfig.GetSharedPolicyConfig(srp.Config)
+				if err != nil {
+					logger.Error("error getting dynamic value mapping service policy config", slog.String("error", err.Error()))
```
jakedoublev: Nit: prefer `slog.Any("error", err)`

service/policy/db/subject_mappings.go
```diff
@@ -266,6 +266,13 @@ func (c PolicyDBClient) DeleteAllUnmappedSubjectConditionSets(ctx context.Contex
 // If a new subject condition set is provided, it will be created. The existing subject condition set id takes precedence.
 func (c PolicyDBClient) CreateSubjectMapping(ctx context.Context, s *subjectmapping.CreateSubjectMappingRequest) (*policy.SubjectMapping, error) {
 	attributeValueID := s.GetAttributeValueId()
+
+	// Enforce no-coexistence: a value-level subject mapping cannot be created on a
+	// definition that already has a dynamic value entitlement mapping.
+	if err := c.ensureNoDynamicValueMappingCoexistence(ctx, attributeValueID); err != nil {
```
jakedoublev, line 272: Two questions here:
1. Is this called within a transaction from the service layer?
2. Is there any path to update an existing Subject Mapping on its own or somehow through updating an Attr Value that would lead to it moving under an Attr Value on a different Attr Definition where a DVM could exist?

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
```diff
@@ -0,0 +1,158 @@
...
+		if len(mappings) == 0 {
+			continue
+		}
+
+		segment, err := resourceValueSegment(valueFQN, attributeAndValue.GetValue())
+		if err != nil {
+			return nil, err
+		}
+
+		// mappings on the same definition are OR-ed together
```
jakedoublev, line 70: Do we have this OR behavior adequately documented in the protos, migration, and docs? Is this the same behavior as with Subject Mappings?

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin_test.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// TestEvaluateDynamicValueMappings_InContains covers the substring operator.
+func TestEvaluateDynamicValueMappings_InContains(t *testing.T) {
+	const def = "https://acme.co/attr/group"
+	const valueFQN = "https://acme.co/attr/group/value/team"
+	mapping := dvemMapping(def, ".groups[]", policy.SubjectMappingOperatorEnum_SUBJECT_MAPPING_OPERATOR_ENUM_IN_CONTAINS, nil, "read")
+	byDef := DynamicValueMappingsByDefinitionFQN{def: {mapping}}
+
+	got, err := EvaluateDynamicValueMappingsWithActions(byDef, dvemDecisionable(def, valueFQN, "team"), dvemEntityRep(t, map[string]interface{}{"groups": []interface{}{"prefix-team-suffix"}}), slog.Default())
```
jakedoublev, line 104: Does this IN_CONTAINS operator seem particularly dangerous here? If the substring was `admin`, `admin-readonly`, `org-admin`, `superadmin`, `admin-ou123` etc would all match by design. I don't know if it's actually possible to leak data by crafting FQNs that result in access decisions, but it definitely seems possible to accidentally leak it to one of those `admin` parties while only intending to give access to one of them. 🤔

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
```diff
@@ -0,0 +1,158 @@
...
+func evaluateValueResolver(resolver *policy.DynamicValueResolver, entity flattening.Flattened, segment string) (bool, error) {
+	operator := resolver.GetOperator()
+	var match func(entityValue string) bool
+	switch operator {
+	case policy.SubjectMappingOperatorEnum_SUBJECT_MAPPING_OPERATOR_ENUM_IN:
+		match = func(entityValue string) bool { return entityValue == segment }
+	case policy.SubjectMappingOperatorEnum_SUBJECT_MAPPING_OPERATOR_ENUM_IN_CONTAINS:
+		match = func(entityValue string) bool { return strings.Contains(entityValue, segment) }
+	case policy.SubjectMappingOperatorEnum_SUBJECT_MAPPING_OPERATOR_ENUM_NOT_IN:
+		return false, errors.New("NOT_IN is unsupported for dynamic value resolution")
```
jakedoublev: Can we please make these named errors so the auth service / PDP layer are able to catch them and log appropriately that something is wrong within Policy if this is encountered?

service/internal/access/v2/pdp.go
```diff
@@ -355,6 +434,7 @@ func (p *PolicyDecisionPoint) GetDecisionRegisteredResource(
 		p.allRegisteredResourceValuesByFQN,
 		p.allEntitleableAttributesByValueFQN,
 		p.allAttributesByDefinitionFQN, /*action, */
+		p.dynamicMappingsByDefinitionFQN,
```
jakedoublev, line 437: Sanity checking my understanding: Since the Registered Resource's (RR) associated Action-Attribute-Values (AAVs) _are_ the entitlements of a registered resource when it's treated as the entity in a decision request flow, Dynamic Value Mappings (DVMs) are irrelevant to the decision? 

This seems like a pretty tricky edge case for Registered Resources with Dynamic Value Mappings, because:
1. Registered Resources (as entities in the decision request) are entitled to their associated Action-Attribute-Values. There is no entity representation, and therefore nothing for a Dynamic Value Mapping to be evaluated against. 
2. When the Registered Resource is the resource FQN in the decision flow, the associated Action-Attribute-Values, filtered by the decision request Action, are the required entitlements for a PERMIT decision. 

With the logic in this PR, the Attribute Values mapped to the RR via AAVs might be dynamically entitled, which means they're only relevant to a decision flow where the RR is the resource but not the entity. This seems really confusing. 

What do you think about drawing a line in the sand that you cannot add any Attribute Values under an Attribute Definition with a DVM to a Registered Resource's Action-Attribute-Values list? This is a similar logic to what we're doing with Subject Mappings on the Definition's Values.

jakedoublev: cc @ttschampel

service/authorization/v2/cache.go
```diff
@@ -200,6 +217,12 @@ func (c *EntitlementPolicyCache) Refresh(ctx context.Context) error {
 		return errors.Join(ErrFailedToSet, err)
 	}
 
+	err = c.cacheClient.Set(ctx, dynamicValueMappingsCacheKey, dynamicValueMappings, authzCacheTags)
```
jakedoublev: Should this be conditional on `c.allowDynamicValueMappings` so we don't set an empty slice to the cache?

service/policy/db/dynamic_value_mappings.go
alkalescent: removed

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
alkalescent: SM also AND multiple subject sets — added a comment

service/policy/dynamicvaluemapping/dynamic_value_mapping.go
alkalescent: fixed

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
alkalescent: OR inter/across mappings, AND intra/within a mapping's subject sets — same as subject mappings.
Documented in the policy ADR 0005 and [these protos](https://github.com/opentdf/platform/pull/3745).

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin.go
alkalescent: Created named errors (`ErrDynamicResolverOperator*`) so the PDP/authz layer can check them and log.

service/authorization/v2/cache.go
alkalescent: added a check to avoid setting an empty slice if `c.allowDynamicValueMappings` is false

service/internal/subjectmappingbuiltin/dynamic_value_mapping_builtin_test.go
alkalescent: Def have to be careful using the IN_CONTAINS operator, but it is keeping parity with SM. 
Documented in the policy ADR 0005 and https://github.com/opentdf/platform/pull/3745.

service/policy/db/subject_mappings.go
alkalescent: 1. Yes, the `CreateSubjectMapping` wraps the coexistence check + insert in `RunInTx`.
2. No, `UpdateSubjectMapping` never changes `attribute_value_id`, and a value can't change its definition, so a mapping can't migrate onto a DVM definition after creation.

service/internal/access/v2/pdp.go
alkalescent: Agree with keeping parity with SM. Added a guard rejecting attribute values whose definition has a DVM from being added to a RR's AAVs, mirroring the value-level SM coexistence rule.

service/policy/db/migrations/20260618000000_add_dynamic_value_mappings.sql
jakedoublev, line 1: Can you please update this to include a companion doc for the migration that links to the ADR so it follows the existing convention, and run `make policy-erd-gen` to regenerate the ERD?

service/integration/dynamic_value_mappings_test.go
```diff
@@ -0,0 +1,332 @@
...
+	})
+	s.Require().NoError(err)
+
+	regRes, err := s.db.PolicyClient.CreateRegisteredResource(s.ctx, &registeredresources.CreateRegisteredResourceRequest{
+		NamespaceId: s.f.GetNamespaceKey("example.com").ID,
+		Name:        "dvem_rr_coexist_res",
+	})
+	s.Require().NoError(err)
+
+	// A value under a definition with a dynamic mapping must not be added to a registered resource's
```
jakedoublev, line 170: Can you please add tests for any paths from the other direction, i.e. where the attribute value is already on a RR and a DVM is added to it?

jakedoublev (approved):

service/policy/db/migrations/20260618000000_add_dynamic_value_mappings.sql
alkalescent: added doc and regen'd ERD

service/integration/dynamic_value_mappings_test.go
alkalescent: added a check and test

Issue #3335, opened 2026-04-20
Decompose subject mapping operators into comparison, quantifier, and modifier axes


jrschumacher:
## Problem

The current `SubjectMappingOperatorEnum` conflates three concerns into a single enum:

| Current operator | String comparison | Quantifier | Negation |
|---|---|---|---|
| `IN` | exact match | any-of | no |
| `NOT_IN` | exact match | none-of | yes |
| `IN_CONTAINS` | substring | any-of | no |

This creates several issues:

1. **Missing operators** — there's no `STARTS_WITH` or `ENDS_WITH`, so there's no safe way to match email domains. `IN_CONTAINS` with `@acme.com` would also match `user@acme.com.badactor.ru`.
2. **Enum sprawl** — adding `STARTS_WITH`, `ENDS_WITH`, and their `NOT_` and `CASE_INSENSITIVE_` variants would require 20+ enum values.
3. **No quantifier control** — there's no way to express "the entity must have ALL of these values" without chaining multiple AND'd single-value conditions.

## Proposal

Decompose the condition into three orthogonal axes:

### 1. Comparison operator (how to compare two strings)

```protobuf
enum ConditionComparisonOperator {
  CONDITION_COMPARISON_OPERATOR_EQUALS = 0;
  CONDITION_COMPARISON_OPERATOR_CONTAINS = 1;
  CONDITION_COMPARISON_OPERATOR_STARTS_WITH = 2;
  CONDITION_COMPARISON_OPERATOR_ENDS_WITH = 3;
}
```

### 2. Quantifier (how to aggregate across value lists)

```protobuf
enum ConditionQuantifier {
  CONDITION_QUANTIFIER_ANY = 0;   // at least one match (current IN behavior)
  CONDITION_QUANTIFIER_ALL = 1;   // every expected value is matched
  CONDITION_QUANTIFIER_NONE = 2;  // no matches (current NOT_IN behavior)
}
```

### 3. Case sensitivity modifier (bool flag)

```protobuf
bool case_insensitive = 7;
```

### Updated Condition message

```protobuf
message Condition {
  string subject_external_selector_value = 1;
  repeated string subject_external_values = 2;

  // New decomposed fields
  ConditionComparisonOperator comparison = 5;
  ConditionQuantifier quantifier = 6;
  bool case_insensitive = 7;

  // Deprecated — old conflated operator, kept for backward compat
  SubjectMappingOperatorEnum operator = 3 [deprecated = true];
}
```

### Combination examples

[...]

PR #3580, opened 2026-06-05
feat(policy): DSPX-2754 DynamicValueMapping protos + generated code

alkalescent:
### Proposed Changes

* Protocol-first half of the dynamic attribute value entitlement feature (DSPX-2754). Adds the `DynamicValueMapping` / `DynamicValueResolver` messages and `DynamicValueOperatorEnum` to `objects.proto`, plus a dedicated `DynamicValueMappingService` in a new `policy.dynamicvaluemapping` package. Includes regenerated `protocol/go` and OpenAPI/gRPC docs.
* No service implementation, DB, SDK client, or PDP wiring here. Those land in the consumer PR.

**Why split:** the dedicated service introduces a new `protocol/go` package. The per-module `go mod tidy` CI check does not use the go workspace, so consumer modules cannot tidy against a package that is not yet in a released `protocol/go`. Landing and releasing `protocol/go` first lets the consumer PR require the new version and pass CI.

### Checklist

### Testing Instructions

- `cd protocol/go && go build ./... && GOFLAGS=-mod=mod go mod tidy` (clean)
- `buf lint service`

### Related

- Jira: https://virtru.atlassian.net/browse/DSPX-2754 (spike), https://virtru.atlassian.net/browse/DSPX-3498 (implementation)
- Consumer PR (service + sdk + DB + PDP): #3568 (rebase + bump protocol/go after this releases)

## Release Notes

**New Features**
* Added Dynamic Value Mapping service with full CRUD operations (Create, Read, Get, List, Update, Delete)
* Dynamic Value Mappings enable runtime attribute evaluation using custom resolvers and operators
* List operations support pagination and filtering by namespace and attribute definition
* Update operations allow optional modification of resolvers, conditions, and permitted actions


service/policy/dynamicvaluemapping/dynamic_value_mapping.proto
```diff
@@ -0,0 +1,173 @@
...
+    id: "optional_uuid_format"
+    message: "Optional field must be a valid UUID"
+    expression: "size(this) == 0 || this.matches('[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}')"
+  }];
+  string namespace_fqn = 8 [
+    (buf.validate.field).string = {
+      min_len: 0
+      uri: true
+    }
+  ];
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/dynamicvaluemapping/dynamic_value_mapping.proto
```diff
@@ -0,0 +1,171 @@
...
+  // Required: Attribute Definition ID or FQN to scope the mapping to
+  option (buf.validate.message).oneof = {
+    fields: ["attribute_definition_id", "attribute_definition_fqn"]
+    required: true
+  };
+  // Optional: Namespace ID or FQN to scope the mapping to
+  option (buf.validate.message).oneof = {
+    fields: ["namespace_id", "namespace_fqn"]
+    required: false
+  };
```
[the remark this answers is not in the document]
alkalescent: repo convention

service/policy/dynamicvaluemapping/dynamic_value_mapping.proto
```diff
@@ -0,0 +1,171 @@
...
+
+message DynamicValueMappingsSort {
+  SortDynamicValueMappingsType field = 1 [(buf.validate.field).enum.defined_only = true];
+  policy.SortDirection direction = 2 [(buf.validate.field).enum.defined_only = true];
+}
+
+message ListDynamicValueMappingsRequest {
+  // Optional
+  // Namespace ID, or Attribute Definition ID to filter by
+  string namespace_id = 1 [(buf.validate.field).cel = {
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/objects.proto
```diff
@@ -180,6 +180,21 @@ enum ConditionBooleanTypeEnum {
...
 
+// Operators for dynamic, definition-level value entitlement. Unlike
+// SubjectMappingOperatorEnum, whose right-hand operand is a static list authored into
+// policy, a DynamicValueOperatorEnum's right-hand operand is the requested resource's
+// attribute value segment, supplied at decision time. Each value is the inversion of its
+// static SubjectMappingOperatorEnum counterpart.
+enum DynamicValueOperatorEnum {
+  DYNAMIC_VALUE_OPERATOR_ENUM_UNSPECIFIED = 0;
+  // true when the requested resource value segment equals one of the values resolved by
+  // the selector against the entity representation (inversion of IN)
```
jakedoublev: I'm not totally understanding the note of the `inversion`. I wonder if we could clarify this some more?

jakedoublev: Have you considered making this change at the same time, which would make the operators more explicit and durable? https://github.com/opentdf/platform/issues/3335

service/policy/objects.proto
```diff
@@ -204,6 +219,53 @@ message SubjectMapping {
...
+  // the Attribute Definition whose values are entitled dynamically
+  Attribute attribute_definition = 2;
+
+  // the dynamic resolver matched against the requested resource value segment
+  DynamicValueResolver value_resolver = 3;
+
+  // optional static pre-gate on the entity, evaluated with normal SubjectConditionSet
+  // semantics (no dynamic overload). When present, both the gate and the resolver must
+  // pass for entitlement.
+  SubjectConditionSet subject_condition_set = 4;
```
jakedoublev, line 275: I like this idea of letting the dynamic value mapping also be static. That's a great design suggestion not included in the original ADR. Did you intend to make it a oneof in the service layer so you could only add a static or dynamic condition and not both, or should we add a oneof in the protovalidate layer here?

jakedoublev (dismissed):

service/policy/objects.proto
alkalescent: made #3335 decomposition changes

service/policy/objects.proto
alkalescent: No, I don't believe we need a oneof because `value_resolver` is required on create, so every mapping is dynamic. Basically, `subject_condition_set` is optional and gets ANDed with the resolver. With the #3335 decomposition changes: now the static gate and resolver share the same operators, so we can combine them vs making them mutually exclusive.

service/policy/objects.proto
```diff
@@ -204,6 +230,58 @@ message SubjectMapping {
...
+
+  // how the requested resource value segment is compared against each value the selector
+  // resolves from the entity representation
+  ConditionComparisonOperatorEnum comparison = 2 [
+    (buf.validate.field).enum.defined_only = true,
+    (buf.validate.field).required = true
+  ];
+
+  // when true, comparison is case-insensitive (default false = case-sensitive)
+  bool case_insensitive = 3;
```
jakedoublev: I think we should probably use `google.protobuf.BoolValue` here. We've run into issues before with zero values/defaults using just `bool`, and the explicit nature of defined and false vs undefined is useful if we want to make case insensitivity a default later on. Most of the strings throughout policy are lowercased/insensitive.

service/policy/objects.proto
```diff
@@ -213,16 +291,27 @@ message Condition {
...
   repeated string subject_external_values = 3 [(buf.validate.field).repeated.min_items = 1];
+
+  // how each selector result is compared to subject_external_values entries
+  ConditionComparisonOperatorEnum comparison = 4 [(buf.validate.field).enum.defined_only = true];
+
+  // how matches are aggregated across subject_external_values
+  ConditionQuantifierEnum quantifier = 5 [(buf.validate.field).enum.defined_only = true];
+
+  // when true, comparison is case-insensitive (default false = case-sensitive)
+  bool case_insensitive = 6;
```
jakedoublev: Same here with `google.protobuf.BoolValue`

service/policy/objects.proto
```diff
@@ -213,16 +291,27 @@ message Condition {
   // from idP/LDAP)
   string subject_external_selector_value = 1 [(buf.validate.field).required = true];
 
-  // the evaluation operator of relation
+  // Deprecated: use comparison + quantifier (+ case_insensitive) instead. Normalized to the
```
jakedoublev, line 295: So the plan is for the service implementation to map the old operators to the new ones? Full backwards compatibility with existing conditions?

service/policy/objects.proto
jakedoublev: Interesting! I didn't think of it that way. I was thinking you'd maybe want a static mapping at the definition level to entitle every value under that definition (only the SCS instead of the resolver). It sounds like you're thinking the employment of SCSs as you've designed would be more of a precursor to the dynamic value? Do you have an example of a use case?

service/policy/objects.proto
alkalescent: updated

service/policy/objects.proto
alkalescent: Thanks for the catch, updated.

service/policy/objects.proto
alkalescent: Yes. The deprecated operator field is retained, and the service PR (#3568) normalizes it into comparison + quantifier when those are unset (e.g. IN → EQUALS + ANY, NOT_IN → EQUALS + NONE, IN_CONTAINS → CONTAINS + ANY). New payloads set the decomposed fields directly, and existing SCS work without changes. So, we should have full backwards compatibility.

service/policy/objects.proto
alkalescent: Yeah, I was thinking of SCS as a static precondition ANDed with the dynamic resolver instead of a standalone definition-level grant. 
Example (from my understanding of [your ADR](https://github.com/virtru-corp/adr/blob/main/docs/decisions/dspx-2559-opentdf-policy-dynamic-attribute-value-entitlements.md)): 
- a definition `https://hospital.co/attr/mrn` whose values are individual patient IDs/MRNs. The resolver entitles each requester to the IDs present in their own `.patientAssignments[]` claim (resource value ∈ entity-resolved values), and the optional SCS gates who participates at all, e.g. `.role` EQUALS `clinician`, so a non-clinician never gets dynamic entitlement even on a matching ID.

jakedoublev (approved):

service/policy/objects.proto
jakedoublev: Cool! I like the expressiveness.

PR #3581, opened 2026-06-08
feat(sdk): DSPX-3848 add DPoP client support with HTTP RoundTripper

dmihalcik-virtru:
## Summary

Adds RFC 9449 DPoP (Demonstrating Proof-of-Possession) support to the Go SDK, so HTTP/ConnectRPC calls carry a DPoP proof and can use sender-constrained tokens. Part of the Keycloak v26 / DPoP effort tracked in [DSPX-62](https://virtru.atlassian.net/browse/DSPX-62).

## What changed

**SDK**
- New `DPoPTransport` (`http.RoundTripper`) that adds a DPoP proof to every request — token endpoint and resource calls alike. Composes with any `http.Client`.
- Handles the `DPoP-Nonce` challenge (RFC 9449 §8): on a `401` carrying a nonce, cache it, re-sign, and retry once; refresh the cached nonce from `2xx` responses. Nonce cache is per-origin and thread-safe.
- RFC 9449 URI normalization for `htu` (lowercase scheme/host, strip default ports, drop query/fragment, preserve escaped path bytes).
- Key resolution via `WithDPoPKeyPEM` / `WithDPoPAlgorithm` / JWK; defaults to an ephemeral EC P-256 (ES256) key when none is supplied.
- `SupportedFeatures()` reports `["dpop", "dpop_nonce_challenge", "connectrpc"]` so the xtest harness can probe capabilities.

**otdfctl**
- DPoP is negotiated automatically for authenticated calls — no flags.
- Credential validation runs through the DPoP-bound path; the printed/reusable token stays a plain bearer so it works outside otdfctl (DSPX-3998).
- `--version --json` now includes `supported_features`.

**service**
- Host-header origin normalization for DPoP `htu` validation.

## Testing

- `go test ./...` and `cd sdk && go test -run TestREADMECodeBlocks` pass; lint clean.
- Cross-SDK e2e (go/java/web) via `opentdf/tests` xtest: 🟢 https://github.com/opentdf/tests/actions/runs/29624686479

**Jira**: [DSPX-3397](https://virtru.atlassian.net/browse/DSPX-3397)

[DSPX-3397]: https://virtru.atlassian.net/browse/DSPX-3397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSPX-3397]: https://virtru.atlassian.net/browse/DSPX-3397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DSPX-62]: https://virtru.atlassian.net/browse/DSPX-62?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/otdfctl/pkg/auth/auth.go
+++ b/otdfctl/pkg/auth/auth.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
 	"strconv"
@@ -184,7 +185,9 @@ func ValidateProfileAuthCredentials(ctx context.Context, profile *profiles.Otdfc
 	case "":
 		return ErrProfileCredentialsNotFound
 	case profiles.AuthTypeClientCredentials:
-		_, err := GetTokenWithClientCreds(ctx, profile.GetEndpoint(), c.ClientID, c.ClientSecret, profile.GetTLSNoVerify(), c.Scopes)
+		// Validation exercises the DPoP-bound path so it succeeds against a
+		// DPoP-enforcing token endpoint; the token is discarded.
+		_, err := GetTokenWithClientCredsDPoP(ctx, profile.GetEndpoint(), c.ClientID, c.ClientSecret, profile.GetTLSNoVerify(), c.Scopes)
 		if err != nil {
 			return err
 		}
@@ -204,6 +207,8 @@ func GetTokenWithProfile(ctx context.Context, profile *profiles.OtdfctlProfileSt
 
 	switch c.AuthType {
 	case profiles.AuthTypeClientCredentials:
+		// print or reuse path: return a plain bearer token (not DPoP sender-constrained)
+		// so it stays usable outside otdfctl. See DSPX-3998.
 		return GetTokenWithClientCreds(ctx, profile.GetEndpoint(), c.ClientID, c.ClientSecret, profile.GetTLSNoVerify(), c.Scopes)
 	case profiles.AuthTypeAccessToken:
 		return buildToken(&c), nil
@@ -212,12 +217,30 @@ func GetTokenWithProfile(ctx context.Context, profile *profiles.OtdfctlProfileSt
 	}
 }
 
-// Uses the OAuth2 client credentials flow to obtain a token.
+// GetTokenWithClientCreds uses the OAuth2 client credentials flow to obtain a
+// bearer token with no sender constraint.
 func GetTokenWithClientCreds(ctx context.Context, endpoint string, clientID string, clientSecret string, tlsNoVerify bool, scopes []string) (*oauth2.Token, error) {
+	return getTokenWithClientCreds(ctx, endpoint, clientID, clientSecret, tlsNoVerify, scopes, false)
+}
+
+// GetTokenWithClientCredsDPoP uses sender-constrained tokens, which are not currently exportable.
+func GetTokenWithClientCredsDPoP(ctx context.Context, endpoint string, clientID string, clientSecret string, tlsNoVerify bool, scopes []string) (*oauth2.Token, error) {
+	return getTokenWithClientCreds(ctx, endpoint, clientID, clientSecret, tlsNoVerify, scopes, true)
+}
+
+func getTokenWithClientCreds(ctx context.Context, endpoint string, clientID string, clientSecret string, tlsNoVerify bool, scopes []string, dpop bool) (*oauth2.Token, error) {
+	httpClient := utils.NewHTTPClient(tlsNoVerify)
+	if dpop {
+		var err error
+		httpClient, err = sdk.NewDPoPValidationHTTPClient(httpClient)
+		if err != nil {
+			return nil, err
+		}
+	}
 	rp, err := newOidcRelyingParty(ctx, endpoint, tlsNoVerify, oidcClientCredentials{
 		clientID:     clientID,
 		clientSecret: clientSecret,
-	})
+	}, httpClient)
 	if err != nil {
 		return nil, err
 	}
@@ -337,14 +360,14 @@ func RevokeAccessToken(ctx context.Context, endpoint, clientID, refreshToken str
 	rp, err := newOidcRelyingParty(ctx, endpoint, tlsNoVerify, oidcClientCredentials{
 		clientID: clientID,
 		isPublic: true,
-	})
+	}, utils.NewHTTPClient(tlsNoVerify))
 	if err != nil {
 		return err
 	}
 	return oidcrp.RevokeToken(ctx, rp, refreshToken, "refresh_token")
 }
 
-func newOidcRelyingParty(ctx context.Context, endpoint string, tlsNoVerify bool, clientCreds oidcClientCredentials) (oidcrp.RelyingParty, error) {
+func newOidcRelyingParty(ctx context.Context, endpoint string, tlsNoVerify bool, clientCreds oidcClientCredentials, httpClient *http.Client) (oidcrp.RelyingParty, error) {
 	if clientCreds.clientID == "" {
 		return nil, errors.New("client ID is required")
 	}
@@ -367,6 +390,6 @@ func newOidcRelyingParty(ctx context.Context, endpoint string, tlsNoVerify bool,
 		clientCreds.clientSecret,
 		"",
 		nil,
-		oidcrp.WithHTTPClient(utils.NewHTTPClient(tlsNoVerify)),
+		oidcrp.WithHTTPClient(httpClient),
 	)
 }
--- a/sdk/auth/dpop_transport.go
+++ b/sdk/auth/dpop_transport.go
@@ -1,0 +1,417 @@
+package auth
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/google/uuid"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/lestrrat-go/jwx/v2/jwt"
+)
+
+// DPoPTransport wraps each go standard net/http RoundTripper request with DPoP (RFC 9449)
+// proof tokens. These proofs are for both token endpoint (IdP, etc) calls and
+// resource (i.e. KAS or policy service, for the SDK) endpoint calls,
+// handling server-issued nonces with automatic retry.
+type DPoPTransport struct {
+	// Base is the underlying transport. If nil, http.DefaultTransport is used.
+	Base http.RoundTripper
+
+	// DPoPKey is the private key used to sign DPoP proofs.
+	DPoPKey jwk.Key
+
+	// TokenSource provides access tokens for resource requests.
+	// For resource requests (any URL other than TokenEndpoint), the transport
+	// sets Authorization: DPoP <token> and includes the ath claim binding the
+	// proof to the access token. Requests to TokenEndpoint get neither.
+	TokenSource AccessTokenSource
+
+	// TokenEndpoint is the OAuth token endpoint URL.
+	// Requests to this endpoint are treated as token requests
+	// and do not include the ath claim.
+	TokenEndpoint string
+
+	// tokenFetchTimeout bounds the internal access-token fetch performed while
+	// adding the ath claim to resource requests. It mirrors the configured
+	// client's Timeout so a hung IdP cannot stall the request indefinitely.
+	tokenFetchTimeout time.Duration
+
+	nonceOnce         sync.Once
+	nonceMu           sync.RWMutex
+	nonceCache        map[string]string
+	cachedTokenURL    *url.URL
+	cachedTokenURLStr string
+}
+
+var _ http.RoundTripper = (*DPoPTransport)(nil)
+
+// RoundTrip implements http.RoundTripper, adding DPoP proofs to requests.
+func (t *DPoPTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.DPoPKey == nil {
+		return nil, errors.New("DPoP transport has no signing key")
+	}
+
+	base := t.Base
+	if base == nil {
+		base = http.DefaultTransport
+	}
+
+	// NewDPoPHTTPClient initializes the cache; this Once covers a directly
+	// constructed transport without paying a write-lock on every request.
+	t.nonceOnce.Do(t.initNonceCache)
+
+	// Avoid modifying the original
+	req2 := cloneRequest(req)
+
+	// Buffer the body and install GetBody on the clone so a DPoP-Nonce retry
+	// can replay it. ConnectRPC/gRPC clients set Body and ContentLength but
+	// not GetBody, so without this the retry path would send an empty body
+	// against a non-zero ContentLength and net/http would abort the request.
+	if err := bufferRequestBody(req2); err != nil {
+		return nil, err
+	}
+
+	isTokenRequest := t.isTokenEndpointRequest(req2.URL)
+
+	origin := getOrigin(req2.URL)
+	nonce := t.getCachedNonce(origin)
+
+	if err := t.addDPoPProof(req2, base, nonce, isTokenRequest); err != nil {
+		return nil, fmt.Errorf("failed to add DPoP proof: %w", err)
+	}
+
+	resp, err := base.RoundTrip(req2)
+	if err != nil {
+		return resp, err
+	}
+
+	// Handle DPoP-Nonce challenge (RFC 9449 §8).
+	if resp.StatusCode == http.StatusUnauthorized ||
+		(resp.StatusCode == http.StatusBadRequest && resp.Header.Get("DPoP-Nonce") != "") {
+		retryResp, retried, err := t.retryWithNonce(req2, base, resp, origin, nonce, isTokenRequest)
+		if err != nil {
+			return nil, err
+		}
+		if retried {
+			resp = retryResp
+		}
+	}
+
+	// Handle DPoP-Nonce updates (RFC 9449 §8.2).
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		if newNonce := resp.Header.Get("DPoP-Nonce"); newNonce != "" {
+			t.setCachedNonce(origin, newNonce)
+		}
+	}
+
+	return resp, nil
+}
+
+// retryWithNonce handles a DPoP-Nonce server challenge. It returns the retried
+// response and true when a retry was performed, or the original response and
+// false when no retry was needed.
+//
+// A retry happens once per request whenever the server supplies a DPoP-Nonce
+// that differs from the one we already sent (RFC 9449 §8). Requiring a *different*
+// nonce both covers the initial challenge (we sent none) and a server that rotates
+// its nonce after previously accepting one, while preventing a retry loop when the
+// server keeps returning the same nonce we just used. The single retry is returned
+// as-is even if it is itself a 401.
+func (t *DPoPTransport) retryWithNonce(
+	req *http.Request, base http.RoundTripper,
+	resp *http.Response, origin, nonce string, isTokenRequest bool,
+) (*http.Response, bool, error) {
+	newNonce := resp.Header.Get("DPoP-Nonce")
+	if newNonce == "" || newNonce == nonce {
+		return resp, false, nil
+	}
+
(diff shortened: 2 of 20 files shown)


sdk/auth/dpop_transport.go
```diff
@@ -0,0 +1,275 @@
...
+			}
+
+			// Cache the new nonce
+			t.setCachedNonce(origin, newNonce)
+
+			// Close the failed response body
+			resp.Body.Close()
+
+			// Clone the original request again for retry
+			req3 := cloneRequest(req)
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed. Lines 98–104 now call `req.GetBody()` to reset the request body before retry.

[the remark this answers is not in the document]
dmihalcik-virtru: Already implemented — lines 98-104 reset the body via `GetBody` before retry:

```go
if req.GetBody != nil {
    body, err := req.GetBody()
    if err != nil {
        return nil, fmt.Errorf("failed to reset request body for retry: %w", err)
    }
    req3.Body = body
}
```

sdk/auth/dpop_transport.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	// For resource requests (not token endpoint), add ath claim
+	var accessToken string
+	if !isTokenRequest && t.TokenSource != nil {
+		at, err := t.TokenSource.AccessToken(req.Context(), nil)
+		if err != nil {
+			return fmt.Errorf("failed to get access token: %w", err)
+		}
+		accessToken = string(at)
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed. `addDPoPProof` now passes `&http.Client{Transport: base}` to `AccessToken`, preserving the configured transport chain while avoiding infinite recursion.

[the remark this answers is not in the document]
dmihalcik-virtru: Already implemented — lines 146-147 pass `&http.Client{Transport: base}` to `AccessToken`:

```go
client := &http.Client{Transport: base}
at, err := t.TokenSource.AccessToken(req.Context(), client)
```

This preserves custom transport config and avoids recursion back through `DPoPTransport`.

sdk/auth/dpop_transport.go
```diff
@@ -0,0 +1,275 @@
...
+		return false
+	}
+	tokenURL, err := url.Parse(t.TokenEndpoint)
+	if err != nil {
+		return false
+	}
+	return u.Scheme == tokenURL.Scheme &&
+		u.Host == tokenURL.Host &&
+		u.Path == tokenURL.Path
+}
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed. `isTokenEndpointRequest` now caches the parsed `*url.URL` in `cachedTokenURL`/`cachedTokenURLStr` using double-checked locking via the existing `nonceMu`. The struct fields were added at lines 41–42.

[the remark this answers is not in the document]
dmihalcik-virtru: Already implemented — `isTokenEndpointRequest` (lines 202-235) uses the double-checked locking pattern with `cachedTokenURL`/`cachedTokenURLStr` struct fields to cache the parsed URL across requests.

sdk/auth/dpop_transport.go
```diff
@@ -0,0 +1,275 @@
...
+		host = host[:strings.LastIndex(host, ":")]
+	}
+
+	return fmt.Sprintf("%s://%s%s", scheme, host, u.Path)
+}
+
+// getOrigin returns the origin (scheme://host:port) from a URL.
+func getOrigin(u *url.URL) string {
+	return fmt.Sprintf("%s://%s", u.Scheme, u.Host)
+}
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed. `getOrigin` now applies `strings.ToLower` to ensure consistent cache lookups regardless of URL casing.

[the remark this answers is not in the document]
dmihalcik-virtru: Already implemented — `getOrigin` (line 256) wraps the result in `strings.ToLower`:

```go
func getOrigin(u *url.URL) string {
    return strings.ToLower(fmt.Sprintf("%s://%s", u.Scheme, u.Host))
}
```

pflynn-virtru (dismissed):
no security concerns found

dmihalcik-virtru:
@pflynn-virtru I made a couple of small changes 

- Use a named string alias for JWA algorithms: https://github.com/opentdf/platform/pull/3581/changes/d54317467e00cf9aa2de4e3ff136f443ff9dea3e
- Pre-allocate copyTo buffer in unary grpc request flow: https://github.com/opentdf/platform/pull/3581/changes/630f0cda8af4bc4ad14f48d313d67d53ef439481

pflynn-virtru (dismissed):

jakedoublev:
Should we have clients upgrade to DPoP at the SDK level if the platform requires it? Some kind of well-known or published state that our SDKs are smart enough to handle?

otdfctl/cmd/common/common.go
```diff
@@ -86,7 +86,7 @@ func InitProfile(c *cli.Cli) *profiles.OtdfctlProfileStore {
 // TODO make this a preRun hook
 //
 //nolint:nestif // separate refactor [https://github.com/opentdf/otdfctl/issues/383]
-func NewHandler(c *cli.Cli) handlers.Handler {
+func NewHandler(c *cli.Cli, extraSDKOpts ...sdk.Option) handlers.Handler {
```
jakedoublev: This is an exported function that is consumed when the CLI is used as a library, so this is a breaking change to some consumers.

otdfctl/docs/man/encrypt/_index.md
```diff
@@ -37,6 +37,16 @@ command:
     - name: with-assertions
       description: >
         EXPERIMENTAL: JSON string or path to a JSON file of assertions to bind metadata to the TDF. See examples for more information. WARNING: Providing keys in a JSON string is strongly discouraged. If including sensitive keys, instead provide a path to a JSON file containing that information.
+    - name: dpop
```
jakedoublev: If these indeed are added as CLI flags, should they be relevant globally to every command or just encrypt/decrypt?

dmihalcik-virtru: These may apply to all authenticated connectRPC calls. The client will request DPoP enabled tokens when it hits the oauth server's access token endpoint.

otdfctl/cmd/common/common.go
dmihalcik-virtru: you are correct, go is persnickety about this

otdfctl/docs/man/encrypt/_index.md
jakedoublev: Is there a way we can get that state into the SDK without a flag, then? Or is the flag mostly for testing purposes? I ask because it doesn't seem very user-friendly to have to pass a `--dpop` flag when I would want my CLI to figure that out for me as its own concern, and I would also want it to be done correctly on all API calls instead of just encrypt/decrypt if they would fail otherwise. Does that make sense?

dmihalcik-virtru:
> Should we have clients upgrade to DPoP at the SDK level if the platform requires it? Some kind of well-known or published state that our SDKs are smart enough to handle?

I've removed the otdfctl dpop parameters. Now the SDK unconditionally creates ES256 DPoP keys if the backend + IdP support it. If we desire it (probably as part of ML-DSA support) we can expose mechanism preference as part of the SDK configuration and wire it through to the frontend

To let the tests skip if the feature is not implemented, I've exposed the `supported_features` array on the `otdfctl --version --json` command options. See [this test code](https://github.com/opentdf/tests/pull/529/changes#diff-d6dbd423a65b9ef82c2011c322ccfe5620818a80d01778089ad4cd6d1fe8ec89)

otdfctl/pkg/handlers/sdk.go
```diff
@@ -31,17 +31,17 @@ type handlerOpts struct {
 	sdkOpts []sdk.Option
 }
 
-type handlerOptsFunc func(handlerOpts) handlerOpts
+type HandlerOptsFunc func(handlerOpts) handlerOpts
```
jakedoublev: 🙏

dmihalcik-virtru: Thanks for the reminder! I've un-exported this type.

sdk/auth/dpop_transport.go
```diff
@@ -0,0 +1,395 @@
...
+func bufferRequestBody(req *http.Request) error {
+	if req.Body == nil || req.Body == http.NoBody || req.ContentLength < 0 {
+		return nil
+	}
+	// Pre-size from ContentLength (guaranteed >= 0 here) so this is a single exact
+	// allocation instead of io.ReadAll's grow-by-doubling. ConnectRPC unary payloads
+	// are already fully buffered in memory, so this is a straight copy of a
+	// known-length body. ReadFrom still reads to EOF, so behavior is unchanged if
+	// ContentLength ever under-/over-states the actual body.
+	buf := bytes.NewBuffer(make([]byte, 0, req.ContentLength))
```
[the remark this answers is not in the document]
dmihalcik-virtru: It does compile, but yes it will panic if too big. That said, fine we are already screwed at this point. Somebody put a value larger than maxint here and it is already in memory, good job, caller, you played yourself.

otdfctl/docs/man/encrypt/_index.md
dmihalcik-virtru: Good call — I removed the `--dpop`/`--dpop-key` flags entirely. otdfctl now negotiates DPoP automatically for every authenticated ConnectRPC call (the SDK requests DPoP-bound tokens at the OAuth token endpoint), so it is no longer per-command and there is nothing for the user to remember. Credential validation runs through the same DPoP-bound path (`GetTokenWithClientCredsDPoP`), and the only deliberate exception is the printed/reusable token, which stays a plain bearer so it is usable outside otdfctl (tracked in DSPX-3998).

sdk/auth/dpop_transport_test.go
```diff
@@ -0,0 +1,963 @@
...
+	for range n {
+		wg.Go(func() {
+			resp, err := client.Do(mustReq(t, server.URL))
+			if err != nil {
+				errs <- err
+				return
+			}
+			resp.Body.Close()
+		})
+	}
```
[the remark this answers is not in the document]
dmihalcik-virtru: `sync.WaitGroup.Go` was added in Go 1.25, and this module is on `go 1.25`. The package builds and `go vet ./sdk/auth/` passes, so this compiles as written.

sdk/version.go
```diff
@@ -9,3 +9,13 @@ const (
...
 )
+
+// SupportedFeatures returns a list of optional features supported by this SDK build.
+// Used by xtest integration harness for feature detection.
+func SupportedFeatures() []string {
+	return []string{
+		"dpop",                 // RFC 9449 DPoP (Demonstrating Proof-of-Possession)
+		"dpop_nonce_challenge", // RFC 9449 §8 server-issued DPoP-Nonce challenge/retry
+		"connectrpc",           // Connect RPC protocol support
+	}
```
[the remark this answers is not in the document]
dmihalcik-virtru: Advertising `dpop_nonce_challenge` is intentional (xtest probes it separately from base `dpop`). Updating the PR description to match rather than dropping the feature.

otdfctl/pkg/auth/auth.go
```diff
@@ -212,12 +217,36 @@ func GetTokenWithProfile(ctx context.Context, profile *profiles.OtdfctlProfileSt
 	}
 }
 
-// Uses the OAuth2 client credentials flow to obtain a token.
+// GetTokenWithClientCreds uses the OAuth2 client credentials flow to obtain a plain
+// bearer token with no sender constraint, so it remains usable outside otdfctl (e.g.
+// printed for reuse). Making an exported token reusable *under* DPoP (key export and/or
+// OS secure storage) is tracked in DSPX-3998.
```
dmihalcik-virtru:
```suggestion
// GetTokenWithClientCreds uses the OAuth2 client credentials flow to obtain a plain
// bearer token with no sender constraint. This is useful for flows that export the token for later use.
```

sdk/auth/dpop_transport.go
```diff
@@ -0,0 +1,417 @@
...
+	// TokenSource provides access tokens for resource requests.
+	// For resource requests (any URL other than TokenEndpoint), the transport
+	// sets Authorization: DPoP <token> and includes the ath claim binding the
+	// proof to the access token. Requests to TokenEndpoint get neither.
+	TokenSource AccessTokenSource
+
+	// TokenEndpoint is the OAuth token endpoint URL.
+	// Requests to this endpoint are treated as token requests
+	// and do not include the ath claim.
+	TokenEndpoint string
```
sujankota, line 53: `TokenEndpoint` is a public field read in `isTokenEndpointRequest` without the nonce mutex, while `cachedTokenURLStr` is protected by it. The double-checked lock on `cachedTokenURLStr != t.TokenEndpoint` is racy if a caller mutates `TokenEndpoint` after the transport is first used. `NewDPoPHTTPClient` pre-parses and caches it at construction, so mutation after that point is already a misuse — but the field doc comment does not say so. Worth adding: "TokenEndpoint must not be mutated after the transport is first used."

sdk/auth/dpop_transport_test.go
```diff
@@ -0,0 +1,963 @@
...
+	return AccessToken(m.token), nil
+}
+
+func (m *mockTokenSource) MakeToken(_ func(jwk.Key) ([]byte, error)) ([]byte, error) {
+	// Not used in transport tests
+	return nil, nil
+}
+
+func generateTestKey(t *testing.T) jwk.Key {
```
sujankota, line 69: `generateTestKey` returns an RSA-2048 key, so `TestDPoPTransport_AddsProofToRequests`, `TestDPoPTransport_NonceRetry`, and several other primary tests exercise RSA but not the ES256/P-256 path — which is both the production default and the recommended algorithm per `WithDPoPAlgorithm`. `TestDPoPTransport_ProofAlgorithms` does cover ES256 but as a secondary matrix test. Consider switching `generateTestKey` to ES256 so the main proof path matches production. A regression in the EC signing path would otherwise go undetected in the primary test.

sdk/version.go
```diff
@@ -9,3 +9,13 @@ const (
 	// The three-part semantic version number of this SDK
 	Version = "0.27.0" // x-release-please-version
 )
+
+// SupportedFeatures returns a list of optional features supported by this SDK build.
+// Used by xtest integration harness for feature detection.
+func SupportedFeatures() []string {
```
sujankota, line 20: Once `opentdf/tests` xtest starts gating tests on these strings (`"dpop"`, `"dpop_nonce_challenge"`, `"connectrpc"`), removing or renaming one will silently cause those tests to be `SKIPPED` rather than failing — a quiet regression. Worth adding a short stability contract to the doc comment, e.g.: "These strings are part of the stable API surface. Removing or renaming a feature string must be coordinated with opentdf/tests before merging."

service/internal/auth/authn.go
```diff
@@ -388,13 +390,50 @@ func (e *DPoPProofError) Error() string { return e.err.Error() }
...
+			slog.String("host", host),
+			slog.String("error", err.Error()))
+		return scheme + "://" + host
+	}
+	h := strings.ToLower(u.Hostname())
+	if strings.Contains(h, ":") {
+		h = "[" + h + "]" // re-bracket IPv6 literal stripped by Hostname()
+	}
+	if port := u.Port(); port != "" &&
+		!(scheme == httpsScheme && port == "443") && //nolint:staticcheck // QF1001: written as it would be understood (don't append port if the port is expected for its scheme)
```
sujankota, line 417: The `//nolint:staticcheck // QF1001` suppresses a quickfix suggestion to simplify `!(scheme == httpsScheme && port == "443")`. If the intent is readability, `scheme != httpsScheme || port != "443"` is logically equivalent, equally readable, and does not need a linter suppression. Suppression comments tend to accumulate and confuse future readers about whether the lint failure is real or intentional.

service/internal/auth/dpop_uri.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func isUnreserved(c byte) bool {
+	return c >= 'a' && c <= 'z' || c >= 'A' && c <= 'Z' || c >= '0' && c <= '9' ||
+		c == '-' || c == '.' || c == '_' || c == '~'
+}
+
+// removeDotSegments implements RFC 3986 section 5.2.4 without collapsing
+// repeated slashes, which are significant path data.
+func removeDotSegments(input string) string {
```
sujankota, line 83: `output += input[:n]` inside the loop allocates on every iteration (O(n²) for paths with many segments). Real HTTP paths are short so this will not matter in practice, but a `strings.Builder` would be idiomatic Go. Easy swap:

```go
var output strings.Builder
// ... replace output += with output.WriteString(...)
return output.String()
```

sdk/auth/dpop_transport.go
dmihalcik-virtru: Good catch — added a note to the `TokenEndpoint` field doc: it must not be mutated after the transport is first used, since `isTokenEndpointRequest` caches the parsed URL (and `NewDPoPHTTPClient` pre-parses it at construction). (dc8cc81c)

sdk/auth/dpop_transport_test.go
dmihalcik-virtru: Done. `generateTestKey` now returns the production-default ES256/P-256 key, so the primary tests exercise the shipped proof path. `TestDPoPTransport_AddsProofToRequests` and `TestDPoPTransport_NonceRetry` now run as subtests over both RS256 and ES256, so the core behaviors keep RSA coverage while the default path matches production. (dc8cc81c)

sdk/version.go
dmihalcik-virtru: Added a stability-contract note to the `SupportedFeatures` doc comment: these strings are stable API surface, and removing/renaming one must be coordinated with opentdf/tests first since xtest silently SKIPs (rather than fails) on an unknown feature string. (dc8cc81c)

service/internal/auth/authn.go
dmihalcik-virtru: Keeping the explicit `!(scheme == httpsScheme && port == "443")` form here intentionally — it reads as "don't append the port when it's the default for the scheme," which maps to the intent more directly than the De Morgan rewrite. The `//nolint:staticcheck // QF1001` documents that the quickfix is knowingly declined rather than an unnoticed lint failure, so leaving as-is.

service/internal/auth/dpop_uri.go
dmihalcik-virtru: Done — switched `removeDotSegments` to `strings.Builder`, removing the O(n^2) accumulation. Thanks. (dc8cc81c)

sdk/auth/dpop_transport.go
```diff
@@ -0,0 +1,422 @@
...
+			return fmt.Errorf("failed to get access token: %w", err)
+		}
+		accessToken = string(at)
+
+		// Calculate ath = base64url(SHA-256(access_token))
+		h := sha256.New()
+		h.Write([]byte(accessToken))
+		ath := base64.RawURLEncoding.EncodeToString(h.Sum(nil))
+		builder = builder.Claim("ath", ath)
+	}
```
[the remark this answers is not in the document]
dmihalcik-virtru: Good catch — fixed in ecee503.

`DPoPTransport` now resolves the resource-request credential up front via the optional `AccessTokenCredentialSource` interface (the same one the Connect interceptor uses). When the token source reports a non-DPoP scheme, the transport sets `Authorization: Bearer <token>`, adds no `ath`, and sends **no** DPoP proof — matching the interceptor's behavior instead of forcing every resource request onto DPoP. Sources that don't implement the interface still default to DPoP, so existing behavior is preserved; token-endpoint requests are unchanged (proof, no `ath`).

Added tests: a bearer credential source (asserts `Authorization: Bearer` and no `DPoP` header), a DPoP credential source (still signs proof + `ath`), and a unit test that the bearer path strips any inherited `DPoP` header on a retry clone.

sujankota (approved):

ryanulit (approved):

PR #3582, opened 2026-06-08
feat(core): Adds comprehensive DPoP (RFC 9449) support

dmihalcik-virtru:
Implements comprehensive DPoP (Demonstrating Proof-of-Possession) support per [RFC 9449](https://www.rfc-editor.org/rfc/rfc9449.html) for the OpenTDF platform service.

> Note: Providing DPoP support at the application server may not fit your deployment strategy! If you want a high-availability or multi-region support, consider implementing DPoP with an application gateway or proxy

## Summary

This PR adds full RFC 9449 DPoP support to the platform authentication middleware, enabling:

- ✅ Both `Authorization: Bearer` and `Authorization: DPoP` token schemes
- ✅ Complete DPoP proof validation (typ, alg, jwk, signature, htm, htu, ath, jkt)  
- ✅ Server-issued DPoP-Nonce challenges with rotation (RFC 9449 §8)
- ✅ gRPC support (htm=POST, full service paths as htu)
- ✅ Feature detection via wellknown service (`supports_dpop`)
- ✅ Comprehensive unit tests
- ✅ Extensive debug logging, which should be reduced after clients are up to spec

Since we have an existing, partial implementation, some fallback protection is in place:

- Continue to allow DPoP requiring JWTs in `Authorization: Bearer [jwt]` headers.
  - Emit WARN server message if a Bearer token has the `cnf` claim.
- Continue to support `htu` claims with missing origin, e.g. `/kas.AccessService/Rewrap` will match
   - To require full `htu`, set `server.auth.dpop.strict_htu: true`

### DPoP Proof Validation (RFC 9449 §4.3 + §7.1)

The middleware validates all required DPoP proof claims:
- **typ**: Must be `dpop+jwt`
- **alg**: Only asymmetric algorithms (ES256/384/512, RS256/384/512, PS256/384/512)
- **jwk**: Public key embedded in header
- **Signature**: Verified against embedded JWK
- **htm**: HTTP method must match request
- **htu**: Normalized URI must match request
- **ath**: SHA-256 hash of access token (base64url)
- **jkt**: RFC 7638 JWK thumbprint in access token's `cnf.jkt` claim

### DPoP-Nonce Support (RFC 9449 §8)

Server-issued nonces prevent replay attacks:
- **Configuration**: `server.auth.dpop.require_nonce` (default: false), `server.auth.dpop.nonce_expiration` (default: 5m)
- **Challenge flow**: Missing/invalid nonce → 401 with `DPoP-Nonce` header and `WWW-Authenticate: DPoP error="use_dpop_nonce"`
- **Validation window**: Current + previous nonce (allows graceful rotation)
- **Rotation**: Automatic based on expiration interval
- **HTTP & gRPC**: Works in both MuxHandler (HTTP) and ConnectUnaryServerInterceptor (gRPC)

### gRPC Support

DPoP works seamlessly with gRPC/Connect:
- **htm**: Always `POST` for gRPC calls
- **htu**: Full procedure path (e.g., `/kas.AccessService/Rewrap`)
- **DPoP-Nonce**: Propagated via response headers/trailers

### Feature Detection

Registers `supports_dpop: true` in the wellknown service for integration with xtest feature gates (`pfs.skip_if_unsupported("dpop")`).

### Unit Tests (`dpop_nonce_test.go`)

Comprehensive test coverage:
- Nonce generation, rotation, validation window
- Proof validation (signature, htm, htu, ath, jkt)
- Algorithm restrictions (asymmetric only)
- Error types and detection
- Token expiration handling

### Integration Tests

Existing DPoP tests in `test/integration/oauth/oauth_test.go` continue to pass and validate end-to-end flows with real Keycloak.

### End-to-End Tests

Validated end-to-end with the xtest DPoP scenario: https://github.com/opentdf/tests/pull/529.

> Note: The default `docker-compose.yaml` stays on Keycloak 25 so downstream consumers that aren't DPoP-ready are unaffected. The `start-up-with-containers` action overlays Keycloak 26.2 (+ `admin-fine-grained-authz:v1`) onto the compose file only when invoked with `dpop-challenge-enabled: true`. Wiring a CI job to pass that flag is left for a follow-up step.

## Related

- **Parent Jira**: [DSPX-3397](https://virtru.atlassian.net/browse/DSPX-3397)
- **Test scenario**: `xtest/scenarios/DSPX-3397.yaml`
- **SDK implementations**: platform-go-sdk, java-sdk, web-sdk (sibling PRs in this feature)

### All PRs:

- https://github.com/opentdf/platform/pull/3582
- https://github.com/opentdf/platform/pull/3581
- https://github.com/opentdf/java-sdk/pull/374
- https://github.com/opentdf/web-sdk/pull/939

[DSPX-3397]: https://virtru.atlassian.net/browse/DSPX-3397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added Demonstration of Proof-of-Possession (DPoP) nonce validation support for enhanced token security.
  * Enabled configurable DPoP settings including nonce enforcement, expiration policies, and validation modes.

* **Improvements**
  * Enhanced error classification and diagnostic logging for access control decisions and token operations.

### DIFF
--- a/sdk/auth/token_adding_interceptor.go
+++ b/sdk/auth/token_adding_interceptor.go
@@ -64,6 +64,11 @@ func (i TokenAddingInterceptor) AddCredentials(
 		return status.Error(codes.Unauthenticated, err.Error())
 	}
 
+	slog.DebugContext(
+		ctx, "preparing dpop for grpc request",
+		slog.String("grpc_method", method),
+		slog.String("dpop_htm", http.MethodPost),
+	)
 	dpopTok, err := i.GetDPoPToken(method, http.MethodPost, string(accessToken))
 	if err == nil {
 		newMetadata = append(newMetadata, "DPoP", dpopTok)
@@ -99,6 +104,12 @@ func (i TokenAddingInterceptor) AddCredentialsConnect() connect.UnaryInterceptor
 			req.Header().Set("Authorization", fmt.Sprintf("DPoP %s", accessToken))
 
 			// Add DPoP header if possible
+			slog.DebugContext(
+				ctx, "preparing dpop for connect request",
+				slog.String("procedure", req.Spec().Procedure),
+				slog.String("dpop_htm", http.MethodPost),
+				slog.Any("stream_type", req.Spec().StreamType),
+			)
 			dpopTok, err := i.GetDPoPToken(req.Spec().Procedure, http.MethodPost, string(accessToken))
 			if err == nil {
 				req.Header().Set("DPoP", dpopTok)
@@ -146,6 +157,11 @@ func (i TokenAddingInterceptor) GetDPoPToken(path, method, accessToken string) (
 		h.Write([]byte(accessToken))
 		ath := h.Sum(nil)
 
+		slog.Debug(
+			"building dpop token",
+			slog.String("htm", method),
+			slog.String("htu", path),
+		)
 		dpopTok, err := jwt.NewBuilder().
 			Claim("htu", path).
 			Claim("htm", method).
--- a/service/internal/auth/authn.go
+++ b/service/internal/auth/authn.go
@@ -362,10 +362,17 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 				origin = "http://" + strings.TrimSuffix(origin, ":80")
 			}
 		}
-		accessTok, ctx, err := a.checkToken(r.Context(), header, receiverInfo{
+		ri := receiverInfo{
 			u: []string{normalizeURL(origin, r.URL)},
 			m: []string{r.Method},
-		}, dp)
+		}
+		log.DebugContext(
+			r.Context(), "dpop receiverInfo set (http handler)",
+			slog.Any("expected_htm", ri.m),
+			slog.Any("expected_htu", ri.u),
+			slog.String("request_method", r.Method),
+		)
+		accessTok, ctx, err := a.checkToken(r.Context(), header, ri, dp)
 		if err != nil {
 			// Check if this is a nonce error requiring a challenge
 			var nonceErr *DPoPNonceError
@@ -484,6 +491,13 @@ func (a Authentication) ConnectAuthNInterceptor() connect.UnaryInterceptorFunc {
 				},
 				m: []string{req.HTTPMethod()},
 			}
+			log.DebugContext(
+				ctx, "dpop receiverInfo set (connect interceptor)",
+				slog.Any("expected_htm", ri.m),
+				slog.Any("expected_htu", ri.u),
+				slog.String("procedure", procedure),
+				slog.String("connect_http_method", req.HTTPMethod()),
+			)
 
 			header := req.Header()["Authorization"]
 			if len(header) < 1 {
@@ -884,6 +898,12 @@ func (a Authentication) validateDPoP(accessToken jwt.Token, acessTokenRaw string
 		return nil, errors.New("`htm` claim invalid format in DPoP JWT")
 	}
 
+	a.logger.Debug(
+		"dpop htm validation",
+		slog.String("received_htm", htm),
+		slog.Any("expected_htm", dpopInfo.m),
+		slog.Bool("match", slices.Contains(dpopInfo.m, htm)),
+	)
 	if !slices.Contains(dpopInfo.m, htm) {
 		return nil, fmt.Errorf("incorrect `htm` claim in DPoP JWT; received [%v], but should match [%v]", htm, dpopInfo.m)
 	}
@@ -962,13 +982,20 @@ func (a Authentication) ipcReauthCheck(ctx context.Context, path string, header 
 
 			// Validate the token and create a JWT token
 			ipcHost := header.Get("Host")
-			token, ctxWithJWK, err := a.checkToken(ctx, authHeader, receiverInfo{
+			ri := receiverInfo{
 				u: []string{
 					"http://" + ipcHost + path,
 					"https://" + ipcHost + path,
 				},
 				m: []string{http.MethodPost},
-			}, header["Dpop"])
+			}
+			a.logger.DebugContext(
+				ctx, "dpop receiverInfo set (ipc reauth)",
+				slog.Any("expected_htm", ri.m),
+				slog.Any("expected_htu", ri.u),
+				slog.String("path", path),
+			)
+			token, ctxWithJWK, err := a.checkToken(ctx, authHeader, ri, header["Dpop"])
 			if err != nil {
 				return nil, connect.NewError(connect.CodeUnauthenticated, errors.New("unauthenticated"))
 			}
--- a/service/internal/auth/authn.go
+++ b/service/internal/auth/authn.go
@@ -15,6 +15,7 @@ import (
 	"net/url"
 	"path"
 	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -225,6 +226,20 @@ func NewAuthenticator(ctx context.Context, cfg Config, logger *logger.Logger, we
 	// Register DPoP support feature
 	if err := wellknownRegistration("supports_dpop", true); err != nil {
 		logger.Warn("failed to register dpop support", slog.Any("error", err))
+	}
+
+	// Register supported DPoP JWT signing algorithms (RFC 9449 §5.1 dpop_signing_alg_values_supported)
+	supportedAlgs := make([]string, 0, len(allowedSignatureAlgorithms))
+	for alg := range allowedSignatureAlgorithms {
+		supportedAlgs = append(supportedAlgs, alg.String())
+	}
+	sort.Strings(supportedAlgs)
+	if err := wellknownRegistration("dpop_supported_alg_values", supportedAlgs); err != nil {
+		logger.Warn("failed to register dpop supported alg values", slog.Any("error", err))
+	}
+
+	if err := wellknownRegistration("dpop_nonce_required", cfg.DPoP.RequireNonce); err != nil {
+		logger.Warn("failed to register dpop nonce required", slog.Any("error", err))
 	}
 
 	return a, nil
--- a/service/internal/auth/authn.go
+++ b/service/internal/auth/authn.go
@@ -439,8 +439,13 @@ func (a Authentication) ConnectAuthNInterceptor() connect.UnaryInterceptorFunc {
 
 			log := a.logger
 
+			procedure := req.Spec().Procedure
+			host := req.Header().Get("Host")
 			ri := receiverInfo{
-				u: []string{req.Spec().Procedure},
+				u: []string{
+					"http://" + host + procedure,
+					"https://" + host + procedure,
+				},
 				m: []string{http.MethodPost},
 			}
 
@@ -920,8 +925,12 @@ func (a Authentication) ipcReauthCheck(ctx context.Context, path string, header 
 			}
 
 			// Validate the token and create a JWT token
+			ipcHost := header.Get("Host")
 			token, ctxWithJWK, err := a.checkToken(ctx, authHeader, receiverInfo{
-				u: []string{path},
+				u: []string{
+					"http://" + ipcHost + path,
+					"https://" + ipcHost + path,
+				},
 				m: []string{http.MethodPost},
 			}, header["Dpop"])
 			if err != nil {
--- a/service/internal/auth/authn.go
+++ b/service/internal/auth/authn.go
@@ -18,6 +18,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"connectrpc.com/connect"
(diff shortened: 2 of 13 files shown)


service/internal/auth/authn.go
```diff
@@ -95,16 +100,88 @@ type Authentication struct {
...
+		panic(fmt.Sprintf("failed to generate nonce: %v", err))
+	}
+	nm.currentNonce = hex.EncodeToString(nonce)
+	nm.lastRotation = time.Now()
+}
+
+func (nm *dpopNonceManager) getCurrentNonce() string {
+	nm.mu.RLock()
+	if time.Since(nm.lastRotation) <= nm.expiration {
+		defer nm.mu.RUnlock()
```
jentfoo:
```suggestion
		nm.mu.RUnlock()
```

defer looks a little funny right before the return.

service/internal/auth/authn.go
```diff
@@ -95,16 +100,88 @@ type Authentication struct {
...
+	if _, err := rand.Read(nonce); err != nil {
+		panic(fmt.Sprintf("failed to generate nonce: %v", err))
+	}
+	nm.currentNonce = hex.EncodeToString(nonce)
+	nm.lastRotation = time.Now()
+}
+
+func (nm *dpopNonceManager) getCurrentNonce() string {
+	nm.mu.RLock()
+	if time.Since(nm.lastRotation) <= nm.expiration {
```
jentfoo: Putting the rotation time and nonce into an atomic could avoid the read lock and double locking

service/internal/auth/authn.go
```diff
@@ -95,16 +100,88 @@ type Authentication struct {
...
+		nm.lastRotation = time.Now()
+	}
+
+	return nm.currentNonce
+}
+
+func (nm *dpopNonceManager) validateNonce(nonce string) bool {
+	if !nm.requireNonce {
+		return true
+	}
```
jentfoo, line 180: You need to verify the nonce is not empty here, otherwise you may mark as validated to an empty / initial state.

test/start-up-with-containers/action.yaml
```diff
@@ -126,9 +141,9 @@ runs:
     - name: Download latest init-temp-keys.sh, docker-compose.yaml, and watch.sh
       shell: bash
       run: |
-        curl -sSfL https://raw.githubusercontent.com/opentdf/platform/refs/tags/pqc-enabled/.github/scripts/init-temp-keys.sh > otdf-test-platform/.github/scripts/init-temp-keys.sh
-        curl -sSfL https://raw.githubusercontent.com/opentdf/platform/refs/tags/pqc-enabled/docker-compose.yaml > otdf-test-platform/docker-compose.yaml
-        curl -sSfL https://raw.githubusercontent.com/opentdf/platform/refs/tags/pqc-enabled/.github/scripts/watch.sh > otdf-test-platform/.github/scripts/watch.sh
+        curl -sSfL https://raw.githubusercontent.com/opentdf/platform/refs/tags/feat-kc26-dpop/.github/scripts/init-temp-keys.sh > otdf-test-platform/.github/scripts/init-temp-keys.sh
+        curl -sSfL https://raw.githubusercontent.com/opentdf/platform/refs/tags/feat-kc26-dpop/docker-compose.yaml > otdf-test-platform/docker-compose.yaml
+        curl -sSfL https://raw.githubusercontent.com/opentdf/platform/refs/tags/feat-kc26-dpop/.github/scripts/watch.sh > otdf-test-platform/.github/scripts/watch.sh
```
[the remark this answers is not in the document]
dmihalcik-virtru: Good catch. The three downloads now use `${{ inputs.platform-ref }}` instead of the hardcoded `feat-kc26-dpop` tag, which aligns with how the action is invoked (`platform-ref: ${{ github.event.pull_request.head.sha || github.sha }}`) and avoids breakage once the feature tag is deleted.

service/internal/auth/authn_test.go
```diff
@@ -749,6 +758,142 @@ func (s *AuthSuite) TestInvalid_DPoP_Cases() {
...
+	})
+
+	_, _, err = s.auth.checkToken(
+		context.Background(),
+		[]string{"DPoP " + string(signedTok)},
+		receiverInfo{
+			u: []string{"/kas.AccessService/PublicKey"},
+			m: []string{http.MethodGet},
+		},
+		[]string{dpopToken},
```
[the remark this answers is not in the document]
dmihalcik-virtru: Done. The GET test now uses a full-origin `htu` (`https://localhost:8080/kas.AccessService/PublicKey`) and the `receiverInfo.u` list includes both the http and https origins, so it exercises `htm=GET` independently of strict/loose HTU mode.

service/internal/auth/dpop_nonce_test.go
```diff
@@ -0,0 +1,307 @@
...
+			AuthNConfig: AuthNConfig{
+				EnforceDPoP: true,
+				Issuer:      s.server.URL,
+				Audience:    "test",
+				DPoPSkew:    time.Hour,
+				TokenSkew:   time.Minute,
+				DPoP: DPoPConfig{
+					RequireNonce:    true,
+					NonceExpiration: 5 * time.Minute,
+				},
```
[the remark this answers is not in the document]
dmihalcik-virtru: Done. `newAuthWithNonce()` now pins `StrictHTU: false` explicitly so the nonce tests stay deterministic and keep exercising nonce validation rather than failing on HTU mode if the default changes.

test/start-up-with-containers/action.yaml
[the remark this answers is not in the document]
dmihalcik-virtru: Reverting this one — we'd rather keep an explicit, stricter pin for the downloaded scripts/compose than follow `platform-ref` (which floats to `main` by default). The `feat-kc26-dpop` tag stays for now; it'll be repointed/cleaned up as part of merging this feature line.

service/internal/auth/authn.go
elizabethhealy, line 825: claude comment: Missing jti replay tracking. RFC 9449 §9.1 requires the server to record recently-seen jti values to prevent proof replay. Right now, any captured DPoP proof can be reused for the full DPoPSkew window (default: 1 hour). The nonce mechanism helps but require_nonce defaults to false and even with nonces, a proof is replayable within the nonce's 5-minute window.

sdk/auth/token_adding_interceptor.go
```diff
@@ -64,6 +64,11 @@ func (i TokenAddingInterceptor) AddCredentials(
 		return status.Error(codes.Unauthenticated, err.Error())
```
elizabethhealy, line 64: claude comment: The SDK interceptors never retry on a 401 use_dpop_nonce from the resource server. If require_nonce: true is set on the server, all SDK clients will hard-fail on the first request (they'll get a 401 with DPoP-Nonce header and have no code path to extract it and retry). The GetAccessToken/DoTokenExchange functions do handle the token endpoint nonce retry (correctly checking 400), but resource server challenges (401) are unhandled

service/internal/auth/config.go
```diff
@@ -31,6 +31,23 @@ type AuthNConfig struct { //nolint:revive // AuthNConfig is a valid name
 	CacheRefresh string        `mapstructure:"cache_refresh_interval" json:"cache_refresh_interval"`
 	DPoPSkew     time.Duration `mapstructure:"dpopskew" json:"dpopskew" default:"1h"`
```
elizabethhealy, line 32: do we still need the 1h skew?

service/internal/auth/authn.go
```diff
@@ -725,10 +982,21 @@ func (a Authentication) ipcReauthCheck(ctx context.Context, path string, header
...
 				m: []string{http.MethodPost},
-			}, header["Dpop"])
+			}
+			a.logger.DebugContext(
+				ctx, "dpop receiverInfo set (ipc reauth)",
+				slog.Any("expected_htm", ri.m),
+				slog.Any("expected_htu", ri.u),
+				slog.String("path", path),
+			)
+			token, ctxWithJWK, err := a.checkToken(ctx, authHeader, ri, header["Dpop"])
```
elizabethhealy, line 1031: existing issue: this differs from the muxhandler impl which detects if err is a nonce err and handles that case so the sender can act accordingly

service/internal/auth/authn.go
```diff
@@ -95,16 +102,89 @@ type Authentication struct {
...
 	// Custom Logger
 	logger *logger.Logger
+	// DPoP nonce management
+	dpopNonceManager *dpopNonceManager
 
 	// Used for testing
 	_testCheckTokenFunc func(ctx context.Context, authHeader []string, dpopInfo receiverInfo, dpopHeader []string) (jwt.Token, context.Context, error)
 }
 
+// dpopNonceManager manages server-issued DPoP nonces per RFC 9449 §8
```
elizabethhealy: nit: is this intended for the type declaration below?

service/internal/auth/authn.go
```diff
@@ -565,7 +795,12 @@ func (a *Authentication) checkToken(ctx context.Context, authHeader []string, dp
 	}
 	dpopKey, err := a.validateDPoP(accessToken, tokenRaw, dpopInfo, dpopHeader)
 	if err != nil {
-		a.logger.Warn("failed to validate dpop", slog.Any("err", err))
+		var nonceErr *DPoPNonceError
+		if errors.As(err, &nonceErr) {
+			a.logger.DebugContext(ctx, "dpop nonce challenge issued", slog.String("reason", nonceErr.Message))
+		} else {
+			a.logger.Warn("failed to validate dpop", slog.Any("err", err))
```
elizabethhealy: nit: warnContext

elizabethhealy (dismissed):
overall lgtm, claude had a few comments that seemed outside the scope of the PR but worth noting, the rest are non blocking

service/internal/auth/authn.go
dmihalcik-virtru: Added server-side replay protection in this PR. New `dpopReplayCache` (`service/internal/auth/dpop_replay.go`) remembers each proof's `jti` for the acceptance window and `validateDPoP` now rejects a reused `jti` ("DPoP proof replay detected; `jti` already used"). The check runs as the **last** step in `validateDPoP`, so only otherwise-valid proofs populate the cache (no cache-poisoning DoS). TTL tracks `DPoPSkew`, since a proof older than that is already rejected by the `iat` check; entries are pruned lazily to bound memory. Covered by unit tests in `dpop_replay_test.go` and a suite test (`Test_CheckToken_RejectsReplayedDPoP`) asserting first-use succeeds and a replay fails.

sdk/auth/token_adding_interceptor.go
dmihalcik-virtru: Good catch — confirmed gap. The token-endpoint nonce retry (400 `use_dpop_nonce`) is handled in `GetAccessToken`/`DoTokenExchange`, but the resource-server 401 challenge is not yet handled in the SDK interceptors. Since `require_nonce` defaults to `false`, this doesn't break the default configuration, so we'll track the interceptor 401-retry (extract `DPoP-Nonce` from the response header, rebuild the proof with the nonce claim, retry once) as a follow-up rather than expanding this PR. Server-side replay protection is being added here in the meantime.

service/internal/auth/config.go
dmihalcik-virtru: Keeping 1h for now. `DPoPSkew` bounds the `iat` freshness window and is deliberately generous to tolerate client clock skew during rollout. The replay exposure it implied is now closed by the `jti` replay cache added in this PR (cache TTL == `DPoPSkew`), so a captured proof can no longer be reused within the window. Once SDK clocks are confirmed tight we can tighten this default in a follow-up; doing it now risks rejecting legitimate clients.

service/internal/auth/authn.go
dmihalcik-virtru: Acknowledged — this is the IPC reauth path, which collapses errors to a generic `unauthenticated` and doesn't surface nonce challenges the way the mux handler does. That's a pre-existing divergence and intentional for now: IPC reauth doesn't issue DPoP nonce challenges, so there's no nonce error to propagate on this path. I'll leave it out of scope for this PR and track aligning the two error paths as a follow-up.

service/internal/auth/authn.go
dmihalcik-virtru: Good eye — the doc comment was sitting on `nonceState`. Moved the `// dpopNonceManager manages server-issued DPoP nonces per RFC 9449 §8` comment onto `dpopNonceManager` and gave `nonceState` its own comment.

service/internal/auth/authn.go
dmihalcik-virtru: Done — switched to `a.logger.WarnContext(ctx, ...)` since `ctx` is in scope here.

service/internal/auth/authn.go
dmihalcik-virtru: ^ Claude wrote that. One thing to note is that we generally have historically had very high skew tolerance due to the fact that clients (especially, but not only, web clients) will have clock skew that is outside the user's control.

sdk/auth/token_adding_interceptor.go
dmihalcik-virtru: Actual human response: The SDK work is being applied in separate PRs (e.g. #3581 for go-sdk). Since SDKs don't already support nonce challenges, they will all have to be updated to support first, then the feature flag (`use_dpop_nonce`) may be enabled if desired, and applicable to the KAS deployment

service/internal/auth/authn.go
dmihalcik-virtru: Human response: IPC doesn't need dpop

service/internal/auth/config.go
dmihalcik-virtru: Human response: I'll file a bug to reduce this and the JWT skew, and possibly combine them into one config option or struct

elizabethhealy (dismissed):

elizabethhealy (approved):

PR #3590, opened 2026-06-09
feat(core)!: Resolve keys.

c-r33d:
## Summary

  Adds authz v2 support for URI-scoped KAS `GetKey` authorization.

  ## Changes

  - Register a KAS registry authz resolver for `GetKey`
  - Resolve `kas_uri` for KAS key requests
  - Cache the resolved KAS key in resolver context so `GetKey` can reuse it
  - Leave `ListKeys` without granular URI resolution for now
  - Keep attribute resolver registration disabled for this KAS-only scope
  - Update authz v2 BDD coverage for KAS key access
  - Update the authz v2 BDD template to grant URI-specific access only for `GetKey`
  - Add `client:` prefix to client_ids

  ## Behavior Covered

  - `opentdf-admin` can `GetKey` and `ListKeys`
  - `opentdf-standard` cannot `GetKey` or `ListKeys`
  - `kas-a` can `GetKey` only for `https://kas-a.example.com`
  - `kas-b` can `GetKey` only for `https://kas-b.example.com`
  - `kas-a` and `kas-b` cannot access each other’s keys
  - URI-scoped `kas-a` / `kas-b` roles are denied `ListKeys`

  ## Verification

  ```sh
  go test ./internal/auth/...
  go test ./policy/kasregistry/...
  go test ./policy/attributes/...
  PLATFORM_IMAGE=DEBUG go test ./tests-bdd -v --tags=cukes --godog.random --godog.format=pretty --godog.tags='@authz-v2' ./features
```

>[!WARNING]
>The breaking change is that I set the `client:` prefix for client_ids.
>
>For current policy, customers will need to change their custom policy to include the `client:` prefix.

### DIFF
--- a/service/policy/attributes/attributes.go
+++ b/service/policy/attributes/attributes.go
@@ -92,18 +92,6 @@ func NewRegistration(ns string, dbRegister serviceregistry.DBRegister) *servicer
 				}
 				as.cache = resolverCache
 
-				// Register authz resolvers per-method
-				// Each resolver extracts authorization dimensions from the request, performing DB lookups as needed.
-				// The resolver is called by the auth interceptor before the handler.
-				if srp.AuthzResolverRegistry != nil {
-					srp.AuthzResolverRegistry.MustRegister("CreateAttribute", as.createAttributeAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("GetAttribute", as.getAttributeAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("GetAttributeValuesByFqns", as.getAttributeValuesByFqnsAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("ListAttributes", as.listAttributesAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("UpdateAttribute", as.updateAttributeAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("DeactivateAttribute", as.deactivateAttributeAuthzResolver)
-				}
-
 				return as, nil
 			},
 		},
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: service/policy/attributes/attributes.go [hunk 1]
```diff
@@ -92,18 +92,6 @@ func NewRegistration(ns string, dbRegister serviceregistry.DBRegister) *servicer
...
-				// Register authz resolvers per-method
-				// Each resolver extracts authorization dimensions from the request, performing DB lookups as needed.
-				// The resolver is called by the auth interceptor before the handler.
-				if srp.AuthzResolverRegistry != nil {
-					srp.AuthzResolverRegistry.MustRegister("CreateAttribute", as.createAttributeAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("GetAttribute", as.getAttributeAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("GetAttributeValuesByFqns", as.getAttributeValuesByFqnsAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("ListAttributes", as.listAttributesAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("UpdateAttribute", as.updateAttributeAuthzResolver)
-					srp.AuthzResolverRegistry.MustRegister("DeactivateAttribute", as.deactivateAttributeAuthzResolver)
```
alkalescent, line 104: Are we removing these because we're not supporting attributes right now or because this is handled a different way?

c-r33d: Not supporting right now, decided to slim down what we are launching for the initial release.

alkalescent (approved):

PR #3592, opened 2026-06-09
feat(authz): split connect token claims and enforcement

jakedoublev:
### Proposed Changes

* Separate token validation (authn) and casbin authorization (authz) middlewares
* Unify public route middleware check and set context state if public route for downstream interceptors/middlewares
* Reduce token parsing and claims handling by exposing context handlers to get roles, subject, client ID from context that was set by upstream authn interceptor

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Return immediate permission-denied responses when role/claim validation fails; fail earlier for missing Authorization headers.

* **New Features**
  * Public-route detection computed once and propagated through request handling.
  * Support dot-notation nested JWT claim paths for role/group extraction.
  * Authorization context now carries client ID, subject and groups for downstream checks.
  * Accept multiple/auth interceptor slices and external interceptor factories via config.

* **Tests**
  * Added comprehensive unit and integration tests covering authn/authz chaining, role resolution, and context propagation.


service/internal/auth/casbin.go
```diff
@@ -190,28 +190,59 @@ func (e *Enforcer) Enforce(ctx context.Context, token jwt.Token, req authz.RoleR
 	return result, ErrPermissionDenied
 }
 
+func (e *Enforcer) ContextWithClaims(ctx context.Context, t jwt.Token, req authz.RoleRequest) (context.Context, error) {
+	roles, err := e.roleProvider.Roles(ctx, t, req)
+	if err != nil {
+		return ctx, err
+	}
+	claims, _ := authz.ClaimsFromContext(ctx)
```
c-r33d, line 198: Not something to worry about in this PR, but eventually I like the idea of all the items in the `RequestClaims` or at least: `subject`, `roles`, `domain` being pluggable. Meaning we can implement multiple engines for providing a subject, or pieces of a subject.

service/internal/auth/authn.go
```diff
@@ -274,15 +287,17 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
...
+// ConnectAuthNInterceptor authenticates Connect requests and enriches the
+// request context with configured token claims needed by later middleware.
+func (a Authentication) ConnectAuthNInterceptor() connect.UnaryInterceptorFunc {
 	interceptor := func(next connect.UnaryFunc) connect.UnaryFunc {
 		return connect.UnaryFunc(func(
 			ctx context.Context,
 			req connect.AnyRequest,
 		) (connect.AnyResponse, error) {
-			// if the token is already in the context, skip the interceptor
-			if ctxAuth.GetAccessTokenFromContext(ctx, a.logger) != nil {
```
c-r33d, line 285: Is there a reason to remove this? Was this ever possible?

service/internal/auth/authn.go
```diff
@@ -331,30 +340,71 @@ func (a Authentication) ConnectUnaryServerInterceptor() connect.UnaryInterceptor
...
+			}
+
+			token := ctxAuth.GetAccessTokenFromContext(ctx, a.logger)
+			if token == nil {
+				return nil, connect.NewError(connect.CodeUnauthenticated, errors.New("missing access token in context"))
+			}
+
+			p := strings.Split(req.Spec().Procedure, "/")
+			resource := path.Join(p[1], p[2])
+			action := getAction(p[2])
```
c-r33d: Not a big deal, but this is duplicated in the AuthN interceptor

service/internal/auth/authn_test.go
```diff
@@ -418,6 +560,24 @@ func (s *AuthSuite) Test_UnaryServerInterceptor_When_Authorization_Header_Missin
...
 
+func (s *AuthSuite) Test_ConnectAuthNInterceptor_RequiresHeaderWithExistingContextToken() {
+	interceptor := s.auth.ConnectAuthNInterceptor()
+
+	next := func(_ context.Context, _ connect.AnyRequest) (connect.AnyResponse, error) {
+		return connect.NewResponse[string](nil), nil
+	}
+
+	req := connect.NewRequest[string](nil)
+	ctx := ctxAuth.ContextWithAuthNInfo(context.Background(), nil, jwt.New(), "raw-token")
```
c-r33d:
```suggestion
	ctx := ctxAuth.ContextWithAuthNInfo(s.T().Context(), nil, jwt.New(), "raw-token")
```

service/internal/auth/casbin_test.go
```diff
@@ -494,6 +504,60 @@ func (s *AuthnCasbinSuite) Test_ExternalRoleProvider() {
...
+	provider := &countingProvider{roles: []string{"role:admin"}}
+	enforcer, err := NewCasbinEnforcer(CasbinConfig{
+		PolicyConfig: policyCfg,
+		RoleProvider: provider,
+	}, logger.CreateTestLogger())
+	s.Require().NoError(err)
+
+	req := authz.RoleRequest{Resource: "policy.attributes.List", Action: "read"}
+	tok := jwt.New()
+	ctx, err := enforcer.ContextWithClaims(context.Background(), tok, req)
```
c-r33d:
```suggestion
	ctx, err := enforcer.ContextWithClaims(s.T().Context(), tok, req)
```

service/internal/auth/casbin_test.go
```diff
@@ -494,6 +504,60 @@ func (s *AuthnCasbinSuite) Test_ExternalRoleProvider() {
...
+	provider := &countingProvider{roles: []string{"role:admin"}}
+	enforcer, err := NewCasbinEnforcer(CasbinConfig{
+		PolicyConfig: policyCfg,
+		RoleProvider: provider,
+	}, logger.CreateTestLogger())
+	s.Require().NoError(err)
+
+	req := authz.RoleRequest{Resource: "policy.attributes.List", Action: "read"}
+	tok := jwt.New()
+	ctx := authz.ContextWithClientID(context.Background(), "client-123")
```
c-r33d:
```suggestion
	ctx := authz.ContextWithClientID(context.T().Context(), "client-123")
```

c-r33d (dismissed):

service/internal/auth/casbin.go
jakedoublev: Like extending the role provider interface?

service/internal/auth/authn.go
jakedoublev: It shouldn't have been possible, no, so was dead code from what I could tell (why I removed it). Our passing e2e tests give me pretty good confidence.

jakedoublev: The git blame looks like it was added around the same batch of changes as the gRPC gateway addition, but that's now deprecated and removed, so I'm guessing that's why it's dead code now.

service/internal/auth/authn.go
jakedoublev: Added a builder helper: b8da2dee2157db209b6646d22d353605ef80cb13

service/internal/auth/casbin.go
c-r33d: Yeah, to something like `PrincipalProvider` which gives us information about the identity of the requestor, such as groups/identity/other context.

c-r33d (dismissed):

c-r33d (approved):

service/internal/auth/casbin.go
jakedoublev: Yeah, parsing the token claims multiple times in chained interceptors isn't ideal. We also don't always know what token claims integrators will need, which varies implementation to implementation (custom claims) and idP to idP. I think things that are standard across the OIDC spec (subject, groups, etc) being configurable with dotnotation and custom claims requiring an extra parse is an okay middleground for the time being, but this is a good thing to continue thinking about.

PR #2095, opened 2025-04-23
WIP: Novel Key and Security Providers

dmihalcik-virtru:
### Proposed Changes

*

### Checklist

- [ ] I have added or updated unit tests
- [ ] I have added or updated integration tests (if appropriate)
- [ ] I have added or updated documentation

PR #3597, opened 2026-06-10
fix(sdk): align policy binding encoding with spec, keep legacy compat

### LINKED ISSUE #3578: SDK creates policyBinding that does not conform to the Spec
The spec says [here](https://opentdf.io/spec/schema/opentdf/key_access_object#policy-binding-object-keyaccesspolicybinding) about the hash field: "A Base64 encoding of HMAC(KEY, POLICY)". [What the SDK actually does is](https://github.com/opentdf/platform/blob/626ce47dd662cb8ff16898e3b6727001a4753d92/sdk/tdf.go#L590):

    policyBindingHash := hex.EncodeToString(ocrypto.CalculateSHA256Hmac(symKey, base64PolicyObject))
    pbstring := string(ocrypto.Base64Encode([]byte(policyBindingHash)))

Note the call to `hex.EncodeToString`. Because of this, the resulting hash is incompatible with independent 3rd party implementations.

biscoe916:
Fixes [3578](https://github.com/opentdf/platform/issues/3578)

### Proposed Changes

* Gate policy binding writer in `sdk/tdf.go` on the existing `tdfConfig.useHex` flag. Default (spec >= 4.3.0) emits `Base64(HMAC)` per spec; `WithTargetMode("<4.3.0")` keeps legacy `Base64(hex(HMAC))` for byte-identical compatibility
* Extract shared `decodePolicyBinding` helper in `service/kas/access/rewrap.go`. Length-detects the encoding (32 bytes raw vs 64 bytes hex after base64 decode) so KAS dual-accepts both formats — no manifest version trust needed
* Fix latent bug in the production rewrap path where the raw branch fed an untrimmed buffer to `dek.VerifyBinding`
* Replace two near-duplicate inline decode blocks with the shared helper
* Apply the same gating to the chunked segment writer (`sdk/chunked_writer.go`), which landed after this branch was cut and shares the `createPolicyBinding` helper — it otherwise inherits the same always-hex bug

### Checklist

### Testing Instructions

**Automated:**
- `go test ./sdk/... ./service/kas/... -race` — round-trip tests cover both encodings; `TestDecodePolicyBinding` covers raw/hex/invalid base64; `TestCreatePolicyBinding` pins a known-answer vector for each target mode
- `TestChunkedLegacyTargetMode` / `TestChunkedCurrentTargetMode` pin the binding encoding alongside the root/segment signatures they already check
- `Test_SimpleTDF` verifies the binding *value* (`HMAC(payload key, base64 policy)`), not just its shape
- `golangci-lint run sdk/... service/...` — no new issues

**Manual interop check (recommended before rollout):**
1. Encrypt a TDF with this branch's SDK (default mode → raw binding) and rewrap against this branch's KAS — should succeed
2. Encrypt a TDF with `WithTargetMode("4.2.2")` (legacy hex binding) and rewrap against this branch's KAS — should succeed
3. Decrypt an existing pre-fix TDF (hex binding) against this branch's KAS — should succeed

### Rollout Notes

- **No released KAS can accept a raw binding.** Dual-accept was attempted in #2095, but the raw branch passes an untrimmed 33-byte buffer (`base64.DecodedLen(44)`) to `VerifyBinding`, which compares it length-sensitively against a 32-byte HMAC — so it always fails. Verified in released `service/v0.9.0` and reproduced by xtest (`v0.9.0`/`main` platform × `go@pull-3597` encrypt). A KAS carrying this PR's fix must be deployed *before* any SDK in the fleet defaults to raw; until then pin writers to `WithTargetMode("<4.3.0")`. See the rebase comment below for the full analysis and the merge-sequencing options
- Hex acceptance must be retained indefinitely — TDFs are archival
- Java and JS SDKs have the same writer bug and need analogous fixes (out of scope here)
- `sdk/experimental/tdf/key_access.go` has its own unconditional-hex `createPolicyBinding`; that package has no target-mode concept yet, so it is left alone here
- Spec doc should note `Base64(hex(HMAC))` as deprecated-but-MUST-accept so third-party KAS implementations don't break on archival TDFs

### Note on this update

Rebased onto `main` (was 226 commits behind, conflicting). The conflict was with the `createPolicyBinding` helper extracted on main; `useHex` is now a parameter to it. The chunked writer work above is new in this rebase.

### DIFF
--- a/sdk/tdf.go
+++ b/sdk/tdf.go
@@ -587,8 +587,15 @@ func (s SDK) prepareManifest(ctx context.Context, t *TDFObject, tdfConfig TDFCon
 		symKeys = append(symKeys, symKey)
 
 		// policy binding
-		policyBindingHash := hex.EncodeToString(ocrypto.CalculateSHA256Hmac(symKey, base64PolicyObject))
-		pbstring := string(ocrypto.Base64Encode([]byte(policyBindingHash)))
+		// Spec (>= 4.3.0) requires Base64(HMAC); pre-4.3.0 TDFs used Base64(hex(HMAC)).
+		// useHex tracks the same threshold as segment/root signatures via WithTargetMode.
+		hmacBytes := ocrypto.CalculateSHA256Hmac(symKey, base64PolicyObject)
+		var pbstring string
+		if tdfConfig.useHex {
+			pbstring = string(ocrypto.Base64Encode([]byte(hex.EncodeToString(hmacBytes))))
+		} else {
+			pbstring = string(ocrypto.Base64Encode(hmacBytes))
+		}
 		policyBinding := PolicyBinding{
 			Alg:  "HS256",
 			Hash: pbstring,
@@ -1247,7 +1254,7 @@ func createRewrapRequest(_ context.Context, r *Reader) (map[string]*kas.Unsigned
 			invalidPolicy = !ok
 			alg, ok = policyBinding["alg"].(string)
 			invalidPolicy = invalidPolicy || !ok
-		case (PolicyBinding):
+		case PolicyBinding:
 			hash = policyBinding.Hash
 			alg = policyBinding.Alg
 		default:
--- a/sdk/tdf_test.go
+++ b/sdk/tdf_test.go
@@ -524,6 +524,7 @@ func (s *TDFSuite) Test_SimpleTDF() {
 				WithSessionKeyType(ocrypto.EC256Key),
 				WithKasAllowlist([]string{s.kasTestURLLookup["https://d.kas/"]}),
 			},
+			expectedSize: 1970,
 		},
 		{
 			name: "target-mode-0",
@@ -657,6 +658,24 @@ func (s *TDFSuite) Test_SimpleTDF() {
 				} else {
 					s.Require().Error(err1)
 					s.Require().Error(err2)
+				}
+
+				// check that the policy binding matches the same hex/raw scheme as the
+				// other signatures. Spec >= 4.3.0 emits Base64(HMAC); pre-4.3.0 emits
+				// Base64(hex(HMAC)).
+				s.Require().NotEmpty(r.Manifest().KeyAccessObjs)
+				pb, ok := r.Manifest().KeyAccessObjs[0].PolicyBinding.(map[string]any)
+				s.Require().True(ok, "expected PolicyBinding to deserialize as map")
+				pbHash, ok := pb["hash"].(string)
+				s.Require().True(ok, "expected PolicyBinding.hash to be a string")
+				decodedPB, err := ocrypto.Base64Decode([]byte(pbHash))
+				s.Require().NoError(err)
+				if config.useHex {
+					s.Len(decodedPB, hex.EncodedLen(sha256.Size), "legacy policy binding should be hex-encoded HMAC")
+					_, err = hex.DecodeString(string(decodedPB))
+					s.Require().NoError(err)
+				} else {
+					s.Len(decodedPB, sha256.Size, "spec-compliant policy binding should be raw HMAC bytes")
 				}
 
 				// check version is present if usehex is false
@@ -927,7 +946,7 @@ func (s *TDFSuite) Test_TDFWithAssertion() {
 			},
 			verifiers:                    nil,
 			disableAssertionVerification: false,
-			expectedSize:                 2689,
+			expectedSize:                 2656,
 		},
 		{
 			assertions: []AssertionConfig{
@@ -990,7 +1009,7 @@ func (s *TDFSuite) Test_TDFWithAssertion() {
 				DefaultKey: defaultKey,
 			},
 			disableAssertionVerification: false,
-			expectedSize:                 2689,
+			expectedSize:                 2656,
 		},
 		{
 			assertions: []AssertionConfig{
@@ -1039,7 +1058,7 @@ func (s *TDFSuite) Test_TDFWithAssertion() {
 				},
 			},
 			disableAssertionVerification: false,
-			expectedSize:                 2988,
+			expectedSize:                 2955,
 		},
 		{
 			assertions: []AssertionConfig{
@@ -1079,7 +1098,7 @@ func (s *TDFSuite) Test_TDFWithAssertion() {
 				},
 			},
 			disableAssertionVerification: false,
-			expectedSize:                 2689,
+			expectedSize:                 2656,
 		},
 		{
 			assertions: []AssertionConfig{
@@ -1096,7 +1115,7 @@ func (s *TDFSuite) Test_TDFWithAssertion() {
 				},
 			},
 			disableAssertionVerification: true,
-			expectedSize:                 2180,
+			expectedSize:                 2147,
 		},
 	} {
 		expectedTdfSize := test.expectedSize
@@ -1333,7 +1352,7 @@ func (s *TDFSuite) Test_TDFWithAssertionNegativeTests() {
 					SigningKey: defaultKey,
 				},
 			},
-			expectedSize: 2689,
+			expectedSize: 2656,
 		},
 		{
 			assertions: []AssertionConfig{
@@ -1381,7 +1400,7 @@ func (s *TDFSuite) Test_TDFWithAssertionNegativeTests() {
 					},
 				},
 			},
-			expectedSize: 2988,
+			expectedSize: 2955,
 		},
 		{
 			assertions: []AssertionConfig{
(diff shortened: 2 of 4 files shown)


pflynn-virtru (dismissed):

sujankota (dismissed):

pflynn-virtru:
## Rebased onto `main` — plus a compatibility finding that needs a decision

### What changed in this update

Rebased onto `main` (was 226 commits behind, `CONFLICTING`, now `MERGEABLE`). The conflict was with the `createPolicyBinding` helper extracted on `main`; `useHex` is now a parameter to it rather than an inline branch.

One addition: the chunked segment writer (#3940) landed after this branch was cut and shares `createPolicyBinding`, so it silently inherited the same always-hex bug. Its `useHex` is now threaded through too, and its fake KAS was taught to dual-accept (it only accepted hex, so it would have passed while every real KAS rejected the output). Also picked up the outstanding CodeRabbit nit — `Test_SimpleTDF` now verifies the binding *value*, not just its shape.

Local: `go test ./sdk/... ./service/kas/... -race` green, `golangci-lint` clean on all touched files, `make fmt` clean.

### The finding

The xtest matrix now covers `go@pull-3597` as the **encrypting** SDK against older platforms — a cell that did not exist when this PR was last run. Both new cells fail reproducibly (the first round was a `proxy.golang.org` flake; on re-run the infra failures cleared and these two stayed):

| platform | encrypt SDK | result |
|---|---|---|
| `pull-3597` | `go@pull-3597` | ✅ pass |
| `main` | `go@pull-3597` | ❌ fail |
| `v0.9.0` | `go@pull-3597` | ❌ fail |

88 × `"failure to verify policy binding"` in the KAS logs. Every test that *encrypts* with this branch's SDK fails against an older KAS, regardless of the decrypting SDK. (The negative tests — `test_tdf_with_altered_policy_binding`, `test_tdf_with_unbound_policy` — pass, because they expect a rejection anyway.)

### Why — the rollout note in the description is not accurate

The description says:

> Platform KAS dual-accept has been in place since May 2025 (#2095).

Dual-accept was *attempted*, but the raw branch has never actually worked. From the released `service/v0.9.0` (`service/kas/access/rewrap.go:597`):

```go
policyBinding := make([]byte, base64.StdEncoding.DecodedLen(len(policyBindingB64Encoded)))
n, err := base64.StdEncoding.Decode(policyBinding, []byte(policyBindingB64Encoded))
// ...
if n == 64 { // hex path: policyBinding is REPLACED with a correctly-sized slice
    dehexed := make([]byte, hex.DecodedLen(n))
    _, err = hex.Decode(dehexed, policyBinding[:n])
    if err == nil { policyBinding = dehexed }
}
// raw path (n == 32): policyBinding is NEVER trimmed to n
dek.VerifyBinding(ctx, []byte(req.GetPolicy().GetBody()), policyBinding)
```

For a raw binding the base64 is 44 chars, so `DecodedLen(44)` allocates **33** bytes while `n` is **32**. The hex branch replaces the slice with a right-sized one; the raw branch passes the untrimmed 33-byte buffer straight through. `AESProtectedKey.VerifyBinding` does `hmac.Equal(actualHMAC /*32*/, policyBinding /*33*/)`, which is length-sensitive and therefore always false.

This is precisely the "latent bug ... untrimmed buffer" this PR already claims to fix — but the consequence is bigger than the description implies: **no released KAS can accept a raw binding.** The fix has to be deployed on the KAS side *before* any SDK in the fleet starts emitting raw bindings. `pull-3597 × go@pull-3597` passes only because that platform carries this PR's own fix.

### What this means for merging

The KAS half of this PR (`decodePolicyBinding`) is strictly a fix and safe to land on its own. The SDK half flips the default writer encoding, and that is a breaking change against every currently-deployed KAS — so it is a rollout-sequencing decision, not a code-review one.

I've left the change intact rather than making that call in a rebase. Worth deciding between:

1. **Split** — land the KAS dual-accept fix now, flip the SDK default in a follow-up once a fixed KAS is broadly deployed. Makes xtest green immediately.
2. **Hold** — keep the PR as-is and block merge until fixed-KAS deployment is confirmed, treating the two red cells as an intentional signal.

Same caveat applies to the Java and JS writers when they're fixed, and the same untrimmed-buffer bug should be checked for in any third-party KAS.

pflynn-virtru:
## On the title: "keep legacy compat" and the missing breaking-change marker

Follow-up to my rebase comment above — this is about how the change is *labelled* rather than what it does.

### The title claims the wrong thing

Current title: `fix(sdk): align policy binding encoding with spec, keep legacy compat`

(I changed the scope from `tdf` → `sdk` to get `pull-request-checks` passing — `tdf` isn't in the allowed scope list, which is why that check had been red since this PR was opened. The rest of the title is unchanged.)

"Keep legacy compat" is **not false**. Both things it implies are genuinely delivered:

1. Archival TDFs still decrypt — `decodePolicyBinding` accepts hex, and the PR commits to keeping that indefinitely. Covered by `TestDecodePolicyBinding`.
2. Legacy writing is still reachable via `WithTargetMode("<4.3.0")`.

The problem is that there are three compatibility directions here, and the title advertises the two that hold while staying silent on the one that doesn't:

| direction | status |
|---|---|
| old TDF → new KAS (archival reads) | ✅ preserved |
| new SDK, legacy target mode → any KAS | ✅ preserved |
| **new SDK, default mode → any deployed KAS** | ❌ **broken (see analysis above)** |

The broken direction is the *default* path — what every caller gets without opting into anything. And it's the one a reader is least likely to infer from "legacy compat," since that phrase idiomatically means "old things still work," not "new output still lands on old servers."

So the title doesn't misdescribe the code. It offers an unwarranted reassurance about what's safe to do with it.

### The `!` marker is missing, and this repo uses it for exactly this

Repo convention is `!` in the title (no `BREAKING CHANGE:` footers anywhere in recent history). Two close precedents:

- `feat(core)!: conform hybrid PQ/T key formats to IETF drafts (#3563)` — structurally identical to this PR: "our wire format doesn't match the spec, conform it." Marked breaking, with an explicit "⚠️ Wire-format break" callout. Notably its own justification reads *"No on-disk material in the old format was deployed, so no migration tooling is needed"* — i.e. it took the marker with **zero** blast radius.
- `fix(sdk)!: reclassify KAS 400 errors — distinguish tamper from misconfiguration (#3166)` — an SDK↔KAS behavioural change, also marked.

This PR changes a wire format where the blast radius is every deployed KAS and every TDF written after the upgrade, and carries no marker.

Worth noting why that matters more than usual here: `release-please-config.sdk.json` sets `"versioning": "always-bump-patch"`, so the SDK goes `0.32.0` → `0.32.1` regardless. The version number will never warn anyone. The `!` and its changelog entry are the only signal a consumer gets that a routine patch bump can make newly-written TDFs undecryptable against their own KAS.

### This reads as a wrong-premise problem, not a sloppy-description one

To be fair to the original framing: the claim was internally consistent. *If* dual-accept had really worked since #2095, the only exposure would have been pre-May-2025 deployments — and the rollout notes **did** flag exactly that, along with the `WithTargetMode` pinning workaround. The reasoning was sound; the premise underneath it wasn't. The untrimmed-buffer bug is what turns "a few stale deployments" into "all of them."

### Suggestion

The notable fact about this change is the new requirement it *imposes*, not the old behaviour it retains. Depending on which sequencing option is chosen:

- **Keeping the flip in this PR:**
  `fix(sdk)!: emit spec-compliant policy binding; requires KAS with dual-accept fix`
- **Splitting:**
  - this PR → `fix(kas): accept raw (spec-compliant) policy bindings` — a pure fix, no marker needed
  - follow-up → `fix(sdk)!: emit Base64(HMAC) policy binding per spec`

Either way the `!` and the KAS-version requirement belong in the title, and "keep legacy compat" belongs in the body where it can be stated precisely — which two encodings are accepted, in which direction, and for how long — rather than as a blanket claim.

I've deliberately not retitled beyond the scope fix, since where the `!` lands depends on the split-vs-hold decision.

pflynn-virtru:
Opened #4081 as Phase 1 of this work, per the analysis above.

It carries the KAS half only — `decodePolicyBinding` with the buffer trimmed, plus deletion of the dead `verifyPolicyBinding` that made the broken dual-accept look correct. On the SDK side the spec-compliant encoding is available as an opt-in (`WithSpecCompliantPolicyBinding`), but the default is unchanged, so #4081 alters no output and is safe to merge and deploy on its own.

That leaves this PR as Phase 2: flipping the default. It should rebase onto #4081 once that lands, and wants a `!` marker at that point — see the title comment above.

The opt-in also gives you a way to validate #3578's third-party interop concern before the default moves.

PR #3599, opened 2026-06-12
fix(core): Fix malicious escaping of dimensions.

c-r33d:
## Summary

Fixes KAS URI dimension injection in authz v2 by escaping dimension values before Casbin matching, then decoding them during dimension parsing. This prevents harmful values containing dimension parser tokens from being interpreted as additional authorization dimensions.

## Changes

- Escape only dimension grammar conflicts during serialization:
  - `%` as `%25`
  - `&` as `%26`
  - literal `*` as `%2A`
- Decode escaped request and policy dimension values with percent-decoding before comparison.
- Keep values readable where possible: `=`, `+`, URI schemes, paths, and query keys remain unchanged.
- Add unit coverage for:
  - URI values containing `&` and `=`
  - literal `+`, spaces, `%`, and `*`
  - serialize/parse round trips
  - `dimensionMatch` behavior with escaped policy dimensions and wildcard values
- Update authz-v2 cukes to verify a `kas-a` scoped client cannot access KAS keys whose URI injects a trailing `kas_uri=https://kas-a.example.com`.

## Security Impact

Previously, a malicious KAS URI like:

```text
https://kas-b.example.com?foo=bar&kas_uri=https://kas-a.example.com
```

could be serialized into dimensions in a way that allowed the trailing `kas_uri` pair to overwrite the original value during parsing. The fix keeps the full URL as one dimension value, so policy matching compares against the actual KAS URI.

## Policy Authoring Note

Raw policy dimensions remain readable. Authors only need percent-encoding for values containing characters that conflict with the dimension grammar, especially literal `&` inside a value:

```text
kas_uri=https://kas.example.com?foo=bar%26baz=qux
```

Raw `*` remains the wildcard policy value; use `%2A` for a literal `*`.

## 💡 Follow-up Recommendation: Structured Policy Dimensions

> 🚧 **Recommendation:** Consider moving policy dimensions to a structured representation before matching, such as `map[string][]string` per RPC/path. This would avoid delimiter ambiguity in raw Casbin dimension strings and reduce the need for authors to percent-encode parser tokens.

### DIFF
--- a/service/internal/auth/authz/casbin/casbin.go
+++ b/service/internal/auth/authz/casbin/casbin.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net/url"
 	"sort"
 	"strings"
 
@@ -555,10 +556,11 @@ func serializeDimensions(ctx *authz.ResolverContext) (string, error) {
 	}
 	sort.Strings(keys)
 
-	// Build canonical string
+	// Build canonical string, URL-encoding values to prevent injection via
+	// special characters like '&' and '=' that are used as separators.
 	parts := make([]string, 0, len(keys))
 	for _, k := range keys {
-		parts = append(parts, fmt.Sprintf("%s=%s", k, allDims[k]))
+		parts = append(parts, fmt.Sprintf("%s=%s", k, url.QueryEscape(allDims[k])))
 	}
 
 	return strings.Join(parts, "&"), nil
@@ -671,7 +673,11 @@ func parseDimensions(dims string) (map[string]string, bool) {
 		if !isValidDimensionKey(kv[0]) {
 			return nil, false
 		}
-		result[kv[0]] = kv[1]
+		unescapedVal, err := url.QueryUnescape(kv[1])
+		if err != nil {
+			return nil, false
+		}
+		result[kv[0]] = unescapedVal
 	}
 	return result, true
 }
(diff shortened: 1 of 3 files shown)


c-r33d:
@coderabbitai review

service/internal/auth/authz/casbin/casbin.go
```diff
@@ -636,6 +638,14 @@ func dimensionMatch(reqDims, policyDims string) bool {
 		if !isValidDimensionKey(key) {
 			return false
 		}
+		policyWildcard := policyVal == "*"
+		if !policyWildcard {
+			unescapedVal, err := url.QueryUnescape(policyVal)
```
elizabethhealy: claude comment -- this turns "+" into a space, not sure if thats a case we need to account for, if so we could use url.PathUnescape instead

elizabethhealy (dismissed):
lgtm, if the above comment isnt a case we need to handle

service/internal/auth/authz/casbin/casbin.go
c-r33d: I actually decided to change the method for how we are doing the escaping instead. Before, I was expecting customers to have to url escape their FQNs but that would be unreadable. I was serializing all values to prevent malicious actors from taking advantage of the policy dimension parsing (`&`). Instead now I just require that the following special characters are escaped if used in the dimension value literally: [`&`, `*`, `%`].

elizabethhealy (approved):

PR #3632, opened 2026-06-17
feat(policy): DSPX-2541 add GetKeyMappingsByFqns API

alkalescent:
### Proposed Changes

* Add `GetKeyMappingsByFqns`, a narrow client-side key-split read API on the attributes service: proto, regenerated `protocol/go`, handler, DB projection, and tests, all in one PR (per repo convention for adding an RPC, e.g. 4c3d53d5f).
* Per requested attribute value FQN it returns the attribute `rule` and the effective KAS keys, resolved server-side with **value > definition > namespace** precedence over the `SimpleKasKey` model (mirroring `sdk/granter.go` `newGranterFromService`). Replaces the full attribute+value graph that `GetAttributeValuesByFqns` returns for the encrypt/splits path.

Part of DSPX-2541. The companion `GetEntitleableAttributesByFqns` API is a stacked follow-up. Deprecating `GetAttributeValuesByFqns` and the SDK/decisioning migrations are tracked separately.

Limitation: values configured only with legacy `KeyAccessServer` grants (no `kas_keys`) return an empty key set; `kas_keys` is the forward model.

### Checklist

- [x] I have added or updated unit tests
- [x] I have added or updated integration tests (if appropriate)
- [ ] I have added or updated documentation

### Testing Instructions

* `go test ./service/policy/db/ -run TestResolveEffectiveKasKeys`
* `go test ./service/integration/ -run TestAttributesSuite/Test_GetKeyMappingsByFqns`
* `golangci-lint run ./policy/db/... ./policy/attributes/...` — 0 issues.

### Related

* [DSPX-2541](https://virtru.atlassian.net/browse/DSPX-2541)

[DSPX-2541]: https://virtru.atlassian.net/browse/DSPX-2541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3633, opened 2026-06-17
feat(policy): DSPX-2541 add GetEntitleableAttributesByFqns API

alkalescent:
### Proposed Changes

* Add `GetEntitleableAttributesByFqns`, a narrow server-side decisioning read API on the attributes service: proto, regenerated `protocol/go`, handler, DB method, lean SQL, and tests, all in one PR (per repo convention for adding an RPC).
* Per requested attribute value FQN it returns the attribute `rule`, the value id, the definition's ordered value FQNs (for hierarchy rule logic), and the value-level subject mappings. Subject mappings are fetched via a new `getSubjectMappingsByValueFqns` query in a single roundtrip, instead of the full `ListAllSubjectMappings` load the entitlement path uses today.

**Stacked on #3632** (base `DSPX-2541-key-mappings-api`). Migrating the v1/v2 decisioning + PDP consumers onto this API and adding namespace-scoped fetch is a separate PR (it depends on the still-open tenancy model under epic DSPX-2184).

### Checklist

### Testing Instructions

* `go test ./service/integration/ -run TestAttributesSuite/Test_GetEntitleableAttributesByFqns`
* `golangci-lint run ./policy/db/... ./policy/attributes/...` — 0 issues.

### Related

* [DSPX-2541](https://virtru.atlassian.net/browse/DSPX-2541)
* Key API PR: #3632

[DSPX-2541]: https://virtru.atlassian.net/browse/DSPX-2541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


alkalescent:
Combined into a single PR #3634 (both narrow attribute read APIs), per discussion: they edit the same files, match the repo convention of bundling related RPCs (4c3d53d5f), and release as one protocol/go version.

PR #3634, opened 2026-06-17
feat(policy): add narrow attribute read API protos and generated code

alkalescent:
## Proposed Changes

Proto and generated-code portion of the narrow attribute read APIs, split out so the heavy service implementation lands separately and `protocol/go` releases cleanly.

Adds two RPCs to `AttributesService`:
- **`GetKeyMappingsByFqns`** — per value FQN: the attribute rule and effective KAS keys (value > definition > namespace precedence), for client-side key splits.
- **`GetEntitleableAttributesByFqns`** — per value FQN: rule, value id, the definition's ordered value FQNs (hierarchy), and value-level subject mappings, for server-side decisioning.

Includes the `.proto`, regenerated `protocol/go`, gRPC/OpenAPI docs, and the generated `sdkconnect` client wrapper. The wrapper and the `AttributesServiceClient` test mocks ship here because CI regenerates them from the proto (`make connect-wrapper-generate`) and requires a clean tree. Server methods are `Unimplemented` stubs.

## PR stack
Replaces the previous combined PR per review (separate proto/generated from service logic). Merge order:
1. **this PR** — proto + generated + docs + sdkconnect wrapper
2. #3697 — service implementation (DB, SQL, real handlers, tests)

Follow-up feature branch: switch the SDK granter to consume `GetKeyMappingsByFqns` and add an xtest proving key-split parity with the previous SDK.


service/policy/db/attribute_fqn.go
```diff
@@ -205,6 +206,148 @@ func (c *PolicyDBClient) GetAttributesByValueFqns(ctx context.Context, r *attrib
...
+		return nil, fmt.Errorf("failed to marshal standard actions: %w", err)
+	}
+	customActionsBytes, err := json.Marshal(row.CustomActions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal custom actions: %w", err)
+	}
+	actions := []*policy.Action{}
+	if err := unmarshalAllActionsProto(stdActionsBytes, customActionsBytes, &actions); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/db/queries/subject_mappings.sql
```diff
@@ -202,6 +202,66 @@ ORDER BY
...
+    JOIN actions a ON sma.action_id = a.id
+    LEFT JOIN attribute_namespaces ans ON ans.id = a.namespace_id
+    LEFT JOIN attribute_fqns ans_fqns ON ans_fqns.namespace_id = ans.id AND ans_fqns.attribute_id IS NULL AND ans_fqns.value_id IS NULL
+    GROUP BY sma.subject_mapping_id
+)
+SELECT
+    sm.id,
+    fqns.fqn AS value_fqn,
+    sa.standard_actions,
+    sa.custom_actions,
```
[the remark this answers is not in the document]
alkalescent: fixed

jakedoublev:
@dmihalcik-virtru Would you mind taking a look as the author of the `autoconfigure` logic in the SDK today to build key splits? Any thoughts on this new API dedicated to that purpose?

service/policy/attributes/attributes.proto
```diff
@@ -340,6 +340,73 @@ message GetAttributeValuesByFqnsResponse {
...
+}
+message GetKeyMappingsByFqnsResponse {
+  message AttributeKeyMapping {
+    // the attribute value FQN this mapping is for
+    string fqn = 1;
+    // the attribute rule, which governs how splits combine (any_of / all_of / hierarchy)
+    AttributeRuleTypeEnum rule = 2;
+    // effective KAS keys resolved for this value (value > definition > namespace
+    // precedence, key mappings preferred over legacy grants); may be empty
+    repeated policy.SimpleKasKey keys = 3;
```
jakedoublev: cc @dmihalcik-virtru 

I think we may need to clarify in the response whether there are keys assigned at the attribute value level, definition level, and namespace level for each of the requested attribute value FQNs. If I have a key on only some of the attribute values under a single definition, it seems like it wouldn't be possible to easily tell from this API response if the specific keys I got in this list were for the value, definition, or namespace, unless it's more of a tuple with 3 indexes than a list?

service/policy/attributes/attributes.proto
```diff
@@ -340,6 +340,73 @@ message GetAttributeValuesByFqnsResponse {
...
+message GetEntitleableAttributesByFqnsRequest {
+  // Required
+  // Fully Qualified Names of attribute values (i.e. https://<namespace>/attr/<attribute_name>/value/<value_name>), normalized to lower case.
+  repeated string fqns = 1 [(buf.validate.field).repeated = {
+    min_items: 1
+    max_items: 250
+  }];
+}
+message GetEntitleableAttributesByFqnsResponse {
+  message EntitleableAttribute {
```
jakedoublev: Some ideas about this response:
1. Would it be an over-optimization if we only provide the entire list of values back for hierarchy attributes? If we have an any_of or all_of attribute, when decisioning in the PDP, we only need to know the rule and compare against the actual attributes in the decision request (on the TDF), such as /attr/project/value/abc and attr/project/value/xyz. If an entity is entitled to attr/project/value/jkl, it doesn't matter to the decision for any_of and all_of attributes, but that logic doesn't hold up for hierarchy attributes where entitlement to one of the values that is not present on the resource in the decision request could result in entitlement to the resource. 
2. Should we restructure the response so that we only return all of the values for each definition once? This way, if we have an attribute definition with 10k values and have to make a decision on a request with anywhere from 1-10k attribute values under that single definition, we only get the definition with all of its values 1 time instead of `n` times in the response back?

service/policy/attributes/attributes.proto
c-r33d: Dave will definitely be able to tell you, but the granter does need to be able to tell what split to make; which is derived from where the key lives. (value, def, namespace)

dmihalcik-virtru: In the key assignment algorithm, the most specific key is picked - for example, if there is a key on a value, that is preferred to keys on the attribute or namespace. IIRC if assigned two values from one attribute, and one has a key and the other does not, I think then both keys from the attr (or namespace) and value are used. For example if we have assignments:

```
a -> k1
a/x -> k2
a/y -> none
```

If `a` is an `and`, you'll get a split over k1 and k2, and if `a` is an `or` you'll get a share over k1 and k2.

alkalescent: Thanks, Dave. I'll keep the flat `repeated SimpleKasKey keys` since `resolveEffectiveKasKeys` returns the most specific keys per value (value > definition > namespace), and each entry carries the attribute rule.

service/policy/attributes/attributes.proto
alkalescent: These are some good ideas to fine-tune the entitlements response; I've incorporated both into the changes.

service/integration/attributes_test.go
```diff
@@ -1523,6 +1524,169 @@ func (s *AttributesSuite) Test_UnsafeUpdateAttribute_WithNewName() {
...
+	s.Equal(lowFqn, def.GetValues()[2].GetFqn())
+	// The mid value carries the subject mapping; siblings have none.
+	s.Len(def.GetValues()[1].GetSubjectMappings(), 1)
+	s.Empty(def.GetValues()[0].GetSubjectMappings())
+
+	e := resp.GetFqnEntitleableAttributes()[lowFqn]
+	s.Require().NotNil(e)
+	s.Equal(defFqn, e.GetDefinitionFqn())
+	s.Equal(lowFqn, e.GetValue().GetFqn())
+}
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/db/attribute_fqn.go
```diff
@@ -205,6 +205,182 @@ func (c *PolicyDBClient) GetAttributesByValueFqns(ctx context.Context, r *attrib
...
+// governing attribute rule and the effective KAS keys needed to build key
+// splits. Keys are resolved with value > definition > namespace precedence using
+// the mapped key model (SimpleKasKey), mirroring the client-side granter
+// resolution. Values configured only with legacy KeyAccessServer grants (no
+// kas_keys) return an empty key set; migrate such policy to keys to use this API.
+func (c *PolicyDBClient) GetKeyMappingsByFqns(ctx context.Context, r *attributes.GetKeyMappingsByFqnsRequest) (map[string]*attributes.GetKeyMappingsByFqnsResponse_AttributeKeyMapping, error) {
+	ctx, span := c.Start(ctx, "DB:GetKeyMappingsByFqns")
+	defer span.End()
+
+	pairs, err := c.GetAttributesByValueFqns(ctx, &attributes.GetAttributeValuesByFqnsRequest{Fqns: r.GetFqns()})
```
elizabethhealy: GetEntitleableAttributesByFqns below looks like it normalizes the fqns first, should we do that here as well?

service/policy/db/attribute_fqn_test.go
```diff
@@ -47,3 +48,62 @@ func TestDefinitionFqnFromValueFqn(t *testing.T) {
...
+		{
+			name:  "no keys at any level returns nil",
+			value: &policy.Value{},
+			attr:  def(nil, nil),
+			want:  nil,
+		},
+	}
+
+	for _, tc := range tests {
+		tc := tc
```
elizabethhealy: unnecessary decleration

service/policy/attributes/attributes.proto
```diff
@@ -340,6 +340,90 @@ message GetAttributeValuesByFqnsResponse {
...
+   Narrow read API for the client-side encrypt flow. Given attribute value FQNs,
+   returns only the information needed to build key splits: the governing
+   attribute rule and the effective KAS keys resolved server-side. It does not
+   return subject mappings, resource mappings, obligations, sibling values, or
+   metadata. Prefer this over GetAttributeValuesByFqns when building splits.
+*/
+message GetKeyMappingsByFqnsRequest {
+  // Required
+  // Fully Qualified Names of attribute values (i.e. https://<namespace>/attr/<attribute_name>/value/<value_name>), normalized to lower case.
+  repeated string fqns = 1 [(buf.validate.field).repeated = {
```
elizabethhealy, line 355: what is the expected behavior if one of the fqns dne? claude says current behavior is "it's silently absent from the response map", is that expected? i think the getattrvalbyfqn endpoint fails if any dne
either way can we add a test to make sure were validating that expected behavior

service/policy/attributes/attributes.proto
```diff
@@ -340,6 +340,90 @@ message GetAttributeValuesByFqnsResponse {
...
+    min_items: 1
+    max_items: 250
+  }];
+}
+message GetKeyMappingsByFqnsResponse {
+  message AttributeKeyMapping {
+    // the attribute value FQN this mapping is for
+    string fqn = 1;
```
c-r33d: Is this needed if you are returning a map<fqn, AttributeKeyMapping>?

service/policy/attributes/attributes.proto
```diff
@@ -340,6 +340,90 @@ message GetAttributeValuesByFqnsResponse {
...
+    // identity of the attribute value
+    string value_id = 2;
+    // subject mappings used to resolve entitlements for this value
+    repeated policy.SubjectMapping subject_mappings = 3;
+  }
+  // An attribute definition referenced by one or more requested value FQNs,
+  // returned once regardless of how many of its values were requested.
+  message EntitleableDefinition {
+    // the attribute definition FQN
+    string fqn = 1;
```
c-r33d: Is this needed, since the map is, map<def fqn, DefInfo>

service/integration/attributes_test.go
```diff
@@ -1523,6 +1524,178 @@ func (s *AttributesSuite) Test_UnsafeUpdateAttribute_WithNewName() {
 	s.Contains(val.GetFqn(), updated.GetName())
 }
 
+func (s *AttributesSuite) Test_GetKeyMappingsByFqns() {
```
c-r33d: Should we add more coverage here?

Maybe the following:
1.) If no definition key, assert we get a namespace key.
2.) Multiple attribute values return multiple keys
3.) If no namespace key, we return the base key?

service/integration/attributes_test.go
```diff
@@ -1523,6 +1524,178 @@ func (s *AttributesSuite) Test_UnsafeUpdateAttribute_WithNewName() {
...
+		s.NotNil(sm.GetSubjectConditionSet())
+	}
+
+	e2 := resp.GetFqnEntitleableAttributes()[fqn2]
+	s.Require().NotNil(e2)
+	s.Equal(defFqn, e2.GetDefinitionFqn())
+	s.Equal(value2.ID, e2.GetValue().GetValueId())
+	s.GreaterOrEqual(len(e2.GetValue().GetSubjectMappings()), 1)
+	for _, sm := range e2.GetValue().GetSubjectMappings() {
+		s.Equal(fqn2, sm.GetAttributeValue().GetFqn())
```
c-r33d: Should we check the same thing as above in this iterative loop? Maybe a chance to refactor to a common function.

service/policy/db/attribute_fqn.go
```diff
@@ -205,6 +205,182 @@ func (c *PolicyDBClient) GetAttributesByValueFqns(ctx context.Context, r *attrib
 	return list, nil
 }
 
+// GetKeyMappingsByFqns returns, for each requested attribute value FQN, the
+// governing attribute rule and the effective KAS keys needed to build key
+// splits. Keys are resolved with value > definition > namespace precedence using
+// the mapped key model (SimpleKasKey), mirroring the client-side granter
+// resolution. Values configured only with legacy KeyAccessServer grants (no
+// kas_keys) return an empty key set; migrate such policy to keys to use this API.
+func (c *PolicyDBClient) GetKeyMappingsByFqns(ctx context.Context, r *attributes.GetKeyMappingsByFqnsRequest) (map[string]*attributes.GetKeyMappingsByFqnsResponse_AttributeKeyMapping, error) {
```
c-r33d: Maybe not in this PR, but I somewhat think we want to consider returning a base key when no keys are found from the requested fqns

service/policy/db/attribute_fqn_test.go
alkalescent: fixed

service/policy/attributes/attributes.proto
alkalescent: Both RPCs should error now with tests.

service/policy/db/attribute_fqn.go
alkalescent: normalized

service/policy/attributes/attributes.proto
alkalescent: removed FQN from `AttributeKeyMapping`

service/policy/attributes/attributes.proto
alkalescent: removed FQN from `EntitleableDefinition`

service/policy/db/attribute_fqn.go
alkalescent: not a bad idea, but I'll defer for now

service/integration/attributes_test.go
alkalescent: I added 1) and 2)

service/integration/attributes_test.go
alkalescent: refactored to `assertValueEntry` helper

service/integration/attributes_test.go
```diff
@@ -1523,6 +1524,253 @@ func (s *AttributesSuite) Test_UnsafeUpdateAttribute_WithNewName() {
...
+		assertSingleKey(mappings[fqnGamma], policy.AttributeRuleTypeEnum_ATTRIBUTE_RULE_TYPE_ENUM_ANY_OF, key1)
+	})
+
+	s.Run("errors when a requested fqn does not exist", func() {
+		_, err := s.db.PolicyClient.GetKeyMappingsByFqns(s.ctx, &attributes.GetKeyMappingsByFqnsRequest{
+			Fqns: []string{"https://keymap-dne.example/attr/nope/value/nope"},
+		})
+		s.Require().Error(err)
+	})
+}
```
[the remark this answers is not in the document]
alkalescent: added

service/policy/attributes/attributes.proto
```diff
@@ -340,6 +340,86 @@ message GetAttributeValuesByFqnsResponse {
...
+    max_items: 250
+  }];
+}
+message GetKeyMappingsByFqnsResponse {
+  message AttributeKeyMapping {
+    // the attribute rule, which governs how splits combine (any_of / all_of / hierarchy)
+    AttributeRuleTypeEnum rule = 1;
+    // effective KAS keys resolved for this value (value > definition > namespace
+    // precedence, key mappings preferred over legacy grants); may be empty
+    repeated policy.SimpleKasKey keys = 2;
```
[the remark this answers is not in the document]
alkalescent: fixed

service/policy/db/attribute_fqn.go
```diff
@@ -205,6 +205,186 @@ func (c *PolicyDBClient) GetAttributesByValueFqns(ctx context.Context, r *attrib
...
+	defer span.End()
+
+	fqns := r.GetFqns()
+	normalized := make([]string, len(fqns))
+	for i, fqn := range fqns {
+		normalized[i] = strings.ToLower(fqn)
+	}
+
+	pairs, err := c.GetAttributesByValueFqns(ctx, &attributes.GetAttributeValuesByFqnsRequest{Fqns: normalized})
+	if err != nil {
```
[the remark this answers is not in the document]
alkalescent: fixed

c-r33d (dismissed):
Approved, but i would do a couple things before merging:

1.) I'd break up the protos and the service-level changes so we don't run into an issue with releases.
2.) I would make the SDK changes in another feature branch and add a test to xtest to show that given the changes, new service/sdk will produce the same split logic as the older previous sdk.

elizabethhealy (dismissed):

elizabethhealy (approved):

service/policy/attributes/attributes.go
```diff
@@ -172,6 +172,18 @@ func (s *AttributesService) GetAttributeValuesByFqns(ctx context.Context,
 	return connect.NewResponse(rsp), nil
 }
 
+func (s *AttributesService) GetKeyMappingsByFqns(_ context.Context, _ *connect.Request[attributes.GetKeyMappingsByFqnsRequest]) (*connect.Response[attributes.GetKeyMappingsByFqnsResponse], error) {
```
c-r33d, line 175: I think these could still have issues during release since it requires the protos to have been generated.

alkalescent: It's unavoidable bc of the "Gencode Up-to-date" check which requires `sdk/sdkconnect/attributes.go` with the two new wrapper methods. Then, we can release protocol/go and then update service/sdk go.mod. I believe anything breaking would cause checks to fail.

c-r33d: Makes sense.

c-r33d (approved):

PR #3645, opened 2026-06-18
fix(ers): coerce attempted_strategies []string to []interface{} for structpb

damonbanks:
## Summary

`Service.ResolveEntity` in `service/entityresolution/multi-strategy/service.go` stores `attempted_strategies` as a raw `[]string` in `result.Metadata`. The v2 `ResolveEntities` handler later iterates that metadata into a `map[string]interface{}` and calls `structpb.NewStruct`. `structpb.NewValue` accepts `string | float64 | bool | nil | map | []interface{}` only — a raw `[]string` is rejected with `proto: invalid type: []string`, the successful entity resolution is silently dropped via `continue` in the loop, and the response goes back with empty `entityRepresentations`.

Downstream effect: any caller that exercises the v2 `ResolveEntities` path with multi-strategy ERS gets no resolved entities, no entitlements, and (in the case of KAS rewrap) `PermissionDenied`.

One-line fix: coerce to `[]interface{}` at the point of insertion so all downstream consumers (v1 + v2 `ResolveEntities`) can serialize it.

## Commits — classic TDD shape

Two commits on this branch. Each is self-contained and reviewable in isolation.

| SHA | Type | Contents |
|---|---|---|
| `6ddbf70f` | `test(ers)` | Adds 7 failing regression tests. Fails against `main` with `proto: invalid type: []string`. |
| `eed88865` | `fix(ers)` | The point-of-insertion `[]string → []interface{}` coercion + pre-existing test's type-assertion update. Makes all 7 new tests pass. |

Reviewers can `git checkout 6ddbf70f && go test ./service/entityresolution/multi-strategy/... ./service/entityresolution/integration/` to see the RED state, then advance to `eed88865` to see it go GREEN.

## Reproduction (before this PR)

The failure is only visible in the full v2 PDP flow (`CreateEntityChainsFromTokens` → `ResolveEntities` on the resolved Entity → flatten `AdditionalProps` for `subject_external_selector_value` matching), not in the `Entity_Claims` direct-pass path the existing examples exercise.

Minimal config (LDAP provider, two-strategy chain so both ERS calls have a matching strategy):

```yaml
services:
  entityresolution:
    mode: multi-strategy
    failure_strategy: fail-fast
    providers:
      corporate_ldap:
        type: ldap
        connection: { host: ldap, port: 3890, use_tls: false, bind_dn: "uid=admin,...", bind_password: "..." }
    mapping_strategies:
      - name: ldap_lookup_from_jwt        # call 1: token claims (preferred_username from Okta/Keycloak)
        provider: corporate_ldap
        entity_type: subject
        conditions: { jwt_claims: [ { claim: preferred_username, operator: exists } ] }
        input_mapping: [ { jwt_claim: preferred_username, parameter: username, required: true } ]
        ldap_search: &search
          base_dn: "ou=people,dc=demo,dc=com"
          filter: "(&(objectClass=inetOrgPerson)(mail={username}))"
          scope: subtree
          attributes: [uid, mail, cn, clearance, nationality, needtoknow]
        output_mapping: &out
          - { source_attribute: uid,          claim_name: username }
          - { source_attribute: clearance,    claim_name: clearance }
          - { source_attribute: nationality,  claim_name: nationality, transformation: array }
          - { source_attribute: needtoknow,   claim_name: needtoknow,  transformation: array }
      - name: ldap_lookup_from_entity      # call 2: claims map derived from resolved Entity (userName camelCase)
        provider: corporate_ldap
        entity_type: subject
        conditions: { jwt_claims: [ { claim: userName, operator: exists } ] }
        input_mapping: [ { jwt_claim: userName, parameter: username, required: true } ]
        ldap_search:
          <<: *search
          filter: "(&(objectClass=inetOrgPerson)(uid={username}))"
        output_mapping: *out
```

**Before this PR:** authz returns `PermissionDenied`; dsp logs show `failed to create result struct ... error="proto: invalid type: []string"` for the second `ResolveEntities` call.

**After this PR:** the same call returns the full claim map in `additionalProps`; PDP flattens and matches against `subject_condition_set` selectors; KAS rewraps the key; `tructl decrypt` succeeds.

## Why this wasn't caught earlier

The reference multi-strategy ERS runbook validates resolution only via the Web Admin *Entity Resolution* tool, which posts a JWT wrapped as `Entity_Claims` and renders `additionalProps` for human inspection. The full v2 PDP flow (used by KAS rewrap) is the only path that re-resolves the resolved Entity from call 1; that's what triggers the `[]string`/`structpb` collision in metadata serialization. There is no encrypt→decrypt round-trip in the runbook, so the bug surfaces only when a real authz decision is requested against multi-strategy-resolved attributes.

The existing `service/entityresolution/integration/` suite covers **chain construction** (`CreateEntityChainsFromTokens` shape) and **strategy selection** (which strategy matches which JWT), but never asserts that resolved entities survive `structpb` serialization back to the caller. This PR's new integration tests close that gap.

### New test coverage (7 tests)

| Location | Test | Layer | Locks in |
|---|---|---|---|
| `multi-strategy/service_test.go` | `TestResolveEntity_MetadataIsStructpbSerializable` | Unit | **Type-hygiene contract** — every value in `EntityResult.Metadata` must pass `structpb.NewValue`. Guards the whole class of bug. |
| `multi-strategy/service_test.go` | `TestResolveEntity_AttemptedStrategiesStructpbSerializable` | Unit | Direct assertion on the specific field. |
| `multi-strategy/service_test.go` | `TestResolveEntity_AttemptedStrategiesAccumulatesAcrossContinue` | Unit | Accumulation across `continue` path stays serializable (guards a fix that only handles single-strategy success). |
| `multi-strategy/v2/registration_test.go` (new file) | `TestERSV2_ResolveEntities_PopulatesRepresentations` | Handler | **Direct symptom test** — v2 handler end-to-end, real service, in-process. Response must contain `EntityRepresentations` with `metadata_attempted_strategies` as a `ListValue`. Fails today with empty response. |
| `integration/multistrategy_v2_serialization_test.go` (new file) | `TestIntegration_ResolveEntities_ReturnsPopulatedRepresentation` | Integration | Full v2 pipeline via `Entity_Claims` payload. |
| `integration/multistrategy_v2_serialization_test.go` (new file) | `TestIntegration_TwoCallHandshake` | Integration | Exact `JustInTimePDP.resolveEntitiesFromToken` call pattern — `CreateEntityChainsFromTokens` then `ResolveEntities` on each chain entity. **The production flow that silently drops entities today.** |
| `integration/multistrategy_v2_serialization_test.go` (new file) | `TestIntegration_FailureIsolation_MixedBatch` | Integration | Mixed success/error batch — the handler's `continue` on structpb failure must not swallow legitimate successes. Empty response on old code (1/3 entities), all 3 on fixed code. |

### Verification

- **RED:** `git checkout 6ddbf70f && go test ./service/entityresolution/multi-strategy/... ./service/entityresolution/integration/` — all 7 new tests fail with `proto: invalid type: []string`.
- **GREEN:** `git checkout eed88865 && go test ./service/entityresolution/multi-strategy/... ./service/entityresolution/integration/` — all pass, including the pre-existing `TestMultiStrategyService_FailureStrategyContinue` (assertion updated from `[]string` to `[]interface{}` to reflect the new contract).
- **End-to-end:** TDF encrypt → decrypt verified locally against a DSP build vendoring this patch, with a Connect-RPC LDAP provider returning multi-valued ABAC attributes and an LDAP-only canary claim (`ldap_source_marker`) that survives end-to-end.

### Provider coverage note

New tests use the **claims provider** (in-process, no external deps) because the bug lives at `service.go:119` — after any provider returns — and is provider-agnostic. The existing `multistrategy_comprehensive_test.go` covers LDAP/SQL via testcontainers-go; those paths hit the same coerced line, so a separate test wouldn't catch a different failure mode.

## Checklist

- [x] Unit tests pass — 3 new unit tests + 1 handler test; pre-existing `TestMultiStrategyService_FailureStrategyContinue` assertion updated to reflect the new contract
- [x] Integration tests added — 3 new tests in `service/entityresolution/integration/multistrategy_v2_serialization_test.go` covering the exact `JustInTimePDP` call pattern the bug fires in
- [x] `golangci-lint` clean on changed files
- [x] `gofumpt` clean on changed files
- [x] No public-API change in the proto / connect surface (only the Go type stored in an opaque metadata map)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved multi-strategy entity resolution so `metadata_attempted_strategies` is reliably serialized and preserved in responses.
  - Ensured resolved entity representations include correctly mapped additional properties alongside the attempted-strategy metadata.
  - Enhanced mixed-batch handling: when configured to continue on failure, both successful and unresolved entities are returned with appropriate error details.

- **Tests**
  - Added new integration and unit coverage for serialization compatibility, two-step resolution flow behavior, strategy registration, and mixed-batch failure isolation.

### DIFF
--- a/service/entityresolution/integration/multistrategy_comprehensive_test.go
+++ b/service/entityresolution/integration/multistrategy_comprehensive_test.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net"
 	"path/filepath"
-	"reflect"
 	"runtime"
 	"strings"
 	"testing"
@@ -551,14 +550,20 @@ func TestMultiStrategy_ClaimsThenLDAPFallback(t *testing.T) {
 		t.Fatalf("Expected LDAP search_filter for alice, got %v", got)
 	}
 
-	attemptedStrategies, ok := result.Metadata["attempted_strategies"].([]string)
+	expectedStrategies := []string{"claims_direct_lookup", "ldap_directory_lookup"}
+	attemptedStrategies, ok := result.Metadata["attempted_strategies"].([]interface{})
 	if !ok {
-		t.Fatalf("Expected attempted_strategies to be []string, got %T", result.Metadata["attempted_strategies"])
+		t.Fatalf("Expected attempted_strategies to be []interface{}, got %T", result.Metadata["attempted_strategies"])
 	}
 
-	expectedStrategies := []string{"claims_direct_lookup", "ldap_directory_lookup"}
-	if !reflect.DeepEqual(attemptedStrategies, expectedStrategies) {
-		t.Fatalf("Expected attempted_strategies %v, got %v", expectedStrategies, attemptedStrategies)
+	if len(attemptedStrategies) != len(expectedStrategies) {
+		t.Fatalf("Expected %d attempted_strategies, got %d: %v", len(expectedStrategies), len(attemptedStrategies), attemptedStrategies)
+	}
+	for i, expected := range expectedStrategies {
+		got, isStr := attemptedStrategies[i].(string)
+		if !isStr || got != expected {
+			t.Fatalf("Expected attempted_strategies[%d] = %q, got %v", i, expected, attemptedStrategies[i])
+		}
 	}
 }
 
--- a/service/entityresolution/integration/multistrategy_v2_serialization_test.go
+++ b/service/entityresolution/integration/multistrategy_v2_serialization_test.go
@@ -1,0 +1,349 @@
+package integration
+
+import (
+	"context"
+	"testing"
+
+	"connectrpc.com/connect"
+	"github.com/opentdf/platform/protocol/go/entity"
+	entityresolutionV2 "github.com/opentdf/platform/protocol/go/entityresolution/v2"
+	"github.com/opentdf/platform/service/entityresolution/multi-strategy/types"
+	multistrategyv2 "github.com/opentdf/platform/service/entityresolution/multi-strategy/v2"
+	"github.com/opentdf/platform/service/logger"
+	"google.golang.org/protobuf/types/known/anypb"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// Integration tests targeting the exact response-marshalling path where the
+// []string -> structpb bug lives. The existing multistrategy_contract_test.go
+// suite covers chain construction and strategy selection but never asserts
+// that resolved entities survive serialization back to the caller — this
+// file fills that gap.
+//
+// These tests exercise the real ERSV2 handler (no HTTP layer) with real
+// providers, mirroring the JustInTimePDP call patterns from production.
+
+// serializationTestConfig builds a config with a claims provider and
+// strategies covering the shapes needed to exercise the two v2 call paths:
+//   - user_strategy: matches on `sub` (JWT-cased, as CreateEntityChainsFromTokens
+//     sees it on the first call).
+//   - user_mirror_strategy: matches on `userName` (proto-cased, as the default
+//     ResolveEntities path derives it from an Entity_UserName payload — the
+//     second call in the JustInTimePDP handshake).
+//   - client_strategy: environment entity matched via `azp`.
+//
+// The mirror strategy is what the spec's "Related concern" section flags as
+// commonly missing in real configs — including it here means the two-call
+// handshake test actually hits the success path on call 2, which is where
+// the []string -> structpb bug fires.
+func serializationTestConfig() types.MultiStrategyConfig {
+	return types.MultiStrategyConfig{
+		FailureStrategy: types.FailureStrategyContinue,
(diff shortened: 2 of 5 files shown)


pflynn-virtru (dismissed):

khvirtru:
Hey Damon — I was running the integration tests locally and found that `TestMultiStrategy_ClaimsThenLDAPFallback` fails after your structpb coercion fix. The test at `multistrategy_comprehensive_test.go:554` asserts `attempted_strategies` is `[]string`, but your fix changes it to `[]interface{}`.

Error:
```
Expected attempted_strategies to be []string, got []interface {}
```

I opened #3760 targeting this branch with a fix — the assertion now accepts both types. Could you take a look and merge it in?

Thanks!

pflynn-virtru (approved):

PR #3691, opened 2026-06-29
fix(kas): verify rewrap SRT with its actual JWS algorithm

dmihalcik-virtru:
verifySRTSignature assumed RS256, so a signed request token signed with an EC DPoP key (ES256/384/512) failed verification with 'unable to verify request token'. No shipping SDK hit this yet because clients decrypt via Bearer, but a DPoP-bound rewrap with an EC key cannot succeed.

Read the algorithm from the SRT's JWS header, validate it against an asymmetric allowlist (matching the DPoP proof allowlist in the auth package), and verify with that algorithm against the trusted DPoP key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved verification of signed request tokens to validate using the signature algorithm declared in the token (with a secure allowlist), rather than assuming a single algorithm.
  * Added safer fallback behavior when token metadata cannot be inspected, to avoid unexpected verification issues.
  * Updated behavior so requests signed with unsupported token algorithms are rejected with a clear error.
* **Tests**
  * Expanded signature verification coverage across multiple supported asymmetric algorithms and added rejection checks for unsupported algorithms.


elizabethhealy (dismissed):

dmihalcik-virtru:
Ugh I fixed cryptographically signing and somehow disabled --signoff, so I've had to re-push with signed-off commits

elizabethhealy (approved):

PR #3699, opened 2026-06-30
feat(sdk): resolve key splits via GetKeyMappingsByFqns

alkalescent:
## Proposed Changes

Switch the SDK granter to resolve key splits through the new `GetKeyMappingsByFqns` RPC instead of walking the full attribute set from `GetAttributeValuesByFqns`.

- `newGranterFromService` now calls `GetKeyMappingsByFqns` and feeds the returned per-value rule + effective KAS keys (server-side value > definition > namespace precedence) into the existing split planner.
- **Fallbacks preserve behavior:**
  - Values that return no mapped keys (e.g. legacy-grant-only policy) fall back per-value to `GetAttributeValuesByFqns`.
  - Older platforms that do not expose the RPC (`Unimplemented`) fall back entirely.
- Adds `TestReasonerKeyMappingFallback` covering both fallback paths; existing reasoner tests now exercise the new path (mapped keys) and the fallback (grants) and still pass.

## Stack
Depends on #3634 (proto + generated + `sdkconnect` wrapper). Base is the #3634 branch; rebase onto `main` and bump `sdk/go.mod` to the released `protocol/go` once #3634 merges.

## Follow-up
An xtest in `opentdf/tests` will prove the new go SDK produces the same TDF key splits as a previous released SDK (cross-SDK parity).

## Summary by CodeRabbit

* **New Features**
  * Prefer server-side key mappings (value → definition → namespace) when resolving grants into planning inputs.
  * Automatically cache and reuse mapped wrapping keys for consistent planning.
  * Fall back to full attribute-based resolution for legacy or unmapped values, including when mapping data is unavailable.

* **Bug Fixes**
  * Ensure correct planning when key-mapping responses are unimplemented or omit requested entries.
  * Maintain consistent behavior for policies that mix mapped-key values and legacy grant-based values.

* **Tests**
  * Added scenarios covering fallback for unimplemented/empty/partial mapping responses.
  * Added coverage for mixed mapping and legacy grant inputs producing the expected combined plan.

### DIFF
--- a/sdk/granter.go
+++ b/sdk/granter.go
@@ -447,24 +447,26 @@ func newGranterFromService(ctx context.Context, logger *slog.Logger, keyCache *k
 		// attribute lookup for all values.
 		if connect.CodeOf(err) == connect.CodeUnimplemented {
 			if e := grants.addGrantsFromAttributeValues(ctx, logger, keyCache, as, fqnsStr); e != nil {
-				return grants, e
+				return granter{}, e
 			}
 			return grants, nil
 		}
 		return granter{}, err
 	}
 
+	// Iterate the requested FQNs (not the response map) so a value the server
+	// omits still falls back rather than being silently dropped. The response is
+	// keyed by normalized FQN, and the protobuf getters are nil-safe, so an absent
+	// or nil entry yields empty keys and resolves via the fallback below.
+	mappings := km.GetFqnKeyMappings()
 	var fallback []string
-	for fqnstr, m := range km.GetFqnKeyMappings() {
-		fqn, err := NewAttributeValueFQN(fqnstr)
-		if err != nil {
-			return grants, err
-		}
+	for _, fqn := range fqns {
+		m := mappings[fqn.key]
 		keys := m.GetKeys()
 		if len(keys) == 0 {
-			// No mapped keys (legacy-grant-only or unconfigured); resolve via the
-			// full attribute lookup below.
-			fallback = append(fallback, fqnstr)
+			// No mapped keys (legacy-grant-only, unconfigured, or omitted); resolve
+			// via the full attribute lookup below.
+			fallback = append(fallback, fqn.key)
 			continue
 		}
 		// Only the rule and FQN are read downstream (constructAttributeBoolean /
@@ -492,7 +494,7 @@ func newGranterFromService(ctx context.Context, logger *slog.Logger, keyCache *k
 
 	if len(fallback) > 0 {
 		if err := grants.addGrantsFromAttributeValues(ctx, logger, keyCache, as, fallback); err != nil {
-			return grants, err
+			return granter{}, err
 		}
 	}
 
--- a/sdk/granter.go
+++ b/sdk/granter.go
@@ -10,6 +10,8 @@ import (
 	"sort"
 	"strings"
 
+	"connectrpc.com/connect"
+
 	"github.com/opentdf/platform/lib/ocrypto"
 	"github.com/opentdf/platform/protocol/go/policy"
 	"github.com/opentdf/platform/protocol/go/policy/attributes"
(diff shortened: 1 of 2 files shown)


sdk/granter.go
```diff
@@ -419,55 +421,121 @@ func (r granter) byAttribute(fqn AttributeValueFQN) *keyAccessGrant {
...
+	av, err := as.GetAttributeValuesByFqns(ctx, &attributes.GetAttributeValuesByFqnsRequest{Fqns: fqns})
+	if err != nil {
+		return err
+	}
+	for fqnstr, pair := range av.GetFqnAttributeValues() {
+		fqn, err := NewAttributeValueFQN(fqnstr)
+		if err != nil {
+			return err
+		}
 		def := pair.GetAttribute()
```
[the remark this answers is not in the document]
alkalescent: unnecessary

sdk/granter.go
```diff
@@ -419,55 +421,121 @@ func (r granter) byAttribute(fqn AttributeValueFQN) *keyAccessGrant {
...
+		}
+	}
+
+	if len(fallback) > 0 {
+		if err := grants.addGrantsFromAttributeValues(ctx, logger, keyCache, as, fallback); err != nil {
+			return grants, err
+		}
+	}
+
+	return grants, nil
```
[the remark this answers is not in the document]
alkalescent: fixed

c-r33d (commented):
I don't see any issues. Are there any tests that mix grants and mappings together? I looked through the old `granter_test.go` code and don't really see any. Do you think it would be useful to add at least one test where we have 2 fqns with mappings and 2 with grant only?

c-r33d (dismissed):

sdk/granter_test.go
```diff
@@ -812,6 +814,110 @@ func (*mockAttributesClient) GetAttributeValuesByFqns(_ context.Context, req *at
...
+	i := 0
+	plan, err := reasoner.plan([]string{kasUs}, func() string {
+		i++
+		return strconv.Itoa(i)
+	})
+	require.NoError(t, err)
+
+	got := make(map[string]bool)
+	for _, s := range plan {
+		got[s.KAS] = true
```
c-r33d: Nit: try to use map[string]struct{}, in golang empty struct's allocate zero bytes.

c-r33d (dismissed):

sdk/granter_test.go
alkalescent: learned something new
reads less clear to me but it's apparently idiomatic 😄

sdk/granter.go
```diff
@@ -419,55 +421,123 @@ func (r granter) byAttribute(fqn AttributeValueFQN) *keyAccessGrant {
...
+			// via the full attribute lookup below.
+			fallback = append(fallback, fqn.key)
+			continue
+		}
+		// Only the rule and FQN are read downstream (constructAttributeBoolean /
+		// ruleToOperator), so a minimal attribute is sufficient.
+		attr := &policy.Attribute{Fqn: fqn.Prefix().String(), Rule: m.GetRule()}
+		for _, sk := range keys {
+			if err := grants.addMappedKey(fqn, sk); err != nil {
+				logger.Debug("failed to add mapped key",
```
dmihalcik-virtru: Consider logging this as WARN or otherwise expose the error in a side channel somehow. We shouldn't rely on end users to use DEBUG logging to discover configuration issues

dmihalcik-virtru (dismissed):

sdk/granter.go
alkalescent: changed to WARN and added a `kas` field to make the log more consistent w the rest of the file

sdk/granter_test.go
```diff
@@ -812,6 +814,110 @@ func (*mockAttributesClient) GetAttributeValuesByFqns(_ context.Context, req *at
...
+		i++
+		return strconv.Itoa(i)
+	})
+	require.NoError(t, err)
+
+	got := make(map[string]struct{})
+	for _, s := range plan {
+		got[s.KAS] = struct{}{}
+	}
+	assert.ElementsMatch(t, []string{evenMoreSpecificKas, kasAu, kasCa}, slices.Collect(maps.Keys(got)))
```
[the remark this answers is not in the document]
alkalescent: fixed

elizabethhealy (approved):

Issue #3744, opened 2026-07-10
Unbounded attribute values in ListAttributes and GetAttribute responses


eugenioenko:
## Summary

The `ListAttributes` and `GetAttribute` RPCs return **all** attribute values for each attribute with no pagination or limit. If an attribute definition accumulates a large number of values, this can cause excessive memory usage, large response payloads, and high latency.

### Root Cause

Both `ListAttributes` and `GetAttribute` use the same `hydrateAttribute` path, which deserializes the full `valuesJSON` blob from the database query. The underlying SQL (`listAttributesDetail`, `getAttribute`) aggregates values via `JSON_AGG` with no limit:

```sql
-- service/policy/db/attributes.sql.go (listAttributesDetail)
JSON_AGG(
    JSON_BUILD_OBJECT(
        'id', avt.id,
        'value', avt.value,
        'active', avt.active,
        'fqn', CONCAT(fqns.fqn, '/value/', avt.value)
    ) ORDER BY ARRAY_POSITION(ad.values_order, avt.id)
) FILTER (WHERE avt.id IS NOT NULL) AS values
```

The `LEFT JOIN` on `attribute_values` (lines 787–795 in `attributes.sql.go`) has **no filter or limit** on the number of values joined per attribute definition.

### Affected RPCs

| RPC | Impact |
|-----|--------|
| `ListAttributes` | Every attribute in the paginated response carries **all** of its values. Pagination only controls the number of attribute definitions returned, not the values within each. |
| `GetAttribute` | Returns a single attribute with **all** of its values. |
| `ListAttributesByFqns` | Same `hydrateAttribute` path — all values included. |

### No Alternative Endpoint

The `ListAttributeValues` RPC was **deprecated and removed** — it returns `Unimplemented`:

```go
// service/policy/attributes/attributes.go:309
func (s *AttributesService) ListAttributeValues(_ context.Context, _ *connect.Request[attributes.ListAttributeValuesRequest]) (*connect.Response[attributes.ListAttributeValuesResponse], error) {
    return nil, connect.NewError(connect.CodeUnimplemented, ErrDeprecatedListAttributeValues)
}
```

This means there is currently **no paginated way** to retrieve attribute values for a given attribute definition.

### Potential Impact

- **Memory pressure on Postgres**: `JSON_AGG` builds the entire JSON array in memory server-side before returning.
- **Large gRPC responses**: An attribute with hundreds of thousands of values produces a response payload in the tens or hundreds of MB.
[...]

PR #3747, opened 2026-07-10
docs(adr): bounded attribute value responses

eugenioenko:
## Summary
- Adds a proposed ADR evaluating five options for addressing unbounded `JSON_AGG` on attribute values in the Attributes Service
- Rejects options 1 (modify v1) and 4 (application-layer cap) as breaking changes
- Remaining options: v2 with inline sub-pagination, v2 with dedicated ListAttributeValues RPC, or do nothing

## Related
- Issue: #3744

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture decision record proposing paginated retrieval for attribute values.
  * Documents a future approach that keeps attribute listing responses bounded while preserving compatibility with existing consumers.
  * Compares alternative designs and outlines considerations for performance, pagination, authorization, and platform alignment.

### DIFF
--- a/adr/decisions/2026-07-10-attributes-v2-bounded-value-responses.md
+++ b/adr/decisions/2026-07-10-attributes-v2-bounded-value-responses.md
@@ -1,0 +1,110 @@
+---
+status: proposed
+date: 2026-07-10
+tags:
+ - policy
+ - attributes
+ - pagination
+ - performance
+driver: Eugene Yakhnenko
+---
+# Bounded attribute value responses in the Attributes Service
+
+## Context and Problem Statement
+
+The `AttributesService` RPCs `ListAttributes`, `GetAttribute`, and `ListAttributesByFqns` return **all** attribute values inline via unbounded `JSON_AGG` in SQL. There is no pagination or limit on the number of values returned per attribute.
+
+An attribute with a large number of values (thousands or more) causes:
+- High memory pressure on Postgres during aggregation
+- Oversized gRPC responses that stress client and server memory
+- High latency on what should be lightweight list/get operations
+
+The dedicated `ListAttributeValues` RPC was removed in [PR #3108](https://github.com/opentdf/platform/pull/3108) as "redundant," leaving no paginated way to retrieve values independently.
+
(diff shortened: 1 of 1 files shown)


adr/decisions/2026-07-10-attributes-v2-bounded-value-responses.md
```diff
@@ -0,0 +1,108 @@
...
+### Option 3: v2 with dedicated ListAttributeValues RPC
+
+v2 `ListAttributes` and `GetAttribute` return attributes without values. Separate `ListAttributeValues` RPC with state/sort/search filters, accepts `repeated attribute_ids`.
+
+* 🟩 **Good**, attribute queries fully decoupled from values: no `JSON_AGG`
+* 🟩 **Good**, follows `ListRegisteredResourceValues` pattern
+* 🟩 **Good**, multi-attribute fetch in one paginated call avoids N+1
+* 🟩 **Good**, v2 versioning follows authorization v2 pattern, non-breaking
+* 🟥 **Bad**, consumers need two calls for attribute + values
+* 🟥 **Bad**, new SQL queries partially duplicate existing ones (minus values join)
```
pflynn-virtru, line 67:
```suggestion
* 🟥 **Bad**, new SQL queries partially duplicate existing ones (minus values join)
* 🟥 **Bad**, offset/limit pagination means deep offsets get expensive
```

eugenioenko: Thank you for taking a look!  That's technically 100% correct. 

Adding it here though would unfairly weight against this option when Options 1 and 2 would have the same issue. 
offset/limit to a deep value is an inherent problem of pagination that is applied in the rest of list endpoints as well. 

We could mitigate it by providing next cursor pagination but that should be done for all pagination lists and sounds like a separate (and very valid) ADR

pflynn-virtru (approved):
Decision 3

eugenioenko (commented):
The current API doesn't return RANK either and attribute values cannot be sorted by rank. This might be a separate issue but V2 implementation needs to include RANK and ability to sort by rank. This way if there is 1M attribute values and you want to get the one from lowers rank instead of highest.

PR #3752, opened 2026-07-16
fix(policy)!: GetPC does not respect manager field.

c-r33d:
>[!CAUTION]
>THIS IS A BREAKING CHANGE

1.) Add the new `manager_name` selector and deprecate the `name` only selector.
2.) Change otdfctl to match the api changes.

## Summary by CodeRabbit

* **New Features**
  * Provider configurations can now be retrieved using a name + key manager implementation pair.
  * Added `--manager` (`-m`) support to the provider retrieval command.
* **Bug Fixes**
  * Improved handling of incomplete/invalid lookup options, including clearer validation and error mapping for deprecated identifier usage.
  * Provider configurations with the same name can now be correctly distinguished by manager.
* **Documentation**
  * Updated the `otdfctl` man page with `--manager` details and examples.
* **Tests**
  * Expanded end-to-end and integration coverage for name+manager retrieval and failure scenarios.


service/policy/keymanagement/key_management.go
```diff
@@ -159,7 +159,7 @@ func (ksvc Service) GetProviderConfig(ctx context.Context, req *connect.Request[
 		return nil, connect.NewError(connect.CodeInvalidArgument, nil)
 	}
 
-	pc, err := ksvc.dbClient.GetProviderConfig(ctx, req.Msg.GetIdentifier())
+	pc, err := ksvc.dbClient.GetProviderConfig(ctx, req.Msg)
```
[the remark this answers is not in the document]
c-r33d: Added note to the protos.

service/policy/keymanagement/key_management.proto
```diff
@@ -33,9 +33,12 @@ message GetProviderConfigRequest {
   oneof identifier {
     option (buf.validate.oneof).required = true;
     string id = 2 [(buf.validate.field).string.uuid = true];
+    // Provider config name. When using name, manager SHOULD also be set because
+    // provider config uniqueness is based on (name, manager). Omitting manager
+    // can match multiple provider configs and returns a non-deterministic result.
```
jakedoublev: Should we send back a bad request error here if this is the case?

jakedoublev: Like don't try to optimistically get a provider config and return one of them, but instead require both parameters to successfully return any result, or otherwise make sure if one is found by name alone that it is globally unique?

jakedoublev (dismissed):

service/policy/keymanagement/key_management.proto
c-r33d: Went ahead and returned a deprecation error from the `GetProviderConfiguration` rpc when `name` is the selector.

alkalescent (dismissed):

jakedoublev (dismissed):

alkalescent (approved):

jakedoublev (approved):

PR #3760, opened 2026-07-21
fix(ers): update test assertion for attempted_strategies type after structpb coercion

khvirtru:
## Summary

- `TestMultiStrategy_ClaimsThenLDAPFallback` fails after the `[]string` → `[]interface{}` coercion in the service code because the test assertion only accepts `[]string`
- Updated the assertion to accept both `[]string` and `[]interface{}` for `attempted_strategies` metadata

## Context

Found while running integration tests locally on macOS (Colima). The structpb coercion fix in this PR correctly changes the internal type, but the comprehensive test at `multistrategy_comprehensive_test.go:554` was not updated to match.

## Test plan

- [x] `TestMultiStrategy_ClaimsThenLDAPFallback` passes locally (was failing before this fix)
- [x] All 23 multi-strategy integration tests pass (0 skipped, 0 failed)

### DIFF
--- a/service/entityresolution/integration/multistrategy_comprehensive_test.go
+++ b/service/entityresolution/integration/multistrategy_comprehensive_test.go
@@ -551,14 +551,23 @@ func TestMultiStrategy_ClaimsThenLDAPFallback(t *testing.T) {
 		t.Fatalf("Expected LDAP search_filter for alice, got %v", got)
 	}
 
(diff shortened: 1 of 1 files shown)


damonbanks (dismissed):
Great catch. Thanks

damonbanks (approved):
GTG

PR #3762, opened 2026-07-21
fix(sdk): preserve base path when building KAS rewrap client URL

elizabethhealy:
### Proposed Changes
`parseBaseURL` in `sdk/kas_client.go` reduced a KAS URL to `scheme://host[:port]`, discarding the URL path. The result is passed to `kasconnect.NewAccessServiceClient` as the Connect base URL, and Connect builds each request as `baseURL + "/" + procedure`. For a KAS served under a base path (e.g. `https://host/kas`), dropping the path sends `Rewrap` to `https://host/kas.AccessService/Rewrap` instead of `https://host/kas/kas.AccessService/Rewrap`, so the request never reaches the service.
  * preserve `u.Path` (trailing slash trimmed to avoid a doubled `//`) when constructing
    the base URL
  * update `TestParseBaseUrl` cases that asserted the path was stripped, and add coverage
    for base-path and trailing-slash inputs

#### Background
`parseBaseURL` began life as `getGRPCAddress`, which returned a gRPC dial target (`host:port`) where a path is meaningless — so stripping was correct. The Connect migration (#2200) renamed it and repurposed the return value as an HTTP base URL, but kept the host+port-only logic. A Connect base URL must retain the path, so the path-stripping became a regression. Host-root KAS deployments have no path to lose, which is why it went unnoticed.

#### Symptom
Because the mis-routed request is rejected by an upstream proxy/LB with a plain HTTP `403`, `connect-go` maps it to the `permission_denied` code — surfacing as `kao unwrap failed ... permission_denied: 403 Forbidden` during decrypt, which looks like an access-policy denial rather than a routing error.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed connectivity for KAS deployments hosted under a URL sub-path.
  * Preserved configured base paths while removing trailing slashes to prevent malformed request URLs.


elizabethhealy:
leaving as is

PR #3775, opened 2026-07-22
feat(sdk): add AccessToken accessor to SDK

alkalescent:
### Proposed Changes

* Add a public `AccessToken(ctx context.Context) (auth.AccessToken, error)` method on the `SDK` that returns a valid access token from the configured `tokenSource`, passing the SDK's own HTTP client (the same client the auth interceptor uses).
* Return a new `ErrNoAccessTokenSource` sentinel when the SDK was constructed without credentials (nil token source) instead of panicking. Uncredentialed clients are a supported configuration (`buildIDPTokenSource` returns nil in that case).

### Checklist

### Testing Instructions

- `cd sdk && go test . -race -run TestAccessToken`

## Summary by CodeRabbit

* **New Features**
  * Added access-token retrieval through the SDK’s authentication interface.
  * Supports configured token sources and reports clear errors when credentials are unavailable or invalid.

* **Tests**
  * Added coverage for successful token retrieval, missing token sources, and empty tokens.

### DIFF
--- a/sdk/sdk.go
+++ b/sdk/sdk.go
@@ -45,6 +45,7 @@ const (
 	ErrPlatformTokenEndpointNotFound = Error("token_endpoint not found in well-known idp configuration")
 	ErrPlatformEndpointNotFound      = Error("platform_endpoint not found in well-known configuration")
 	ErrAccessTokenInvalid            = Error("access token is invalid")
+	ErrNoAccessTokenSource           = Error("no access token source configured; SDK was created without credentials")
 	ErrWellKnowConfigEmpty           = Error("well-known configuration is empty")
 	ErrAttributeNotFound             = Error("attribute not found")
 )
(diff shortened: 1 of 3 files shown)


jakedoublev:
@marythought thoughts on this being a top-level method on the SDK instance vs nested somewhere? It seems like we want to be cautious about how many methods are on the SDK client itself, right?

sdk/sdk.go
```diff
@@ -305,6 +306,15 @@ func (s SDK) Close() error {
 	return nil
 }
 
+// AccessToken returns a valid access token for the SDK's configured credentials.
+// It returns ErrNoAccessTokenSource if the SDK was constructed without credentials.
+func (s *SDK) AccessToken(ctx context.Context) (auth.AccessToken, error) {
+	if s.tokenSource == nil {
```
bnrobinson93: Can we also check `s.tokenSource.AccessToken(...) == ""` to ensure that's gated?

marythought:
> @marythought thoughts on this being a top-level method on the SDK instance vs nested somewhere? It seems like we want to be cautious about how many methods are on the SDK client itself, right?

@alkalescent Rather than a top-level method, could we put this behind an Auth() accessor on the SDK — so the call reads `s.Auth().AccessToken(ctx)`? It's a departure from the current pattern, but it keeps the client's public surface lean and lines up better with how we group things in the Virtru SDK. Happy to talk it through if that's more than you signed up for.

```
  func (s *SDK) Auth() *AuthClient { return &AuthClient{sdk: s} }

  type AuthClient struct{ sdk *SDK }

  func (a *AuthClient) AccessToken(ctx context.Context) (auth.AccessToken, error) {
      if a.sdk.tokenSource == nil {
          return "", ErrNoAccessTokenSource
      }
      return a.sdk.tokenSource.AccessToken(ctx, a.sdk.httpClient)
  }

  Call site:
  token, err := s.Auth().AccessToken(ctx)
```

sdk/sdk.go
alkalescent: done

alkalescent:
@marythought moved to `Auth` accessor

marythought (approved):
thank you for doing this!

PR #3782, opened 2026-07-23
feat(sdk): add ChunkedWriter for out-of-order segment TDF creation

ronelliott:
## Proposed Changes

Adds ChunkedWriter, an SDK entry point that produces a TDF from segments
supplied one at a time. Callers encrypt and upload each segment
independently — off-thread or in parallel — then call Finalize to seal
the archive. Contrast with `SDK.CreateTDF`, which requires the full
plaintext buffered up front.

The writer exposes injection seams (`Clock`, `SegmentCipher`,
`ArchiveWriterFactory`, `KeySplitter`) so tests can substitute
deterministic or in-memory implementations without touching global
state. The default splitter is single-KAS only; callers with
attribute-based multi-KAS grants inject their own via
`WithChunkedKeySplitter`.

No reader is included — produced TDFs round-trip through the mainline
`SDK.LoadTDF` path.

### Commits in this PR

1. **`feat(sdk): add ChunkedWriter for out-of-order segment TDF creation`** — core interface, options, default segment cipher, single-KAS splitter, archive factory, initial round-trip / keep-segments / assertion-rejection tests.
2. **`feat(sdk): plumb Clock through ChunkedWriter to zipstream for deterministic ZIP output`** — threads `sdk.Clock` through `ArchiveWriterFactory` into `zipstream.WithClock`, so payload/manifest `ModTime` and per-segment `Written` all stamp from the injected clock instead of stdlib `time.Now`. Unblocks byte-deterministic ZIP headers for xtest fixtures.
3. **`test(sdk): expand ChunkedWriter coverage (out-of-order, clock, errors)`** — adds coverage for out-of-order writes with index-order concat, clock threading verified via `archive/zip` entry `ModTime`, and every error path on `WriteSegment` / `Finalize` / `WithChunkedSegments`.

### New files (all under `sdk/`)

- `chunked_writer.go`  — ChunkedWriter interface + concrete writer
- `chunked_options.go` — `WithChunked*` options
- `clock.go`           — Clock interface + `SystemClock` / `FixedClock`
- `segment.go`         — SegmentCipher interface + AES-256-GCM default
- `archive_writer.go`  — `ArchiveWriterFactory` (now Clock-aware) + ZIP64 default
- `key_splitter.go`    — KeySplitter interface + single-KAS default
- `chunked_test.go`    — 12 tests covering round-trip, out-of-order, keep-segments, clock threading, and error paths

### Modified files

- `sdk/internal/zipstream/{writer,segment_writer,zip_primitives}.go` — accept an injected `Now func() time.Time` (defaults to `time.Now`); `NewSegmentMetadata` takes the time source explicitly; `SegmentEntry.Written` stamps from it.

Jira: DSPX-2604

## Checklist

- [x] Signed commits with DCO sign-off
- [x] Unit tests added (12 total on the ChunkedWriter surface)
- [x] `make fmt` clean
- [x] `sdk` package tests pass with `-race`

## Testing Instructions

```
cd sdk && go test -run TestChunked -race .
```

Covers:
1. `TestChunkedRoundTrip` — 3 in-order segments, decrypt via mainline `SDK.LoadTDF` against in-process RSA-2048 fake KAS.
2. `TestChunkedKeepSegments` — write 3, keep prefix `[0, 1]`, decrypt only retained.
3. `TestChunkedFinalizeRejectsAssertions` — assertions option returns sentinel error (unsupported).
4. `TestChunkedOutOfOrderWrites` — write in `[2, 0, 3, 1]` order, concat by index, verify plaintext round-trip.
5. `TestChunkedClockThreadedToZipHeaders` — every ZIP entry `Modified` matches the injected `FixedClock`.
6. `TestChunkedDuplicateSegment`, `TestChunkedNegativeIndex`, `TestChunkedWriteAfterFinalize`, `TestChunkedDoubleFinalize`, `TestChunkedKeepSegmentsNonContiguous`, `TestChunkedKeepSegmentsUnwritten` — sentinel error contracts.
7. `TestChunkedGetManifestBeforeFinalize` — snapshot before Finalize, frozen manifest after.

Cross-SDK behavior is covered by `platform-xtest` in the checks below.

## Summary by CodeRabbit

- **New Features**
  - Added chunked TDF writing with independently encrypted, out-of-order segments.
  - Added configurable encryption, integrity protection, key splitting, metadata, MIME types, timestamps, and retained segments.
  - Added manifest snapshots before and after finalization.
  - Added AES-GCM encryption, single-KAS key splitting, encrypted metadata, and ZIP64 archives.
  - Added legacy and current format support.
- **Bug Fixes**
  - Added validation for invalid writes and segment retention requests.
  - Improved EC and hybrid key-wrapping reliability.


strantalis:
@ronelliott The experimental tdf package supports out of order segment writes. What is this pr addressing that currently doesn't exist?

dmihalcik-virtru:
## Perf review: measured A/B of `CreateTDF` on the ChunkedWriter

Someone raised the concern that moving `CreateTDF` onto the chunked writer regresses performance. I built a local A/B harness and measured it rather than guessing. **Short version: no wall-clock or peak-memory regression. There is one real regression, in allocation count.**

Full harness source, raw `benchstat` output, and repro commands are on [DSPX-4372](https://virtru.atlassian.net/browse/DSPX-4372); the summary for this ticket's own tracking is on [DSPX-2604](https://virtru.atlassian.net/browse/DSPX-2604).

### ⚠️ Read this first: these numbers are not about this PR's current head

This PR (`8cf39960`) **does not touch `sdk/tdf.go`.** The `CreateTDF` rewrite that the perf concern is about lives on commits that are not pushed here. So:

- **No CI run on this PR has ever exercised the change in question.**
- The numbers below come from those unpushed commits, measured locally against `main`.
- They need re-running once that code is actually pushed.

If the plan is to land the writer first and the `CreateTDF` switch separately, that is worth saying explicitly in the PR description, because reviewers are currently evaluating a perf question against a diff that does not contain the code.

### Results

Sequential, `benchstat` over `-count 6`, in-process KAS key (no network), so this isolates encrypt + archive:

| Payload | Wall clock | Bytes/op | Allocs/op |
| --- | --- | --- | --- |
| 1 KiB | **−61.84%** faster | **−97.16%** | ⚠️ +16.87% |
| 1 MiB | **−8.44%** faster | **−31.92%** | ⚠️ +18.12% |
| 16 MiB | no significant change (p=0.240) | no change | ⚠️ +23.42% |
| 128 MiB | no significant change (p=0.180) | no change | ⚠️ +44.51% |

Concurrent, 50-way (the shape the CI TDF3 benchmark exercises): 1 KiB **−5.52%**, 1 MiB **−7.77%** wall clock; allocs/op +19.11%.

4 GiB single streaming encrypt: peak `HeapInuse` 22.9–23.1 MiB before vs 20.9–23.3 MiB after, elapsed ~1.08 s both. It streams, and it streams no worse.

The bytes/op win is from replacing the unconditional 2 MiB read buffer with `min(segmentSize, inputSize)`.

### The one real regression: allocation count

Allocations are up +17% to +45% depending on size, and on the 4 GiB run mallocs went 15,690 → 23,900 (+52%). That is roughly **+4 allocations per segment**, so it scales linearly with segment count. Attribution, in rough order of how cheap they are to fix:

1. **`bytes.NewReader(sealed)` per segment** in `WriteSegment` — a heap allocation per segment purely to adapt a `[]byte` to an `io.Reader`.
2. **`map[int]*Segment` insert per segment** — both the `&Segment{}` and the map growth. If segments are usually written in order, a slice with a fallback map for out-of-order would drop this to amortized zero.
3. **`io.MultiReader` for segment 0** — allocates the multiReader plus its slice. Only affects the first segment, so it is a constant, not a scaling cost.
4. **base64 decode per segment at finalize** in `buildManifest`, when building the aggregate hash. Storing the raw digest alongside the base64 form would avoid re-decoding every segment at the end.

None of these showed up as slower wall clock at any size measured, so this is a GC-pressure argument, not a latency one. It is worth addressing before this becomes the default path, since segment count scales with payload size.

### Two behavioral notes for the reviewer

- **Output is byte-identical** to the current implementation for seekable readers. Verified at 1 KiB (3 writes / 2559 B) and 16 MiB (10 writes / 16,779,576 B), identical on both trees.
- **Non-seekable input produces +160 bytes** of output, because the size is unknown so the writer forces ZIP64. `WithInputSize(n)` fully restores the smaller output when the caller knows the length. Since non-seekable input is the entire point of DSPX-2604, this deserves a line in the doc comment.

### Do not use the CI `benchmark tests` job to judge this

Four runs of that job on the same branch reported TDF3 throughput of **100.62 / 105.89 / 114.11 / 174.60 req/s** — a 1.73x spread with no relevant code change between the last two. Its noise band is far wider than any regression worth caring about, it measures round trips through a live KAS over loopback (so SDK encrypt CPU is a small fraction of what is timed), and it reports throughput only — no memory, no allocations, no output size. It cannot confirm or refute a regression here. Fixing that is [DSPX-4372](https://virtru.atlassian.net/browse/DSPX-4372).

### Reproducing this

The harness is two `_test.go` files in package `sdk` (full source on DSPX-4372). It needs no network — it uses the existing `newSDK()` / `fakeWriter` / `mockRSAPublicKey1` test fixtures, plus one line that is easy to miss and produces a confusing nil-deref in `getBaseKey` if omitted:

```go
sdk := newSDK()
sdk.wellknownConfiguration = newMockWellKnownService(createWellKnown(nil), nil)
```

Then, from a checkout of each tree:

```bash
# A/B wall clock, bytes, allocs
cd sdk && go test -run '^$' -bench 'BenchmarkCreateTDFPerf' -benchmem -count 6 . | tee /tmp/<tree>.txt
benchstat /tmp/main.txt /tmp/change.txt

# 4 GiB peak heap + output write shape (note -count=1; results are cached otherwise)
PERF_PEAK=1 go test -count=1 -run 'TestCreateTDFPeakMemory|TestCreateTDFWriteShape' -v .
```

`-count=1` matters — without it Go serves cached results and you get byte-identical output across runs, which looks like a suspiciously stable measurement rather than no measurement at all.

### Suggested actions

**Submitter:**
- Say in the PR description whether the `CreateTDF` switch lands here or in a follow-up, since the perf discussion is about code not in this diff.
- Consider items 1 and 2 above (per-segment `bytes.Reader`, per-segment map insert) — cheapest allocation wins.
- Document the `WithInputSize` / ZIP64 interaction on the exported API.
- Answer @strantalis's open question below about overlap with `sdk/experimental/tdf`. DSPX-2604 is literally written as "graduate the experimental writer to stable," so the ticket text and this approach need reconciling.

**Reviewer:**
- Wall clock and peak memory are fine; the thing to push on is allocation count and the API/scope question, not throughput.
- Do not treat a green or red `benchmark tests` run as evidence either way.


[DSPX-4372]: https://virtru.atlassian.net/browse/DSPX-4372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSPX-2604]: https://virtru.atlassian.net/browse/DSPX-2604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSPX-4372]: https://virtru.atlassian.net/browse/DSPX-4372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ronelliott:
Superseded by https://github.com/opentdf/platform/pull/3940

PR #3783, opened 2026-07-23
feat(sdk): add ForAction helper to authorization/v2

marythought:
## Problem

The `authorization/v2` package provides ergonomic constructor helpers — `ForToken()`, `ForEmail()`, `ForAttributeValues()`, etc. — so consumers can build a `GetDecisionRequest` without importing internal proto packages. There is no equivalent helper for the request's `Action` field, so callers must import the `policy` package directly and hand-build the value:

```go
import (
    authorizationv2 "github.com/opentdf/platform/protocol/go/authorization/v2"
    "github.com/opentdf/platform/protocol/go/policy"
)

req := &authorizationv2.GetDecisionRequest{
    EntityIdentifier: authorizationv2.ForToken(jwt),
    Action:           &policy.Action{Name: "decrypt"},
    Resource:         authorizationv2.ForAttributeValues("https://example.com/attr/classification/value/confidential"),
}
```

This forces an extra import and leaks the `policy` package into consumer code.

## Change

Add `ForAction(name string) *policy.Action`, authored in `protocol/go/internal/authorization/v2/action.go` and copied into the public package via `protocol/codegen` (`action.gen.go`), following the same pattern as `entity_identifier` and `resource`.

```go
req := &authorizationv2.GetDecisionRequest{
    EntityIdentifier: authorizationv2.ForToken(jwt),
    Action:           authorizationv2.ForAction("decrypt"),
    Resource:         authorizationv2.ForAttributeValues("https://example.com/attr/classification/value/confidential"),
}
```

Because `ForAction` returns a `*policy.Action` (a different proto package), the codegen preserves the `policy` import and qualifier in the generated file — only the self-referencing `authorizationv2` alias is stripped.

## Testing

- `make proto-helper-generate` regenerates `action.gen.go`
- `go test ./protocol/go/internal/authorization/v2/...` — pass (new `action_test.go`)
- `go build` / `go vet` / `golangci-lint run` on `protocol/go` authz v2 packages — clean

## Summary by CodeRabbit

* **New Features**
  * Added a convenient helper for creating authorization actions from a name.
  * Supports both named and empty action values.

* **Tests**
  * Added coverage verifying that action names are preserved correctly.


marythought:
Codex review:

LGTM ✅
Nice addition — ForAction(...) rounds out the v2 helper set and keeps call sites cleaner by avoiding direct policy imports for common decision requests.
What I checked:
Helper implementation in generated and internal packages matches existing helper patterns.
Unit tests cover both normal and empty-string inputs.
Focused package tests pass:go test ./protocol/go/internal/authorization/v2
go test ./protocol/go/authorization/v2

No blocking issues from my review.

protocol/go/internal/authorization/v2/action_test.go
```diff
@@ -0,0 +1,22 @@
...
+import (
+	"testing"
+)
+
+func TestForAction(t *testing.T) {
+	name := "decrypt"
+	a := ForAction(name)
+
+	if a.GetName() != name {
+		t.Errorf("name = %q, want %q", a.GetName(), name)
```
elizabethhealy, line 12: nit: assert.Equal

protocol/go/internal/authorization/v2/action_test.go
```diff
@@ -0,0 +1,22 @@
...
+	}
+}
+
+func TestForAction_EmptyString(t *testing.T) {
+	a := ForAction("")
+
+	if a.GetName() != "" {
+		t.Errorf("name = %q, want empty string", a.GetName())
```
elizabethhealy, line 20: nit: assert.empty, or require.empty idr which it is

elizabethhealy (approved):

protocol/go/internal/authorization/v2/action_test.go
marythought: claude says testify is not a dependency here (yet) so I'll skip this for now so as not to blow up scope, but I appreciate the suggestion!

PR #3794, opened 2026-07-28
fix(ers): pass inline claims through multi-strategy context

### LINKED ISSUE #3790: fix(ers): ResolveEntities does not populate JWTClaimsContextKey for claims provider
## Bug

`ResolveEntities` in `service/entityresolution/multi-strategy/v2/registration.go` (line 92) passes bare `ctx` to `ers.service.ResolveEntity()` without populating `types.JWTClaimsContextKey`. The claims provider (`providers/claims/claims_provider.go` line 38) requires this key and returns an error when missing.

In contrast, `CreateEntityChainsFromTokens` (same file, line 240) correctly populates the context:
```go
ctxWithClaims := context.WithValue(ctx, types.JWTClaimsContextKey, jwtClaims)
```

## Impact

- **Claims provider always fails** on the `ResolveEntities` RPC path (used by authorization decisions via `GetDecision` → `resolveEntitiesFromEntityChain`)
- The bug is masked when `failure_strategy: "continue"` is used because the LDAP strategy succeeds after the claims provider silently fails
- Under `failure_strategy: "fail-fast"`, the claims provider failure aborts resolution immediately, preventing fallback to LDAP
- `Entity_Claims` entities (which carry claims inline) cannot be resolved at all through the `ResolveEntities` path

## Reproduction

PR #3791 includes a BDD test (`@claims-only-ers` in `tests-bdd/features/claims-only-ers.feature`) that reproduces this bug directly:

- **diana_claims** has inline claims `{"userName":"diana","department":"engineering"}` — expects PERMIT but gets **DENY** because the claims provider fails with "JWT claims not found in context"
- **bob_claims** has inline claims `{"userName":"bob","department":"marketing"}` — expects DENY and gets DENY (passes, but for the wrong reason — claims provider fails, not because policy correctly denied)

The `@multi-strategy-ers-failfast` test in the same PR also demonstrates the bug: alice gets DENY under fail-fast because `claims_passthrough` always fails before LDAP can execute. The error log confirms:

```
failed to resolve entity ... strategy execution failed with global fail-fast policy:
  JWT claims not found in context
```

**Once this bug is fixed, the `@claims-only-ers` diana scenario in PR #3791 will turn green.**

## Suggested Fix

One-line change in `registration.go`:

```go
// line 91-92, change:
result, err := ers.service.ResolveEntity(ctx, entityID, claimsMap)

// to:
ctxWithClaims := context.WithValue(ctx, types.JWTClaimsContextKey, types.JWTClaims(claimsMap))
result, err := ers.service.ResolveEntity(ctxWithClaims, entityID, claimsMap)
```

`claimsMap` is already computed for every entity type (Entity_Claims, Entity_UserName, etc.) at lines 66-89.

## Related

- PR #3791 — BDD tests reproducing this bug (intentionally failing `@claims-only-ers` scenario)
- PR #3767 — initial multi-strategy ERS BDD tests (merged)
- DSPX-4100 — multi-strategy ERS BDD coverage

jrschumacher:
## Summary
- pass inline ResolveEntities claims through `types.JWTClaimsContextKey` before calling the multi-strategy service
- add regression tests for both v1 and v2 claims-only ResolveEntities paths
- normalize result values before building protobuf structs so metadata/claim arrays serialize cleanly

## Testing
- go test ./entityresolution/multi-strategy/...
- golangci-lint run ./entityresolution/multi-strategy/...
- cd sdk && go test -run TestREADMECodeBlocks

Fixes #3790

## Summary by CodeRabbit

* **New Features**
  * Multi-strategy entity resolution now supports inline JWT claims, including claim-to-entity field mapping and metadata.
  * Added BDD support and end-to-end scenarios for claims-based authorization, including permitted and denied access outcomes.

* **Bug Fixes**
  * Improved normalization of claim and metadata values for reliable response formatting.
  * Claims context is only applied when provided by a claims entity.

* **Tests**
  * Added unit and integration coverage for claims resolution, mappings, metadata, and authorization decisions.

### DIFF
--- a/service/entityresolution/multi-strategy/v2/registration_test.go
+++ b/service/entityresolution/multi-strategy/v2/registration_test.go
@@ -1,7 +1,6 @@
 package multistrategy
 
 import (
-	"context"
 	"testing"
 
 	"connectrpc.com/connect"
@@ -137,16 +136,7 @@ func TestERSV2_ResolveEntities_PopulatesRepresentations(t *testing.T) {
 		},
 	})
 
-	// The claims provider reads JWT claims from ctx (see
-	// providers/claims/claims_provider.go). In production, upstream callers
-	// (admin API middleware, JustInTimePDP) populate this before invoking
-	// the handler; we mirror that here so the strategy actually succeeds
-	// and we exercise the serialization path where the bug lives.
-	ctx := context.WithValue(t.Context(), types.JWTClaimsContextKey, types.JWTClaims{
-		"sub":   "alice",
-		"email": "alice@example.com",
-	})
-	resp, err := ers.ResolveEntities(ctx, req)
+	resp, err := ers.ResolveEntities(t.Context(), req)
 	if err != nil {
 		t.Fatalf("ResolveEntities returned error: %v", err)
 	}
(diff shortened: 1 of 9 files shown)


service/entityresolution/multi-strategy/v2/registration_test.go
```diff
@@ -137,16 +136,7 @@ func TestERSV2_ResolveEntities_PopulatesRepresentations(t *testing.T) {
...
-	// The claims provider reads JWT claims from ctx (see
-	// providers/claims/claims_provider.go). In production, upstream callers
-	// (admin API middleware, JustInTimePDP) populate this before invoking
-	// the handler; we mirror that here so the strategy actually succeeds
-	// and we exercise the serialization path where the bug lives.
-	ctx := context.WithValue(t.Context(), types.JWTClaimsContextKey, types.JWTClaims{
-		"sub":   "alice",
-		"email": "alice@example.com",
-	})
-	resp, err := ers.ResolveEntities(ctx, req)
```
jrschumacher, line 149: Why was this removed?

khvirtru (dismissed):

strantalis (dismissed):

strantalis (approved):

service/entityresolution/multi-strategy/v2/registration_test.go
```diff
@@ -189,3 +179,149 @@ func TestERSV2_ResolveEntities_PopulatesRepresentations(t *testing.T) {
...
+	if got := result["metadata_source"]; got != "jwt_claims" {
+		t.Fatalf("expected metadata_source jwt_claims, got %v", got)
+	}
+	if _, hasError := result["error"]; hasError {
+		t.Fatalf("expected successful resolution, got error payload: %v", result["error"])
+	}
+}
+
+func TestResolveEntities_UserNameEntityDoesNotSeedClaimsContext(t *testing.T) {
+	t.Helper()
```
elizabethhealy, line 274: this is marked as a helper but it isnt called anywhere? (see others marked as helpers as well)

service/entityresolution/multi-strategy/v2/registration_test.go
```diff
@@ -189,3 +179,149 @@ func TestERSV2_ResolveEntities_PopulatesRepresentations(t *testing.T) {
...
+	}
+
+	if got := len(resp.Msg.GetEntityRepresentations()); got != 1 {
+		t.Fatalf("expected 1 entity representation, got %d", got)
+	}
+
+	props := resp.Msg.GetEntityRepresentations()[0].GetAdditionalProps()
+	if len(props) != 1 {
+		t.Fatalf("expected 1 additional props entry, got %d", len(props))
+	}
```
elizabethhealy, line 318: nit: assert/require instead of t.fatal (across PR)

service/pkg/protohelper/structpb.go
```diff
@@ -0,0 +1,29 @@
...
+			result[i] = StructPBCompatibleValue(item)
+		}
+		return result
+	case map[string]interface{}:
+		result := make(map[string]interface{}, len(v))
+		for key, item := range v {
+			result[key] = StructPBCompatibleValue(item)
+		}
+		return result
+	default:
```
elizabethhealy, line 26: this only handles []string, JWT claims can also produce []float64, []bool, or []int which will fall through to default, be returned as-is, and panic at structpb.NewStruct i believe. can you add a test to verify the behavior and possibly extend this to support those types if necessary?

elizabethhealy (approved):
lgtm mostly, only concern is around the structpb comment ^, but feel free to address separately if needed

service/pkg/protohelper/structpb.go
jrschumacher: Great call out! I'm going to handle this now.

service/entityresolution/multi-strategy/v2/registration_test.go
jrschumacher: Agreed. The top-level tests should not call `t.Helper()`. I removed those calls in the stacked follow-up #3836 so the existing ERS stack can merge without another restack.

service/entityresolution/multi-strategy/v2/registration_test.go
jrschumacher: Addressed in the stacked follow-up #3836. The tests introduced by this change now use `testify/require` consistently instead of manual `t.Fatalf`/`t.Errorf` checks.

service/pkg/protohelper/structpb.go
jrschumacher: Addressed with regression coverage in the stacked follow-up #3836. `StructPBCompatibleValue` now normalizes `[]float64`, `[]bool`, and `[]int` (including nested values), and the test verifies the normalized result is accepted by `structpb.NewStruct`.

PR #3791, opened 2026-07-27
chore(ers): expand multi-strategy ERS BDD coverage

khvirtru:
## Summary

> **Blocked by #3794** — `@claims-only-ers` scenarios will fail until the #3790 fix lands. The other two features (`@multi-strategy-ers-failfast`, `@claims-ldap-fallback-ers`) pass independently.

Expands multi-strategy ERS BDD test coverage (DSPX-4100) with three new feature files:

- **`@multi-strategy-ers-failfast`** — Demonstrates that `failure_strategy: "fail-fast"` stops entity resolution at the first strategy error, preventing LDAP fallback. Alice gets DENY (vs PERMIT under `continue` in the existing feature).

- **`@claims-ldap-fallback-ers`** — Validates condition-based strategy routing: a claims strategy with condition `department exists` is skipped when the entity lacks department, and the LDAP strategy with condition `userName exists` provides it via fallback. Covers both PERMIT (engineering user via LDAP) and DENY (operations user via LDAP).

- **`@claims-only-ers`** — Validates that `Entity_Claims` entities carrying inline claims can be resolved by a claims-only multi-strategy ERS without LDAP. Currently fails due to #3790; will pass once #3794 merges.

Also adds a DocString-based step definition for creating claims entities with embedded JSON.

### Test results

| Tag | Scenarios | Result |
|-----|-----------|--------|
| `@multi-strategy-ers-failfast` | 1 | ✅ Pass |
| `@claims-ldap-fallback-ers` | 2 | ✅ All pass |
| `@claims-only-ers` | 2 | ⏳ Blocked by #3794 |
| `@multi-strategy-ers` (existing) | 3 | ✅ All pass (no regression) |

### Related

- #3790 — `ResolveEntities` does not populate `JWTClaimsContextKey` for claims provider
- #3794 — Fix for #3790 (once merged, `@claims-only-ers` should pass)
- #3797 — Condition operator BDD coverage (merged)
- PR #3767 — Initial multi-strategy ERS BDD tests (merged)

## Test plan

- [ ] CI: `@multi-strategy-ers-failfast` and `@claims-ldap-fallback-ers` pass
- [ ] CI: `@claims-only-ers` expected to fail until #3794 merges
- [ ] After #3794 merges: rebase, reconcile overlapping step definitions, all scenarios pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end coverage for claims-based access decisions.
[...]

PR #3797, opened 2026-07-29
chore(ers): add BDD scenarios for condition operators equals, contains, regex

khvirtru:
## Summary

Adds BDD test coverage for the three untested condition operators in multi-strategy ERS strategy matching:

- **`@ers-condition-equals`** — 3 scenarios: exact match (alice PERMIT), negative match (bob DENY), case-insensitive matching (`values: ["ALICE"]` matches userName `alice`)
- **`@ers-condition-contains`** — 3 scenarios: substring match (`"ali"` in `"alice"` PERMIT), negative match (bob DENY), AND logic (two `contains` conditions — only `"diana"` matches both)
- **`@ers-condition-regex`** — 3 scenarios: pattern match (`^[a-d].*` matches alice PERMIT), negative match (henry DENY), AND logic (regex + exists conditions)

All 9 scenarios use LDAP strategies with `user_name` entities, avoiding the claims provider bug (#3790). Each feature file is `@stateless` and runs independently.

### Coverage before/after

| Operator | Unit tests | BDD tests (before) | BDD tests (after) |
|----------|-----------|--------------------|--------------------|
| `exists` | Many | Many | Many |
| `equals` | 1 (negative only) | 0 | 3 |
| `contains` | 0 | 0 | 3 |
| `regex` | 0 | 0 | 3 |
| AND logic (multiple conditions) | 0 | 0 | 2 |

### Related

- #3790 — claims provider context bug (why these tests use LDAP, not claims)
- #3791 — fail-fast, fallback, and claims-only BDD tests
- DSPX-4100 — multi-strategy ERS BDD coverage

## Test plan

- [ ] CI: all 9 scenarios should pass (equals, contains, regex)
- [ ] Run locally with `--godog.tags=@ers-condition-equals` / `@ers-condition-contains` / `@ers-condition-regex`

## Summary by CodeRabbit

* **Tests**
  * Added BDD coverage for ERS condition operators: `contains`, `equals` (including case-insensitive and multi-value), `regex`, and mixed-operator AND logic.
  * Validated substring and pattern matching semantics, non-matching behavior, and scenario routing outcomes.
  * Added end-to-end decision requests asserting expected **`PERMIT`** and **`DENY`** results across the new scenarios.

### DIFF
--- a/tests-bdd/features/ers-condition-and-logic.feature
+++ b/tests-bdd/features/ers-condition-and-logic.feature
@@ -14,7 +14,7 @@ Feature: ERS condition — multiple conditions AND logic
         jwt_claims:
           - claim: userName
             operator: contains
-            values: ["an"]
+            values: ["a"]
           - claim: userName
             operator: contains
             values: ["di"]
@@ -34,7 +34,7 @@ Feature: ERS condition — multiple conditions AND logic
       """
     And a local platform with inline ERS configuration
 
-  Scenario: Both AND conditions match — diana contains "an" and "di" gets PERMIT
+  Scenario: Both AND conditions match — diana contains "a" and "di" gets PERMIT
     Given I submit a request to create a namespace with name "and-match.test" and reference id "ns_and_match"
     And I send a request to create an attribute with:
       | namespace_id  | name       | rule  | values                         |
@@ -54,7 +54,7 @@ Feature: ERS condition — multiple conditions AND logic
     Then the response should be successful
     And I should get a "PERMIT" decision response
 
-  Scenario: Only one AND condition matches — alice contains "ali" but not "di" gets DENY
+  Scenario: Only one AND condition matches — alice contains "a" but not "di" gets DENY
     Given I submit a request to create a namespace with name "and-partial.test" and reference id "ns_and_partial"
     And I send a request to create an attribute with:
       | namespace_id    | name       | rule  | values                         |
--- a/tests-bdd/features/ers-condition-and-logic.feature
+++ b/tests-bdd/features/ers-condition-and-logic.feature
@@ -1,0 +1,75 @@
+@ers-condition-and-logic @stateless
+Feature: ERS condition — multiple conditions AND logic
+  Validate that when a strategy has multiple JWT claim conditions, ALL
+  conditions must match for the strategy to be selected (AND logic).
+
+  Background:
+    Given an LDAP directory with test users
+    And an ERS configuration with mode "multi-strategy" and failure strategy "continue"
+    And an ERS provider "ldap_directory" of type "ldap" connected to the LDAP directory
+    And an ERS mapping strategy "multi_condition_strat" using provider "ldap_directory"
+      """
+      entity_type: subject
+      conditions:
+        jwt_claims:
+          - claim: userName
+            operator: contains
+            values: ["an"]
+          - claim: userName
+            operator: contains
+            values: ["di"]
+      ldap_search:
+        base_dn: "ou=users,dc=opentdf,dc=test"
+        filter: "(&(objectClass=inetOrgPerson)(uid={username}))"
+        scope: subtree
+        attributes: ["uid", "mail", "departmentNumber"]
+      input_mapping:
+        - jwt_claim: userName
+          parameter: username
+      output_mapping:
+        - source_attribute: departmentNumber
+          claim_name: department
+        - source_attribute: uid
+          claim_name: username
+      """
+    And a local platform with inline ERS configuration
+
+  Scenario: Both AND conditions match — diana contains "an" and "di" gets PERMIT
+    Given I submit a request to create a namespace with name "and-match.test" and reference id "ns_and_match"
+    And I send a request to create an attribute with:
+      | namespace_id  | name       | rule  | values                         |
+      | ns_and_match  | department | anyOf | engineering,marketing,security |
+    Then the response should be successful
+    Given a condition group referenced as "cg_and_match" with an "or" operator with conditions:
+      | selector_value | operator | values      |
+      | .department    | in       | engineering |
+    And a subject set referenced as "ss_and_match" containing the condition groups "cg_and_match"
+    And I send a request to create a subject condition set referenced as "scs_and_match" containing subject sets "ss_and_match"
+    And I send a request to create a subject mapping with:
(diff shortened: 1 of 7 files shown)


tests-bdd/features/ers-condition-and-logic.feature
```diff
@@ -0,0 +1,75 @@
...
+    And I send a request to create a subject mapping with:
+      | reference_id  | attribute_value                                           | condition_set_name | standard actions | custom actions |
+      | sm_and_match  | https://and-match.test/attr/department/value/engineering  | scs_and_match      | read             |                |
+    Then the response should be successful
+    Given there is a "user_name" subject entity with value "diana" and referenced as "diana_and"
+    When I send a decision request for entity chain "diana_and" for "read" action on resource "https://and-match.test/attr/department/value/engineering"
+    Then the response should be successful
+    And I should get a "PERMIT" decision response
+
+  Scenario: Only one AND condition matches — alice contains "ali" but not "di" gets DENY
```
elizabethhealy: The scenario says "alice contains 'ali' but not 'di'" but the actual conditions in the Background are contains "an" and contains "di". "alice" doesn't contain "an" either, so zero conditions match here -- its not actually exercising the AND short-circuit. You'd need a user that matches exactly one of the two conditions to really test partial-AND. Maybe change the conditions to something like contains "a" + contains "di" so alice hits one but not the other?

tests-bdd/features/ers-condition-and-logic.feature
```diff
@@ -0,0 +1,75 @@
...
+      """
+      entity_type: subject
+      conditions:
+        jwt_claims:
+          - claim: userName
+            operator: contains
+            values: ["an"]
+          - claim: userName
+            operator: contains
+            values: ["di"]
```
elizabethhealy, line 20: nit: This only tests contains + contains on the same claim. The more realistic AND pattern, like the one featured in the docs is mixed operators on different claims (e.g., exists on one claim + equals on another — like the "Contextual Failover" example in the README). Adding a scenario to cover that could be helpful

tests-bdd/features/ers-condition-contains.feature
```diff
@@ -0,0 +1,75 @@
+@ers-condition-contains @stateless
+Feature: ERS condition operator — contains
+  Validate that the "contains" condition operator performs case-insensitive
```
elizabethhealy, line 3: The description here says it validates case-insensitive substring matching but neither scenario actually tests case-insensitivity since everything is lowercase.

tests-bdd/features/ers-condition-contains.feature
```diff
@@ -0,0 +1,75 @@
...
+    And an ERS configuration with mode "multi-strategy" and failure strategy "continue"
+    And an ERS provider "ldap_directory" of type "ldap" connected to the LDAP directory
+    And an ERS mapping strategy "contains_ali_strat" using provider "ldap_directory"
+      """
+      entity_type: subject
+      conditions:
+        jwt_claims:
+          - claim: userName
+            operator: contains
+            values: ["ali"]
```
elizabethhealy: nit: valueContainsIgnoreCase in the implementation has a separate code path for array-typed claim values (like "aud": ["internal", "partner"]). All the tests here use userName which is a string. Might be worth a scenario with an array claim to hit that code branch

tests-bdd/features/ers-condition-equals.feature
```diff
@@ -0,0 +1,75 @@
...
+    And an ERS configuration with mode "multi-strategy" and failure strategy "continue"
+    And an ERS provider "ldap_directory" of type "ldap" connected to the LDAP directory
+    And an ERS mapping strategy "alice_only_strategy" using provider "ldap_directory"
+      """
+      entity_type: subject
+      conditions:
+        jwt_claims:
+          - claim: userName
+            operator: equals
+            values: ["alice"]
```
elizabethhealy, line 20: nit: Every condition across all 5 files uses a single-element values array. The implementation does any-of matching within values[] (i.e., match if the claim equals any of the listed values). Might be worth adding a scenario with values: ["alice", "bob"] somewhere to cover that OR-within-condition semantic

elizabethhealy (dismissed):
looks good, some comments about some cases to add to get fuller coverage + 1 about discrepancy in the description vs what its actually testing. but you could tackle them as a follow on if you want

tests-bdd/features/ers-condition-and-logic.feature
khvirtru: Fixed — changed conditions to `contains "a"` + `contains "di"`. Now alice matches "a" (yes) but not "di" (no) = partial AND → DENY, properly testing the short-circuit. Diana still matches both → PERMIT.

tests-bdd/features/ers-condition-and-logic.feature
khvirtru: Added `ers-condition-and-mixed.feature` with `regex "^[a-d].*"` + `contains "an"` conditions. Diana matches both → PERMIT, alice matches regex but not contains → DENY.

tests-bdd/features/ers-condition-contains.feature
khvirtru: Fixed — changed the Background condition to `values: ["ALI"]` (uppercase) so the scenarios now actually validate case-insensitive substring matching against lowercase userName "alice".

tests-bdd/features/ers-condition-contains.feature
khvirtru: Good catch — the array code path in `valueContainsIgnoreCase` can only be exercised with multi-valued JWT claims (like `aud`). With `user_name` entities, all claims are flat strings. Testing array claims requires `Entity_Claims` entities, which are blocked by #3790 (`JWTClaimsContextKey` not populated). I'll add array claim coverage once #3790 is fixed — tracking this in DSPX-4100.

tests-bdd/features/ers-condition-equals.feature
khvirtru: Added `ers-condition-equals-multi.feature` with `values: ["alice", "diana"]`. Three scenarios: alice matches first value → PERMIT, diana matches second value → PERMIT, bob matches neither → DENY.

elizabethhealy (approved):

Issue #3821, opened 2026-08-03
GetDecisions fails with resource_exhausted when attribute has many values


khvirtru:
## Summary

`GetDecisions` returns HTTP 500 (`resource_exhausted`) when a request includes FQNs from an attribute definition that has a large number of values (e.g. 1000+ needtoknow PIDs). The internal SDK call from authorization → policy (`GetAttributeValuesByFqns`) returns the **entire attribute definition** with all values and their subject mappings, exceeding the 4MB gRPC message size limit.

## Reproduction

1. Create an attribute with 1000 values (e.g. `needtoknow` with pid-0001 through pid-1000), each with a subject mapping
2. Call `GetDecisions` with 9+ FQNs from that attribute:

```json
POST /authorization.AuthorizationService/GetDecisions
{
  "decisionRequests": [{
    "actions": [{"name": "read"}],
    "entityChains": [{"id": "test", "entities": [{"id": "subject", "category": "ENTITY_CATEGORY_SUBJECT", "userName": "some-user"}]}],
    "resourceAttributes": [{"attributeValueFqns": [
      "https://ngsap.mil/attr/needtoknow/value/pid-0001",
      "https://ngsap.mil/attr/needtoknow/value/pid-0002",
      ... (9+ PIDs) ...
      "https://ngsap.mil/attr/classification/value/topsecret",
      "https://ngsap.mil/attr/relto/value/usa"
    ]}]
  }]
}
```

3. Server returns: `{"code":"internal","message":"resource retrieval failed"}`

## Server log

```
level=ERROR msg="request error" namespace=authorization
  error="resource_exhausted: message size 4335321 is larger than configured max 4194304"
```

## Root cause

`retrieveAttributeDefinitions()` in `service/authorization/authorization.go:727` calls `GetAttributeValuesByFqns`, which executes the `listAttributesByDefOrValueFqns` SQL query. This query returns the full attribute definition including **all** values with their subject mappings, resource mappings, grants, and keys — not just the requested FQNs.

With 1000 needtoknow values, each having a subject mapping with condition set JSON, the response is ~4.1MB. The connect-go client uses the default `maxCallRecvMsgSize` of 4MB (`service/internal/server/server.go:103`), so requests with 9+ needtoknow FQNs push the response over the limit.

## Threshold data (medium scale: 1000 PIDs)

| Total FQNs in request | Result |
|----------------------|--------|
| 10 (8 PIDs + class + relto) | 200 OK |
[...]

PR #3832, opened 2026-08-04
fix(authz): Add exponential pagination backoff on subject mapping list for v2 GetDecisions

elizabethhealy:
### Proposed Changes

* Subject mappings can be >= 4mb if they have many scs, this can cause resource exhaustion errors
* Adds an exponential backoff to the pagination list limit to avoid this error
  * basically if it sees a resource exhaustion error, it cuts the list page limit and tries again

### Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved authorization policy retrieval when responses exceed message-size limits by automatically retrying with smaller page sizes.
  - Preserved complete results across paginated requests while continuing to surface unrelated errors.
  - Decision requests now succeed for large subject-mapping configurations when authorization remains valid.

- **Tests**
  - Added coverage for pagination, retry behavior, error handling, and large authorization configurations.
  - Updated test logging to report result counts instead of full response contents.


service/internal/access/v2/policy_store.go
```diff
@@ -33,6 +35,45 @@ var (
...
+	return connect.CodeOf(err) == connect.CodeResourceExhausted
+}
+
+// paginateAll invokes fetchPage until no pages remain (nextOffset <= 0). On resource_exhausted it
+// halves the page size and retries the same offset, keeping the load under the message limit
+// regardless of object size. fetchPage makes the list request and appends items; it must not append
+// when returning an error.
+func paginateAll(fetchPage func(offset, limit int32) (nextOffset int32, err error)) error {
+	var offset int32
+	limit := int32(entitlementPolicyMaxPageSize)
```
alkalescent, line 60: What happens when the policy service max is lower than the hardcoded 2500; does it fail? What if we use `ListRequestLimitDefault` from `service/policy/config/config.go` here as the initial limit instead?

strantalis:
@elizabethhealy do we have more context around where we are seeing this issue? Also does this have to do with us requesting all of policy?

elizabethhealy:
> @elizabethhealy do we have more context around where we are seeing this issue? Also does this have to do with us requesting all of policy?

@strantalis this specific failure hasnt come up from customers, but there was an issue around v1 getdecisions resource exhaustion https://github.com/opentdf/platform/issues/3821 -- we decided to just reccomend upgrading to v2 as a solution there -- but that triggered my investigation into the v2 logic and if resource exhaustion could be triggered there and i found this issue with overly padded subject mappings and resource mappings. ya i believe it does stem from the fact that were pulling all of policy. this is just a short term fix to make sure we dont run into any errors, but as were growing i believe pulling all of policy into memory to do getdecisions may not be feasible anymore -- we may have to rethink the logic here.

elizabethhealy:
moving back to draft until needed

PR #3845, opened 2026-08-06
feat(policy): Add sort functionality on key_status

elizabethhealy:
### Proposed Changes

* Adds the ability to sort on key status

### Checklist

## Summary by CodeRabbit

* **New Features**
  * Added sorting of KAS keys by key status.
  * Supports ascending and descending alphabetical ordering by status name.
  * Added a new key-status sorting option to the API.
  * Updated API documentation to describe the new option and status ordering.

* **Tests**
  * Added coverage validating key-status sorting in ascending and descending order.
  * Verified ordering across active, rotated, and unspecified key statuses.


service/integration/kas_registry_key_test.go
```diff
@@ -2501,6 +2501,46 @@ func (s *KasRegistryKeySuite) Test_ListKeys_SortByUpdatedAt_ASC() {
 	assertIDsInOrder(s.T(), list.GetKasKeys(), func(k *policy.KasKey) string { return k.GetKey().GetId() }, ids[0], ids[1], ids[2])
 }
 
+func (s *KasRegistryKeySuite) Test_ListKeys_SortByKeyStatus_ASC() {
```
c-r33d, line 2504: Should we rename this to SortByKeyStatus and do the ASC/DESC check in the same function?

elizabethhealy: we could, but this is matching the existing test naming, ex: Test_ListKeys_SortByUpdatedAt_ASC above

elizabethhealy: so if we do we should do it across the board

service/integration/kas_registry_key_test.go
```diff
@@ -2501,6 +2501,46 @@ func (s *KasRegistryKeySuite) Test_ListKeys_SortByUpdatedAt_ASC() {
...
+		KasFilter: &kasregistry.ListKeysRequest_KasId{KasId: kasID},
+		Sort: []*kasregistry.KasKeysSort{
+			{Field: kasregistry.SortKasKeysType_SORT_KAS_KEYS_TYPE_KEY_STATUS, Direction: policy.SortDirection_SORT_DIRECTION_DESC},
+		},
+	})
+	s.Require().NoError(err)
+	s.NotNil(list)
+
+	// ROTATED (2) before ACTIVE (1)
+	s.assertKeyStatusesOrdered(list.GetKasKeys(), false)
```
c-r33d: Should we just use s.Equal(t, expected, actual)? Where expected is an ordered list of expected results? You can just have the list be the IDs.

service/policy/kasregistry/key_access_server_registry.proto
```diff
@@ -79,6 +79,8 @@ enum SortKasKeysType {
   SORT_KAS_KEYS_TYPE_KEY_ID = 1;
   SORT_KAS_KEYS_TYPE_CREATED_AT = 2;
   SORT_KAS_KEYS_TYPE_UPDATED_AT = 3;
+  // Orders by policy.KeyStatus numeric value, not status name.
```
c-r33d: Do we think this is what customers expect?

service/integration/kas_registry_key_test.go
elizabethhealy: removed the new helper and switched to assertIDsInOrder to match the rest of the file

service/policy/kasregistry/key_access_server_registry.proto
elizabethhealy: updated based on conversation

c-r33d (approved):

PR #3867, opened 2026-08-12
feat(authz): wrap streaming/bidi handlers in Connect authN/authz interceptors

### LINKED ISSUE #3866: feat(auth): wrap streaming/bidi handlers in the Connect authN/authz interceptors
`ConnectAuthNInterceptor` and `ConnectAuthZInterceptor` (`service/internal/auth/authn.go`) are `connect.UnaryInterceptorFunc`, so they only cover unary RPCs. Add `WrapStreamingHandler` support so streaming/bidi RPCs get the same JWT/DPoP authN and casbin authz at stream establishment — ideally failing closed by default when `server.auth` is enabled.

This would let services register streaming RPCs and have platform auth apply to them the same way it does for unary today.

cshamrick:
Closes #3866.

The Connect authN/authz interceptors were `connect.UnaryInterceptorFunc`, so their streaming wrap methods were no-ops — any streaming/bidi RPC would bypass JWT/DPoP authN and casbin authz. Both now return a full `connect.Interceptor` that also covers streaming/bidi at stream establishment, failing closed by default.

Streaming authz is procedure-level only (no request message exists at establishment, so resource resolvers can't run), mirroring the raw-HTTP `MuxHandler` path.

### Checklist

## Summary by CodeRabbit

- **New Features**
  - Added authentication and authorization support for streaming and bidirectional RPCs.
  - Streams now authenticate at establishment and enforce procedure-level authorization by default.
  - Added optional fine-grained authorization based on each received message’s resource dimensions.
  - Authorization decisions are reused for matching dimensions and rechecked when dimensions change.
  - Unauthorized messages terminate the stream with a permission error.

- **Documentation**
  - Documented streaming authorization behavior, configuration, caching, and enforcement caveats.

### DIFF
--- a/adr/decisions/2026-01-02-authz-fine-grain-resource-support.md
+++ b/adr/decisions/2026-01-02-authz-fine-grain-resource-support.md
@@ -3,7 +3,7 @@
 **Status:** Proposed (active implementation)
 **Authors:** Platform Team
 **Created:** 2024-12-30
-**Last Updated:** 2026-03-05
+**Last Updated:** 2026-08-12
 **Record Date:** 2026-01-02 (ADR filename/index date)
 
 ## Problem Statement
@@ -327,6 +327,8 @@ The authorization interceptor orchestrates the authorization flow as a ConnectRP
 ```
 
 This flow describes **v2 mode**. In **v1 mode**, authorization uses the legacy path+action model and does not invoke resource resolvers.
+
+For **streaming and bidi RPCs**, the interceptor authenticates and authorizes once at stream establishment, before the handler runs. No request message is available at that point, so step 2 (resolver invocation) is skipped and enforcement uses wildcard dimensions (`*`). Streaming authorization is therefore procedure-level only, and dimension-scoped v2 policy fails closed for streams — a stream is authorized only by a wildcard-dimension rule. This mirrors the raw-HTTP `MuxHandler` path.
 
 **Key behaviors:**
 - No resolver registered = empty dimensions serialized as `*` (matches wildcard policies)
--- a/service/internal/auth/authn.go
+++ b/service/internal/auth/authn.go
@@ -633,178 +633,137 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 	})
 }
 
-// ConnectAuthNInterceptor authenticates Connect requests and enriches the
-// request context with configured token claims needed by later middleware.
-func (a Authentication) ConnectAuthNInterceptor() connect.UnaryInterceptorFunc {
-	interceptor := func(next connect.UnaryFunc) connect.UnaryFunc {
-		return connect.UnaryFunc(func(
-			ctx context.Context,
-			req connect.AnyRequest,
-		) (connect.AnyResponse, error) {
-			publicRoute := slices.ContainsFunc(a.publicRoutes, a.isPublicRoute(req.Spec().Procedure)) //nolint:contextcheck // There is no way to pass a context here
-			ctx = ctxAuth.ContextWithPublicRoute(ctx, publicRoute)
-			if publicRoute {
-				return next(ctx, req)
-			}
-
-			log := a.logger
-
-			procedure := req.Spec().Procedure
-			host := req.Header().Get("Host")
-			// Build the acceptable htu values with the same normalization the SDK
-			// applies when signing its proof (lowercased host, default ports
-			// stripped) so the exact-string htu comparison in matchHTU succeeds.
-			// Both schemes are offered because the interceptor cannot observe the
-			// wire scheme here.
-			ri := receiverInfo{
-				u: []string{
-					originFromHost(host, false) + procedure,
-					originFromHost(host, true) + procedure,
-				},
-				m: []string{req.HTTPMethod()},
-			}
-			log.DebugContext(
-				ctx, "dpop receiverInfo set (connect interceptor)",
-				slog.Any("expected_htm", ri.m),
-				slog.Any("expected_htu", ri.u),
-				slog.String("procedure", procedure),
-				slog.String("connect_http_method", req.HTTPMethod()),
-			)
-
-			header := req.Header()["Authorization"]
-			if len(header) < 1 {
-				return nil, connect.NewError(connect.CodeUnauthenticated, errors.New("missing authorization header"))
-			}
-
-			token, ctxWithJWK, err := a.checkToken(
-				ctx,
-				header,
-				ri,
-				req.Header()["Dpop"],
-			)
-			if err != nil {
-				// Check if this is a nonce error requiring a challenge
-				var nonceErr *DPoPNonceError
-				if errors.As(err, &nonceErr) {
-					nonce := a.dpopNonceManager.getCurrentNonce()
-					connectErr := connect.NewError(connect.CodeUnauthenticated, errors.New("unauthenticated"))
-					connectErr.Meta().Set("DPoP-Nonce", nonce)
-					connectErr.Meta().Set("WWW-Authenticate", `DPoP error="use_dpop_nonce"`)
-					return nil, connectErr
-				}
-				// Other DPoP proof failures get an invalid_dpop_proof challenge (RFC 9449 §7.1).
-				var proofErr *DPoPProofError
-				if errors.As(err, &proofErr) {
-					connectErr := connect.NewError(connect.CodeUnauthenticated, errors.New("unauthenticated"))
-					if a.dpopNonceManager.requireNonce {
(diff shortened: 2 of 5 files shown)


service/internal/auth/authn_test.go
```diff
@@ -758,6 +758,96 @@ func (r *authnTestRequest) HTTPMethod() string {
...
+	}
+	conn := newFakeStreamingHandlerConn("/kas.AccessService/Rewrap")
+	conn.RequestHeader().Set("Authorization", "DPoP token")
+
+	err := interceptor.WrapStreamingHandler(next)(s.T().Context(), conn)
+
+	s.Require().NoError(err)
+	s.Require().NotNil(handlerToken, "authenticated token should reach the handler context")
+	s.Equal("client-stream", handlerToken.PrivateClaims()["azp"])
+}
```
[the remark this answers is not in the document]
cshamrick: Done in `03f40f11`. Added `Test_ConnectAuthNInterceptor_Streaming_SetsNonceHeaderWhenRequired`: it enables `RequireNonce`, runs `WrapStreamingHandler` on a non-public route with a valid (stubbed) token, and asserts `conn.ResponseHeader().Get("DPoP-Nonce")` is non-empty on success.

service/internal/auth/authn.go
```diff
@@ -633,178 +633,137 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
...
+	return next
+}
 
-			header := req.Header()["Authorization"]
-			if len(header) < 1 {
-				return nil, connect.NewError(connect.CodeUnauthenticated, errors.New("missing authorization header"))
-			}
+func (i connectInterceptor) WrapStreamingHandler(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
+	return i.stream(next)
+}
```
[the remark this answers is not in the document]
cshamrick: Acknowledged as an accepted, non-blocking follow-up rather than a change in this PR. This PR is server-side middleware and deliberately fails closed for streams. Client-side streaming nonce retry belongs in `sdk/auth/dpop_transport.go`, which today only retries 401/400 unary responses, and is only exercised once (a) a real streaming RPC exists (there are none in-platform today) and (b) a deployment sets `require_nonce: true` (defaults to `false`). The server already emits the standard `use_dpop_nonce` challenge symmetrically for streams, so the fail-closed posture is correct in the meantime. Captured in the README/ADR streaming notes; will open a follow-up for the SDK transport.

service/internal/auth/authn.go
```diff
@@ -1042,13 +1130,48 @@ func (a *Authentication) authorize(
 		authzReq.ResourceContext = resourceCtx
 	}
 
-	// Perform authorization check
+	result := a.decide(ctx, log, authzReq)
+	if result.err == nil {
+		result.resourceContext = resourceCtx // Pass resolved data to handler for cache reuse
+	}
+	return result
```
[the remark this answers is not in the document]
cshamrick: Leaving as-is. As noted, `finalizeAuthzDecision` returns an error on deny, so `resourceContext` never reaches a handler for a denied result today — there is no leak. The attach is keyed off `result.err == nil` intentionally so the cache-reuse assignment stays in one place. Not worth adding a second condition that can only diverge from `finalizeAuthzDecision` over time.

service/internal/auth/interceptor_authz_test.go
```diff
@@ -559,13 +559,123 @@ func (s *InterceptorAuthzSuite) TestAuthorizeV2_UnregisteredResolverProcedureUse
...
+		return nil
+	}
+	conn := newFakeStreamingHandlerConn(requestedRPC)
+
+	err := authn.ConnectAuthZInterceptor().WrapStreamingHandler(next)(ctx, conn)
+
+	s.Require().Error(err)
+	s.Equal(connect.CodePermissionDenied, connect.CodeOf(err))
+	s.False(called, "dimension-scoped policy must fail closed for streams")
+}
```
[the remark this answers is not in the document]
cshamrick: Done in `03f40f11`. `TestStreaming_DimensionScopedPolicyFailsClosed` now registers a recording resolver for `GetAttribute` via `registry.ScopedForService(&attributes.AttributesService_ServiceDesc).Register(...)` and asserts `resolverCalled` is `false` after the stream is rejected — proving the mechanism (resolvers never run for streams; authorization is procedure-level only), not just the deny outcome. The v2 authorizer returns `SupportsResourceAuthorization() == true`, so the resolver would be reachable if the stream path ever tried to resolve.

service/internal/auth/authn.go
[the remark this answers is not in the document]
cshamrick: yes

adr/decisions/2026-01-02-authz-fine-grain-resource-support.md
```diff
@@ -328,6 +328,8 @@ The authorization interceptor orchestrates the authorization flow as a ConnectRP
 
 This flow describes **v2 mode**. In **v1 mode**, authorization uses the legacy path+action model and does not invoke resource resolvers.
 
+For **streaming and bidi RPCs**, the interceptor authenticates and authorizes once at stream establishment, before the handler runs. No request message is available at that point, so step 2 (resolver invocation) is skipped and enforcement uses wildcard dimensions (`*`). Streaming authorization is therefore procedure-level only, and dimension-scoped v2 policy fails closed for streams — a stream is authorized only by a wildcard-dimension rule. This mirrors the raw-HTTP `MuxHandler` path.
```
jakedoublev, line 331: @c-r33d Does this align with your design intentions for the fine-grained casbin authz feature?

ronelliott (dismissed):

adr/decisions/2026-01-02-authz-fine-grain-resource-support.md
c-r33d: I would think ideally no matter if it's a stream or not that our authZ system could react the same, meaning we could handle dimensional-level analysis if needed for streams.

c-r33d: I guess ideally too our MuxHandler could be dimension aware also. I wonder if there is anyway we can force the stream to re-Authz? We can then cache the response and allow the rest of the reqs through without re-authorizing.

cshamrick: @c-r33d ~~streams now re-Authz on the first message and cache the decision per dimension-set, as you suggested. Please take another look?~~ actually we're deferring this for later.

service/internal/auth/authn.go
```diff
@@ -633,178 +633,137 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
...
-				return nil, connect.NewError(connect.CodeInvalidArgument, err)
-			}
-			ctxWithJWK, err = a.subjectExtractor.ContextWithClaims(ctxWithJWK, token, roleReq)
-			if err != nil {
-				log.WarnContext(ctxWithJWK, "role provider error", slog.Any("error", err))
-				return nil, connect.NewError(connect.CodeInternal, errors.New("role provider error"))
+		},
+		stream: func(next connect.StreamingHandlerFunc) connect.StreamingHandlerFunc {
+			return func(ctx context.Context, conn connect.StreamingHandlerConn) error {
+				// Streaming RPCs are always POST and expose no HTTPMethod on the conn.
```
c-r33d, line 695: Maybe a link to connect documentation showing this:

- https://connectrpc.com/docs/protocol/

service/internal/auth/interceptor_authz_test.go
```diff
@@ -559,13 +559,139 @@ func (s *InterceptorAuthzSuite) TestAuthorizeV2_UnregisteredResolverProcedureUse
...
+	conn := newFakeStreamingHandlerConn("/kas.AccessService/PublicKey")
+	ctx := ctxAuth.ContextWithPublicRoute(s.T().Context(), true)
+
+	err := authn.ConnectAuthZInterceptor().WrapStreamingHandler(next)(ctx, conn)
+
+	s.Require().NoError(err, "public routes should bypass stream authorization")
+	s.True(called)
+}
+
+func (s *InterceptorAuthzSuite) TestStreaming_DimensionScopedPolicyFailsClosed() {
```
c-r33d, line 654: We should probably add one test for authorizer v1 also.

c-r33d (dismissed):

service/internal/auth/authn.go
cshamrick: done

service/internal/auth/interceptor_authz_test.go
cshamrick: done

c-r33d (approved):

PR #3873, opened 2026-08-13
refactor(authz): use targeted entitlement lookups

alkalescent:
### Proposed Changes

* replace the v1 authorization service's deprecated attribute lookup with batched entitleable-attribute requests
* match subject mappings before loading policy for the default entitlement Rego path
* preserve the paginated full-policy input for configured custom Rego

### Checklist

### Testing Instructions

- `make fmt`
- `golangci-lint run ./authorization`
- `go test ./authorization`
- `cd service && go test ./... -race` (all packages passed except `service/rttests`, which requires a platform on `localhost:8080`)
- `make lint` reaches all modules and reports existing findings in unchanged SDK, server, and BDD files
- `make test` stops in `lib/fixtures` because Keycloak is unavailable on `localhost:8888`

## Summary by CodeRabbit

- **New Features**
  - Improved authorization entitlement retrieval for custom and built-in policy configurations.
  - Added targeted attribute lookup with batching, scope filtering, deduplication, and hierarchy support.
  - Enhanced handling of entity properties and subject mappings.

- **Bug Fixes**
  - Improved validation and error reporting for missing definitions, invalid values, and mapping failures.
  - Ensured entitlement evaluation works when no selectors are matched.

- **Tests**
  - Added coverage for targeted lookups, batching, scope filtering, pagination, and error scenarios.

### DIFF
--- a/service/authorization/authorization.go
+++ b/service/authorization/authorization.go
@@ -15,6 +15,7 @@ import (
 	"github.com/go-viper/mapstructure/v2"
 	"github.com/open-policy-agent/opa/v1/ast"
 	"github.com/open-policy-agent/opa/v1/rego"
+	"github.com/opentdf/platform/lib/flattening"
 	"github.com/opentdf/platform/protocol/go/authorization"
 	"github.com/opentdf/platform/protocol/go/authorization/authorizationconnect"
 	"github.com/opentdf/platform/protocol/go/common"
@@ -39,6 +40,8 @@ import (
 )
 
 var ErrEmptyStringAttribute = errors.New("resource attributes must have at least one attribute value fqn")
+
+const maxEntitleableFQNsPerRequest = 250
 
 type AuthorizationService struct { //nolint:revive // AuthorizationService is a valid name for this struct
 	sdk    *otdf.SDK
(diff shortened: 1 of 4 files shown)


service/authorization/authorization.go
```diff
@@ -40,6 +41,8 @@ import (
 
 var ErrEmptyStringAttribute = errors.New("resource attributes must have at least one attribute value fqn")
 
+const maxEntitleableFQNsPerRequest = 250
```
elizabethhealy, line 47: whats the reasoning for 250? is this defined in the proto?

service/authorization/authorization.go
```diff
@@ -365,6 +397,19 @@ func (as *AuthorizationService) GetEntitlements(ctx context.Context, req *connec
...
 
+	var subjectMappings map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue
+	if as.usesCustomRego() {
+		subjectMappings, err = retrieveFullAttributeMappings(ctx, req.Msg.GetScope(), as.sdk)
+	} else {
+		subjectMappings, err = as.retrieveMatchedAttributeMappings(ctx, ersResp, req.Msg.GetScope())
+	}
+	if err != nil {
+		as.logger.ErrorContext(ctx, "failed to retrieve subject mappings", slog.String("error", err.Error()))
+		return nil, connect.NewError(connect.CodeInternal, err)
```
elizabethhealy: do we want to sanitize this error like the old code flow? ex: connect.NewError(connect.CodeInternal, errors.New("failed to list attributes")) instead of giving the full err stack

service/authorization/authorization.go
```diff
@@ -727,31 +796,68 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
...
+		for _, fqn := range batch {
+			entitleable, ok := resp.GetFqnEntitleableAttributes()[fqn]
+			if !ok || entitleable.GetValue().GetValueId() == "" {
+				return nil, status.Error(codes.NotFound, db.ErrTextNotFound)
+			}
+
+			definitionFQN := entitleable.GetDefinitionFqn()
+			definition, ok := resp.GetDefinitions()[definitionFQN]
+			if definitionFQN == "" || !ok || definition == nil {
+				return nil, connect.NewError(connect.CodeInternal, fmt.Errorf("entitleable attribute %q references missing definition %q", fqn, definitionFQN))
```
elizabethhealy: should we do status.Error or connect.Error? i see before we were doing status.error so this may be an existing issue

service/authorization/authorization.go
```diff
@@ -727,31 +796,68 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
...
+			}
+
+			attribute := &policy.Attribute{
```
elizabethhealy, line 836: should we set name here as well? for the fqn builder

service/authorization/authorization.go
```diff
@@ -727,31 +796,68 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
...
+						Fqn:             value.GetFqn(),
+						SubjectMappings: value.GetSubjectMappings(),
+					})
+				}
+			}
+
+			value := entitleable.GetValue()
+			result[fqn] = &attr.GetAttributeValuesByFqnsResponse_AttributeAndValue{
+				Attribute: attribute,
+				Value: &policy.Value{
```
elizabethhealy, line 855: same here for value.value

service/authorization/authorization.go
elizabethhealy, line 882: since were not populating non-hierarchy arrd defs in the map anymore, this will trigger a warn log for all non hierarchy attrs

service/authorization/authorization.go
```diff
@@ -365,6 +397,19 @@ func (as *AuthorizationService) GetEntitlements(ctx context.Context, req *connec
 		return nil, err
 	}
 
+	var subjectMappings map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue
+	if as.usesCustomRego() {
+		subjectMappings, err = retrieveFullAttributeMappings(ctx, req.Msg.GetScope(), as.sdk)
+	} else {
+		subjectMappings, err = as.retrieveMatchedAttributeMappings(ctx, ersResp, req.Msg.GetScope())
```
elizabethhealy, line 407: ai flagged this -- NOT_IN mappings on absent selectors now dropped
the old path handed the whole attribute/subject-mapping universe to Rego, which evaluated every condition set. NOT_IN initializes `notInResult := true` and only flips false on a match. an empty flattened result means the loop never runs and the condition passes. the new SQL prefilter  never gives Rego the chance

"
The MatchSubjectMappings prefilter changes which mappings can entitle. matchSubjectMappings (queries/subject_mappings.sql:388) filters on scs.selector_values && @selectors, so a mapping only returns if the entity's flattened props contain one of its selectors. But EvaluateCondition (subject_mapping_builtin.go:227-243) returns true for NOT_IN when the selector is absent. So .groups NOT_IN ["contractors"] used to entitle an entity with no .groups claim, and now won't. The len(properties) == 0 early return at :508 makes the extreme case explicit — zero props means zero entitlements with no policy evaluation. Intentional? v2 does the same (just_in_time_pdp.go:334), so it may be deliberate alignment, but it's a silent entitlement loss for existing v1 deployments and isn't in the PR description. Worth an xtest run exercising a NOT_IN mapping.
"

service/authorization/authorization.go
```diff
@@ -727,31 +796,68 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
...
-			return nil, status.Error(codes.NotFound, db.ErrTextNotFound)
+		if _, ok := seen[normalized]; ok {
+			continue
+		}
+		seen[normalized] = struct{}{}
+		normalizedFQNs = append(normalizedFQNs, normalized)
+	}
+
+	result := make(map[string]*attr.GetAttributeValuesByFqnsResponse_AttributeAndValue, len(normalizedFQNs))
+	for start := 0; start < len(normalizedFQNs); start += maxEntitleableFQNsPerRequest {
```
elizabethhealy, line 814: possible future improvement: the batches here could be paralellized

service/authorization/authorization.go
alkalescent: yes, it's the max from the proto

service/authorization/authorization.go
alkalescent: good idea, sanitized

service/authorization/authorization.go
alkalescent: agreed

service/authorization/authorization.go
alkalescent: changed so `getComprehensiveHierarchy` now keys the map on the requested value FQN too
non-hierarchy attrs should resolve without the warn now

service/authorization/authorization.go
alkalescent: we shouldn't need it bc `populateAttrDefValueFqns` only builds FQNs when `value.GetFqn() == ""`, but `GetEntitleableAttributesByFqns` should always be returning FQNs

service/authorization/authorization.go
alkalescent: same as above

service/authorization/authorization.go
alkalescent: I was trying to align v1's default entitlement path with v2. Both flatten the entity's properties and call `MatchSubjectMappings`, so only selectors the entity actually presents reach policy. A mapping whose condition references a selector the entity lacks (`.groups NOT_IN [...]` with no `.groups` claim) is never returned and so can't entitle. v2's `JustInTimePDP.getMatchedSubjectMappings` has always behaved this way.

Also, this only affects the built-in rego path. Configured custom rego still loads the full attribute and subject-mapping set via `retrieveFullAttributeMappings`, so `NOT_IN-on-absent-selector behavior is unchanged there.

Here's the v2 PR: https://github.com/opentdf/platform/pull/3912
xtest run [here](https://github.com/opentdf/tests/actions/runs/32520797420), but decrypt/rewrap path uses the v2 PDP (`AuthorizationV2.GetDecision`) rather than v1 `GetEntitlements`

elizabethhealy (approved):

PR #3889, opened 2026-08-18
feat(core): Expose real-time audit logging

c-r33d:
## Summary

  Expose audit logging support for asynchronous work that outlives the originating RPC.

  ## Changes

  - Added `Audit.Detach(ctx)` to:
    - Preserve request audit attribution and other context values.
    - Detach cancellation from the originating RPC.
    - Create an independent audit transaction without copying pending events.
   

  - Added `Audit.LogPolicyCRUD(...)` to:
    - Emit policy CRUD audit events immediately.
    - Support background jobs outside the RPC interceptor lifecycle.
    - Log event-construction errors internally.
    - Guard against misuse with the detached flag

  - Existing synchronous `LogAuditEvent` guards against detached auditTransactions.

## Summary by CodeRabbit

* **New Features**
  * Added support for immediately recording policy create, update, and delete audit events outside an RPC transaction.
  * Added context cloning that preserves audit attribution while isolating pending events.
* **Bug Fixes**
  * Improved consistency between buffered and immediate audit event records.
  * Preserved JWT claim details and default attribution across cloned contexts.
  * Prevented duplicate or unintended audit event emission.
* **Tests**
  * Expanded coverage for transaction independence, event emission, attribution, and audit payload handling.

### DIFF
--- a/service/logger/audit/logger.go
+++ b/service/logger/audit/logger.go
@@ -144,7 +144,7 @@ func (a *Logger) PolicyCRUDFailure(ctx context.Context, eventParams PolicyEventP
 func (a *Logger) LogPolicyCRUD(ctx context.Context, isSuccess bool, eventParams PolicyEventParams) {
 	tx, ok := ctx.Value(contextKey{}).(*auditTransaction)
 	if !ok || tx == nil || !tx.detached {
-		a.logger.ErrorContext(ctx, "immediate policy CRUD audit logging requires a detached audit context; use Logger.Clone first")
+		a.logger.ErrorContext(ctx, "immediate policy CRUD audit logging requires a detached audit context; use Logger.Detach first")
 		return
 	}
 
--- a/service/logger/audit/logger.go
+++ b/service/logger/audit/logger.go
@@ -178,8 +178,11 @@ func (a *Logger) GetDecisionV2(ctx context.Context, eventParams GetDecisionV2Eve
 
 func LogAuditEvent(ctx context.Context, verb Verb, event *EventObject) {
 	tx, ok := ctx.Value(contextKey{}).(*auditTransaction)
-	if !ok {
+	if !ok || tx == nil {
 		panic("audit transaction missing from context")
+	}
+	if tx.detached {
+		panic("cannot buffer an audit event on a detached transaction; use Logger.LogPolicyCRUD")
 	}
 	if event == nil {
 		panic("nil audit event provided")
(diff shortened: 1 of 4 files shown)


strantalis:
@c-r33d Did we find audit logging is blocking or causing performance regressions in our downstream services?

c-r33d:
> @c-r33d Did we find audit logging is blocking or causing performance regressions in our downstream services?

@strantalis 
Audit logging does not work for threads that live outside of the RPC flow because audit events are flushed by the connect interceptor; this work is needed to unblock downstream audit logging for such work.

service/logger/audit/logger.go
```diff
@@ -137,6 +137,27 @@ func (a *Logger) PolicyCRUDFailure(ctx context.Context, eventParams PolicyEventP
 	a.policyCrudBase(ctx, false, eventParams)
 }
 
+// LogPolicyCRUD creates and immediately emits a policy CRUD audit event. It is
+// intended for work whose lifetime is not owned by an RPC audit transaction.
+// If you are within a synchronous policy operation you should use: PolicyCRUDSuccess/Failure.
+// ! Use this carefully to avoid duplication of audit events from being recorded.
+func (a *Logger) LogPolicyCRUD(ctx context.Context, isSuccess bool, eventParams PolicyEventParams) {
+	tx, ok := ctx.Value(contextKey{}).(*auditTransaction)
+	if !ok || tx == nil || !tx.detached {
```
alkalescent, line 146: Heads up that `LogAuditEvent` appends to `tx.events`, so buffered audit calls on the cloned context are never emitted because the detached transaction has no close or flush path.

Could we reject detached transactions in the buffered path, or emit the event immediately when `tx.detached`?

c-r33d: I think blocking on the buffered path for tx.detached would be better. The way I would use this upstream is not appending to an in-context slice, but immediately logging via `LogPolicyCRUD`. 

While I think we could immediately log events in LogAuditEvent if the context is detached, that flow seems less obvious than calling a function that actually flushes. 

I think ideally in the long run we would have a different type of logger for a service that is not going to fit within the normal RPC flow, in addition that we make our audit logger more generic so we can pass in different object types.

This is somewhat of a quick-fix, until we determine a longer term solution for out-of-band services,

alkalescent: Blocking sounds fine to me + a test.

c-r33d: Fixed!

alkalescent (approved):

PR #3898, opened 2026-08-19
feat(sdk): cache ListKeyAccessServers response during decryption

### LINKED ISSUE #3897: feat(sdk): cache ListKeyAccessServers response during decryption
## Problem

Every call to `LoadTDF()` (TDF3 decrypt) and `setupKasAllowlist()` (bulk decrypt) triggers a `ListKeyAccessServers` RPC to the platform service to build a KAS allowlist. This happens on both decrypt paths:

- `SDK.LoadTDF()` in `sdk/tdf.go:882` via `getKasAllowList()`
- `SDK.setupKasAllowlist()` in `sdk/bulk.go:128` via `allowListFromKASRegistry()`

The result is never cached. There is no memoization and no deduplication of concurrent requests. If a caller decrypts 100 files, 100 identical RPCs are made to fetch the same KAS server list.

For comparison, KAS public keys *are* cached via `kasKeyCache` on the `SDK` instance (`sdk/kas_client.go:385`) with a 5-minute TTL. The server list has no equivalent caching.

Callers can work around this by passing `WithKasAllowlist()` directly, but the default path always makes the network call.

## Proposed Solution

Add a cache for the `allowListFromKASRegistry()` result on the `SDK` instance. The cache should:

- Be keyed by `platformUrl`
- Use a 5-minute TTL, consistent with `kasKeyCache`
- Deduplicate concurrent in-flight requests for the same platform URL
- Be scoped to the `SDK` instance so it does not leak across unrelated sessions
- Be invalidatable (e.g. `ClearKasAllowlistCache()`) for callers that need a fresh list

This would reduce redundant network calls without sacrificing security, since the KAS registry changes infrequently relative to decrypt operations.

See also: https://github.com/opentdf/web-sdk/issues/998 (same issue in the web SDK).

eugenioenko:
## Motivation

Avoid repeated `ListKeyAccessServers` calls during decryption while letting SDK users control cache freshness. Resolves #3897.

## PR Changes

- Add opt-in `WithKASAllowlistCache(ttl)`; disabled by default and rejects non-positive TTLs.
- Cache allowlists per SDK/platform endpoint with synchronized, deduplicated loads for standard and bulk decrypt paths.
- Add configuration, expiration, concurrency, miss-deduplication, and failure tests.

## Summary by CodeRabbit

* **New Features**
  * Added optional caching for KAS allowlists, configurable with a time-to-live setting.
  * Reuses cached allowlists by platform endpoint to reduce repeated registry requests.
  * Coordinates concurrent requests for the same uncached allowlist.
  * Added validation to reject non-positive cache durations.

* **Bug Fixes**
  * Failed allowlist loads are no longer retained in the cache.

### DIFF
--- a/sdk/kas_client.go
+++ b/sdk/kas_client.go
@@ -372,6 +372,39 @@ func (k *KASClient) getRewrapRequest(reqs []*kas.UnsignedRewrapRequest_WithPolic
 	return &rewrapRequest, nil
 }
 
+type kasAllowlistCache struct {
+	c map[string]timeStampedAllowList
+}
+
+type timeStampedAllowList struct {
+	AllowList
+	time.Time
+}
+
+func newKasAllowlistCache() *kasAllowlistCache {
+	return &kasAllowlistCache{make(map[string]timeStampedAllowList)}
+}
+
+func (c *kasAllowlistCache) clear() {
+	c.c = make(map[string]timeStampedAllowList)
+}
+
+func (c *kasAllowlistCache) get(platformURL string) AllowList {
+	cv, ok := c.c[platformURL]
+	if !ok {
+		return nil
+	}
+	if time.Now().Add(-5 * time.Minute).After(cv.Time) {
+		delete(c.c, platformURL)
+		return nil
+	}
+	return cv.AllowList
+}
+
+func (c *kasAllowlistCache) store(platformURL string, al AllowList) {
+	c.c[platformURL] = timeStampedAllowList{al, time.Now()}
+}
+
 type kasKeyRequest struct {
 	url, algorithm, kid string
 }
--- a/sdk/bulk.go
+++ b/sdk/bulk.go
@@ -124,13 +124,25 @@ func (s SDK) setupKasAllowlist(ctx context.Context, bulkReq *BulkDecryptRequest)
 			if err != nil {
 				return fmt.Errorf("retrieving platformEndpoint failed: %w", err)
 			}
-			// if no kasAllowlist is set, we get the allowlist from the registry
+
+			if s.kasAllowlistCache != nil {
+				if cached := s.kasAllowlistCache.get(platformEndpoint); cached != nil {
+					bulkReq.kasAllowlist = cached
+					bulkReq.TDF3DecryptOptions = append(bulkReq.TDF3DecryptOptions, withKasAllowlist(cached))
+					return nil
+				}
+			}
+
 			allowlist, err := allowListFromKASRegistry(ctx, s.logger, s.KeyAccessServerRegistry, platformEndpoint)
 			if err != nil {
 				return fmt.Errorf("failed to get allowlist from registry: %w", err)
 			}
 			bulkReq.kasAllowlist = allowlist
 			bulkReq.TDF3DecryptOptions = append(bulkReq.TDF3DecryptOptions, withKasAllowlist(bulkReq.kasAllowlist))
+
+			if s.kasAllowlistCache != nil {
+				s.kasAllowlistCache.store(platformEndpoint, allowlist)
+			}
 		} else {
 			s.Logger().Error("no KAS allowlist provided and no KeyAccessServerRegistry available")
 			return errors.New("no KAS allowlist provided and no KeyAccessServerRegistry available")
--- a/sdk/kas_client_test.go
+++ b/sdk/kas_client_test.go
@@ -395,6 +395,46 @@ func TestKasKeyCache_Expiration(t *testing.T) {
 	assert.False(t, exists, "Expired key should be removed from cache")
 }
 
+func TestKasAllowlistCache_Basic(t *testing.T) {
+	cache := newKasAllowlistCache()
+	require.NotNil(t, cache)
+
+	al := AllowList{}
+	require.NoError(t, al.Add("https://kas1.example.org"))
+	require.NoError(t, al.Add("https://kas2.example.org"))
+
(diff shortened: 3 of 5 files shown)


sdk/kas_client.go
```diff
@@ -372,6 +372,39 @@ func (k *KASClient) getRewrapRequest(reqs []*kas.UnsignedRewrapRequest_WithPolic
...
+	}
+	if time.Now().Add(-5 * time.Minute).After(cv.Time) {
+		delete(c.c, platformURL)
+		return nil
+	}
+	return cv.AllowList
+}
+
+func (c *kasAllowlistCache) store(platformURL string, al AllowList) {
+	c.c[platformURL] = timeStampedAllowList{al, time.Now()}
```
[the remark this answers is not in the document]
eugenioenko: Good catch — the existing \`kasKeyCache\` has the same lack of synchronization (plain map, no mutex, shared via pointer through value receivers). Adding a mutex only to the new cache would be inconsistent, and fixing both here is scope creep from the caching feature.

Tracked in #3899.

sdk/kas_client.go
```diff
@@ -372,6 +372,39 @@ func (k *KASClient) getRewrapRequest(reqs []*kas.UnsignedRewrapRequest_WithPolic
...
+
+func newKasAllowlistCache() *kasAllowlistCache {
+	return &kasAllowlistCache{make(map[string]timeStampedAllowList)}
+}
+
+func (c *kasAllowlistCache) clear() {
+	c.c = make(map[string]timeStampedAllowList)
+}
+
+func (c *kasAllowlistCache) get(platformURL string) AllowList {
```
pflynn-virtru: This map isn't synchronized, and unlike `kasKeyCache` (encrypt-path only) it's now hit from the decrypt path on a shared `SDK` — concurrent `LoadTDF` will `fatal error: concurrent map read and map write`. Needs a `sync.RWMutex`, with `get()` under the write lock since it `delete`s on expiry.

sdk/kas_client.go
```diff
@@ -372,6 +372,39 @@ func (k *KASClient) getRewrapRequest(reqs []*kas.UnsignedRewrapRequest_WithPolic
...
+	cv, ok := c.c[platformURL]
+	if !ok {
+		return nil
+	}
+	if time.Now().Add(-5 * time.Minute).After(cv.Time) {
+		delete(c.c, platformURL)
+		return nil
+	}
```
jrschumacher, line 400: @eugenioenko we can't hardcode a cache in our customer environments without making this configurable. The issue we've raised with caching is that we need a central way to configure, manage, and flush. Additionally, this lacks appropriate audit and logs to indicate that the result was a cache hit or miss. 

In this case if the cache has a bad KAS added this will leave clients vulnerable for 5 minutes right? How does this work if the KAS is not added will the client not rewrap?

Let's discuss before moving forward.

eugenioenko: @jrschumacher 
The existing kasKeyCache uses the same pattern: hardcoded 5-min TTL, no configurability, no central flush, no hit/miss logging. This PR intentionally mirrors that for consistency. 

Making caches configurable and adding audit logging are valid improvements. They should be done across both caches, not gate one cache while the other has the same characteristics. This PR is not fixing pre-existing cache issue, it's reducing calls to KAS allowlist.

Issue created to fix hardcoded cache for pre-existing Key Cache and the cache from this PR:
 https://github.com/opentdf/platform/issues/3916

Alternatively issue #3916 can be done first and this PR will be blocked until that's done.
Happy to follow up ones #3916 is resolved

eugenioenko: On the security concern: 

the allowlist doesn't cause clients to contact a KAS, it only permits contact with KASes already named in the TDF manifest. A rogue KAS added to the registry won't receive rewrap requests unless a TDF was encrypted with it (at which point the security issue was at encryption time). 
 
The reverse scenario: 
A KAS is removed from the registry but the cache keeps it allowed for up to 5 minutes.
This is the same tradeoff the key cache already accepts, this is a pre-exisitng issue with cache in the sdk.

dmihalcik-virtru: kasKeyCache is also on a legacy path we are planning to deprecate, so it doesn't get a lot of love; I wouldn't use it as an exemplar of how to do things

eugenioenko: Good point. KasKeyCache isn't the pattern to follow if it's being deprecated.  And more so with the pre-existing concurrency issue #3899

Happy to update this PR to match whatever caching strategy the team lands on (#3916) and fixes (#3899)

Feel free to close in the meantime if you'd prefer to land the cache architecture first.

eugenioenko: As talked about, this was made opt in with ability for setting ttl

sdk/kas_client.go
```diff
@@ -372,6 +373,53 @@ func (k *KASClient) getRewrapRequest(reqs []*kas.UnsignedRewrapRequest_WithPolic
 	return &rewrapRequest, nil
 }
 
+type kasAllowlistCache struct {
+	entries map[string]timeStampedAllowList
+	ttl     time.Duration
+	mu      sync.Mutex
+}
```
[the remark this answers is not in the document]
eugenioenko: Addressed in 4dc9bddff: allowlist loads now use keyed singleflight coordination with a cache recheck and success-only storage. Both standard and bulk decrypt paths use the coordinated loader. The concurrent-miss test runs 100 callers and asserts exactly one registry RPC; failed-load retry is also covered.

eugenioenko:
@coderabbitai full review

PR #3901, opened 2026-08-19
feat(core): add canonical audit recorder

strantalis:
### Proposed Changes

Injected services need a public audit API that works without a request transaction and reports validation or delivery failures. Add synchronous `Recorder.Record(ctx, Event)` and an instance-scoped `Processor`, configurable through `server.WithAuditProcessor`.

PEPs and other consumers configure `AuditProcessor` and `AuditTimeout` through `logger.Config`, passed to `logger.NewLogger` without functional options. `AuditProcessor` is runtime-only and excluded from serialized configuration.

- Apply audit configuration only during setup; copy input mappings once and read them without config locks or per-event cloning.
- Reuse the existing typed event model and retain `EventObject` as a compatibility alias.
- Stamp request attribution and the authenticated principal, validate generic fields, and process under a deadline detached from request cancellation.
- Configure the processing budget through `logger.audit_timeout` in YAML or `OPENTDF_LOGGER_AUDIT_TIMEOUT`. An explicit `server.WithAuditTimeout` overrides those settings; five seconds remains the default, including for non-positive values. Namespace loggers inherit the effective processor and budget.
- Let embedding applications own destination-specific conversion, validation, fan-out, delivery, and recovery.
- Treat each call as an independent event; no attempted/completed lifecycle API is introduced.
- Preserve the existing `level:"AUDIT"`, `msg:<verb>`, `audit:{...}` output with the default processor.

`Record` returns processing errors directly. Default slog acceptance does not guarantee downstream persistence. Buffered compatibility methods are integrated in #3902.

PR 1 of 3 in stack #3976. Base: `main`. Next: #3902.

### Checklist

- [x] Unit tests added or updated
- [x] Documentation added or updated
- [ ] Full local integration validation

### Testing Instructions

- `make fmt`
- `go test -race ./service/logger/... ./service/pkg/config ./service/pkg/server`
- Focused Go lint passed with zero new issues for the combined updates.
- `git diff --check`

Full `make lint` stops at an invalid Buf API token. Full `make test` passes the focused packages but fails in integration suites requiring a working Colima test-container socket and running Platform/Keycloak services.

### DIFF
--- a/docs/Configuring.md
+++ b/docs/Configuring.md
@@ -100,6 +100,7 @@ Root level key `logger`
 | `level`             | The logging level.                                           | `info`   | OPENTDF_LOGGER_LEVEL             |
 | `type`              | The format of the log output.                                | `json`   | OPENTDF_LOGGER_TYPE              |
 | `output`            | Stream output for logs, stderr or stdout                     | `stdout` | OPENTDF_LOGGER_OUTPUT            |
+| `audit_timeout` | Audit processing budget; non-positive values use the default | `5s` | OPENTDF_LOGGER_AUDIT_TIMEOUT |
 | `trace_correlation` | Add the active trace and span IDs to log and audit records   | `true`   | OPENTDF_LOGGER_TRACE_CORRELATION |
 
 Example:
@@ -109,6 +110,7 @@ logger:
   level: debug
   type: text
   output: stderr
+  audit_timeout: 5s
 ```
 
 `trace_correlation` emits fields only when the request carries trace context; see
--- a/service/logger/audit/logger.go
+++ b/service/logger/audit/logger.go
@@ -38,8 +38,7 @@ var logLevelNames = map[slog.Leveler]string{
 type Logger struct {
 	logger        *slog.Logger
 	diagnostics   *slog.Logger
-	encoder       Encoder
-	sink          Sink
+	processor     Processor
 	recordTimeout time.Duration
 	configMu      sync.RWMutex
 	config        Config
@@ -48,20 +47,11 @@ type Logger struct {
 // Option configures an audit logger at construction time.
 type Option func(*Logger)
 
-// WithEncoder configures the canonical event encoder.
-func WithEncoder(encoder Encoder) Option {
+// WithProcessor configures canonical event processing.
+func WithProcessor(processor Processor) Option {
 	return func(logger *Logger) {
-		if encoder != nil {
-			logger.encoder = encoder
-		}
-	}
-}
-
-// WithSink configures the emission sink. The default sink writes through slog.
-func WithSink(sink Sink) Option {
-	return func(logger *Logger) {
-		if sink != nil {
-			logger.sink = sink
+		if processor != nil {
+			logger.processor = processor
 		}
 	}
 }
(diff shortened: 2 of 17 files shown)


service/logger/audit/recorder.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (f ProcessorFunc) Process(ctx context.Context, event Event) error {
+	return f(ctx, event)
+}
+
+// Record stamps and validates an event before processing it under a bounded
+// context detached from request cancellation. Callers must not mutate reference
+// data in event until Record returns.
+func (a *Logger) Record(ctx context.Context, event Event) error {
+	a.stampEvent(ctx, &event)
```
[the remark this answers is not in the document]
strantalis: Won't fix: Record accepts Event by value using ordinary Go value semantics; it does not promise or perform a deep clone. The ownership contract explicitly requires callers not to mutate reference data until Record returns and processors not to mutate or retain it. A mutating-processor test would test prohibited behavior, and deep-copying arbitrary map[string]any would restore the serialization/cloning complexity deliberately removed.

service/logger/audit/recorder.go
```diff
@@ -0,0 +1,135 @@
...
+	defer func() {
+		if recovered := recover(); recovered != nil {
+			processErr = fmt.Errorf("%w: panic: %v", ErrProcessing, recovered)
+		}
+	}()
+
+	if a.processor == nil {
+		return a.defaultProcess(ctx, event)
+	}
+	if err := a.processor.Process(ctx, event); err != nil {
```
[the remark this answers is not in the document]
strantalis: Won't fix: synchronous processing is intentional, and Processor explicitly requires implementations to honor context cancellation. Starting an unbounded goroutine would let delivery outlive Record with no safe way to stop it or determine whether a late result was accepted, reintroducing ambiguous delivery and lifetime hazards. The fixed deadline bounds compliant processors and the default handler.

strantalis:
@coderabbitai review

service/logger/logger.go
```diff
@@ -47,13 +48,39 @@ func (c Config) traceCorrelationEnabled() bool {
...
+		options.auditTimeout = timeout
+	}
+}
+
 const (
 	LevelTrace = slog.Level(-8)
 )
 
-func NewLogger(config Config) (*Logger, error) {
+func NewLogger(config Config, options ...Option) (*Logger, error) {
```
dmihalcik-virtru: Mixing a config and functional options seems like an anti-pattern.

strantalis: Yeah we have inconsistency problems where we use config in some places and functional options in others for public facing apis. Not sure there is an easy way to rectify this. Open to thoughts though.

service/logger/audit/logger.go
```diff
@@ -80,9 +109,24 @@ func (a *Logger) ApplyConfig(cfg Config) error {
...
+
+// Processor returns the configured processor, or nil for default OpenTDF processing.
+func (a *Logger) Processor() Processor {
+	return a.processor
+}
+
+// RecordTimeout returns the configured processing budget, or the five-second default.
+func (a *Logger) RecordTimeout() time.Duration {
+	if a.recordTimeout <= 0 {
+		return defaultRecordTimeout
```
dmihalcik-virtru: We don't want a no-timeout option?

docs/Configuring.md
```diff
@@ -109,6 +110,7 @@ logger:
   level: debug
   type: text
   output: stderr
+  audit_timeout: 5s
```
dmihalcik-virtru: Why 5s? IIRC it was 30s in other products we've had

c-r33d (approved):

c-r33d (approved):

PR #3908, opened 2026-08-20
fix(ers): fix direct entitlements silently dropped on token-identifier decisions

elizabethhealy:
### Proposed Changes

Fixes a regression in the access decision flow where **direct entitlements present on JWT claims were silently dropped** when decisions were made via **token identifiers** (token → entity chain shortcut path), and adds regression coverage across unit + BDD tests.

**Changes:**
- Force ERS hydration when resolved token-chain claims include direct-entitlements claim keys so entitlements are preserved on entity representations.
- Add unit tests to verify hydration is triggered and that both `direct_entitlements` and `directEntitlements` claim keys are handled.
- Add BDD scenarios + step support to mint JWTs with direct entitlements and exercise the token-identifier decision path.

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Direct entitlements are now preserved when resolving permissions through token identifiers.
  * Supports both `direct_entitlements` and `directEntitlements` claim formats.
  * Prevents direct entitlements from being omitted during token-chain processing.
* **Tests**
  * Added regression coverage confirming matching actions return `PERMIT` and mismatched actions return `DENY`.

### DIFF
--- a/service/internal/access/v2/just_in_time_pdp.go
+++ b/service/internal/access/v2/just_in_time_pdp.go
@@ -408,6 +408,21 @@ func filterEntityChain(entityChain *entity.EntityChain, skipEnvironmentEntities 
 	return filteredEntities
 }
 
+// directEntitlementClaimKeys mirrors the claim names the claims ERS recognizes in
+// parseDirectEntitlementsFromClaims. Keep the two in sync.
+var directEntitlementClaimKeys = []string{"direct_entitlements", "directEntitlements"}
+
+// directEntitlementClaimKey reports whether the resolved claims carry direct
+// entitlements, and under which claim name.
+func directEntitlementClaimKey(claimsStruct *structpb.Struct) (string, bool) {
+	for _, key := range directEntitlementClaimKeys {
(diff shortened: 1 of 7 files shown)


service/internal/access/v2/just_in_time_pdp.go
```diff
@@ -408,6 +408,21 @@ func filterEntityChain(entityChain *entity.EntityChain, skipEnvironmentEntities
 	return filteredEntities
 }
 
+// directEntitlementClaimKeys mirrors the claim names the claims ERS recognizes in
+// parseDirectEntitlementsFromClaims. Keep the two in sync.
+var directEntitlementClaimKeys = []string{"direct_entitlements", "directEntitlements"}
```
elizabethhealy: not a fan of hardcoding these here, seems like logic that should be in ERS

elizabethhealy: moved to ERS

alkalescent (approved):

marythought (approved):

dmihalcik (approved):

PR #3920, opened 2026-08-25
fix(core): resolve audit IP through trusted proxies

strantalis:
### Proposed Changes

- Record the socket peer by default; accept `X-Forwarded-For` only from configured trusted proxies.
- Resolve chains right to left, including IPv4 and bracketed IPv6 addresses with ports.
- Ignore `X-Real-IP` and `True-Client-IP`; trusted ingress must sanitize XFF rather than merely preserve it.
- Keep the separate single-IP propagation header restricted to in-process calls.
- Ignore blank trusted-proxy entries while rejecting invalid nonempty CIDRs.
- Document AWS append mode and Google's additional frontend-IP hop.

PR 1 of 2 in stack #4071. Base: `main`. Next: #3902. Independent of the buffered-processing changes.

### Validation

- `make fmt` and `git diff --check`
- `go test -race ./service/internal/server/realip ./service/internal/server ./service/logger/audit ./service/logger ./service/pkg/server ./service/pkg/config`
- Service lint with a fresh cache and `--new-from-rev=origin/main`: zero new issues.
- New behavior tests fail against the previous resolver and pass with the changes. Coverage includes ports, malformed addresses, spoofed prefixes, alternate headers, Google-style chains, and environment variables with trailing commas.

Full `make lint` passes Buf validation but fails on existing Go lint issues outside this patch. Full `make test` fails in integration suites requiring local Keycloak/Platform and Docker.

## Summary by CodeRabbit

* **New Features**
  * Added trusted-proxy configuration for HTTP servers, including CIDR ranges and environment-variable support.
  * Client IPs are resolved securely from forwarding headers only when requests originate from configured trusted proxies.
  * Audit records consistently capture the resolved client IP and request ID.
  * Requests through trusted internal transports can propagate a validated client IP.

* **Documentation**
  * Added configuration guidance, YAML examples, and security recommendations for trusted proxies.


docs/Configuring.md
```diff
@@ -171,6 +171,30 @@ server:
...
+| Field            | Description                                                | Default | Environment Variable                |
+| ---------------- | ---------------------------------------------------------- | ------- | ----------------------------------- |
+| `trustedProxies` | Proxy CIDRs allowed to supply client-IP forwarding headers | `[]`    | `OPENTDF_SERVER_HTTP_TRUSTEDPROXIES` |
+
+Set the environment variable to a comma-separated CIDR list without spaces.
+
+When `trustedProxies` is empty, Platform ignores `X-Forwarded-For`, `X-Real-IP`,
+and `True-Client-IP` and records the direct socket peer. When the peer matches a
+trusted CIDR, Platform resolves `X-Forwarded-For` from right to left, removing
+trusted proxy hops. Configure only ingress networks that overwrite or append
```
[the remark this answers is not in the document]
strantalis: Fixed in ffd704a0. Documented trusted-peer header precedence: X-Forwarded-For, then X-Real-IP, then True-Client-IP. Also documented single-address requirements and fallback to the socket peer without trying lower-priority headers when resolution encounters an invalid value.

strantalis:
@coderabbitai review

service/internal/server/realip/realip.go
```diff
@@ -18,45 +19,139 @@ const (
...
+		if prefix.Contains(ip) {
+			return true
+		}
+	}
+	return false
+}
+
+func parseIP(value string) (netip.Addr, bool) {
+	value = strings.TrimSpace(value)
+	ip, err := netip.ParseAddr(value)
```
pflynn-virtru, line 133: `parseIP` must handle the `ip:port` XFF form that Azure AGW and AWS ALB emit — e.g. `203.0.113.10:5555` and `[2001:db8::1]:5555`. Those fail to parse, so the resolver falls back to the LB's own address. `peerIP` below already does this via `ParseAddrPort`. 
Please add both as cases in the resolver table test.

service/internal/server/realip/realip.go
```diff
@@ -18,45 +19,139 @@ const (
...
+			ctx = context.WithValue(ctx, ClientIP{}, clientIP)
 			return next(ctx, req)
 		})
+	})
+}
+
+func newResolver(trustedProxyCIDRs []string) (*resolver, error) {
+	trustedProxies := make([]netip.Prefix, 0, len(trustedProxyCIDRs))
+	for _, cidr := range trustedProxyCIDRs {
+		prefix, err := netip.ParsePrefix(strings.TrimSpace(cidr))
```
pflynn-virtru, line 72: edge case: a blank entry here is a hard startup abort. `OPENTDF_SERVER_HTTP_TRUSTEDPROXIES="10.0.0.0/8,"` decodes to `[]string{"10.0.0.0/8", ""}` (viper splits on comma with no empty-filtering)
 One line fixes it: `if cidr == "" { continue }` after the TrimSpace.

strantalis:
@coderabbitai review

pflynn-virtru (approved):

PR #3921, opened 2026-08-25
fix(cli): stream encrypt and decrypt instead of buffering whole payload

dmihalcik-virtru:
### Proposed Changes

`otdfctl encrypt` and `otdfctl decrypt` held the entire plaintext **and** the entire
ciphertext in memory at once — peak RSS ran ~3.6x the payload, so a large enough file
OOMs on a machine with plenty of disk for it. The SDK already streams; this was purely a
CLI-layer choice. This makes memory bounded by segment size instead of payload size.

Two commits, each building and testing on its own:

**`fix(otdfctl): stream encrypt instead of buffering whole payload`**

* `handlers.EncryptBytes` → `Encrypt(ctx, out io.Writer, in io.Reader, EncryptOptions)`,
  taking an `io.Reader` now that `CreateTDF` no longer requires a seeker (#3865).
* MIME detection reads the head of the payload with `bufio.Peek`, which does not consume,
  rather than inspecting a full in-memory copy.
* Buffering hides an `*os.File`'s `Seeker`, which would push the SDK onto its unknown-size
  path and force ZIP64. The stat size is passed via `WithInputSize` for regular files so
  output stays comparable with the buffered implementation. Piped input is legitimately ZIP64.
* Output goes to a temp sibling renamed into place on success, so a failed run leaves no
  partial `.tdf`.

**`fix(otdfctl): stream decrypt and inspect instead of buffering whole payload`**

* `handlers.DecryptBytes` → `Decrypt(ctx, out, in, DecryptOptions)`; `InspectTDF` takes an
  `io.ReadSeeker`. `decrypt` previously made a *third* whole-payload copy via `fmt.Print`;
  `inspect` read an entire TDF just to reach its manifest.
* `io.Copy` selects `sdk.Reader`'s `WriteTo`, which decrypts one segment at a time. A
  compile-time `var _ io.WriterTo = (*sdk.Reader)(nil)` assertion guards that — without
  `WriteTo`, `io.Copy` silently falls back to `Read`/`ReadAt` and re-buffers the whole
  payload with no test failure to show for it.
* The manifest lives at the end of the archive, so both commands need a seekable input. A
  file argument is opened directly; piped stdin is spooled to a temp file, trading disk for
  the memory the old read used.
* Drops the 10 GB `MaxFileSize` CLI cap — it existed to bound RAM, and the SDK's real limit
  now applies.
* Deletes `pkg/cli/pipe.go`, whose three whole-file readers `inspect` was the last caller
  of. `ReadFromFile` had an unbounded `io.ReadAll` with no cap at all.

`cli.ExitWithError` calls `os.Exit` and skips deferred functions, so temp-file cleanup is
also invoked explicitly on every error path — and in `inspect`, on the success path too,
since every exit there goes through `os.Exit`.

> **Stacked PR.** Based on #3865, which is itself based on #3782. Both must merge before
> this can. Review the top two commits only; the rest of the diff is the parents'.

### Checklist

Unit tests cover the new output-file, pipe-reader, and spool helpers in
`otdfctl/cmd/tdf/tdf_test.go`. Nine cases were added to `otdfctl/e2e/encrypt-decrypt.bats`
covering pipe round-trips, empty-stdin rejection, no-partial-output-on-failure, and a
GNU-`time`-guarded peak-RSS bound on a 1 GiB payload.

> **Note:** `e2e/encrypt-decrypt.bats` carries a pre-existing file-level `skip`
> ("Temporarily disabled [namespaced-subject-mappings]") unrelated to this change, so the
> new cases will not run until that is lifted.

`spec/DSPX-4499.md` is filled in from the ticket. No flags changed, so no `docs/man`
updates were needed.

### Testing Instructions

Measured against a local platform stack, peak RSS is now flat in payload size:

| payload | encrypt | decrypt |
|---|---|---|
| 1 GiB | 73 MiB | 76 MiB |
| 4 GiB | 78 MiB | 74 MiB |
| *1 GiB, before* | *3754 MiB* | *3808 MiB* |

Both round-trips are byte-identical (`cmp` clean). A file input still produces a non-ZIP64
archive; piped input is ZIP64, since the size cannot be known before the first segment
header is written.

To reproduce:

```bash
# round-trip a large file and watch peak RSS
head -c $((1024*1024*1024)) /dev/urandom > big.bin
/usr/bin/time -v otdfctl encrypt big.bin -o big.tdf     # Linux; use -l on macOS
/usr/bin/time -v otdfctl decrypt big.tdf -o big.out
cmp big.bin big.out

# fully piped round-trip
echo "hello world" | otdfctl encrypt | otdfctl decrypt

# inspect from a pipe
otdfctl encrypt big.bin | otdfctl inspect
```

Cross-SDK e2e passed on this branch:
[opentdf/tests run 32893257650](https://github.com/opentdf/tests/actions/runs/32893257650)
— 7 legacy + 68 standard + 88 ABAC tests ran (confirmed not `SKIPPED`) against java@main
and js@main.

Refs DSPX-4499


dmihalcik-virtru:
From [test performance comparison](https://github.com/opentdf/tests/actions/runs/32900286662#summary-97972398263):

image

otdfctl/cmd/tdf/decrypt.go
```diff
@@ -43,55 +43,60 @@ func decryptRun(cmd *cobra.Command, args []string) {
...
-	if len(bytesToDecrypt) == 0 {
-		cli.ExitWithError("Must provide ONE of the following to decrypt: [file argument, stdin input]", errors.New("no input provided"))
+	// cli.ExitWithError calls os.Exit, which skips deferred functions, so both
+	// the spooled input and the partial output have to be discarded first.
+	fail := func(msg string, err error) {
+		closeIn()
+		if outFile != nil {
+			outFile.Cleanup()
+		}
+		cli.ExitWithError(msg, err)
```
jrschumacher, line 80: @dmihalcik-virtru can you move this into the helper? There are other cases where we pipe results through and it seems like this should be reusable.

otdfctl/cmd/tdf/encrypt.go
```diff
@@ -24,6 +26,38 @@ var (
...
+	detected := mimetype.Detect(head).String()
+	if detected == "application/octet-stream" && fileExt != "" {
+		// Lookup returns nil for an extension it does not know, which is not an
+		// error — octet-stream remains the right answer.
+		if byExt := mimetype.Lookup(fileExt); byExt != nil {
+			detected = byExt.String()
+		}
+	}
+	return detected, in, nil
+}
```
jrschumacher, line 59: Shouldn't this be moved to a helper library rather than embedded in the command?

otdfctl/cmd/tdf/encrypt.go
```diff
@@ -77,70 +111,86 @@ func encryptRun(cmd *cobra.Command, args []string) {
...
-	if e != nil {
-		cli.ExitWithError("Failed to write encrypted data to stdout", e)
+	// cli.ExitWithError calls os.Exit, which skips deferred functions, so the
+	// partial output has to be discarded before handing off to it.
+	fail := func(msg string, err error) {
+		if tdfFile != nil {
+			tdfFile.Cleanup()
+		}
+		cli.ExitWithError(msg, err)
+	}
```
jrschumacher, line 173: Same callout as in the decrypt.

otdfctl/cmd/tdf/tdf.go
```diff
@@ -1,31 +1,161 @@
...
+
+// Cleanup discards the temporary file. It is a no-op after a successful
+// Commit, so it is safe to both defer it and call it directly.
+func (o *outputFile) Cleanup() {
+	if o.committed {
+		return
+	}
+	o.f.Close()
+	os.Remove(o.f.Name())
+}
```
jrschumacher, line 161: Can we move these to the `/pkg` directory as helpers? Also why aren't we updating our shared `readPipedStdin` and instead developing yet another helper?

otdfctl/pkg/cli/pipe.go
```diff
@@ -1,45 +0,0 @@
...
-		ExitWithError("Failed to git open file at path: "+filePath, err)
-	}
-	defer fileToEncrypt.Close()
-
-	bytes, err := io.ReadAll(fileToEncrypt)
-	if err != nil {
-		ExitWithError("Failed to read bytes from file at path: "+filePath, err)
-	}
-	return bytes
-}
```
jrschumacher, line 45: Are we considering if external libraries might be using this package?

PR #3922, opened 2026-08-26
fix(otdfctl): stream TDF encryption

dmihalcik-virtru:
### Proposed Changes

* Stream `otdfctl encrypt` input through `sdk/experimental/tdf.Writer` in bounded 2 MiB segments instead of buffering the plaintext and ciphertext in memory.
* Resolve attribute values and effective KAS keys before output, while preserving wrapping-key selection, assertions, MIME metadata, default KAS behavior, and legacy target modes.
* Stage named outputs in a sibling temporary file and atomically rename on success so failures do not leave partial TDFs.
* Add unit coverage for streaming-before-EOF, MIME sniffing without consumption, stdin handling, attribute/KAS resolution, target-mode signatures, and output cleanup; add file/stdin/stdout BATS cases.

### Checklist

### Testing Instructions

Passing checks:

```shell
cd sdk && go test ./... -race
cd sdk && go test -run TestREADMECodeBlocks
# golangci-lint for changed SDK and otdfctl code: 0 issues
```

Repository-wide commands were also attempted. `make test` stops in the unchanged `lib/fixtures` module because `TestTokenManager_InitialLogin` and `TestTokenManager_CustomTokenBuffer` expect a 120s/60s buffer but receive 30s. `make lint` reports existing baseline findings across unchanged modules; changed-line lint for this PR reports 0 issues.

### DIFF
--- a/otdfctl/cmd/tdf/encrypt.go
+++ b/otdfctl/cmd/tdf/encrypt.go
@@ -1,8 +1,11 @@
 package tdf
 
 import (
+	"bufio"
+	"errors"
 	"io"
 	"log/slog"
+	"mime"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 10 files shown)


jakedoublev:
https://github.com/opentdf/platform/pull/4042 this will help give confidence in working streaming

PR #3923, opened 2026-08-27
refactor(sdk): de-duplicate ChunkedWriter onto existing sdk helpers

dmihalcik-virtru:
## Purpose

Intermediate PR in the stack `main ← #3782 ← **this** ← #3865 ← #3921`.

#3782 introduced `ChunkedWriter` with its own private copies of crypto helpers that already
existed in `sdk/tdf.go` (same package) and in `sdk/experimental/tdf/`. This PR collapses that
three-way duplication before #3865 rewrites `CreateTDF` on top of `ChunkedWriter` — so that
rewrite has one implementation to build on instead of two to reconcile.

Duplication was located with `dupl` (AST/token-based, so it sees through the renames
`wrapKeyWithEC` → `chunkedWrapKeyWithEC`). `jscpd` reported only 1.05% *exact* clones; that gap
is itself the signal that this was a paraphrased copy rather than a literal one, and it's why
CI never flagged it (`dupl` is not in `.golangci.yaml`'s enabled linters).

## Commits

1. **`refactor(sdk): reuse tdf.go crypto helpers in ChunkedWriter`**
   Deletes `chunkedEncryptMetadata`, `chunkedWrapKeyWith{EC,KEM,RSA,PublicKey}`, and
   `chunkedCreatePolicyBinding`, routing key access construction through the existing
   `createKeyAccess` / `encryptMetadata` / `createPolicyBinding`. Also hoists the policy binding
   and metadata encryption out of the per-KAS loop, which had been re-encrypting identical
   metadata once per URL in an OR-group.

2. **`refactor(sdk): point experimental/tdf manifest and assertion types at sdk`**
   Replaces the manifest and assertion types in `sdk/experimental/tdf` with aliases onto their
   `sdk` counterparts. Source-compatible — `experimental/tdf` is exported public API with
   likely downstream consumers, so nothing is deleted from its surface. `IntegrityAlgorithm`
   keeps its own defined type because `sdk`'s is `= int` and cannot carry the `String()` method.

3. **`fix(sdk): support legacy hex signatures in ChunkedWriter`** — see below.

## Bug fixes

**EC key access was unusable.** Copied verbatim from `experimental/tdf/key_access.go`,
`ChunkedWriter` emitted key type `"eccWrapped"` (the KAS only accepts `"ec-wrapped"`,
`rewrap.go:713`) and XOR-wrapped the DEK with the HKDF output where the unwrap path expects
AES-GCM. EC-KAS TDFs from `ChunkedWriter` could not be rewrapped. It went unnoticed because
`experimental/tdf/reader.go` is a stub and `chunked_test.go`'s fake KAS is RSA-only. Fixed in
both packages; `TestChunkedECKeyAccess` covers it and fails on the pre-fix code.

**`WithChunkedExcludeVersion` produced unreadable TDFs.** The option omitted `schemaVersion` —
which is exactly how a reader detects a pre-4.3.0 TDF and switches to expecting hex-then-base64
signatures — while both signature sites hardcoded `isLegacyTDF: false`. Every such TDF failed
root signature verification.

The two settings have to travel together, and `useHex` is consumed during `WriteSegment`, long
before a `Finalize` option is seen. So this adds `WithChunkedTargetMode(mode string)` at
construction, mirroring mainline `WithTargetMode` and reusing the package's existing
`isLessThanSemver` / `hexSemverThreshold`. `WithChunkedExcludeVersion` on its own now fails with
`ErrChunkedVersionHexMismatch` rather than emitting a TDF no reader can verify.

Legacy hex readers are still deployed, so the doubly-encoded form remains fully supported —
`TestChunkedLegacyTargetMode` round-trips a `4.2.2`-mode TDF and asserts the root and every
segment signature decode to 64 bytes.

## Testing

- `make lint` — clean apart from the pre-existing `SA1019` at `sdk/kas_client_test.go:188`
- `make test` — passes across `sdk`, `otdfctl`, `service`
- New: `TestChunkedECKeyAccess`, `TestChunkedKAOShape`, `TestChunkedLegacyTargetMode`,
  `TestChunkedCurrentTargetMode`, `TestChunkedExcludeVersionRequiresLegacyMode`,
  `TestChunkedTargetModeInvalid`
- `ChunkedWriter` is only reachable end-to-end through #3865 + #3921, so cross-SDK xtest is run
  against the tip of the stack.

## Follow-ups (not in scope here)

- Enable `dupl` for `sdk/` in `.golangci.yaml` — it would have blocked the original PR.
- `SplitResult.KASPublicKeys` could be `map[string]KASInfo`, deleting the exported
  `KASPublicKey` entirely. That's a public-contract change to the new `KeySplitter` interface
  and belongs in its own PR.
- Invert the relationship fully: mark `sdk/experimental/tdf` deprecated in its godoc pointing at
  the `sdk` equivalents.

### DIFF
--- a/sdk/chunked_options.go
+++ b/sdk/chunked_options.go
@@ -1,6 +1,7 @@
 package sdk
 
 import (
+	"fmt"
 	"io"
 
 	"github.com/opentdf/platform/protocol/go/policy"
@@ -132,9 +133,43 @@ func WithChunkedEncryptedMetadata(metadata string) ChunkedFinalizeOption {
 
 // WithChunkedExcludeVersion omits the schemaVersion field from the
 // produced manifest for compatibility with older readers.
+//
+// A reader treats a missing schemaVersion as "predates 4.3.0" and so
+// expects hex-then-base64 signatures. Those are written during
+// WriteSegment, before this option is seen, so on its own this option
+// makes Finalize fail with [ErrChunkedVersionHexMismatch]. Pass
+// [WithChunkedTargetMode] at construction instead; it sets both.
 func WithChunkedExcludeVersion() ChunkedFinalizeOption {
 	return func(c *ChunkedFinalizeConfig) error {
 		c.excludeVersion = true
+		return nil
+	}
+}
+
+// WithChunkedTargetMode targets a specific TDF spec version, given as
+// a semver string such as "4.2.2".
+//
+// Below 4.3.0 the writer emits the legacy wire format: segment, root,
+// and assertion signatures are hex-encoded before base64 (yielding the
+// doubly-encoded values pre-4.3.0 readers expect), and schemaVersion is
+// omitted from the manifest, which is how those readers detect it. The
+// two travel together -- a manifest carrying one without the other
+// cannot be verified by any reader.
+//
+// An empty mode selects the current format.
+func WithChunkedTargetMode(mode string) ChunkedWriterOption {
+	return func(c *ChunkedWriterConfig) error {
+		if mode == "" {
+			c.useHex = false
+			c.excludeVersion = false
+			return nil
+		}
+		legacy, err := isLessThanSemver(mode, hexSemverThreshold)
+		if err != nil {
+			return fmt.Errorf("target mode %q: %w", mode, err)
+		}
+		c.useHex = legacy
+		c.excludeVersion = legacy
 		return nil
 	}
 }
--- a/sdk/chunked_test.go
+++ b/sdk/chunked_test.go
@@ -4,6 +4,9 @@ import (
 	"archive/zip"
 	"bytes"
 	"context"
+	"crypto/ecdsa"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"io"
@@ -423,7 +426,7 @@ func (k *chunkedFakeKAS) Rewrap(_ context.Context, in *connect.Request[kaspb.Rew
 		policyResult := &kaspb.PolicyRewrapResult{PolicyId: req.GetPolicy().GetId()}
 		for _, kaoReq := range req.GetKeyAccessObjects() {
 			kao := kaoReq.GetKeyAccessObject()
-			if kao.GetKeyType() != "wrapped" {
+			if kao.GetKeyType() != kWrapped {
 				return nil, fmt.Errorf("unsupported key type %q", kao.GetKeyType())
 			}
 			share, err := dec.Decrypt(kao.GetWrappedKey())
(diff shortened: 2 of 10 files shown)


dmihalcik-virtru:
xtest cross-SDK run against the tip of the stack (`DSPX-4499-streaming-codec`, which includes #3782 → #3923 → #3865 → #3921): **all green** — https://github.com/opentdf/tests/actions/runs/33077938349

Per the plan, I grepped the job logs rather than trusting the check mark, since both fixes here are only reachable end-to-end through the stack.

**EC key access** (`eccWrapped` → `ec-wrapped`, XOR → AES-GCM) — `ztdf-ecwrap` ran and passed, not skipped, for all three decrypt SDKs:

```
PASSED test_tdfs.py::test_tdf_roundtrip[...-go@BR-ztdf-ecwrap]
PASSED test_tdfs.py::test_tdf_roundtrip[...-java@main-ztdf-ecwrap]
PASSED test_tdfs.py::test_tdf_roundtrip[...-js@main-ztdf-ecwrap]
```

plus the `test_policytypes.py` or/and/hierarchy cases under `ztdf-ecwrap`. These assert `kao.type == "ec-wrapped"` and round-trip through a real rewrap, so they exercise exactly the path that was broken.

**Legacy hex / doubly-encoded signatures** — `test_tdf_spec_target_422` passed for all three pairs:

```
PASSED test_tdfs.py::test_tdf_spec_target_422[...-go@BR-...]
PASSED test_tdfs.py::test_tdf_spec_target_422[...-java@main-...]
PASSED test_tdfs.py::test_tdf_spec_target_422[...-js@main-...]
```

That test writes a `4.2.2`-mode TDF from the go SDK and reads it back with java/js, which is the deployed-legacy-reader compatibility this PR is meant to preserve.

Worth noting for reviewers: the `looks_like_422` branch inside `test_tdf_roundtrip` did **not** run, because `select_target_version` returns `4.3.0` whenever both SDKs report `hexless` (semver ≥ 0.4.39) — which every current java/js release does. So the manifest-shape assertions on legacy output come from `test_tdf_spec_target_422` and from the new unit test `TestChunkedLegacyTargetMode`, not from the roundtrip matrix.

**16 skips**, none related: 12 DPoP (Keycloak realm not configured for DPoP; js SDK lacks `dpop`; platform 0.25.0 lacks `dpop_nonce_challenge`) and 4 obligations (java SDK gap). Totals: 88 + 68 + 7 passed across the three steps.

`ec_tdf_enabled` is confirmed on — `otdf-local` sets it for key-management KAS instances (`otdf_local/services/kas.py:79`), and the km1/km2 KAS instances started in this run.

ronelliott (approved):

PR #3935, opened 2026-09-01
chore(sdk): add direct tests for createKeyAccess, encryptMetadata and tdfSalt

dmihalcik-virtru:
> **Part 06 of 20** in the DSPX-2604 re-cut. Base branch: `main`.
>

### Proposed Changes

createKeyAccess builds the manifest key access object and picks the
wrapping scheme from the KAS key algorithm. Today the only thing
exercising it is TDFSuite, which asserts that a round trip against its
own fake KAS succeeds -- so it cannot pin which KAO fields are populated,
cannot catch an error path that silently returns a zero value, and says
nothing about encryptMetadata or tdfSalt, which have no direct coverage.

Adds, test-only:

  - the RSA KAO field shape, including the deliberately empty
    ephemeralPublicKey and the schemaVersion the KAS dispatches on
  - empty and malformed KAS public keys are rejected
  - an EC-wrapped DEK unwrapped the way service/kas/access/rewrap.go
    does it, so the envelope is checked to be AES-GCM over the HKDF
    output rather than merely tagged "ec-wrapped"
  - encryptMetadata's envelope shape, key sensitivity, empty input, and
    unusable key
  - tdfSalt determinism and length

The three hybrid KEM schemes already round-trip in tdf_hybrid_test.go
and are not repeated.

Landing this ahead of the DSPX-2604 work that deletes the duplicate
key-wrapping implementation in sdk/experimental/tdf makes that deletion
checkable as a de-duplication rather than a coverage loss.

### Checklist

### Testing Instructions

```
cd sdk && go test ./... -race
```

<details>

| # | PR | Based on |
|---|----|----------|

the KAS wire format.

clears those annotations.

</details>

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for RSA and EC key-access creation.
  * Added validation for invalid or unusable public keys and incorrect decryption keys.
  * Added tests for metadata encryption and decryption, including empty metadata.
  * Added coverage for encrypted metadata structure and EC interoperability.
  * Added deterministic TDF salt generation tests.

### DIFF
--- a/sdk/key_access_test.go
+++ b/sdk/key_access_test.go
@@ -1,0 +1,191 @@
+package sdk
+
+import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/x509"
+	"encoding/json"
+	"encoding/pem"
+	"strings"
+	"testing"
+
+	"github.com/opentdf/platform/lib/ocrypto"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// These cover createKeyAccess and its helpers directly, without going through
+// CreateTDF. TDFSuite exercises the same code but only asserts that a round
+// trip succeeds against its own fake KAS, so it cannot pin the KAO field shape
+// or the error paths, and encryptMetadata and tdfSalt have no direct coverage
+// at all.
+
+const (
+	testKAS1URL = "https://kas1.example.com/"
+
+	// A real RSA-2048 public key. Generating one per test run is slow enough
+	// to notice across this many subtests.
+	testRSAPublicKey = `-----BEGIN PUBLIC KEY-----
+MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAtQ2ZuyT/p32SFmWTj+wQ
+huQwR4IJSzlJ7CqZ4fOXw90rA2joK27dIGiHrtkQHGhS4SK1mvkYyJaREoppMFRc
+AyZWCgixbSdwYJS/KN0hjLIdhtkdBlZDaZN2ayTf2sZjWzOLL2cYzzVsAy9tGL8a
+bMqf91DEHv+l58fPxmbJ/i6YFFQoOEsyWnPhXdiExe6poQDCHJFYYOp6iu5kOPWr
+jKFj9eGXuFR/CJQ/uxTSM+8/7Ejmi8Oa52TQAUhMPH0U1CRFm/NuiFoFissa0jJC
+J3k6syxvf45mPrbtlhcELskXrquDtJOpIMQmEwfuV4j8iLNwVlsR2tAbClJi6UOy
+SQIDAQAB
+-----END PUBLIC KEY-----`
+
+	testMetadata = "test metadata content"
+)
+
+func testSymKey(t *testing.T) []byte {
+	t.Helper()
+	symKey := make([]byte, kKeySize)
+	_, err := rand.Read(symKey)
+	require.NoError(t, err)
+	return symKey
+}
+
+// TestCreateKeyAccessRSA pins the shape of the default (RSA) wrapping path:
+// which manifest fields are populated and which are deliberately left empty.
+func TestCreateKeyAccessRSA(t *testing.T) {
+	symKey := testSymKey(t)
+
+	kao, err := createKeyAccess(
+		KASInfo{URL: testKAS1URL, PublicKey: testRSAPublicKey, KID: "test-kid", Algorithm: "rsa:2048"},
+		symKey, PolicyBinding{Alg: "HS256", Hash: "test-binding"}, "encrypted-metadata", "split-1",
+	)
+	require.NoError(t, err)
+
+	assert.Equal(t, kWrapped, kao.KeyType)
+	assert.Equal(t, testKAS1URL, kao.KasURL)
+	assert.Equal(t, "test-kid", kao.KID)
+	assert.Equal(t, kKasProtocol, kao.Protocol)
+	assert.Equal(t, "split-1", kao.SplitID)
+	assert.Equal(t, keyAccessSchemaVersion, kao.SchemaVersion)
+	assert.Equal(t, "encrypted-metadata", kao.EncryptedMetadata)
+	assert.NotEmpty(t, kao.WrappedKey)
+	assert.Empty(t, kao.EphemeralPublicKey, "RSA wrapping emits no ephemeral key")
+}
+
+func TestCreateKeyAccessRejectsBadPublicKey(t *testing.T) {
+	symKey := testSymKey(t)
+
+	for _, tc := range []struct{ name, pubKey string }{
+		{"empty", ""},
+		{"malformed", "invalid-pem-data"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := createKeyAccess(
+				KASInfo{URL: testKAS1URL, PublicKey: tc.pubKey, Algorithm: "rsa:2048"},
+				symKey, PolicyBinding{}, "", "split-1",
+			)
+			require.Error(t, err)
+		})
+	}
+}
+
+// TestCreateKeyAccessECUnwrap wraps a DEK to an EC KAS key and then unwraps it
+// exactly the way service/kas/access/rewrap.go does for "ec-wrapped".
+// Asserting only on KeyType would not catch the envelope being, say, a raw XOR
+// of the HKDF output instead of AES-GCM.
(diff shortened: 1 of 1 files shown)


sdk/key_access_test.go
```diff
@@ -0,0 +1,218 @@
...
+// CreateTDF. TDFSuite exercises the same code but only asserts that a round
+// trip succeeds against its own fake KAS, so it cannot pin the KAO field shape
+// or the error paths, and encryptMetadata and tdfSalt have no direct coverage
+// at all.
+
+const (
+	testKAS1URL = "https://kas1.example.com/"
+
+	// A real RSA-2048 public key. Generating one per test run is slow enough
+	// to notice across this many subtests.
```
pflynn-virtru, line 28: is this true?  I guess RSA is that slow

pflynn-virtru (approved):

sdk/key_access_test.go
dmihalcik-virtru: While yes, I think these are about tens of ms on most modern CPUs. But I also would rather keep randomness and additional dependencies out of the test fixture.

sujankota (commented):
Reviewed the diff statically against `main` (I did not check out the branch or run the tests). **This looks good to me** — the tests are meaningfully stronger than what they replace, and everything I spot-checked holds up. Comments below are non-blocking follow-ups.

What I verified independently:

- The `tdfSalt()` hex pin `aa17cf44…8197` is exactly `sha256("TDF")`.
- The EC unwrap really does mirror the KAS — see the inline note on `TestCreateKeyAccessECUnwrap`.
- Both bad-key cases genuinely error (`pem.Decode` fails inside `ocrypto.FromPublicPEM`, `lib/ocrypto/asym_encryption.go:73`), and `encryptMetadata([]byte{}, …)` errors via `aes.NewCipher`'s `KeySizeError`.
- No identifier collisions with the existing `sdk` package tests, despite sharing constant names with `experimental/tdf/key_access_test.go`.

One nit on the PR description rather than the code: it says the test covers "tdfSalt determinism and length", but the hex pin is strictly stronger than either. The description undersells it.

sdk/key_access_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	assert.Equal(t, kWrapped, kao.KeyType)
+	assert.Equal(t, testKAS1URL, kao.KasURL)
+	assert.Equal(t, "test-kid", kao.KID)
+	assert.Equal(t, kKasProtocol, kao.Protocol)
+	assert.Equal(t, "split-1", kao.SplitID)
+	assert.Equal(t, keyAccessSchemaVersion, kao.SchemaVersion)
+	assert.Equal(t, "encrypted-metadata", kao.EncryptedMetadata)
+	assert.Equal(t, PolicyBinding{Alg: "HS256", Hash: "test-binding"}, kao.PolicyBinding)
+	assert.NotEmpty(t, kao.WrappedKey)
```
sujankota, line 92: The RSA path is asserted but never round-tripped: this checks a non-empty string was produced, whereas `TestCreateKeyAccessECUnwrap` below proves the KAS can actually recover the DEK. The asymmetry exists because `testRSAPublicKey` is a bare public key with no matching private key.

The fixture to close it already exists in this package — `sdk/tdf_test.go:91` defines `mockRSAPublicKey1` (a CERTIFICATE PEM, which `FromPublicPEM` handles via `getPublicPart`) alongside `mockRSAPrivateKey1` at `sdk/tdf_test.go:108`. Switching to that pair would delete the new 9-line constant *and* let this test decrypt `kao.WrappedKey` back to `symKey`, giving RSA the same strength as EC.

sdk/key_access_test.go
```diff
@@ -0,0 +1,218 @@
...
+func TestCreateKeyAccessRejectsBadPublicKey(t *testing.T) {
+	symKey := testSymKey(t)
+
+	for _, tc := range []struct{ name, pubKey string }{
+		{"empty", ""},
+		{"malformed", "invalid-pem-data"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := createKeyAccess(
+				KASInfo{URL: testKAS1URL, PublicKey: tc.pubKey, Algorithm: "rsa:2048"},
```
sujankota, line 105: Two gaps in this table, the second more interesting than the first:

1. Only the RSA branch is exercised. The EC and KEM branches parse `PublicKey` through their own paths and have no malformed-input case.

2. There's a silent-failure path here that's squarely in this PR's stated scope. `createKeyAccess` (`sdk/tdf.go:679`) dispatches on `kasInfo.Algorithm`, but `generateWrapKeyWithRSA` → `ocrypto.FromPublicPEM` dispatches on the *PEM contents*. Pass an EC PEM with `Algorithm: "rsa:2048"` — or an empty `Algorithm`, plausible if a KAS omits it — and the default branch returns an `ECEncryptor` (`lib/ocrypto/asym_encryption.go:112-117`). The result is a KAO carrying an ECIES envelope with `KeyType` still `"wrapped"` and `EphemeralPublicKey` empty. Encrypt succeeds; rewrap fails much later with nothing pointing back here.

That's exactly the "error path that silently returns a zero value" the description targets. Worth a case pinning the current behaviour, or a bug filed if it should be rejected outright.

sdk/key_access_test.go
```diff
@@ -0,0 +1,218 @@
...
+			)
+			require.Error(t, err)
+		})
+	}
+}
+
+// TestCreateKeyAccessECUnwrap wraps a DEK to an EC KAS key and then unwraps it
+// exactly the way service/kas/access/rewrap.go does for "ec-wrapped".
+// Asserting only on KeyType would not catch the envelope being, say, a raw XOR
+// of the HKDF output instead of AES-GCM.
```
sujankota, line 116: Confirming the claim in this comment, since it's the load-bearing one: `service/kas/access/rewrap.go:688-755` runs the identical `GetECKeySize` → `ECSizeToMode` → `pem.Decode` → `ParsePKIXPublicKey` → `CompressedECPublicKey` sequence, and `service/internal/security/standard_crypto.go:478` calls `ocrypto.NewSaltedECDecryptor(sk, TDFSalt(), nil)` with the same salt. This is faithful, and it's the strongest test in the file.

sdk/key_access_test.go
```diff
@@ -0,0 +1,218 @@
...
+	pub, err := x509.ParsePKIXPublicKey(block.Bytes)
+	require.NoError(t, err)
+	ecPub, ok := pub.(*ecdsa.PublicKey)
+	require.True(t, ok)
+	compressed, err := ocrypto.CompressedECPublicKey(mode, *ecPub)
+	require.NoError(t, err)
+
+	priv, err := ocrypto.ECPrivateKeyFromPem([]byte(kasPrivPEM))
+	require.NoError(t, err)
+	dec, err := ocrypto.NewSaltedECDecryptor(priv, tdfSalt(), nil)
```
sujankota, line 153: Minor: this uses the `tdfSalt()` under test on both sides of the round trip, so a change to the salt would move the test with the code rather than fail it. `TestTdfSalt` pins the value independently, so the coverage is there overall — just noting the two tests are load-bearing together.

sdk/key_access_test.go
```diff
@@ -0,0 +1,218 @@
...
+	require.NotEmpty(t, encMeta.Cipher)
+	require.NotEmpty(t, encMeta.Iv)
+
+	iv, err := ocrypto.Base64Decode([]byte(encMeta.Iv))
+	require.NoError(t, err)
+	require.Len(t, iv, ocrypto.GcmStandardNonceSize)
+
+	ciphertext, err := ocrypto.Base64Decode([]byte(encMeta.Cipher))
+	require.NoError(t, err)
+	require.GreaterOrEqual(t, len(ciphertext), ocrypto.GcmStandardNonceSize)
```
sujankota, line 67: The real floor is nonce + GCM tag (12 + 16 = 28), not the nonce alone — a truncated or tagless envelope would pass this.

sdk/key_access_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	t.Run("errors on an unusable key", func(t *testing.T) {
+		_, err := encryptMetadata([]byte{}, testMetadata)
+		require.Error(t, err)
+	})
+}
+
+func TestTdfSalt(t *testing.T) {
+	assert.Equal(t,
+		"aa17cf44585fe15fd634c27b9512d842b42af1bac6178d92161edb4e2abf8197",
```
sujankota, line 215: Worth a comment noting this is `sha256("TDF")`, so whoever next sees this fail can regenerate the value rather than copy it out of the diff.

PR #3938, opened 2026-09-01
fix(cli): stream encrypt instead of buffering the whole payload

dmihalcik-virtru:
> **Part 09 of 20** in the DSPX-2604 re-cut. Base branch: `dspx-2604-08-streamio`.
>

### Proposed Changes

`otdfctl encrypt` read the entire plaintext into memory, handed the slice to
EncryptBytes, which accumulated the entire ciphertext in a bytes.Buffer, and
only then copied it to the destination. Peak RSS was therefore roughly twice
the payload, and a 10 GB cap existed solely to keep that from taking the
process down.

The payload now streams from the input to the destination. The SDK already
takes an io.ReadSeeker and writes incrementally, so this is a matter of not
getting in its way: the file argument is opened directly and passed through,
and piped stdin is spooled to a temporary file because the SDK seeks to the
end of the payload to size it. Spooling trades disk for the memory a
whole-payload read cost; once CreateTDF accepts a plain io.Reader the spool
goes away.

Handler.EncryptBytes becomes Handler.Encrypt(ctx, out, in, EncryptOptions).
The options struct replaces eight positional parameters, and passing the
command's context through means an interrupted encrypt is now cancellable.

MIME detection no longer needs the payload in memory either. detectMimeType
reads the first megabyte -- which is all mimetype inspects, given SetLimit --
and rewinds, so the whole payload still reaches the encoder.

Two behaviors worth calling out:

Fixes a panic. The extension fallback called mimetype.Lookup(fileExt), but
Lookup takes a MIME type string, not an extension: it parses its argument with
mime.ParseMediaType, which fails on a bare extension, so it returned nil for
every extension and .String() on that nil dereferenced. Any file whose contents
mimetype could not classify and whose name had an extension crashed the CLI.
The stdlib mime.TypeByExtension is the extension lookup, and an unknown
extension now leaves application/octet-stream in place.

Output to a file is atomic. Encryption writes to a temporary sibling of the
destination and renames it into place only on success, so a failed run no
longer leaves a truncated .tdf where a complete one is expected. Output to
stdout cannot offer this; a failure part-way through has already written bytes
downstream. Because cli.ExitWithError calls os.Exit and skips deferred
functions, every exit path discards the temporary output and the spool
explicitly.

### Checklist

### Testing Instructions

```
```

Worth exercising by hand, since the panic this fixes needed a real file:

```
head -c 64 /dev/urandom > junk.bin && otdfctl encrypt junk.bin   # panicked on main
```

<details>

| # | PR | Based on |
|---|----|----------|

the KAS wire format.

clears those annotations.

</details>

## Summary by CodeRabbit

- **New Features**
  - Encryption now streams input and output, enabling more efficient handling of large files with lower memory usage.

- **Bug Fixes**
  - Improved MIME type detection using file content and extension fallbacks.
  - Unknown file extensions now receive a safe default MIME type instead of causing an error.
  - Empty inputs are correctly identified as plain text.
  - Input content is preserved after MIME detection, preventing truncated encrypted files.

### DIFF
--- a/otdfctl/cmd/tdf/encrypt.go
+++ b/otdfctl/cmd/tdf/encrypt.go
@@ -1,8 +1,10 @@
 package tdf
 
 import (
+	"errors"
 	"io"
 	"log/slog"
+	"mime"
 	"os"
 	"path/filepath"
 	"strings"
@@ -11,8 +13,9 @@ import (
 	"github.com/opentdf/platform/lib/ocrypto"
 	"github.com/opentdf/platform/otdfctl/cmd/common"
 	"github.com/opentdf/platform/otdfctl/pkg/cli"
+	"github.com/opentdf/platform/otdfctl/pkg/handlers"
 	"github.com/opentdf/platform/otdfctl/pkg/man"
-	"github.com/opentdf/platform/otdfctl/pkg/utils"
+	"github.com/opentdf/platform/otdfctl/pkg/streamio"
 	"github.com/spf13/cobra"
 )
 
@@ -23,6 +26,39 @@ var (
 	encryptDoc = man.Docs.GetCommand("encrypt", man.WithRun(encryptRun))
 	EncryptCmd = &encryptDoc.Command
 )
+
+// detectMimeType sniffs the payload's type from its head and rewinds, so the
+// whole payload still reaches the encoder.
+//
+// Detection needs only the first megabyte, which is what mimetype is limited to
+// anyway, so this reads a bounded prefix rather than the whole payload.
+func detectMimeType(in io.ReadSeeker, fileExt string) (string, error) {
+	mimetype.SetLimit(Size1MB) // limit to 1MB
+
+	head := make([]byte, Size1MB)
+	// A payload shorter than the sniff window is the common case, not an error.
+	n, err := io.ReadFull(in, head)
+	if err != nil && !errors.Is(err, io.EOF) && !errors.Is(err, io.ErrUnexpectedEOF) {
+		return "", err
+	}
+	if _, err := in.Seek(0, io.SeekStart); err != nil {
+		return "", err
+	}
+
+	// defaults to application/octet-stream if nothing is recognized
+	detected := mimetype.Detect(head[:n]).String()
+	if detected == "application/octet-stream" && fileExt != "" {
+		// mime.TypeByExtension is the extension lookup. mimetype.Lookup takes a
+		// MIME type string, so passing it a bare extension always returned nil
+		// and dereferencing that panicked — which is what happened for any file
+		// whose contents were unrecognizable and whose name had an extension.
+		// An extension with no known type leaves octet-stream in place.
+		if byExt := mime.TypeByExtension("." + fileExt); byExt != "" {
+			detected = byExt
+		}
+	}
+	return detected, nil
+}
 
 func encryptRun(cmd *cobra.Command, args []string) {
 	c := cli.New(cmd, args, cli.WithPrintJSON())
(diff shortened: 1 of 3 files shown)


otdfctl/cmd/tdf/encrypt.go
```diff
@@ -76,71 +115,84 @@ func encryptRun(cmd *cobra.Command, args []string) {
...
 		if err != nil {
+			cleanup()
 			cli.ExitWithError("Failed to write encrypted file "+out, err)
 		}
-		defer tdfFile.Close()
+		defer tdfFile.Cleanup()
 		dest = tdfFile
 	} else {
 		dest = os.Stdout
 	}
```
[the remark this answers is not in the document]
dmihalcik-virtru: no, generally this is acceptable and we want the output to available faster.

otdfctl/cmd/tdf/encrypt_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestDetectMimeTypeRewindsForTheEncoder(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		content string
+		want    string
+	}{
+		{name: "text", content: "hello, world\n", want: "text/plain; charset=utf-8"},
```
jakedoublev, line 19: should we use a test PDF, PNG, or JPEG file here as well, or is there a reason more mime types are already proven to work as expected?

otdfctl/pkg/handlers/tdf.go
```diff
@@ -38,69 +38,105 @@ type TDFInspect struct {
...
+func (h Handler) Encrypt(ctx context.Context, out io.Writer, in io.ReadSeeker, o EncryptOptions) error {
+	switch o.TDFType {
 	// Encrypt the data as a ZTDF
 	case "", tdf.TypeTDF3, tdf.TypeZTDF:
 		opts := []sdk.TDFOption{
-			sdk.WithDataAttributes(attrValues...),
+			sdk.WithDataAttributes(o.Attributes...),
 			sdk.WithKasInformation(sdk.KASInfo{
-				URL: h.platformEndpoint + kasURLPath,
+				URL: h.platformEndpoint + o.KASURLPath,
```
jakedoublev, line 66: @elizabethhealy should we still be passing in this info directly from the CLI?

otdfctl/pkg/handlers/tdf.go
```diff
@@ -38,69 +38,105 @@ type TDFInspect struct {
...
+		return errors.New("unknown TDF type")
 	}
 }
 
+// Deprecated: holds both the plaintext and the finished TDF in memory, so peak
+// use is proportional to the payload. Use Encrypt, which streams from an
+// io.ReadSeeker to an io.Writer and stays bounded by the SDK's segment size.
+//
+// Retained for downstream consumers of this module; nothing in otdfctl calls it.
+func (h Handler) EncryptBytes(
```
jakedoublev: We're good to delete this.

otdfctl/cmd/tdf/encrypt_test.go
dmihalcik-virtru: Mostly just that I don't want to start doing more to test functionality that isn't really added in this PR. This is sufficient to validate that the mime detector is doing *something*

otdfctl/pkg/handlers/tdf.go
elizabethhealy: probably not, i dont see this flag used anywhere and we should probably be letting autoconfigure + the well-known base key config handle this
but we can address that separately from this change, this looks like it keeps the existing behavior properly

jakedoublev (approved):

PR #3940, opened 2026-09-01
feat(sdk): add a chunked segment writer (experimental)

dmihalcik-virtru:
> **Part 11 of 20** in the DSPX-2604 re-cut. Base branch: `dspx-2604-base-11`.
>

### Proposed Changes

Adds `ChunkedWriter`, a TDF creation path that accepts segments in any
order. Callers encrypt and upload each segment independently -- typically
off-thread or in parallel -- then call `Finalize` to get the ZIP closing
bytes. Contrast with `SDK.CreateTDF`, which needs the whole plaintext up
front behind an `io.ReadSeeker`. This is a rewrite of 
https://github.com/opentdf/platform/blob/main/sdk/experimental/tdf/writer.go,
absent several features that are added later (including assertions).

This is the first half of DSPX-2604. It adds the implementation only; the
public face of out-of-order writing stays `sdk/experimental/tdf`, which is
rewired onto this writer in a later PR. Graduating `ChunkedWriter` to
supported API -- dropping the `Experimental:` markers and adding the
`SDK.NewChunkedWriter` method -- is deliberately deferred to the end of the
stack.

What is exported, and why that set

Go forces a compromise here. `sdk/experimental/tdf` is a separate package,
so it cannot reach unexported symbols in `sdk`, and the writer cannot move
under `sdk/internal/` because it needs `Manifest`, `KeyAccess`, `Segment`,
`createKeyAccess` and `calculateSignature`. So: export the minimum bridge
the adapter needs, mark all of it experimental, unexport everything else.

Exported and `Experimental:`-marked: `ChunkedWriter`, `NewChunkedWriter`,
the two result structs, the two config structs and option types, the
caller-facing `WithChunked*` options, the sentinel errors, and the
`KeySplitter` / `Split` / `SplitResult` / `KASPublicKey` group that an
out-of-package splitter has to implement.

Unexported: the test seams. The clock, the segment cipher and its factory,
the archive writer factory, the entropy source, and the four options that
inject them. `archiveWriterFactory` in particular returns a
`zipstream.SegmentWriter` from `internal/`, so no external package could
have implemented it even when the type was exported -- an exported symbol
no caller can satisfy should not be exported. Those three files (`clock.go`,
`segment.go`, `archive_writer.go` -- 28, 29 and 23 lines) are folded into
`chunked_writer.go` now that nothing outside the package can see them.

Correctness guards included rather than deferred

Three small guards ship with the code they protect, since splitting "add
new code with a known hole" from "fix it" across two PRs of brand-new code
is churn with no review value:

- `Finalize` rejects a write set missing segment 0
  (`ErrChunkedMissingSegmentZero`). Only segment 0 emits the payload's ZIP
  local file header and every recorded offset is measured from it, so a set
  without it silently produces a corrupt archive. It cannot be synthesized
  at `Finalize` time -- by then the caller has already encrypted and shipped
  the bytes.
- The default splitter rejects a KAS key whose algorithm has no wrapping
  scheme (`ErrSplitterUnsupportedAlgorithm`) instead of letting the empty
  string reach `createKeyAccess`, where it selects the RSA branch while
  `ocrypto.FromPublicPEM` sniffs the PEM and wraps anyway -- producing a KAO
  that claims keyType "wrapped" with no ephemeral public key, i.e. a TDF
  nothing can decrypt.
- The injection-seam options reject nil rather than storing it. A stored nil
  is indistinguishable from an unset field, so no default is installed and
  the nil surfaces as a panic partway through -- for the key splitter, not
  until `Finalize`.

`WriteSegment` also reserves an index with a negative-size placeholder and
rolls the reservation back if encryption, signing or the archive write
fails, so a failed attempt cannot leave a placeholder that blocks a retry
or that `Finalize` mistakes for a written segment. Release matches on
pointer identity and on the placeholder still being unwritten, so it can
never discard a segment another call has since completed.

Deferred to the next PR: the `GetManifest`-splits-under-RLock question,
which is a behavior change to an exported method rather than a hole in
what lands here.

Integrity algorithms are fixed, not configurable

Rebasing onto #4030 (DSPX-4703) removed `calculateSignature` and
`integrityAlgorithmString`, which this writer called. Rather than port the
call sites onto the new typed helpers and keep the two options, both
`WithChunkedIntegrityAlgorithm` and `WithChunkedSegmentIntegrityAlgorithm`
are gone and the algorithms are hardcoded: `RootHS256` for the root
signature, `SegmentGMAC` for segment hashes. Neither was a meaningful
choice.

- **Root.** #4030's whole point: the aggregate hash is manifest data the
  AEAD never processed, so a "GMAC" root is just a copy of the last segment
  hash, forgeable by an attacker with no key. `WithChunkedIntegrityAlgorithm(GMAC)`
  accepted exactly that, producing a file `validateRootSignature` now
  refuses. `RootIntegrityAlg` has one value and the option had nothing left
  to express.
- **Segments.** HS256 over a segment is only meaningful when the bytes are
  not AEAD output — a plaintext segment has no tag to read out. There is no
  plaintext payload path anywhere in `sdk/`: all three writers hardcode
  `IsEncrypted: true`, the cipher is always AES-256-GCM, and
  `enableEncryption` in `tdf_config.go` is a dead field nothing sets or
  reads. So GMAC always applies, and HS256 would only re-MAC bytes the
  cipher already authenticated.

Hardcoding `SegmentGMAC` also changes this writer's segment default, which
was `HS256`. GMAC is what `CreateTDF` defaults to (`tdf_config.go`) and what
all eight golden TDFs in the cross-SDK corpus declare, so this is the
alignment that lets #3946 rewrite `CreateTDF` on top of this writer without
changing its manifest output.

Side effect worth noting for reviewers of the doc comments: `tdf_config.go`
and `sdk/experimental/tdf/options.go` both claim `SegmentHS256` covers
plaintext segments "which the experimental writer can produce and the stable
one cannot". That is not true today — `experimental/tdf/writer.go` also
hardcodes `IsEncrypted: true` and always encrypts. Left alone here; it is
#4030's text, not this PR's.

Tests cover round-trip, out-of-order and sparse writes, segment trimming,
KAO shape for RSA and EC, legacy and current target modes, assertion
signing, deterministic ZIP timestamps, the error contracts above,
archive-failure rollback and retry, the two fixed integrity algorithms
(including that a segment hash really is the AEAD tag and not an HMAC over
the same bytes), and concurrent `WriteSegment` calls to distinct and to
duplicate indices under `-race`.

### Checklist

### Testing Instructions

```
cd sdk && go test ./... -race
```

`sdk/chunked_test.go` is the bulk of the diff. The concurrency cases
(`WriteSegment` to distinct and to duplicate indices) are the ones that want
`-race` specifically.

<details>

| # | PR | Based on |
|---|----|----------|

the KAS wire format.

clears those annotations.

</details>

## New Features

- Added an experimental SDK for creating encrypted TDFs from independently written segments, including out-of-order and concurrent segment handling.
- Added finalization and manifest retrieval with configurable metadata, assertions, MIME types, retained segments, and target formats.
- Added configurable key splitting and default KAS settings, including support for RSA and EC public keys.
- Added segment-level results with encryption sizes, hashes, and plaintext sizes.
- Added validation for invalid sequencing, duplicate segments, missing required segments, and unsupported configurations.

## Bug Fixes

- Improved retry-safe handling for recoverable write and finalization failures, preventing inconsistent output after unrecoverable archive errors.


sdk/chunked_writer.go
```diff
@@ -0,0 +1,814 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("encrypt metadata for split %s: %w", split.ID, err)
+			}
+			encMeta = m
+		}
+		for _, url := range split.KASURLs {
+			pk, ok := splits.KASPublicKeys[url]
+			if !ok {
+				continue
+			}
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed in #3942

sdk/chunked_options.go
```diff
@@ -0,0 +1,255 @@
...
+// makes Finalize fail with [ErrChunkedVersionHexMismatch]. Pass
+// [WithChunkedTargetMode] at construction instead; it sets both.
+//
+// Experimental: not part of the stable SDK API; may change or be removed.
+func WithChunkedExcludeVersion() ChunkedFinalizeOption {
+	return func(c *ChunkedFinalizeConfig) error {
+		c.excludeVersion = true
+		return nil
+	}
+}
```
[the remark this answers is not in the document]
dmihalcik-virtru: Used in a later PR down the stack

sujankota:
## Review — chunked segment writer

Reviewed at `c6ba59f` across correctness, concurrency, error handling, tests, comments and type design. Findings are ordered by severity; everything below was verified against source, and a handful of candidate issues were dropped after checking (listed at the end) so they don't cost you time.

Context taken as given: this is part 11 of a 20-PR stack, and `GetManifest`-splits-under-RLock is explicitly deferred to #3941. I've only raised that where I found a harm distinct from what the description already covers.

---

### Blocking

**1. `buildChunkedKeyAccessObjects` silently drops an entire split**

[`sdk/chunked_writer.go#L805-L812`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L805-L812)

```go
pk, ok := splits.KASPublicKeys[url]
if !ok {
    continue                                              // silent
}
if pk.PEM == "" {
    return nil, fmt.Errorf(... errKasPubKeyMissing)       // loud
}
```

A `KeySplitter` supplied through `WithChunkedKeySplitter` — the exported, documented multi-KAS path — that returns a split whose only KAS URL is absent from `KASPublicKeys` contributes **zero** KAOs. A typo, a trailing-slash mismatch, a resolver that partially failed. The `len(out) == 0` guard at [L820](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L820) doesn't fire as long as some other split produced one, so `Finalize` returns success.

The reason this is worth blocking on rather than filing: nothing downstream catches it either. The reader builds `knownSplits` from the KAOs *present in the manifest* ([`sdk/tdf.go#L1354`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/tdf.go#L1354)), so the missing split is invisible to the completeness check at [L1399](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/tdf.go#L1399). It XORs one share, reconstructs a wrong DEK, and surfaces at [L1418](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/tdf.go#L1418) as `ErrRootSignatureFailure` — which sends whoever debugs it looking for tampering, not for a missing KAO, and by then the plaintext is gone.

`CreateTDF` hard-errors on exactly this condition ([`sdk/tdf.go#L559-L561`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/tdf.go#L559-L561)). Making `!ok` return the same `errKasPubKeyMissing` as the `PEM == ""` branch three lines below is a three-line diff.

**2. `release()` runs before `CleanupSegment()`**

[`sdk/chunked_writer.go#L538-L553`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L538-L553)

`release()` deletes `w.segments[index]` under `mu` and returns; the archive rollback happens after the lock is dropped. Another goroutine can claim the index in that window, encrypt, write it to the archive successfully (the archive has no record of the first attempt, so no duplicate), and commit — and then the first goroutine's `CleanupSegment(index)` deletes *the winner's* `metadata.Segments[index]` and subtracts its size from `payloadEntry` ([`internal/zipstream/segment_writer.go#L260-L282`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/internal/zipstream/segment_writer.go#L260-L282)). `Finalize` then emits a data descriptor short by that segment, a `TotalCRC32` omitting its CRC, and manifest/central-directory offsets that land early. Every call returned `nil`.

What makes me fairly confident this is a real bug and not just theoretical: with the default writer, `AddSegment` is atomic and `payloadEntry` is only updated after it succeeds ([`internal/zipstream/zip_primitives.go#L78-L96`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/internal/zipstream/zip_primitives.go#L78-L96)), so a failed archive write never leaves anything of *its own* to clean up. That `CleanupSegment` call is therefore either a no-op or actively destructive — there is no case where it does the thing the comment describes. Swapping the two statements fixes it.

---

### Worth fixing in this PR

**3. Reservation placeholders leak into `segmentOrderLocked`** — [L754-L759](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L754-L759), [L680-L683](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L680-L683)

`written` is built from every key in `w.segments`, including the `&Segment{Size: -1}` placeholders from [L514](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L514). `buildManifest` then rejects them with a bare `fmt.Errorf("segment %d not written; cannot finalize")`.

So `GetManifest` — documented at [L146](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L146) as "a snapshot built from currently-written segments" and taking only `RLock`, i.e. presented as *the* safe concurrent read — hard-fails whenever any `WriteSegment` is in flight. That's the one method a caller would legitimately call while writers run.

You clearly saw this for index 0: [L429](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L429) filters `Size < 0` with the comment "a reservation in flight does not count". The same filter is just missing for every other index. Skipping `Size < 0` in `segmentOrderLocked` fixes `GetManifest`, `Finalize`'s opaque error, and the inflated count in the [L764](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L764) message together.

**4. `GetManifest` before `Finalize` is not a preview of what `Finalize` writes** — [L471-L489](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L471-L489)

Every call re-runs the full `buildManifest`: a fresh `uuid.NewString()` policy UUID ([L829](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L829)), a fresh `splitter.Split`, fresh KAO wrapping with fresh ephemeral keys, fresh assertion signatures. Two consecutive `GetManifest` calls differ from each other and neither matches `Finalize`'s output. A caller that registers the preview's policy UUID with a service finds it never appears in the TDF.

This reads as distinct from the lock question deferred to #3941 — it's a correctness property of the returned value, not of the locking. If the intent is "shape preview only", the doc should say so; otherwise the policy wants generating once at construction. (Secondary: a network-backed splitter does a real round trip under `RLock`, blocking all `WriteSegment` calls, and deadlocks if it calls back into the writer.)

**5. The reachable `Finalize` failure is the undocumented one** — [L446-L449](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L446-L449)

`zipstream` `Finalize` runs `centralDir.AddFile(*sw.payloadEntry)` at [`segment_writer.go#L214`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/internal/zipstream/segment_writer.go#L214) and can then fail at L227, L242, L246 or L250 — all before `sw.finalized = true` at L253. `chunkedWriter.Finalize` sets neither `finalized` nor `closeFailed` on that path, so the writer looks retryable, and a retry appends the payload entry to the central directory a second time.

Worth noting the inversion this creates. `ErrChunkedCloseFailed` gets a sentinel, a struct field, three guard branches and four comments — for a state that cannot occur, since `baseWriter.Close` ([`internal/zipstream/writer.go#L200`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/internal/zipstream/writer.go#L200)) returns `nil` unconditionally and the only other implementation is a test seam. Meanwhile the failure that *is* reachable is unfenced and undocumented. At minimum, a clause on the `ErrChunkedCloseFailed` docs noting the default writer's `Close` cannot fail would save the next maintainer reasoning about an impossible state.

**6. `Finalize` doesn't fence in-flight archive writes** — [L446](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L446) vs [L579](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L579)

`WriteSegment` calls the archive without holding `w.mu`; only zipstream's own mutex serializes the two. With `WithChunkedSegments` excluding the in-flight index — so finding 3's check never fires — a segment can land in the archive between `buildManifest` and `archiveWriter.Finalize`, inflating the data descriptor and every offset. The late `WriteSegment` then returns a `nil` error *after* finalization, contrary to the `ErrChunkedAlreadyFinalized` contract, and the caller has no way to know.

`NewChunkedWriter`'s doc spells out the WriteSegment-vs-WriteSegment contract carefully but says nothing about Finalize. Stating "Finalize must happen-after every WriteSegment returns" would at least make the requirement explicit.

**7. Split-set invariants go unchecked at the one chokepoint that sees caller-supplied data** — [L785](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L785)

Beyond finding 1, `buildChunkedKeyAccessObjects` validates only `pk.PEM != ""`:

- **Duplicate or empty `Split.ID`.** Documented as "empty when only one split", unenforced. Two shares sharing an ID means `sid` is omitted from the manifest (`omitempty`, [`manifest.go#L84`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/manifest.go#L84)); on read, the second hits the `foundSplits` `continue` at [`tdf.go#L1355`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/tdf.go#L1355) and is never XOR'd in, while the completeness check still passes.
- **`Algorithm` is never validated** — and this one isn't hypothetical. `keysplit.formatAlgorithm` returns the literal `"unknown"` on failure ([`experimental/tdf/keysplit/attributes.go#L189-L195`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/experimental/tdf/keysplit/attributes.go#L189-L195)) rather than erroring, and that flows into `KASPublicKey.Algorithm`. Adapt `XORSplitter` onto this interface — which is the interface's stated purpose — and you get exactly the failure [`key_splitter.go#L127-L134`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/key_splitter.go#L127-L134) describes: RSA `default:` branch, `FromPublicPEM` sniffs the PEM and wraps the EC key anyway, KAO claims `"type":"wrapped"` with no `ephemeralPublicKey`, undecryptable.

  That comment is the best one in the PR and the guard it justifies sits in `singleKASSplitter` — the one implementation that will never be the problem. The trusted splitter validates; the injected one is trusted completely. It's backwards.
- **`pk.URL` vs the map key.** [L813](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L813) builds the KAO from `pk.toKASInfo()`, so a `KASPublicKey` with a correct map key but an empty `URL` field emits `"url": ""` in every KAO. `singleKASSplitter` guards this for itself ([L119-L125](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/key_splitter.go#L119-L125), with a comment explaining the consequence); the chokepoint doesn't.

A single `SplitResult.Validate()` — every `KASURLs` entry resolves, IDs unique and non-empty when multi-split, `Algorithm` a recognized `ocrypto.KeyType`, `URL`/`PEM` non-empty — called at the top of `buildChunkedKeyAccessObjects` retires finding 1 and all of finding 7 at one site. Exporting it also gives third-party splitter authors a self-check the package currently has no way to offer.

---

### Tests

Credit where it's due first: nine tests assemble the real bytes and push them through `SDK.LoadTDF` against an in-process KAS doing a real rewrap — round-trip, out-of-order, sparse, keep-segments, assertions, both target modes, archive failure, concurrency. `TestChunkedKeepSegmentsRequiresDroppedBytesAppended` and `TestChunkedArchiveFailureKeepsManifestHonest` assert *negative* end-to-end properties rather than manifest shape, and `TestChunkedIntegrityAlgorithmsAreFixed` correctly picks segment 1 to keep the ZIP header out of the hashed bytes. This is well above the usual bar and the description's claims hold up.

The gaps that matter:

- **`ErrChunkedCloseFailed`: zero occurrences** in the test file. The distinction that justifies the whole `closeFailed` flag — that a subsequent call returns it rather than `ErrChunkedAlreadyFinalized` — is unpinned, and the seam to test it already exists.
- **No failing or panicking cipher is ever injected.** `withChunkedCipherFactory` appears only in the nil-rejection table, so both the `EncryptInPlace`-error path (where the defer must `release` but must *not* `CleanupSegment`) and the panic-unwind that [L536-L553](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L536-L553) explicitly claims to cover are untested.
- **Multi-split is untested.** Every test uses `DefaultKeySplitter`; `WithChunkedKeySplitter` appears only in the nil test. Findings 1 and 7 live entirely inside that untested space. A 2-split XOR splitter against two fake KASes, finalized and round-tripped, would be the highest-value test to add.
- **Attributes never reach a test.** `WithChunkedAttributes`, `WithChunkedInitialAttributes` and `WithChunkedDefaultKASForFinalize` have zero occurrences, so every produced TDF carries an empty `dataAttributes`. This also hides a precedence wart at [L633](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L633): `WithChunkedAttributes(nil)` cannot express "clear the writer-level attributes".
- `TestChunkedWriteSegmentCleansUpArchiveOnFailure` proves the retry doesn't error and that `CleanupSegment` fired, but never finalizes or decrypts — so an arithmetically wrong rollback would still pass while offsets silently drift. Adding segment 0, `Finalize` and a `LoadTDF` round-trip is three lines and makes the assertion load-bearing.
- `TestChunkedSegmentsNotStartingAtZero` accepts three outcomes and `t.Logf`+`return`s on two, so it can pass without asserting anything. The behavior is deterministic today and `TestChunkedFinalizeRequiresSegmentZero` already pins it.
- Everything runs on sub-32-byte payloads, so the CRC32 combination and ZIP64 size arithmetic only ever see toy values. One test at a realistic segment size would be cheap insurance.

---

### Non-blocking

**Types**

- `ChunkedWriterConfig` and `ChunkedFinalizeConfig` are exported with entirely unexported fields, no methods and no constructor — an out-of-package option body can only `return nil`. Go lets an exported func type reference an unexported one, so `type ChunkedWriterOption func(*chunkedWriterConfig) error` compiles, keeps every `WithChunked*` value passable across packages, and stops advertising two types with no usable surface. It's the same reasoning you applied to `clock` / `segmentCipher` / `archiveWriterFactory`.
- `ChunkedSegmentResult.TDFData io.Reader` wraps three already-allocated slices, so it saves nothing and costs re-readability — exactly what a retryable parallel upload path wants. There's also a unit trap between siblings: `EncryptedSize` is `len(sealed)`, but `TDFData` for segment 0 additionally carries the ZIP local file header, so the natural `io.CopyN(dst, res.TDFData, res.EncryptedSize)` truncates segment 0 by the header length. `[]byte` plus a `Reader()` method would give both, or at minimum document that the two aren't the same length.
- `Finalize` returns the same `*Manifest` it stores in `w.manifest` while `GetManifest` goes to the trouble of cloning. Clone on both paths or neither.
- `KASPublicKey.Algorithm` could be `ocrypto.KeyType` — free, `ocrypto` is already imported. Doesn't fix `createKeyAccess` (`KASInfo.Algorithm` is also a bare string) but it makes `Algorithm: "unknown"` a visible conversion rather than an assignment.
- `Segment{Size: -1}` as an in-band sentinel on the JSON-tagged public manifest type. No escape path today — all three read sites guard, and `-1` over `0` is right given zero-length segments are legal — but a `segmentSlot{seg Segment; written bool}` costs the same and lets `Size` go back to meaning size.
- `Split` / `SplitResult` / `KASPublicKey` duplicate `keysplit`'s field-for-field, and the two copies' doc comments already disagree on the `Algorithm` vocabulary (`"rsa:2048"` here vs `"rsa"` in [`keysplit/types.go#L36`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/experimental/tdf/keysplit/types.go#L36)). The cycle forcing the duplication rests on a single symbol: `sdk.PolicyAlgorithmToKeyType` is `keysplit`'s only `sdk.` reference ([`attributes.go#L190`](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/experimental/tdf/keysplit/attributes.go#L190)). Moving it to `lib/ocrypto`, where `KeyType` already lives, would let three of these four new types be deleted rather than maintained in parallel. Probably worth costing before the stack ships a second vocabulary.
- Both `slices` and `sort` are imported; `slices.Sort` covers L759.

**Comments**

The comment density is high and mostly earns its keep — `key_splitter.go:127-134` and the `committed` rationale at L531-L536 explain *why* and name the concrete consequence, which is the good kind. Four claims don't hold:

- [L26-L28](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L26-L28) — "no external package could implement it even if the type were exported". Go interfaces are satisfied structurally; `internal/` blocks *naming* `zipstream.SegmentWriter`, not implementing its method set. The conclusion is still right (the factory signature is unspellable from outside, and the option is unexported anyway), but the stated reason isn't the one doing the work.
- [L79-L81](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L79-L81) — "it grows as more segments arrive". Nothing grows; `ExpectedCount` stays 1 forever. It works only because `Finalize` always derives and sets an explicit order first, which makes `IsComplete` take the Order branch.
- `segmentOrderLocked`'s "every written segment" — it returns reservations too (finding 3).
- [L840](https://github.com/opentdf/platform/blob/c6ba59f753aeaca0b5c554fa475cea7bf2883012/sdk/chunked_writer.go#L840) — "shallow-deep copy": it's a shallow copy with three slices cloned; nested pointer contents stay shared.

Two small additions worth their line: `segmentCipher` implementations must produce nonce-prefixed AEAD output with a trailing 16-byte tag (`readAEADTag`'s precondition), since `WriteSegment` runs GMAC over `nonce||ciphertext` unconditionally; and `EncryptInPlace` does *not* encrypt in place, so `WriteSegment` neither retains nor mutates `data` — given the name, saying so is worth more than several of the comments that are there.

---

### Checked and dropped

Noting these so they don't get re-raised: `EncryptInPlace` does not mutate the caller's buffer (`gcm.Seal(nil, …)`) despite the name; sparse and variable-sized segment sets read back correctly, since `Segment` carries no `omitempty` and `ReadAt` walks per-segment sizes rather than `DefaultSegmentSize`; zero-length segments resolve correctly in both first and non-first position; there's no data race in the memory-model sense; no integer overflow at the `uint64`/`int64` conversions; `segmentOrderLocked`'s prefix validation picks the right error branch for reordering, gaps, duplicates and over-long lists; the discarded `CleanupSegment` error and the bare `crc.Write` error returns are unreachable branches rather than silent failures; and the `if first, ok :=` at L697 is dead defensive code, not a silent skip.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,856 @@
...
+		release()
+		if archiveWriteAttempted {
+			// The archive may have partially recorded the write before
+			// failing; CleanupSegment undoes that so a retry starts
+			// from a state indistinguishable from never having been
+			// attempted (see zipstream.SegmentWriter's contract). The
+			// concrete writer's CleanupSegment cannot itself fail; a
+			// custom archiveWriterFactory's failure here is best-effort
+			// and does not change what this call returns.
+			_ = w.archiveWriter.CleanupSegment(index)
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed. `CleanupSegment` now runs before `release()`, so the reservation is held for the whole rollback and a retry cannot slip in between the two. Covered by `TestChunkedWriteSegmentCleansUpArchiveOnFailure`, which was extended from an error assertion into a full round trip: write segment 0, fail segment 1, retry segment 1 with a deliberately different length, write segment 2, `Finalize`, then `LoadTDF` and compare the plaintext. The length change is the point — if the failed attempt's cleanup had not fully unwound the archive's `TotalSize` and payload entry, the retry's offsets would not line up and the round trip would fail rather than merely returning a plausible-looking error.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,856 @@
...
+// buildChunkedKeyAccessObjects wraps each split share to each KAS
+// listed by the splitter.
+func buildChunkedKeyAccessObjects(splits *SplitResult, policyBytes []byte, metadata string) ([]KeyAccess, error) {
+	if splits == nil || len(splits.Splits) == 0 {
+		return nil, errors.New("no splits produced")
+	}
+	base64Policy := ocrypto.Base64Encode(policyBytes)
+
+	var out []KeyAccess
+	for _, split := range splits.Splits {
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed, though wider than the suggested per-split counter. The underlying problem is that anything reaching `buildChunkedKeyAccessObjects` from an injected `WithChunkedKeySplitter` is caller-supplied data that had no contract enforced on it, and "split with no KAS" is one of several ways to produce a TDF that is accepted at creation and fails at decryption with `ErrRootSignatureFailure` — which reads as tampering, by which point the plaintext is gone. So the check moved into `SplitResult.Validate()` and `buildChunkedKeyAccessObjects` now calls it as its first statement. It rejects: no splits; an empty share; shares of unequal length (the reader XORs only as many bytes as each share carries, so a short one leaves part of the DEK unmasked); an empty or duplicated split id when there is more than one split (`sid` is what the reader groups on); a split naming no KAS; a named KAS with no entry in `KASPublicKeys`; a resolved key with no PEM, no `URL`, a `URL` disagreeing with its map key, or an algorithm `ocrypto.ParseKeyType` rejects. `KASPublicKey.Algorithm` is now `ocrypto.KeyType` rather than `string`, so the last of those is much harder to get wrong in the first place. Tests: `TestSplitResultValidate` (16 cases) and `TestDefaultKeySplitterResultValidates` in `sdk/key_splitter_test.go`. The same enforcement is carried into #3946's rewritten `buildKeyAccessObjects`, which otherwise kept only the aggregate check.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,856 @@
...
+		// The archive is terminally finalized internally regardless of
+		// this error, so a retry can only ever hit the same failure.
+		// Mark the writer unusable rather than leaving it looking
+		// retryable.
+		w.closeFailed = true
+		return nil, fmt.Errorf("%w: %w", ErrChunkedCloseFailed, err)
+	}
+
+	w.finalized = true
+	w.manifest = manifest
```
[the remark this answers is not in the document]
dmihalcik-virtru: Fixed — `Finalize` returns `cloneChunkedManifest(manifest)`. `cloneChunkedManifest`'s doc now spells out what the clone does and does not cover: the struct is copied and `Segments`, `KeyAccessObjs` and `Assertions` are cloned, so a caller can append to or overwrite elements freely, but the element clones are shallow, so reaching into a pointer or map inside an element still shares state. Nothing in the package hands out a manifest a caller has reason to mutate that deeply. Covered by `TestChunkedFinalizeManifestIsIndependent`.

dmihalcik-virtru:
Thank you — this is the most useful review I've had on this stack. The "checked and dropped" section in particular saved real time, and findings 1 and 7 turned out to be the same bug seen from two angles, which is how they got fixed.

Everything below landed by amending the commit rather than stacking a fixup, so `dspx-2604-11-chunked-writer` is now `e11da29` and 3941–3949 have been rebased onto it.

---

### Blocking

**1 + 7 — silent split drop / unvalidated split set.** Both fixed, and your suggested shape is what I took: one exported `SplitResult.Validate()`, called as the first statement of `buildChunkedKeyAccessObjects`. It enforces at least one split; non-empty share data; all shares the same length (a short share leaves the tail of the DEK unmasked, since the reader XORs only as many bytes as the share carries); non-empty and unique `ID` when there is more than one split; every `KASURLs` entry present in `KASPublicKeys`; and per key, non-empty `PEM`, non-empty `URL`, `URL` equal to the map key, and an `Algorithm` that `ocrypto.ParseKeyType` accepts.

The missing-entry case now returns the same `errKasPubKeyMissing` as the `PEM == ""` branch, so the two conditions are indistinguishable to a caller, as they should be. `Validate` is exported so third-party splitter authors get the self-check you noted the package couldn't offer; `TestDefaultKeySplitterResultValidates` closes the loop by asserting the shipped splitter satisfies the contract the writer enforces on injected ones. `TestSplitResultValidate` is a 16-case table covering each rule and its error text.

You're right that the guard was backwards — the validation lived in `singleKASSplitter`, the one implementation that can't be wrong. It now runs at the chokepoint for every splitter. I kept the `singleKASSplitter` check as well, because only it can name `ErrSplitterUnsupportedAlgorithm` and the offending `policy.Algorithm`; the generic check sees a `KeyType` string and can't.

**2 — `release()` before `CleanupSegment()`.** Fixed by swapping the two statements. `TestChunkedCleanupRunsBeforeReleasingIndex` pins it by reading `w.segments[index]` from inside a fake `CleanupSegment`, which observes the ordering directly rather than needing a goroutine to lose the race often enough to be reliable. It fails against the old ordering.

On "that `CleanupSegment` call is therefore either a no-op or actively destructive" — agreed for the default writer, and that's exactly why the fix is the swap rather than deleting the call: the archive writer is an injection seam, and a seam whose rollback contract is "we call cleanup, correctly ordered" is easier to hold than one that depends on `AddSegment` being atomic. `TestChunkedWriteSegmentCleansUpArchiveOnFailure` now finalizes and round-trips through `LoadTDF` (your suggestion below), and deliberately retries index 1 with a different-length payload so an arithmetically wrong rollback shows up as a corrupt payload instead of being masked by identical accounting.

---

### Worth fixing in this PR

**3 — reservation placeholders leak into `segmentOrderLocked`.** Fixed, and by removing the sentinel rather than filtering it: `w.segments` is now `map[int]*segmentSlot` where `segmentSlot` is `{seg Segment; written bool}`, and `written` flips only after the archive has accepted the bytes. `segmentOrderLocked` only ever names written slots, whether it derived the order itself or validated a caller-supplied one. That retires the `Segment{Size: -1}` type-design item in the same change, and lets `Size` go back to meaning size. `TestChunkedGetManifestDuringInFlightWrite` pins that `GetManifest` succeeds with a write in flight.

**4 — `GetManifest` before `Finalize` is not a preview of what `Finalize` writes.** Documented as shape-preview only, on the interface:

> Before Finalize it is a preview of the manifest's *shape* — segment count, sizes, hashes, which KASes appear — and nothing more. It is not byte-stable and is not what Finalize will produce: each call re-runs the whole manifest build, minting a fresh policy UUID, re-splitting the DEK (so ephemeral wrapping keys, and therefore every key access object, differ), and re-signing any assertions. Two back-to-back calls with identical options disagree, and so will Finalize. Use it to inspect progress, never to predict or cache the final manifest.

I went with documenting rather than generating the policy once at construction because the policy body is built from `cfg.attributes`, which `WithChunkedAttributes` can still change at `Finalize` — pinning the UUID at construction would freeze an identity for a policy whose contents aren't decided yet. The caller you describe (registering the preview's UUID with a service) is now told not to.

Your secondary point — a network-backed splitter doing a round trip under `RLock` — is the subject of #3941, which moves the split out from under the lock entirely and adds `TestChunkedGetManifestDoesNotBlockWriteSegment`.

**5 — the reachable `Finalize` failure is the undocumented one.** Fixed. There's now an `ErrChunkedFinalizeFailed` sentinel; a failed `archiveWriter.Finalize` marks the writer unusable, so the retry that would append the payload entry to the central directory a second time is refused. And `ErrChunkedCloseFailed`'s doc now carries the clause you asked for — that the default writer's `Close` returns `nil` unconditionally, so the state is unreachable outside a test seam. The table at `chunked_test.go:1388` pins both sentinels, which closes the "zero occurrences of `ErrChunkedCloseFailed`" gap below.

**6 — `Finalize` doesn't fence in-flight archive writes.** Documented on the interface rather than fixed with a lock, and explicitly including the failure mode you identified:

> Finalize must happen-after every WriteSegment call returns. Calling it while a write is still in flight is not an error and does not block: the writer simply finalizes without that segment, since a segment counts only once the archive has accepted its bytes. Sequencing that is the caller's job — wait on the errgroup, close the channel, join the pool — and getting it wrong costs you the segment silently.

Holding `w.mu` across `archiveWriter.Finalize` would serialize finalization against every in-flight write, which is the one thing the design is trying to avoid; making the caller's happens-before obligation explicit is the honest trade. If you think that's too weak I'll take a second look, but I'd rather not add the lock on this PR.

---

### Tests

Every gap you listed is covered:

- **`ErrChunkedCloseFailed`** — `TestChunkedTrailerFailureFencesWriter`, a table over the finalize-fails and close-fails seams, asserting the correct sentinel and that the writer stays fenced afterwards.
- **Failing / panicking cipher** — `TestChunkedCipherFailureReleasesIndex` and `TestChunkedCipherPanicReleasesIndex`, both injected through `withChunkedCipherFactory`. Each asserts the index is free again afterwards, so the `EncryptInPlace`-error path and the panic unwind the defer explicitly claims to cover are now pinned.
- **Multi-split** — `TestChunkedMultiSplitRoundTrip`: a 2-split XOR splitter against two fake KASes, finalized and round-tripped through `LoadTDF`. `TestChunkedFinalizeRejectsUnresolvedKAS` and `TestChunkedKeyAccessRejectsShareWithNoKAS` cover the two ways finding 1 manifested.
- **Attributes** — `TestChunkedAttributes` decodes the base64 policy body and checks the writer-level default, the `Finalize`-level override (replace, not merge), and the `Finalize`-level KAS. On the precedence wart: `WithChunkedAttributes(nil)` still can't clear writer-level attributes, and that's now deliberate and documented — dropping attributes silently loosens the policy on the data, which is the one mistake here that can't be detected after the fact. Two subtests pin that nil and `[]*policy.Value{}` both keep the initial set; the doc points callers at omitting `WithChunkedInitialAttributes` instead.
- **`TestChunkedWriteSegmentCleansUpArchiveOnFailure`** — now finalizes and round-trips, as described above.
- **`TestChunkedSegmentsNotStartingAtZero`** — removed. You're right that it could pass without asserting anything, and `TestChunkedFinalizeRequiresSegmentZero` already pins the deterministic behavior.
- **Realistic sizes** — `TestChunkedRealisticSegmentSizes` runs at production segment sizes so CRC32 combination and the ZIP64 arithmetic see non-toy values.

---

### Types

- **`ChunkedWriterConfig` / `ChunkedFinalizeConfig`** — both unexported now, with the option types still exported (`type ChunkedWriterOption func(*chunkedWriterConfig) error`). Same reasoning as `clock` / `segmentCipher` / `archiveWriterFactory`, as you say.
- **`Finalize` vs `GetManifest` cloning** — clone on both paths.
- **`KASPublicKey.Algorithm`** — now `ocrypto.KeyType`. The conversion in the `experimental/tdf` keysplit adapter is explicit and commented, so `"unknown"` is a visible conversion rather than a silent assignment — and `Validate` rejects it.
- **`Segment{Size: -1}`** — gone, see finding 3.
- **Both `slices` and `sort` imported** — `sort` dropped.
- **`ChunkedSegmentResult.TDFData`** — kept as `io.Reader`, but the unit trap is now documented at both fields: `EncryptedSize`'s doc says it is *not* the number of bytes `TDFData` yields for segment 0, and tells callers to size an upload from `TDFData` itself or from `len(header)+EncryptedSize`. I'd rather make the `[]byte` + `Reader()` shape change when this graduates out of experimental than churn the signature mid-stack; the hazard you named is the part that needed fixing today, and it's called out where a caller will actually hit it. Happy to do the re-type if you'd prefer it now.
- **`Split` / `SplitResult` / `KASPublicKey` duplicating `keysplit`** — not taken here, and I agree it's worth costing. Moving `PolicyAlgorithmToKeyType` to `lib/ocrypto` crosses a module boundary and would touch every consumer of the `keysplit` package, which is more than this PR should carry. Filing it separately rather than letting it ride.

### Comments

All four corrections made:

- The archive-factory seam now says it's the *signature* that closes the seam, not the return type's implementability — a struct with the right method set satisfies `zipstream.SegmentWriter` structurally; what nobody outside can spell is `func(clock) zipstream.SegmentWriter`.
- The `ExpectedCount` comment no longer claims anything grows. It now says `ExpectedCount` stays 1 and is ignored, because `Finalize` always derives and sets an explicit `Order` first, which makes `IsComplete` take the Order branch.
- `segmentOrderLocked`'s doc no longer claims "every written segment" while returning reservations — it can't return them any more.
- "shallow-deep copy" is now stated as what it is: a shallow copy with three slices cloned, nested pointer contents shared.

Both additions are in: the `segmentCipher` interface now states the nonce-prefixed-AEAD-with-trailing-16-byte-tag contract and names the consequence of violating it, and `EncryptInPlace`'s doc says plainly that nothing is encrypted in place, that the name is inherited from `ocrypto.AesGcm`, and that implementations must neither retain nor modify `data`.

---

### Stack

- #3942 (the `!ok` → error fix, which you found independently) is closed and its test folded in here, since `SplitResult.Validate` subsumes it.
- #3941 moves the split out from under the lock, per finding 4's secondary point.
- #3946 rewrites the key-access seam; the `Validate` call and a per-share "names no KAS" guard were carried across so finding 1 doesn't regress at the top of the stack, and the guard also covers `CreateTDF`'s template path, which bypasses `SplitResult` entirely.

sdk/chunked_options.go
```diff
@@ -0,0 +1,253 @@
...
+// default and the nil is dereferenced during writing -- for the
+// splitter, not until Finalize, long after the caller has encrypted
+// every segment.
+
+// The slice-valued options below clone what they are given. Each is retained
+// for the lifetime of the writer or of one Finalize call, and each determines
+// something the caller cannot inspect afterwards -- the written policy, the
+// signed assertions, the segments the manifest names. Keeping the caller's
+// backing array would let a mutation after the call silently change any of
+// them, including loosening the policy on data already encrypted.
```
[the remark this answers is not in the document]
dmihalcik-virtru: Addressed in 546457a8.

The finding is right, and it mattered more than "trivial" suggests: the existing paragraph justified cloning by saying an un-cloned slice would let a caller "loosen the policy on data already encrypted" -- while the clone it was justifying does not actually prevent that. `WithChunkedAttributes` clones `[]*policy.Value`, `buildChunkedPolicy` dereferences each element at `Finalize`, so mutating a retained value rewrites the bound policy regardless. The comment claimed a boundary the code does not hold.

Rather than the suggested wording I stated which half of the hole the clone closes and which it does not, and named the two call sites (`buildChunkedPolicy`, `signAssertions`) so the claim is checkable:

```
// The clone stops at the slice, though, so it only closes the half of that
// hole reachable by reslicing or reassigning an element. What the elements
// point at stays shared: buildChunkedPolicy reads each *policy.Value at
// Finalize, and signAssertions signs with whatever AssertionKey.Key holds.
// Mutating a retained policy value still rewrites the bound policy, which is
// the loosening described above -- cloning the slice does not prevent it.
// Callers must treat everything they hand these options as frozen until
// Finalize returns.
```

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+	// writes still in flight are skipped rather than waited on or
+	// reported. Unlike Finalize it produces nothing durable, so a
+	// snapshot one segment short is a correct snapshot of that instant
+	// -- which is why, unlike Finalize, it does not refuse them.
+	//
+	// It is safe to call concurrently with WriteSegment, but it is not
+	// free: GetManifest and Finalize both hold the writer lock across
+	// KeySplitter.Split, so a splitter that does I/O stalls concurrent
+	// WriteSegment calls at their commit step for its duration.
+	GetManifest(ctx context.Context, opts ...ChunkedFinalizeOption) (*Manifest, error)
```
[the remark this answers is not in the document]
dmihalcik-virtru: Correct on the facts, but already handled one PR downstream -- and the suggested wording would be wrong by the time it landed.

#3941 (`dspx-2604-12-getmanifest-lock`, based on this branch) moves the split out from under the lock entirely. `GetManifest` now snapshots segment metadata under `RLock`, releases it, and calls `buildManifest` -- and therefore `KeySplitter.Split` -- unlocked, precisely because an `RWMutex` bars new readers once a writer is queued, so holding `RLock` across KAS network I/O stalled every `WriteSegment` commit behind it.

So the proposed line "GetManifest holds the lock for reading only" describes a shape that is about to stop existing. The underlying requirement survives the change -- it gets stronger, since two `Split` calls can now genuinely overlap -- and #3941 puts it on the `KeySplitter` interface itself rather than on `GetManifest`:

```
// Implementations must be safe for concurrent use. The chunked writer calls
// Split with its lock released, so a caller that runs GetManifest alongside
// another GetManifest or a Finalize has two Splits in flight on the same
// splitter at once -- and both are documented as safe to do. A splitter
// holding per-call state in a field rather than on the stack corrupts one of
// the two manifests, and the damage is silent: the manifest is well-formed,
// carries splits that do not reconstruct the DEK, and fails only at decrypt.
```

It also rewrites these exact `GetManifest` lines and adds `TestChunkedGetManifestDoesNotBlockWriteSegment`, which deadlocks on the current shape. Patching the text here would conflict with that and then need reverting, so resolving in favor of #3941.

sujankota (commented):
Read the whole diff and traced the concurrency, fencing, rollback and manifest paths against `main` (including #4030's typed integrity helpers and the current `zipstream` internals). **No blocking defects.** The hard parts hold up:

- The reservation/commit split is sound, and the ordering constraints are documented at exactly the sites where a "simplification" would break them. I checked the two that would be silent: `CleanupSegment` on an unwritten index is a genuine no-op in `SegmentMetadata`, so the benign-failure path can't spuriously fence; and `writeLocalFileHeader` writes only into the caller's buffer, so `writeSegmentErr`'s "no path out of WriteSegment leaves the writer half-changed" is true rather than aspirational.
- The lock-nesting warning in the `WriteSegment` defer is accurate — `Finalize` really does hold `w.mu` across `archiveWriter.Finalize`, so hoisting the lock above `CleanupSegment` would close the cycle.
- Segment and root integrity match what the reader recomputes (`tdf.go:1408-1432` decodes each `Segment.Hash` and concatenates, same as `buildManifest`), and `DefaultSegmentSize == DefaultEncryptedSegSize - 28` satisfies the `ErrSegSizeMismatch` gate for AES-GCM's 12+16 framing.
- `Error.Mutated` is the right shape for the retry-vs-fence decision, and `TestSegmentWriterErrorMutated` pins both sides of the boundary for the two sentinels returned from both sides — which is the whole reason the flag can't be inferred. `TestSegmentWriterNonMutatingFinalizeIsRetryable` then proves the promise instead of just asserting the flag, which is the version of that test that's actually worth having.
- `Validate` + `VerifyReconstruction` close a real hole: a bad injected splitter previously surfaced as a root-signature failure at decrypt, reported as tampering, after the plaintext was gone. Constant-time compare and the deliberate refusal to report the mismatch offset are both the right call.
- Failure-path coverage is better than the code it covers — cipher panic, cipher failure, archive failure with retry, cleanup failure, close failure, mutating vs non-mutating finalize, in-flight writes, concurrent duplicate indices under `-race`. Every sentinel has a test that provokes it.

Eight comments inline. Two are worth settling before #3946 lands `CreateTDF` on this writer:

1. **Non-uniform segment sizes** (`chunked_writer.go:1014`) — nothing requires non-final segments to match segment 0's size, and the manifest advertises segment 0's as `segmentSizeDefault`. Go's reader maps offsets from cumulative per-segment sizes so it copes; readers that seek by `index * segmentSizeDefault` won't. `TestChunkedRoundTrip` bakes the non-uniform shape in as the happy path.
2. **`sealed` is a wasted full-segment copy** (`chunked_writer.go:799`) — it rebuilds the exact buffer `ocrypto` already allocated, to read its last 16 bytes. Now that the algorithm is hardcoded `SegmentGMAC`, `ciphertext` gives the identical tag.

The rest are cleanups: a config field no option can set, a policy-JSON divergence from `createPolicyObject` on the empty-attribute case, a hardcoded literal where the package constant exists, one tautological error message, two unreachable unwrapped error returns, and a needless package-level var.

One more that has no single line to hang off: `chunked_test.go` uses `context.Background()` throughout, while `segment_writer_test.go` in this same PR uses `t.Context()`. It matters most in `TestChunkedConcurrentWrites`, `TestChunkedGetManifestDuringInFlightWrite` and `TestChunkedFinalizeRejectsInFlightWrite` — those deliberately park a write inside a blocking cipher, so a writer that wedges currently hangs to the package timeout instead of failing where it wedged.

One note on the doc comments, since there are a lot of them and they're load-bearing: I spot-checked the claims that would be expensive to get wrong — the `expectedSegments: 1` "harmless because Finalize always derives an order" argument, the `EncryptInPlace` "nothing is encrypted in place" claim (`gcm.Seal(nil, ...)`, so correct — `data` really is neither retained nor modified), and the `TDFData` "second read yields zero bytes and no error" claim. All three check out against the code they describe.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+		decoded, err := ocrypto.Base64Decode([]byte(slot.seg.Hash))
+		if err != nil {
+			return nil, totals, fmt.Errorf("decode segment %d hash: %w", idx, err)
+		}
+		aggregate.Write(decoded)
+	}
+
+	if len(order) > 0 {
+		if first, ok := w.segments[order[0]]; ok {
+			encInfo.DefaultEncryptedSegSize = first.seg.EncryptedSize
```
sujankota, line 1003: **Non-uniform segment sizes produce a TDF only the Go reader maps correctly.**

These two fields come from whichever segment sorts first, and nothing anywhere requires the remaining segments (bar the highest index) to match it. `WriteSegment`'s doc doesn't mention it either, and the API positively invites varying sizes — a caller reserving a block of indices per upload part and filling each partially is the documented use case.

The Go reader is fine: `tdf.go:1064-1085` deliberately derives segment extents from cumulative per-segment `Size` rather than the manifest default, precisely so "a writer that emits varying sizes" isn't mapped onto the wrong segments. But `segmentSizeDefault` in the spec is the size of every segment but the last, and a reader that seeks by `index * segmentSizeDefault` lands mid-segment.

`TestChunkedRoundTrip` bakes the non-uniform shape in as the happy path — `"hello, "` (7), `"chunked "` (8), `"world!"` (6) produces a manifest declaring `segmentSizeDefault: 7` over an 8-byte middle segment. Every varying-size test in the file has that shape, so the corpus this writer establishes is the one other SDKs would choke on.

Two ways out, increasing strictness:

- Document on `WriteSegment` that every segment except the highest index must carry equal plaintext length for cross-SDK readability, and make at least one round-trip test uniform so the golden shape is the uniform one.
- Or reject it in `Finalize` — three lines inside the loop that already walks `order`, comparing each non-final `slot.seg.Size` against the first.

Given #3946 rewrites `CreateTDF` on top of this (where uniformity is currently guaranteed by construction) and #15/#17 want a cross-SDK xtest run, the validating form looks like the cheaper one to own. Failing at `Finalize` is also strictly better than failing at someone else's reader: the caller still has the plaintext.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+			cleanupErr = w.archiveWriter.CleanupSegment(index)
+		}
+		release(cleanupErr)
+	}()
+
+	ciphertext, nonce, err := w.block.EncryptInPlace(data)
+	if err != nil {
+		return nil, fmt.Errorf("encrypt segment %d: %w", index, err)
+	}
+	sealed := make([]byte, 0, len(nonce)+len(ciphertext))
```
sujankota: **`sealed` is a full extra copy of every segment, used only for its last 16 bytes.**

This reconstructs precisely the buffer `ocrypto.AesGcm.EncryptInPlace` already allocated — it does `gcm.Seal(nil, nil, data, nil)` and then reslices the result into `nonce, cipherText` (`lib/ocrypto/aes_gcm.go:48`), so `nonce` and `ciphertext` are already adjacent in one backing array that looked exactly like `sealed` a moment ago.

And `sealed` is used for exactly two things: the `segmentIntegrity` call below and `len(sealed)`. Under the now-hardcoded `SegmentGMAC`, `segmentIntegrity` dispatches to `readAEADTag`, which reads the trailing `kGMACPayloadLength` bytes — identical whether handed `sealed` or `ciphertext`, since `sealed` only prefixes 12 nonce bytes and GCM output is always at least the 16-byte tag. So:

```go
sig, err := segmentIntegrity(ciphertext, w.dek, SegmentGMAC, w.useHex)
...
encryptedSize := int64(len(nonce) + len(ciphertext))
```

That drops one full-segment allocation and copy per segment. At the 2 MiB default that's 2 MiB of garbage per segment, times however many goroutines are writing concurrently — on the hot path of the writer whose entire reason for existing is payloads too large to hold at once.

Worth a line of comment that the equivalence rests on the algorithm being GMAC: an `HS256` segment path MACs the whole segment, where whether the nonce is included very much matters, so reinstating that option means reinstating the concatenation. `TestChunkedIntegrityAlgorithmsAreFixed` already pins the tag-vs-HMAC distinction this relies on, so it's the test to watch after the change.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+	defaultKAS *policy.SimpleKasKey
+
+	// encryptedMetadata is opaque metadata AES-GCM-encrypted on each
+	// KAO with the split share.
+	encryptedMetadata string
+
+	// excludeVersion omits the schemaVersion field from the manifest
+	// for compatibility with older readers. Defaults to the writer's
+	// setting; see WithChunkedTargetMode.
+	excludeVersion bool
```
sujankota, line 412: **A knob nothing can turn.** `applyFinalizeOptions` seeds this from `w.excludeVersion` and `buildManifest` reads it, but no `ChunkedFinalizeOption` sets it — and per the comment at `chunked_options.go:187-192`, deliberately so: it has to travel with `useHex`, which is fixed at construction because `WriteSegment` reads it.

So the field is writer state routed through a config struct that can't alter it. `buildManifest` can read `w.excludeVersion` directly. Dropping the field also removes the shape that invites a future option to set it in isolation, which is exactly what that comment is warning against — better enforced by the field not existing than by a comment asking people not to.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+			}
+			out = append(out, kao)
+		}
+	}
+	return out, nil
+}
+
+// buildChunkedPolicy composes the TDF Policy document from attribute
+// values.
+func buildChunkedPolicy(values []*policy.Value) ([]byte, error) {
```
sujankota, line 1152: **Policy JSON diverges from `createPolicyObject` on the empty-attribute case.**

This pre-sizes both slices, so zero attributes emit `"dataAttributes": [], "dissem": []`. The existing `createPolicyObject` (`tdf.go:766`) appends inside the loop and assigns `Dissem` inside it too, so zero attributes emit `null` / `null`. For one or more attributes the two agree.

Self-consistent either way — the policy binding is an HMAC over whatever bytes come out — but it means an attribute-less chunked TDF has different policy bytes, and therefore a different policy binding, than the same input through `CreateTDF`. Worth reconciling or calling out explicitly in the PR body, since #3946's claim is that rewriting `CreateTDF` on this writer doesn't change its manifest output.

Not suggesting reuse of `createPolicyObject` itself: it takes `[]AttributeValueFQN` where this takes `[]*policy.Value`, and the empty-FQN rejection here is a genuine improvement over silently emitting `{"attribute": ""}` — good catch, and the error text explaining *why* an empty FQN fails closed-but-unhelpfully is the right level of detail.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+	// whose empty argument is taken literally, writing a manifest with no MIME
+	// type rather than the default one.
+	if len(cfg.attributes) == 0 && len(w.initialAttributes) > 0 {
+		cfg.attributes = w.initialAttributes
+	}
+	if cfg.defaultKAS == nil && w.initialDefaultKAS != nil {
+		cfg.defaultKAS = w.initialDefaultKAS
+	}
+	if cfg.mimeType == "" {
+		cfg.mimeType = "application/octet-stream"
```
sujankota: `defaultMimeType = "application/octet-stream"` already exists in this package at `tdf.go:35`, and `CreateTDF` uses it for the same fallback. One identifier away.

`TestChunkedEmptyMimeTypeFallsBackToDefault` asserts the literal too — asserting the constant there means the test keeps agreeing with `CreateTDF` if the default ever moves, which is the property #3946 will want.

sdk/key_splitter.go
```diff
@@ -0,0 +1,308 @@
...
+// reader at another.
+func (k KASPublicKey) validate(mapKey string) error {
+	if k.PEM == "" {
+		return fmt.Errorf("chunked: kas:[%s]: %w", mapKey, errKasPubKeyMissing)
+	}
+	if k.URL == "" {
+		return fmt.Errorf("chunked: kas %q has an empty URL field; every key access object built from it would name no endpoint", mapKey)
+	}
+	if k.URL != mapKey {
+		return fmt.Errorf("chunked: kas %q is keyed under %q but carries URL %q; key access objects are built from the URL field", mapKey, mapKey, k.URL)
```
sujankota: `mapKey` is passed twice, so this renders as `kas "https://kas" is keyed under "https://kas" but carries URL "https://other"` — the first clause states the tautology and the actual mismatch has to be inferred from the third. Drop the first argument and the first `%q`:

```go
return fmt.Errorf("chunked: kas keyed under %q carries URL %q; key access objects are built from the URL field", mapKey, k.URL)
```

The check itself is a good one to have — a split naming one KAS while the manifest points a reader at another is not something the reader could diagnose.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+	// the cipher just produced -- segmentIntegrity's precondition.
+	sig, err := segmentIntegrity(sealed, w.dek, SegmentGMAC, w.useHex)
+	if err != nil {
+		return nil, fmt.Errorf("segment %d signature: %w", index, err)
+	}
+	hash := string(ocrypto.Base64Encode([]byte(sig)))
+	encryptedSize := int64(len(sealed))
+
+	crc := crc32.NewIEEE()
+	if _, err := crc.Write(nonce); err != nil {
```
sujankota: These two returns hand back `err` unwrapped, losing the `segment %d` context that every other error in `WriteSegment` carries — and they're unreachable: `hash.Hash.Write` never returns an error, which is why `CreateTDF` just calls `crc32.ChecksumIEEE(cipherData)` (`tdf.go:259`).

Simplest is to match it — `crc := crc32.ChecksumIEEE(...)` over the joined bytes — though that reintroduces a concatenation you're otherwise trying to shed. `crc32.Update(crc32.ChecksumIEEE(nonce), crc32.IEEETable, ciphertext)` keeps both buffers separate and drops both error branches. Either way, two unreachable returns with the wrong error shape are worth removing rather than fixing.

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1208 @@
...
+	// T is the wall-clock time to return from Now.
+	T time.Time
+}
+
+// Now returns the pinned time.
+func (c fixedClock) Now() time.Time { return c.T }
+
+// defaultRand is the production entropy source used by the chunked writer when
+// no other io.Reader is injected.
+var defaultRand io.Reader = rand.Reader
```
sujankota: This package-level var is read once (in `NewChunkedWriter`'s config default) and overridden by nothing — the actual test seam is `withChunkedRand`, which is the right design. A mutable package-level `io.Reader` is a shared-state hazard in exchange for an indirection nobody uses; `rand: rand.Reader` in the config literal says the same thing and can't be reassigned from another file in the package.

sdk/chunked_writer.go
dmihalcik-virtru: Nonuniform segment sizes is a supported feature of TDF

sdk/chunked_writer.go
dmihalcik-virtru: These are not exported so IDK man. This behavior is called out in the field comment, and I'd rather just keep it. I miss 'final'

sdk/chunked_writer.go
dmihalcik-virtru: Looks like createPolicyObject is wrong here, so changing this is out of scope

sdk/chunked_writer.go
```diff
@@ -0,0 +1,1193 @@
...
+			cleanupErr = w.archiveWriter.CleanupSegment(index)
+		}
+		release(cleanupErr)
+	}()
+
+	ciphertext, nonce, err := w.block.EncryptInPlace(data)
+	if err != nil {
+		return nil, fmt.Errorf("encrypt segment %d: %w", index, err)
+	}
+	sig, err := segmentIntegrity(ciphertext, w.dek, SegmentGMAC, w.useHex)
```
sujankota, line 799: Thanks for taking the `sealed` copy out. One follow-up on the same line: the equivalence that makes this correct is invisible here. Passing `ciphertext` where the tag is computed over `nonce||ciphertext` is only the same answer because `SegmentGMAC` dispatches to `readAEADTag`, which reads the trailing 16 bytes. A segment algorithm that MACs the whole segment -- `SegmentHS256`, if that option ever comes back -- would silently MAC without the nonce, and the failure would surface as a root-signature mismatch on someone else's reader.

One comment pins it:

```suggestion
	// SegmentGMAC reads the trailing AEAD tag, so hashing ciphertext alone is
	// equivalent to hashing nonce||ciphertext -- which is why there is no
	// concatenation here. An algorithm that MACs the whole segment (HS256)
	// would need the nonce prepended back.
	sig, err := segmentIntegrity(ciphertext, w.dek, SegmentGMAC, w.useHex)
```

`TestChunkedIntegrityAlgorithmsAreFixed` already guards the direction that matters today, so this is purely for whoever reintroduces the option.

sujankota (approved):

PR #3942, opened 2026-09-01
fix(sdk): reject a chunked split naming a KAS with no resolved public key

dmihalcik-virtru:
> **Part 13 of 20** in the DSPX-2604 re-cut. Base branch: `dspx-2604-12-getmanifest-lock`.
>

### Proposed Changes

`buildChunkedKeyAccessObjects` skipped any KAS URL that a split named but
`SplitResult.KASPublicKeys` had no entry for. That silently drops the only
key access object that would have let that KAS unwrap the share. In an
OR-group the TDF still decrypts, just against a smaller set of KAS servers
than the policy asked for -- and nothing in the output says so. If every
URL on a split is unresolved the share is unrecoverable outright, and the
failure surfaces much later as a generic "no valid key access objects
generated", or not at all if some other split still produced KAOs.

Treats a missing entry the same way an entry with an empty PEM was already
treated: `errKasPubKeyMissing`, naming the split and the URL. A splitter
that cannot resolve a key it referenced has a bug, and creation time is
where it should be reported.

### Checklist

### Testing Instructions

```
cd sdk && go test ./... -race
```

<details>

| # | PR | Based on |
|---|----|----------|

the KAS wire format.

clears those annotations.

</details>

## Summary by CodeRabbit

- **Bug Fixes**
  - Finalization now fails when a configured split references a KAS without a resolved public key.
  - Error messages identify the unresolved KAS URL and indicate that its public key is missing.
  - Prevents incomplete key access object sets from being created silently.


dmihalcik-virtru:
Closing this one: the fix it carried now lives in #3940.

Reviewers (both CodeRabbit and @sujankota) raised the same class of problem against the chunked writer's key-access path, and the answer that came out of that review is broader than the missing-public-key `continue` this PR converted into an error. #3940 now validates the whole `SplitResult` at the single point where caller-supplied split data becomes manifest content (`SplitResult.Validate`, called from `buildChunkedKeyAccessObjects`): at least one split, every split naming at least one KAS, every named KAS resolving to a usable wrapping key with a parseable algorithm and a URL that matches its map key, share lengths that agree, and split ids the reader can actually group on. An unresolved KAS is one case of that, and it still fails with `errKasPubKeyMissing`, so the sentinel and the assertion style are unchanged.

This PR's regression test moved across unmodified as `TestChunkedFinalizeRejectsUnresolvedKAS` in `sdk/chunked_test.go`, joined by a table (`TestSplitResultValidate` in `sdk/key_splitter_test.go`) covering the rest of the contract. The enforcement is also carried forward into #3946's rewritten `buildKeyAccessObjects`, which previously only had the aggregate "no key access objects generated" check.

Nothing is lost by closing this; the stack has been rebased so #3943 follows #3941 directly.

PR #3943, opened 2026-09-01
chore(sdk): alias experimental/tdf manifest and assertion types

dmihalcik-virtru:
> **Part 14 of 20** in the DSPX-2604 re-cut. Base branch: `dspx-2604-13-unresolved-kas`.
>

### Proposed Changes

`sdk/experimental/tdf` carried its own copies of the manifest and assertion
types -- `Manifest`, `Segment`, `KeyAccess`, `Assertion`, `Statement`,
`AssertionKey` and the rest -- structurally identical to the ones in `sdk`
but distinct to the type system, so anything crossing the boundary needed
conversion. Two copies of the JWT signing and verification logic also had
to be kept in step by hand.

Replaces both files' definitions with type aliases. `sdk` owns the
definitions; this package re-exports them. Every exported name and every
method survives: `Assertion.Sign` / `Verify` / `GetHash`,
`Statement.UnmarshalJSON`, `AssertionKey.IsEmpty` / `Algorithm`,
`AssertionVerificationKeys.Get` / `IsEmpty` and the five `String()` methods
all come along with the aliased types, so importers compile unchanged. A
manifest produced here can now be handed to the stable SDK without
conversion, which is what the follow-up delegation needs.

Two deliberate non-aliases:

`Policy`, `PolicyBody` and `PolicyAttribute` stay local. `sdk.PolicyObject`
declares `Body` as an anonymous struct over an unexported element type, so
there is no nameable sdk equivalent to alias to. Exporting those in `sdk`
first would make the alias possible; that is a separate change.

`IntegrityAlgorithm` stays a distinct `int` type. `sdk.IntegrityAlgorithm`
is itself `= int`, so no method can be attached to it, and aliasing would
silently drop `String()` from this package's public API. The underlying
values match, so the two convert freely.

`kSplitKeyType`, `kPolicyBindingAlg`, `kGMACPayloadLength` and
`calculateSignature` are retained verbatim: this package still builds its
own manifests and they have callers in `writer.go` and `key_access.go`.
The change that removes those callers removes these too.

No behavior change. The package's existing tests pass unmodified.

### Checklist

### Testing Instructions

```
cd sdk && go test ./experimental/... -race
```

No behavior change — the point is that the package's existing tests pass
unmodified against aliased types.

<details>

| # | PR | Based on |
|---|----|----------|

the KAS wire format.

clears those annotations.

</details>

## Summary by CodeRabbit

- **Refactor**
  - Standardized experimental TDF assertion and manifest types with the stable SDK equivalents.
  - Removed duplicate local type definitions and assertion-related implementations.
  - Existing TDF data structures and constants now share the stable SDK’s definitions, improving consistency across the SDK.

- **Bug Fixes**
  - Improved chunked manifest generation so concurrent segment writes can proceed while key splitting is in progress.
  - Ensured manifests consistently reflect the correct snapshot of segment metadata.

- **Documentation**
  - Clarified concurrency and data-handling requirements for key splitters.


sdk/experimental/tdf/assertion.go
```diff
@@ -3,422 +3,125 @@
...
-	"github.com/lestrrat-go/jwx/v2/jwa"
-	"github.com/lestrrat-go/jwx/v2/jws"
-	"github.com/lestrrat-go/jwx/v2/jwt"
-	"github.com/opentdf/platform/lib/ocrypto"
+	"github.com/opentdf/platform/sdk"
 )
 
+// The JWT claim keys sdk writes into an assertion binding. Signing moved to
+// sdk with the aliases below, so these have no caller here beyond the tests
+// that assert against sdk's claim names.
```
dmihalcik-virtru:
```suggestion
```

sujankota (approved):

PR #3946, opened 2026-09-01
chore(sdk): rewrite CreateTDF on top of the chunked writer

dmihalcik-virtru:
> **Part 17 of 20** in the DSPX-2604 re-cut. Base branch: `dspx-2604-base-17`.
>

### Proposed Changes

CreateTDFContext and the chunked writer had grown two full implementations of
the same thing: build a policy, split the DEK, wrap each share to its KAS,
encrypt segments, accumulate an aggregate hash, sign assertions, emit a
manifest. Two copies of TDF construction is one too many — every spec change
has to land twice, and the second copy is the one that gets forgotten.

CreateTDF now delegates. It keeps the parts that are genuinely its own — the
KAO template, autoconfigure, the input-size resolution and read loop from the
previous commit — and hands each segment to the chunked writer, which owns
manifest assembly for both paths from here on.

The two paths differ in when key access is resolved, so that is what the
writer is now parameterized on. The chunked writer defers to a KeySplitter at
Finalize, because a caller may still be adding attributes while segments are
in flight. CreateTDF cannot: it knows its attributes up front and wants an
unreachable KAS to fail the call before a single payload byte reaches the
output writer. Both are expressed as a keyAccessResolver, with the DEK now
injectable so CreateTDF can wrap it ahead of time and hand the writer a
staticKeyAccess.

Everything downstream of that split is shared: resolvePolicyAndKeyAccess and
buildKeyAccessObjects replace prepareManifest's inline loop and the chunked
writer's buildChunkedPolicy/buildChunkedKeyAccessObjects, so both paths now
emit byte-identical policy and key access for the same attributes.

Two things fall out of the unification, both moving the chunked writer onto
the shipped classic behavior:

  - With zero attributes the policy body's "dataAttributes" and "dissem" are
    now null rather than []. createPolicyObjectFromFQNs initializes them
    inside the attribute loop; the deleted buildChunkedPolicy did so
    unconditionally. The classic path has always emitted null here.

  - A KAS named by a split but missing a public key is still rejected rather
    than skipped (the check moved into buildKeyAccessObjects), but the error
    now names the missing PEM rather than the absent map entry — the two cases
    were indistinguishable in practice and only the outcome matters.

The writer also gained an explicit segment size. It previously reported the
first segment's actual length as defaultSegmentSize, which is only correct
when every segment is full; a single-segment TDF would advertise a short
default. CreateTDF knows the configured size and now says so.

TDFObject loses aesGcm and payloadKey, which only ever existed to carry state
between prepareManifest and the encrypt loop.

Since #3940 (revised) fixed the chunked writer's root to HS256 and its
segments to GMAC, there is nothing to plumb through here: CreateTDF stops
passing TDFConfig.rootIntegrityAlg and segmentIntegrityAlg to the writer.
Those were already the only values the defaults could hold and no exported
option set either, so the manifest is unchanged.

Deliberately not in this commit: removing enableEncryption, tdfFormat,
readActionName, and the two now-unread integrity fields, which are dead but
unrelated; they are a separate cleanup (#3947).

Testing: the existing TDFSuite round trips pin byte-level output across
segment sizes, target modes, and multi-KAS splits, and pass unchanged. Also
verified against the streaming-input and input-size
coverage added in the previous commit, the experimental chunked writer suite,
and cross-module builds of examples, otdfctl, service, and tests-bdd.

### Checklist

### Testing Instructions

```
cd sdk && go test ./... -race
```

The existing `TDFSuite` round trips pin byte-level output across segment
sizes, target modes and multi-KAS splits, and pass unchanged — that is the
main assurance here.

This touches key access construction on the shipped path, so it wants a
cross-SDK run before merge. This is the branch to pin it to for the whole
writer-delegation half of the stack: xtest drives the Go side through
`otdfctl` → `SDK.CreateTDF`, and this is the first commit where that call
reaches the chunked writer at all.

```
gh workflow run xtest.yml --repo opentdf/tests --ref main \
  -f platform-ref=dspx-2604-17-createtdf-delegates \
  -f otdfctl-ref=dspx-2604-17-createtdf-delegates \
  -f java-ref=main -f js-ref=main
```

`otdfctl-ref` must name the branch, not `main` — otherwise the CLI is built
against main's `sdk/` and the run passes without executing any of this. The
job label should read `go@dspx-2604-17-createtdf-delegates`.

<details>

| # | PR | Based on |
|---|----|----------|

the KAS wire format. Dispatch it against 17 or 20, never 15 on its own: xtest drives
the Go side through `otdfctl` -> `SDK.CreateTDF`, and 17 is the first commit where
that call reaches the rewritten writer. Set `otdfctl-ref` to the same branch as
`platform-ref` -- it defaults to `main`, which builds the CLI against main's `sdk/`
and makes the run vacuous.

clears those annotations.

</details>


sujankota (approved):

sdk/key_splitter.go
```diff
@@ -320,3 +311,197 @@ func (s *singleKASSplitter) Split(_ context.Context, _ []*policy.Value, dek []by
...
+	id string
+
+	// data is the share itself.
+	data []byte
+
+	// kases are the wrapping targets for this share.
+	kases []KASInfo
+}
+
+// staticKeyAccess returns key access objects resolved ahead of time.
```
dmihalcik-virtru:
```suggestion
// staticKeyAccess returns key access objects resolved ahead of time
```

sujankota (approved):

PR #3951, opened 2026-09-01
feat(kas): Read kas registry from kao

c-r33d:
### Summary

  Use the KAS URI from each rewrap KAO to scope registry key lookup and legacy-key listing, supporting multiple KAS registrations behind a shared endpoint.

  - Adds shared KeyOptions and FindKeyWith for URI-scoped lookups.
  - Uses the indexer’s configured default when no URI is supplied.
  - Keeps in-process key lookup behavior unchanged.
  - Adds table-driven tests covering URI selection and legacy filtering.

  No configuration flag or missing-key fallback is included. Custom KeyIndex implementations must implement the new FindKeyWith method.

  ### Testing

  Focused race tests and scoped lint pass. Full validation remains blocked by unavailable local Keycloak and unrelated existing lint issues.

## Summary by CodeRabbit

- **New Features**
  - Key lookup and listing now support a request-specific KAS URI.
  - Key access operations can select keys from the appropriate KAS endpoint when an override is provided.

- **Bug Fixes**
  - Rewrap and decrypt operations now consistently use the KAS URI associated with each key access request.
  - Requests without a specific KAS URI continue to use the configured default.

### DIFF
--- a/service/internal/security/in_process_provider.go
+++ b/service/internal/security/in_process_provider.go
@@ -62,10 +62,6 @@ func (k *KeyDetailsAdapter) ID() trust.KeyIdentifier {
 	return k.id
 }
 
-func (k *KeyDetailsAdapter) KASURI() string {
-	return ""
-}
-
 func (k *KeyDetailsAdapter) Algorithm() ocrypto.KeyType {
 	return k.algorithm
 }
--- a/service/internal/security/in_process_provider.go
+++ b/service/internal/security/in_process_provider.go
@@ -200,6 +200,10 @@ func (a *InProcessProvider) FindKeyByID(_ context.Context, id trust.KeyIdentifie
 	}, nil
 }
 
+func (a *InProcessProvider) FindKeyByIDWithKASURI(ctx context.Context, id trust.KeyIdentifier, _ string) (trust.KeyDetails, error) {
+	return a.FindKeyByID(ctx, id)
+}
+
 // ListKeys lists all available keys
 func (a *InProcessProvider) ListKeys(ctx context.Context) ([]trust.KeyDetails, error) {
 	return a.ListKeysWith(ctx, trust.ListKeyOptions{LegacyOnly: false})
@@ -236,6 +240,10 @@ func (a *InProcessProvider) ListKeysWith(ctx context.Context, opts trust.ListKey
 	}
 
 	return keys, nil
+}
+
+func (a *InProcessProvider) ListKeysWithKASURI(ctx context.Context, opts trust.ListKeyOptions, _ string) ([]trust.KeyDetails, error) {
+	return a.ListKeysWith(ctx, opts)
 }
 
 // Decrypt implements the unified decryption method for both RSA and EC
--- a/service/internal/security/in_process_provider.go
+++ b/service/internal/security/in_process_provider.go
@@ -62,6 +62,10 @@ func (k *KeyDetailsAdapter) ID() trust.KeyIdentifier {
 	return k.id
 }
 
+func (k *KeyDetailsAdapter) KASURI() string {
+	return ""
+}
+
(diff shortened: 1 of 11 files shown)


c-r33d:
@coderabbit review

service/internal/security/in_process_provider.go
```diff
@@ -200,6 +200,10 @@ func (a *InProcessProvider) FindKeyByID(_ context.Context, id trust.KeyIdentifie
 	}, nil
 }
 
+func (a *InProcessProvider) FindKeyWith(ctx context.Context, id trust.KeyIdentifier, _ trust.FindKeyOptions) (trust.KeyDetails, error) {
+	return a.FindKeyByID(ctx, id)
```
dmihalcik-virtru: Why not run FindKeyByID then filter by the Options? I'm guessing we want to not worry about KAS URI here?

service/trust/key_index.go
```diff
@@ -74,6 +84,9 @@ type KeyIndex interface {
 	// FindKeyByID returns a key with the specified ID
 	FindKeyByID(ctx context.Context, id KeyIdentifier) (KeyDetails, error)
 
+	// FindKeyWith returns a key with the specified ID and options.
+	FindKeyWith(ctx context.Context, id KeyIdentifier, opts FindKeyOptions) (KeyDetails, error)
```
dmihalcik-virtru: Why not include ID as part of the Options struct?

dmihalcik-virtru: This will cause breaks downstream, so needs to be broken out as a separate interface, e.g. KeyIndexFindWith, which can then be cast to

dmihalcik-virtru (changes requested):

service/trust/key_index.go
c-r33d: Will it? We don't currently allow customers to override the key indexer, so implementing it downstream would have no impact.

service/internal/security/in_process_provider.go
c-r33d: Yeah, currently the kas url option are really only needed for the platform indexer.

service/trust/key_index.go
dmihalcik-virtru: It looks like there is some dead/test code in DSP that uses it; I'll send a PR to remove it.

dmihalcik-virtru (approved):

service/trust/delegating_key_service.go
```diff
@@ -159,8 +163,8 @@ func (d *DelegatingKeyService) Name() string {
 	return "DelegatingKeyService"
 }
 
-func (d *DelegatingKeyService) Decrypt(ctx context.Context, keyID KeyIdentifier, ciphertext []byte, ephemeralPublicKey []byte) (ocrypto.ProtectedKey, error) {
-	keyDetails, err := d.index.FindKeyByID(ctx, keyID)
+func (d *DelegatingKeyService) Decrypt(ctx context.Context, keyID KeyIdentifier, kasURI string, ciphertext []byte, ephemeralPublicKey []byte) (ocrypto.ProtectedKey, error) {
+	keyDetails, err := d.FindKeyWith(ctx, FindKeyOptions{KeyOptions: KeyOptions{ID: keyID, KASURI: kasURI}})
```
alkalescent, line 167: Should we be concerned about a cache key collision when two KAS registrations use the same kid? `BasicManager.unwrap` caches only by kid, so could decrypt reuse another registration’s private key and fail?

c-r33d: Yes, this is a possibility. We will need to fix it, thanks for the heads up.

c-r33d: Fixed in https://github.com/opentdf/platform/pull/4056

PR #3955, opened 2026-09-01
feat(policy): DSPX-3888 add internal policy object counts

jakedoublev:
## Summary

- add internal count-only database queries for configured policy object scopes
- resolve supported FQN parents without hydrating policy objects
- add targeted indexes for attribute-value count queries


service/policy/db/obligations.go
```diff
@@ -774,23 +773,8 @@ func (c PolicyDBClient) resolveObligationTriggerActionID(ctx context.Context, ac
 }
 
 func (c PolicyDBClient) getAttributeValueNamespaceID(ctx context.Context, attributeValue *common.IdFqnIdentifier) (string, error) {
```
jakedoublev: avoid this smell

PR #3964, opened 2026-09-02
fix(ers): continue failure_strategy bug - limit multi-strategy token chains to one entity per category

elizabethhealy:
### Proposed Changes

- **First-match-wins is now enforced for both failure strategies** — previously, `continue` 
  kept running all matching strategies after a success and accumulated one entity per 
  strategy into the chain. The [ADR is explicit](https://github.com/opentdf/platform/blob/main/adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md#strategy-configuration-properties:~:text=If%20success%3A%20return,return%20error%20immediately) 
  that on success, execution stops regardless of failure strategy; a token's entity chain 
  always holds exactly one entity
- **`continue` still does what it's supposed to** — the [ADR defines `continue`](https://github.com/opentdf/platform/blob/main/adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md#strategy-configuration-properties:~:text=If%20success%3A%20return,return%20error%20immediately) 
  as error handling only: if a strategy *fails*, fall through to the next matching one. 
  The fix only changes what happens on *success*
- **Strategy order, not failure strategy, determines the winner** — whichever strategy is 
  listed first in config and matches the token is the one that resolves it
- **BDD Scenario 3** was an intentionally-failing test asserting PERMIT against the buggy 
  behavior; it now correctly asserts DENY and acts as a regression guard

### Note for reviewers
The customer pattern of "Claims for routing, LDAP for department" across two separate 
strategies is not supported — multi-entity chain composition is [explicitly deferred in 
the ADR as future work](https://github.com/opentdf/platform/blob/main/adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md#entity-chains-handling). 
The workaround is to put the richer strategy first, or merge both sources in a single 
strategy's `output_mapping`.

### Checklist

## Summary by CodeRabbit

* **Behavior Changes**
  * Multi-strategy entity resolution now stops after the first successful strategy.
  * Each resolved token chain contains exactly one entity from the first matching strategy.
  * Strategy order determines the selected entity; failure handling only controls fallback after errors.

* **Bug Fixes**
  * Corrected multi-strategy resolution to prevent later successful strategies from adding entities to an existing chain.

* **Documentation**
  * Updated guides and acceptance scenarios to describe first-match-wins behavior and current strategy limitations.

### DIFF
--- a/service/entityresolution/integration/entity_chain_comparison_test.go
+++ b/service/entityresolution/integration/entity_chain_comparison_test.go
@@ -11,11 +11,14 @@ import (
 	multistrategyv2 "github.com/opentdf/platform/service/entityresolution/multi-strategy/v2"
 	"github.com/opentdf/platform/service/logger"
 	"github.com/opentdf/platform/service/pkg/cache"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"go.opentelemetry.io/otel/trace/noop"
 )
 
-// TestEntityChainComparison demonstrates the discrepancy between Keycloak (2 entities per chain)
-// and Multi-Strategy (1 entity per chain) entity resolution systems
+// TestEntityChainComparison documents the deliberate difference between Keycloak
+// (2 entities per chain: ENVIRONMENT + SUBJECT) and Multi-Strategy (1 entity per chain,
+// from the first matching strategy, per the multi-strategy ERS ADR).
 func TestEntityChainComparison(t *testing.T) {
 	if testing.Short() {
 		t.Skip("Skipping entity chain comparison tests in short mode")
@@ -75,9 +78,9 @@ func TestEntityChainComparison(t *testing.T) {
 	})
 
 	t.Run("MultiStrategy_EntityChainLength", func(t *testing.T) {
-		// Create Multi-Strategy ERS service with MULTIPLE strategies like Keycloak
+		// Configure two strategies that both match the token to show that only the first runs
 		config := types.MultiStrategyConfig{
-			FailureStrategy: types.FailureStrategyContinue, // Continue to try all strategies
+			FailureStrategy: types.FailureStrategyContinue, // Only governs error handling
 			Providers: map[string]types.ProviderConfig{
 				"jwt_claims": {
 					Type:       "claims",
@@ -165,29 +168,12 @@ func TestEntityChainComparison(t *testing.T) {
 			t.Logf("   - Entity %d: %s (Category: %s)", i+1, getEntityIdentifier(ent), ent.GetCategory())
 		}
 
-		// ✅ EXPECTED: Multi-strategy should now create 2+ entities like Keycloak
-		if actualEntityCount >= 2 {
-			t.Logf("✅ SUCCESS: Multi-strategy creates %d entities per chain (like Keycloak!)", actualEntityCount)
-
-			// Validate entity categories are different (ENVIRONMENT + SUBJECT)
-			categoryCounts := make(map[string]int)
-			for _, ent := range chain.GetEntities() {
-				categoryCounts[ent.GetCategory().String()]++
-			}
-
-			t.Logf("   - Entity Categories: %v", categoryCounts)
-
-			// Check we have both ENVIRONMENT and SUBJECT entities like Keycloak
-			if categoryCounts["CATEGORY_ENVIRONMENT"] >= 1 && categoryCounts["CATEGORY_SUBJECT"] >= 1 {
-				t.Logf("✅ PERFECT: Has both ENVIRONMENT and SUBJECT entities like Keycloak")
-			}
-		} else {
-			t.Logf("⚠️  ISSUE: Multi-strategy creates only %d entity per chain", actualEntityCount)
-			t.Logf("🎯 EXPECTED: Multi-strategy should create 2+ entities like Keycloak:")
-			t.Logf("   - Entity 1: CATEGORY_ENVIRONMENT (client from 'azp' claim)")
-			t.Logf("   - Entity 2: CATEGORY_SUBJECT (user from 'sub' claim)")
-			t.Errorf("❌ MISMATCH: Multi-strategy creates %d entities, but Keycloak creates 2 entities per chain", actualEntityCount)
-		}
+		// Both configured strategies match this token, but per the ADR the first match wins,
+		// so the chain holds a single ENVIRONMENT entity. failure_strategy: continue does not
+		// change this — it only decides whether a *failing* strategy falls through to the next.
+		require.Len(t, chain.GetEntities(), 1, "multi-strategy chains carry only the first matching strategy's entity")
+		assert.Equal(t, entity.Entity_CATEGORY_ENVIRONMENT, chain.GetEntities()[0].GetCategory(),
+			"client_environment_strategy is configured first, so it is the match that wins")
 	})
 
 	t.Run("CompareEntityChainStructures", func(t *testing.T) {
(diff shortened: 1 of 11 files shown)


service/entityresolution/multi-strategy/v2/registration.go
```diff
@@ -294,13 +294,10 @@ func (ers *ERSV2) createEntityChainFromSingleTokenV2(ctx context.Context, token
...
-		// This allows creating chains with multiple entities (e.g., ENVIRONMENT + SUBJECT)
-		// Only break if FailureStrategy is FailFast and we have at least one successful entity
-		if failureStrategy == types.FailureStrategyFailFast {
-			break
-		}
-		// With FailureStrategyContinue, we continue to try more strategies to build richer chains
+		// First match wins, per the ADR: failure_strategy only governs error handling
+		// ("continue" tries the next strategy on failure, "fail-fast" stops immediately).
+		// Neither mode keeps resolving after a success, so a chain holds exactly one entity.
+		break
```
pflynn-virtru, line 300: >  can make every decision fail for environment-first configs.

I am verifying now...

pflynn-virtru: tests to confirm https://github.com/opentdf/platform/pull/3980

service/entityresolution/integration/internal/chain_contract_tests.go
```diff
@@ -13,23 +13,50 @@ import (
...
 )
 
-const (
-	// Test constants for entity chain resolution expectations
-	expectedChainEntityCount = 2
-)
+// ChainShape describes the chain an implementation is expected to build for one token.
+// Entity count and categories are implementation-specific: Keycloak always emits an
+// ENVIRONMENT (client) plus a SUBJECT (user) entity, while multi-strategy ERS is
+// first-match-wins per its ADR and emits exactly one entity from the first matching
```
jakedoublev: I think the ADR should have been written to specify "exactly one _subject_ entity". Does this implementation change the keycloak strategy behavior to only emit one client _or_ one subject entity and not both?

service/entityresolution/integration/internal/chain_contract_tests.go
```diff
@@ -13,23 +13,50 @@ import (
...
+
+// ChainContractTestSuite holds implementation-agnostic entity chain validation tests
 type ChainContractTestSuite struct {
 	TestCases []ContractTestCase
 }
 
-// NewChainContractTestSuite creates a test suite focused on implementation-agnostic multi-entity chain validation
+// NewChainContractTestSuite creates a chain contract suite for an implementation that
+// builds Keycloak-shaped two-entity chains.
 func NewChainContractTestSuite() *ChainContractTestSuite {
```
jakedoublev, line 32: Should we remove this in favor of strictly `NewChainContractTestSuiteWithShape`?

service/entityresolution/integration/internal/chain_contract_tests.go
jakedoublev: I think it looks like the Keycloak strategy implementation remains the same per the chain shape, but leaving this here to double-check and not assume.

service/entityresolution/integration/internal/resolved_token_chain_contract.go
```diff
@@ -22,11 +22,37 @@ type ResolvedTokenChainEntityExpectation struct {
 }
 
 // ResolvedTokenChainExpectation describes the final mapped context expected for one token.
+// Entities are listed in mapping-strategy order; because chain resolution is first-match-wins
```
jakedoublev: Should this file be a `_test.go` file?

service/entityresolution/integration/internal/resolved_token_chain_contract.go
```diff
@@ -60,19 +86,24 @@ func (suite *ResolvedTokenChainContractSuite) RunWithAdapter(t *testing.T, adapt
...
 
-	t.Run(adapter.GetScopeName()+"_EnvironmentThenSubjectPreservesMultiEntityMappedContext", func(t *testing.T) {
-		suite.assertResolvedTokenChains(t, implementation, expectations[:1])
+	// Chain resolution is first-match-wins, so the environment strategy (configured first)
+	// is the only one that runs and the resolved chain carries just its mapped context.
+	t.Run(adapter.GetScopeName()+"_EnvironmentThenSubjectPreservesFirstMatchMappedContext", func(t *testing.T) {
+		suite.assertResolvedTokenChains(t, implementation, narrowExpectations(expectations[:1], ResolvedTokenChainExpectation.firstMatch))
 	})
 
+	// Reversing the strategy list makes the subject strategy the first match, which must be
```
jakedoublev: I'm a little confused about the spirit of this test and if the behavior we really want to assert is that we always get either subject or environment entity for a claims strategy? Is the chain strategy like keycloak in that you always get two entities?

service/entityresolution/integration/entity_chain_comparison_test.go
```diff
@@ -165,29 +168,12 @@ func TestEntityChainComparison(t *testing.T) {
...
-			}
-		} else {
-			t.Logf("⚠️  ISSUE: Multi-strategy creates only %d entity per chain", actualEntityCount)
-			t.Logf("🎯 EXPECTED: Multi-strategy should create 2+ entities like Keycloak:")
-			t.Logf("   - Entity 1: CATEGORY_ENVIRONMENT (client from 'azp' claim)")
-			t.Logf("   - Entity 2: CATEGORY_SUBJECT (user from 'sub' claim)")
-			t.Errorf("❌ MISMATCH: Multi-strategy creates %d entities, but Keycloak creates 2 entities per chain", actualEntityCount)
-		}
+		// Both configured strategies match this token, but per the ADR the first match wins,
+		// so the chain holds a single ENVIRONMENT entity. failure_strategy: continue does not
```
jakedoublev: I think the real goal here is to avoid multiple `subject` entities. If our implementation only returns the environment entity, that seems like a bug because the caller of this is Auth Service, which to date has always discarded environment entities when making decisions?

jakedoublev (commented):
had a couple questions

service/entityresolution/integration/internal/chain_contract_tests.go
elizabethhealy: yes that makes more sense, i took the adr at face value, but we should be able to emit a chain with one of each. i think keycloak strategy still worked here but bc of a special case handling. this has been updated and should now be consistent across multi-strat - 1 entity per category type

service/entityresolution/integration/internal/chain_contract_tests.go
elizabethhealy: I deleted the whole ChainShape thing. Since multi-strategy is back to the same [ENVIRONMENT, SUBJECT] shape as Keycloak, there's nothing to parameterize.

service/entityresolution/integration/internal/resolved_token_chain_contract.go
elizabethhealy: Reverted to main, so it's out of this diff

service/entityresolution/integration/internal/resolved_token_chain_contract.go
elizabethhealy: Reverted in latest changes

service/entityresolution/integration/entity_chain_comparison_test.go
elizabethhealy: i added an env entity to the multi-strat feature, so it should resolve both env and subj entities now -- if it only resolves env it should fail at the decision

service/entityresolution/integration/internal/chain_contract_tests.go
```diff
@@ -154,9 +127,9 @@ func NewChainContractTestSuiteWithShape(shape ChainShape) *ChainContractTestSuit
 					ChainValidation: []EntityChainValidationRule{
 						{
 							EphemeralID:               "consistency-token",
-							EntityCount:               expectedChainEntityCount,
-							EntityTypes:               []string{}, // Implementation-specific entity types allowed
-							EntityCategories:          expectedChainCategories,
+							EntityCount:               expectedChainEntityCount, // Consistent entity count across implementations
+							EntityTypes:               []string{},               // Implementation-specific entity types allowed
+							EntityCategories:          []string{"CATEGORY_ENVIRONMENT", "CATEGORY_SUBJECT"},
```
jakedoublev: Nit: this could be DRY

tests-bdd/features/multi-strategy-ers-multi-success.feature
```diff
@@ -1,14 +1,15 @@
...
   See: adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md
 
-  Both scenarios use two strategies whose conditions match, so the chain must hold
-  exactly one entity from "claims_identity". The first asserts the winner's claims
-  reach the decision; the second asserts the loser's claims do not.
+  All three strategies below match every token, so the chain holds the environment entity
+  from "client_environment" plus one subject entity from "claims_identity" — "ldap_department"
+  is skipped. Authorization discards environment entities, so only the subject's claims count.
 
   This covers Jake's gap analysis row #4.
```
jakedoublev: I think this should be removed. :)

jakedoublev (dismissed):

jakedoublev (approved):

PR #3967, opened 2026-09-03
fix(sdk): DSPX-4590 zip64 conformance and per-segment size defaults

dmihalcik-virtru:
Jira: https://virtru.atlassian.net/browse/DSPX-4590

go-sdk's zip layer disagrees with the other SDKs in a handful of places, and its
manifest reader treats optional per-segment sizes as mandatory. This addresses
all seven findings from the DSPX-4590 investigation.

## Finding 1 (interop): writer switched to ZIP64 at 4 GiB instead of 2 GiB

`Finalize` compared against `^uint32(0)`, so a payload between 2 GiB and 4 GiB
was written as a zip32 archive with a value in the top half of the unsigned
32-bit range. java-sdk widens those central-directory fields *signed*, so
deployed Java clients read the size/offset back as a negative number and cannot
open the container. web-sdk always writes ZIP64, java-sdk (since java-sdk#393)
switches at `Integer.MAX_VALUE`.

- New `maxNonZip64Value = math.MaxInt32` in `zip_primitives.go`, mirroring
  java-sdk's `MAX_NON_ZIP64_VALUE`.
- The rule is applied to the uncompressed size, the compressed size **and**
  `entry.Offset` (the local-header offset), which was previously not checked at
  all — an archive under 2 GiB of payload could still place a later entry's
  header past the boundary.
- The threshold is injectable: `Config.MaxNonZip64Value` plus a
  `WithMaxNonZip64Value` option (clamped to `(0, maxNonZip64Value]`, so it can
  only ever make the writer *more* eager to use ZIP64). This lets the tests
  drive the ZIP64 path with a 1 KiB threshold instead of allocating gigabytes.

## Finding 2: ZIP64 extra field parsed positionally

The reader assumed the ZIP64 extended-information field was the first entry in
the extra-field area and that all three values were always present. Per APPNOTE
4.5.3 the values appear in the order *original size, compressed size, local
header offset*, and each is present **only** when its central-directory
counterpart holds the `0xFFFFFFFF` sentinel. A container whose extra area leads
with, say, an extended-timestamp field (tag `0x5455`) was misparsed.

`parseZip64ExtraField` now walks the whole extra area, skips foreign tags,
reads values in spec order gated on the sentinels, and rejects a field that
claims to run past the end of the area.

## Finding 3: ZIP64 detected from the CD offset alone

`NewReader` only looked at `CentralDirectoryOffset == 0xFFFFFFFF`. An archive
that overflows the entry count (`0xFFFF`) or the central-directory size but not
the offset was read as zip32. `eocdNeedsZip64` now checks all three EOCD
sentinel fields.

## Finding 4: per-entry ZIP64 extra field ignored without a ZIP64 EOCD

A writer may put a ZIP64 extra field on an individual entry while leaving the
EOCD in zip32 form. The reader now consults the extra field whenever the
central-directory header carries a sentinel, independent of the EOCD form.

## Finding 5: central-directory cursor could wrap at uint16

`nextCD` was advanced with uint16 arithmetic and did not include the file
comment. A 65000-byte filename plus a 600-byte extra field wraps to 110 and
the reader walks into the middle of a header. The advance is now done in
uint64 and includes `FileCommentLength`.

## Finding 6: silent truncation when a value does not fit 32 bits

The zip32 paths narrowed with a bare cast. `checkFitsInCentralDirectory` now
returns a new `ErrFieldOverflow` for the compressed size, uncompressed size,
local-header offset, central-directory size/offset and entry count instead of
writing a corrupt archive. (With finding 1 in place this is unreachable in
normal operation; it is a backstop against future callers.)

## Finding 7 (correctness): per-segment sizes are optional, and go-sdk read them as zero

`manifest.schema.json` marks `segmentSizeDefault` and
`encryptedSegmentSizeDefault` required on `integrityInformation` but declares no
required list on `segments/items` — a writer may omit a per-segment size
whenever it equals the default. web-sdk does exactly that for every full-sized
segment, so **every web-sdk container larger than one segment (1 MiB) failed to
decrypt in go-sdk**: the missing keys unmarshalled to 0.

`IntegrityInformation.resolveSegmentSizes` now substitutes the manifest-level
defaults, and is used in all three places that consumed the raw values
independently:

- the `payloadSize` accumulation in `NewReader`/`LoadTDF` (a wrong total
  truncates `Seek` and the `ReadAt` bounds check),
- the `WriteTo` loop,
- the `ReadAt` loop, including the skip-ahead arithmetic.

A segment that resolves to zero or negative is now rejected with a new
`ErrSegSizeUnresolved`. Previously the `len(readBuf) != seg.EncryptedSize`
guard passed vacuously for a zero-length segment and the read failed several
frames later inside the GMAC calculation, with a message that said nothing
about the manifest. `ReadAt` also guards `segmentSizeDefault <= 0` before using
it as a divisor.

## Tests

- `sdk/internal/zipstream/zip64_conformance_test.go` — hand-assembles raw zip
  archives (`buildRawZip`) so the reader can be pointed at containers no Go
  writer would produce: extra field not first, differing compressed and
  uncompressed sizes so the APPNOTE 4.5.3 ordering is actually asserted (a
  fixture with equal sizes passes either way), per-entry ZIP64 under a zip32
  EOCD, a central-directory file comment, the 65646-byte name+extra case that
  wraps to 110 at uint16, ZIP64 implied by the entry count, and a malformed
  extra field. Writer side: `TestWriterSwitchesToZip64AtInjectedThreshold` uses
  the injected 1 KiB threshold, `TestEntryNeedsZip64AtTwoGiB` pins
  `maxNonZip64Value == math.MaxInt32` and covers size / compressed size /
  offset, `TestCentralDirectoryNarrowingGuard` covers finding 6.
- `sdk/internal/zipstream/fuzz_test.go` — two new seeds for the `nextCD`
  overflow and the file-comment case.
- `sdk/tdf_segment_defaults_test.go` — round-trips a 2 MiB payload through a
  manifest rewritten to omit the per-segment sizes (the root signature covers
  the segment hashes, not the JSON encoding, so the rewritten container is
  still internally consistent), asserting both the `WriteTo` and the `ReadAt`
  path plus `payloadSize`. Verified by fault injection: with the fallback
  removed the test fails with `payloadSize` 4242 instead of 2101394 and with
  `[tamper detected] tdf: fail to read payload from tdf`, matching the symptoms
  in the ticket. `TestResolveSegmentSizes` covers the resolution rules directly.

## Follow-up required in opentdf/tests (NOT covered by this PR)

> The xtest cell `test_tdfs.py::test_chunky_roundtrip` currently **SKIPS** for
> go, because `xtest/sdk/go/cli.sh` answers no to `supports chunky`. That shim
> lives in the `opentdf/tests` repo, so merging this PR does **not** flip it —
> the go column will stay skipped and the interop regression will stay
> invisible in CI. When this fix ships in a release, someone needs to
> version-gate the `chunky)` case in `xtest/sdk/go/cli.sh` so it reports
> support at or above that version.

### DIFF
--- a/sdk/internal/zipstream/fuzz_test.go
+++ b/sdk/internal/zipstream/fuzz_test.go
@@ -5,6 +5,7 @@ package zipstream
 import (
 	"bytes"
 	"encoding/base64"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -129,6 +130,24 @@ func FuzzReader(f *testing.F) {
 		"HCALoriwCBQAAAgUAAFBLAQItAC0ACAAAAH11LzEke7o5HwAAAB8AAAAJAAAAAAAAAAAAAA" +
 		"AAAAAAAAAwLnBheWxvYWRQSwECLQAtAAgAAAB9dS8xAuiuLAIE///tBQAADwAAAAAAAAAAA" +
 		"AAAAABWAAAAMC5tYW5pZmVzdC5qc29uUEsFBgAAAAACAAIAdAAAAJUFAAAAAA=="))
+
+	// A central directory file comment: the entry after it only parses if
+	// nextCD accounts for FileCommentLength.
+	f.Add(buildRawZip(f, []rawZipEntry{
+		{name: "0.payload", data: []byte("payload bytes"), comment: "central directory file comment"},
+		{name: "0.manifest.json", data: []byte(`{"m":1}`)},
+	}, false))
+
+	// Filename plus extra field plus header size sums to 65646, which wraps
+	// to 110 if the addition happens at uint16 width.
+	f.Add(buildRawZip(f, []rawZipEntry{
+		{
+			name:        strings.Repeat("n", 65000),
+			data:        []byte("payload bytes"),
+			extraPrefix: bytes.Repeat([]byte{0}, 600),
+		},
+		{name: "0.manifest.json", data: []byte(`{"m":1}`)},
+	}, false))
 
 	f.Fuzz(func(t *testing.T, data []byte) {
 		reader, err := NewReader(bytes.NewReader(data))
--- a/sdk/internal/zipstream/reader.go
+++ b/sdk/internal/zipstream/reader.go
@@ -72,15 +72,19 @@ func NewReader(readSeeker io.ReadSeeker) (Reader, error) {
 	}
 
 	// check if zip is zip64 or zip32 format
+	//
+	// Any of the three EOCD fields that ZIP64 can overflow may carry the
+	// sentinel independently: the entry count (two bytes wide, so its
+	// sentinel is 0xFFFF), the central directory size, and the central
+	// directory offset. An archive with more than 65534 entries needs ZIP64
+	// for the count alone while its central directory still starts below
+	// 4 GiB, so keying off the offset by itself misses it.
 	var entryCount uint64
 	var centralDirectoryStart uint64
-	isZip64 := false
-	if endOfCDRecord.CentralDirectoryOffset != zip64MagicVal { //nolint:nestif // pkzip is complicated
+	if !eocdNeedsZip64(endOfCDRecord) { //nolint:nestif // pkzip is complicated
 		entryCount = uint64(endOfCDRecord.NumberOfCDRecordEntries)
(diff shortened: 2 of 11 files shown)


dmihalcik-virtru:
Merged in  #3979 and #3981

PR #3974, opened 2026-09-03
fix(examples): DSPX-4607 clear goconst and SA1019 lint findings

dmihalcik-virtru:
Part of the [DSPX-4607](https://virtru.atlassian.net/browse/DSPX-4607) lint burndown. golangci-lint v2.13.2 (#3965) surfaced 351 pre-existing findings across the repo; they're being cleared as independent PRs grouped by CODEOWNER. This one covers **`examples/` — 9 findings** (8 `goconst`, 1 `staticcheck` SA1019).

## Changes

**`goconst` (8)** — two strings repeated across the benchmark subcommands are now package constants in `cmd/examples.go`:

- `benchmarkCmdShort = "OpenTDF benchmark tool"` (3 uses)
- `exampleAttrValueFQN = "https://example.com/attr/attr1/value/value1"` (5 uses)

Two redundant aliases folded into the new constant while I was there: the `testAttr` package var in `benchmark_experimental.go` and the `attrValueFQN` local in `benchmark_decision_v2.go` both held the same literal.

**`staticcheck` SA1019 (1)** — `cmd/attributes.go` sets the deprecated `GetAttributeRequest.Id`. Annotated with `//nolint:staticcheck` per the agreed strategy for this burndown (annotate, don't migrate off deprecated proto fields).

> Worth a follow-up: `GetAttributeRequest` does have a non-deprecated `attribute_id` in its `identifier` oneof, and this is *example* code — showing the deprecated path is arguably the wrong thing to demonstrate. Left as-is here to keep the burndown mechanical; happy to migrate it in this PR if reviewers prefer.

No behavior change.

## Testing

- `golangci-lint run` over `examples/` → `0 issues.`, under both the current `.golangci.yaml` and the tuned config from #3968 (so this PR is independent of merge order).
- `golangci-lint fmt ./...` clean.
- `go build ./...` and `go test ./...` pass.

## Related

- #3965 — the linter bump that surfaced these (merged)
- #3968 — `.golangci.yaml` goconst tuning + `gomodguard_v2` migration
- #3969 (`otdfctl`), #3970 (`service/kas`), #3971 (`lib/fixtures`), #3973 (`sdk`) — sibling burndown PRs

[...]

PR #3965, opened 2026-09-03
chore(ci): DSPX-4607 bump golangci-lint to v2.13.2

dmihalcik-virtru:
### Proposed Changes

Fixes flaky lint failures in CI that are a network error, not a real lint finding:

```
[../../.golangci.yaml] validate: compile schema: failing loading
"https://golangci-lint.run/jsonschema/golangci.v2.8.jsonschema.json":
context deadline exceeded (Client.Timeout exceeded while awaiting headers)
Error: Command failed: /home/runner/golangci-lint-2.8.0-linux-amd64/golangci-lint config verify
```

**Root cause:** `golangci/golangci-lint-action` defaults `verify: true`, which runs `golangci-lint config verify`. In golangci-lint **v2.8.0** (what we pinned), `config verify` downloads the config JSONSchema over HTTPS with a **2-second client timeout**, so any slow response from `golangci-lint.run` fails the job. Known upstream complaint: golangci/golangci-lint#4864

**Upstream fix landed in v2.12.0:** `pkg/commands/config_verify.go` switched from an HTTP loader to `jsonsch.NewEmbedLoader()`, so the schema is compiled into the binary via `go:embed`.

| version | schema source |
| --- | --- |
| v2.8.0 | remote URL, 2s timeout |
| v2.9.0, v2.11.0 | remote URL |
| **v2.12.0+** | **embedded** |

This bumps all three `2.8.0` pins to `2.13.2` (current latest):

- `.github/workflows/checks.yaml` — `golangci-lint-action` `version:` in the `go` matrix job
- `.github/workflows/checks.yaml` — `go install ...golangci-lint@v2.13.2` in the proto-generate job
- `Makefile` — `toolcheck` install hint and `required_golangci_lint_version` (still a minimum, not an exact pin)

The action's `verify` input stays at its `true` default. Bumping already removes the network call, so disabling verification would needlessly give up validation of `.golangci.yaml`. No `.golangci.yaml` changes are needed — the existing config validates cleanly under v2.13.2.

Ticket: [DSPX-4607](https://virtru.atlassian.net/browse/DSPX-4607)

#### Notes for reviewers

- `make lint` is never invoked in CI — the only lint gate is the `golangci-lint-action` step in the `go` matrix, which uses `only-new-issues: true`.
[...]

PR #3978, opened 2026-09-04
fix(core): DSPX-4607 clear sloglint, goconst, nolintlint and SA1019 findings

dmihalcik-virtru:
Part of the [DSPX-4607](https://virtru.atlassian.net/browse/DSPX-4607) burndown of the 351 pre-existing findings that golangci-lint v2.13.2 (#3965) surfaced and v2.8.0 never reported. They don't fail CI today — the lint step runs with `only-new-issues: true` — but any future PR touching one of these lines would trip on them.

This PR covers the `service` core slice (60 findings) — everything under `service/` that isn't `service/policy` (#3977) or `service/kas` (#3970). It is independent of the other DSPX-4607 PRs: no file overlap, and it does not touch `.golangci.yaml`.

## Changes

**sloglint (25)**

*Audit event keys are suppressed, not renamed (6).* `service/logger/audit/utils.go` emits `userAgent`, `requestIP`, `requestID` and `actorID`. Those names are a published contract: they're asserted as golden JSON in `logger_test.go` and enforced as reserved paths by `schema.go`, and renaming them would break downstream SIEM consumers. Each `LogValue` gets one function-level `//nolint:sloglint` carrying that rationale.

*The other 17 are ordinary operational keys*, renamed to snake_case: `from group` → `from_group` (entityresolution), `FQN` → `fqn` (evaluate), `supportsResourceAuth` → `supports_resource_auth` (casbin v1 + v2 authorizers), `isDefaultModel` / `isBuiltinPolicy` / `isPolicyExtended` / `isDefaultAdapter` (casbin enforcer), `kasInfo` → `kas_info`, `runMigrations` / `verifyConnection` (db), `prettyPrint` / `maxSizeMB` / `maxBackups` / `maxAgeDays` (otel), `Indexer` → `indexer`.

`SDKConfig.LogValue` in `pkg/config/config.go` also moves from loose key-value pairs to typed `slog.Group`/`slog.String`/`slog.Bool` attrs.

**nolintlint (9 removed)** — 2 stale `//nolint:sloglint // emoji` in `integration/main_test.go` and 7 `// preserve emoji usage` in `internal/fixtures/fixtures.go`. Six of those were reported unused outright. The seventh was genuinely suppressing something — but args-on-separate-lines, not emoji, so the stated reason was wrong. Rather than keep a misleading directive, the three `slog.Error` calls in `(*Fixtures).provision` are split one attr per line.

**goconst (17)** — constants extracted where the repeated literal has a real name:

- `entityresolution/multi-strategy/mapper.go` — `transformationCSVToArray` / `transformationArray` / `transformationString` / `transformationLowercase` / `transformationUppercase`, shared by `mapper.go` and `output_mapper.go`. Named `transformation*` rather than `transform*` to avoid colliding with the existing `(*OutputMapper).transformCSVToArray` method.
- `providers/{claims,ldap,sql}` — one `providerType*` const per package, used by both the provider and its mapper.
- `internal/security` — `pemTypePublicKey = "PUBLIC KEY"`, used by `standard_crypto.go`, `basic_manager.go` and `in_process_provider.go`.

**canonicalheader (1)** — `internal/auth/authn.go`: `r.Header.Values("Dpop")` → `r.Header.Values("DPoP")`. Cosmetic only: `Header.Values` canonicalizes through `textproto.CanonicalMIMEHeaderKey`, so the lookup is unchanged and there is no DPoP interop exposure.

**CodeQL follow-on — DPoP proofs are no longer logged verbatim.** Touching that line made CodeQL attribute a pre-existing `go/clear-text-logging` flow (2 high) to this PR: `dp` is a raw request header and both `WarnContext` calls in the failure path logged it with `slog.Any("dpop", dp)`. The same code is on `main` today — CodeQL only reports it here because the taint source line is in the diff — but rather than suppress it, the two call sites now log `slog.Int("dpop_proof_count", len(dp))`.

That keeps the diagnostic that actually matters (was a proof presented at all?) and drops attacker-controlled header content out of warn-level logs on every unauthenticated request, which was also a log-injection surface. No test asserts on the attribute, and `go test ./internal/auth/...` passes.

**staticcheck SA1019 (9)** — `//nolint:staticcheck` with a per-site reason:

- `authorization/authorization.go` — the v1 authorization API still returns the deprecated `Action.Value` oneof for clients that have not moved to `Action.Name`.
- `integration/keymanagement_test.go` — exercises the deprecated `Manager` field and `Name` identifier, which must keep validating until they're removed.
- `internal/server/memhttp/memhttp.go`, `internal/server/server.go` — `h2c.NewHandler` is deprecated in favour of `http.Server.Protocols`; that migration is a Go 1.24+ server rework and is tracked separately.

## Testing

- `golangci-lint run` over the affected packages — 0 findings, verified under both the tuned config from #3968 and `main`'s config, so merge order doesn't matter.
- `make fmt`
- `cd service && go test ./logger/audit/... -race` — pass. This is the guard for the audit change: the golden-JSON and reserved-path assertions confirm the emitted event schema is byte-identical.
- `cd service && go test ./... -race` — pass, other than pre-existing environmental failures unrelated to this change (`rttests` needs a platform on `:8080`; `integration` and `entityresolution/integration` need testcontainers, and both pass with the Docker socket configured).
- Confirmed zero files under `service/policy/` or `service/kas/` are touched, so this stays disjoint from #3977 and #3970.

## Related

#3965 (linter bump) · #3968 (config tuning) · #3969 · #3970 · #3971 · #3973 · #3974 · #3975 · #3977

[DSPX-4607]: https://virtru.atlassian.net/browse/DSPX-4607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## DSPX-4607 burndown index

- #3965 — golangci-lint v2.13.2 bump (merged)
- #3968 — `.golangci.yaml` goconst tuning + `gomodguard_v2` migration (merged)
- Siblings: #3969 (`otdfctl`), #3970 (`service/kas`), #3971 (`lib/fixtures`), #3973 (`sdk`), #3974 (`examples`), #3975 (`tests-bdd`), #3977 (`service/policy`)

### DIFF
--- a/service/internal/auth/authn.go
+++ b/service/internal/auth/authn.go
@@ -491,7 +491,9 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 			return
 		}
 
-		dp := r.Header.Values("Dpop")
+		// The proofs are attacker-controlled request headers, so only their count is
+		// ever logged; logging the values verbatim trips CodeQL go/clear-text-logging.
+		dp := r.Header.Values("DPoP")
 		log := a.logger
 
 		// Verify the token
@@ -536,7 +538,7 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 					ctxWithAuthX,
 					"unauthenticated",
 					slog.Any("error", err),
-					slog.Any("dpop", dp),
+					slog.Int("dpop_proof_count", len(dp)),
 				)
 				http.Error(w, "unauthenticated", http.StatusUnauthorized)
 				return
@@ -545,7 +547,7 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 				ctxWithAuthX,
 				"unauthenticated",
 				slog.Any("error", err),
-				slog.Any("dpop", dp),
+				slog.Int("dpop_proof_count", len(dp)),
 			)
 			http.Error(w, "unauthenticated", http.StatusUnauthorized)
 			return
--- a/service/authorization/authorization.go
+++ b/service/authorization/authorization.go
@@ -605,6 +605,7 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
 						Decision:      authorization.DecisionResponse_DECISION_DENY,
 						EntityChainId: ec.GetId(),
 						Action: &policy.Action{
+							//nolint:staticcheck // the v1 authorization API still returns the deprecated Action.Value oneof for clients that have not moved to Action.Name
 							Value: &policy.Action_Standard{
 								Standard: policy.Action_STANDARD_ACTION_TRANSMIT,
 							},
@@ -769,6 +770,7 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
 				Decision:      decision,
 				EntityChainId: ec.GetId(),
 				Action: &policy.Action{
+					//nolint:staticcheck // the v1 authorization API still returns the deprecated Action.Value oneof for clients that have not moved to Action.Name
 					Value: &policy.Action_Standard{
 						Standard: policy.Action_STANDARD_ACTION_TRANSMIT,
 					},
--- a/service/entityresolution/keycloak/v2/entity_resolution.go
+++ b/service/entityresolution/keycloak/v2/entity_resolution.go
@@ -373,7 +373,7 @@ func expandGroup(ctx context.Context, groupID string, kcConnector *Connector, kc
 			logger.DebugContext(ctx,
 				"adding members",
 				slog.Int("amount", len(grpMembers)),
-				slog.String("from group", *grp.Name),
+				slog.String("from_group", *grp.Name),
 			)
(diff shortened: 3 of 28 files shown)


service/internal/auth/authn.go
```diff
@@ -536,7 +538,7 @@ func (a Authentication) MuxHandler(handler http.Handler) http.Handler {
 					ctxWithAuthX,
 					"unauthenticated",
 					slog.Any("error", err),
-					slog.Any("dpop", dp),
+					slog.Int("dpop_proof_count", len(dp)),
```
jakedoublev, line 541: Is it a leak that we were logging the signed DPoP proof in the error states prior to this change?

jakedoublev (commented):
LGTM, pending the one DPoP question

PR #3985, opened 2026-09-04
fix(authz): deny decrypt deactivated attribute

elizabethhealy:
### Proposed Changes

*

### Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Deactivated attribute values are now excluded from authorization decisions and entitlement results across direct, dynamic, hierarchy, subject-mapping, and registered-resource access paths.
  * Deactivated attribute definitions now deny access to their values, including dynamically created values.
  * Hierarchy evaluation stops at deactivated values, preventing lower-level values from inheriting access.

* **Tests**
  * Added coverage for deactivated values and definitions across PDP and end-to-end authorization scenarios.


service/internal/access/v2/pdp_deactivated_test.go
```diff
@@ -0,0 +1,455 @@
...
+	}
+
+	pdp, err := NewPolicyDecisionPoint(ctx, s.logger, []*policy.Attribute{attr}, subjectMappings, nil, false, false)
+	s.Require().NoError(err)
+
+	entity := s.createEntityWithProps("entity-both-projects", map[string]interface{}{
+		"project": []interface{}{"active", "inactive"},
+	})
+
+	entitlements, err := pdp.GetEntitlements(ctx, []*entityresolutionV2.EntityRepresentation{entity}, nil, false)
```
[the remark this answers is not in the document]
elizabethhealy: addressed in latest commit

tests-bdd/features/direct-entitlements.feature
```diff
@@ -65,6 +65,30 @@ Feature: Direct entitlements decisioning
...
+    And I deactivate the attribute value "https://example.com/attr/department/value/eng"
+    When I send a decision request for entity chain "alice" for "read" action on resource "https://example.com/attr/department/value/eng"
+    Then the response should be successful
+    And I should get a "DENY" decision response
+
+  # Deactivating a definition does not cascade to its values, so the value is still active here.
+  # The full entitlement policy load drops deactivated definitions outright, so the resource FQN
+  # is unknown; GetDecision treats an unknown FQN as a per-resource deny rather than a request
+  # error, so the response is successful and the decision is DENY.
+  Scenario: Direct entitlement denies a value whose definition is deactivated
```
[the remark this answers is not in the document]
elizabethhealy: addressed in latest commit

service/internal/access/v2/helpers.go
```diff
@@ -25,6 +26,19 @@ var (
 	ErrInvalidDynamicValueMapping     = errors.New("access: invalid dynamic value mapping")
 )
 
+// isExplicitlyInactive reports whether an active state was loaded and is false. An unset state is
+// not inactive: targeted lookups, synthetic values, and in-memory fixtures all leave it unset.
+func isExplicitlyInactive(active *wrapperspb.BoolValue) bool {
+	return active != nil && !active.GetValue()
+}
+
+// isDeactivated reports whether an attribute value, or the definition owning it, is deactivated.
```
jakedoublev, line 35: We have a database trigger that sets values on definitions and definitions + values on namespaces inactive when either parent is deactivated. It should not be possible to have active values on an inactive definition. We can keep this check for defensiveness, but we should log an error about a bad state if we encounter it.

service/internal/access/v2/pdp_deactivated_test.go
```diff
@@ -0,0 +1,455 @@
...
+
+// Deactivating an attribute value must fail closed on every decision path: data already tagged
+// with the value can no longer be decrypted, and no entitlement source may resurrect it.
+// Regression coverage for a TDF remaining decryptable after its value was deactivated.
+
+const deactivatedTestNamespace = "deactivation.example.com"
+
+var (
+	testDeactivatedProjectFQN      = createAttrFQN(deactivatedTestNamespace, "project")
+	testDeactivatedProjectActive   = createAttrValueFQN(deactivatedTestNamespace, "project", "active")
```
jakedoublev: Nit: could use different value names here to avoid confusion with the active state of the values

service/internal/access/v2/pdp_deactivated_test.go
```diff
@@ -0,0 +1,455 @@
...
+	})
+}
+
+// Deactivated values on the direct-entitlement path are covered by the deactivation subtests of
+// Test_GetDecision_DirectEntitlements in pdp_test.go; only the deactivated-definition case below
+// is unique to this file.
+
+// Test_GetDecision_DeactivatedValue_RegisteredResources covers registered resources on both sides
+// of a decision: as the entity being entitled, and as the resource being accessed.
+func (s *PDPTestSuite) Test_GetDecision_DeactivatedValue_RegisteredResources() {
```
jakedoublev, line 207: https://virtru.atlassian.net/browse/DSPX-4630?focusedCommentId=269199 I think we need to get some more alignment clarity on the edge cases here.

tests-bdd/features/deactivated-attribute-values.feature
```diff
@@ -0,0 +1,68 @@
...
+
+  Scenario: HIERARCHY — deactivating the value on the TDF denies decrypt
+    When I encrypt plaintext "internal memo" with attributes "https://demo.com/attr/classification/value/internal" stored as "tdf_class"
+    And using token "alice_tok", decrypt "tdf_class" stored as "plain_class_before"
+    Then the decryption stored as "plain_class_before" should succeed with plaintext "internal memo"
+    When I deactivate the attribute value "https://demo.com/attr/classification/value/internal"
+    And using token "alice_tok", decrypt "tdf_class" stored as "plain_class_after"
+    Then the decryption stored as "plain_class_after" should be denied
+
+  # Deactivating a definition does not cascade to its values in the database — each value keeps
```
jakedoublev: Deactivation of a definition should cascade to its values in the database? https://github.com/opentdf/platform/blob/9deeee89d9a9a49f63228101e8634d91385f6a93/service/policy/db/migrations/20240212000000_add_active_state_enums.sql#L27

tests-bdd/features/direct-entitlements.feature
```diff
@@ -65,6 +65,30 @@ Feature: Direct entitlements decisioning
...
+  Scenario: Direct entitlement denies a deactivated attribute value
+    Given there is a claims subject entity referenced as "alice" with direct entitlements:
+      | attribute_value_fqn                           | actions |
+      | https://example.com/attr/department/value/eng | read    |
+    And I deactivate the attribute value "https://example.com/attr/department/value/eng"
+    When I send a decision request for entity chain "alice" for "read" action on resource "https://example.com/attr/department/value/eng"
+    Then the response should be successful
+    And I should get a "DENY" decision response
+
+  # Deactivating a definition does not cascade to its values, so the value is still active here.
```
jakedoublev: Same comment as other BDD test. Deactivating a definition should indeed cascade to its values.

jakedoublev (commented):
Some open questions around registered resources and clarity of definition deactivation cascading to values below

service/internal/access/v2/helpers.go
elizabethhealy: Kept the check for defensiveness and added the error log. isDeactivated now distinguishes the two cases and only logs when the state is genuinely bad, a definition explicitly inactive with a value explicitly active.

service/internal/access/v2/pdp_deactivated_test.go
elizabethhealy: renamed to alpha/beta so the value name no longer overloads the active state

tests-bdd/features/deactivated-attribute-values.feature
elizabethhealy: fixed

tests-bdd/features/direct-entitlements.feature
elizabethhealy: fixed

service/internal/access/v2/pdp_deactivated_test.go
elizabethhealy: leaving open until we get input from jp/ryan on that jira comment

service/internal/access/v2/helpers.go
elizabethhealy: updated

service/internal/access/v2/pdp_deactivated_test.go
elizabethhealy: updated to use your suggestion, i think it makes sense -- for an entity treat it as over entitled and for a resource treat it like a resource

alkalescent (approved):

ttschampel:
I'm good with this approach ~ DENY on decision for a registered resource (RR) when any of the RR's attached attribute values are deactivated

jakedoublev (approved):

PR #3986, opened 2026-09-05
fix(policy): narrow authorization attribute lookups

strantalis:
### Proposed Changes

`GetEntitleableAttributesByFqns` was using the general attribute query, which hydrated unrelated values, mappings, grants, and keys before issuing its targeted subject-mapping query. Add an authorization-specific SQL query that selects requested values and the ordered active siblings required for hierarchy evaluation.

Preserve normalization, inactive-value errors, missing-value errors, hierarchy order, namespace identity, and allow-traversal behavior. General attribute reads and key mapping lookups continue using their existing queries. No schema migration or public protocol change is needed.

This addresses the default decision path exercised by DSPX-4625. It is separate from the paginated `ListSubjectMappings` optimization.

Layer 1 of 7 in the authorization performance stack. Review and merge from the bottom upward.

### Checklist

- [x] Added or updated unit or integration coverage appropriate to this layer
- [x] Ran focused verification for this layer
- [x] Documented behavior and validation limits below

### Testing Instructions

- Focused policy DB and authorization tests, including race-enabled `GetEntitleableAttributesByFqns` integration tests, passed.
- Used the current DSPX-4625 test fixture through a local Go overlay: 6,011 subject mappings, 6,000 resource mappings, three resources per request, 50 concurrent requests, cache and experimental features disabled. The fixture remains in its separate test task.
- Two local comparisons gave p95 of 4.066s and 4.027s with the old lookup, versus 135.6ms and 133.6ms with the targeted lookup. All 50 requests returned the expected decisions in each run. These are local regression measurements, not a production latency guarantee.
- Generated SQL with `make policy-sql-gen`; `make fmt` passed.

Stack-wide validation:

- Focused race tests, the full verbose policy integration suite with the race detector, SDK README code-block tests, formatting, and service lint restricted to changes since the stack base passed with zero new issues.
- `make test` was attempted with Colima configured. The round-trip suite failed because its required platform server at `127.0.0.1:8080` was not running.
- `make lint` stopped because the configured Buf API token is invalid. Separate `make go-lint` reported existing repository issues; the formatting issue in the new hierarchy test was corrected.
- `make govulncheck` reported vulnerabilities in the existing dependencies and the installed Go 1.26.3 standard library. This stack changes no dependency versions.
- GitHub CI is still running. Keep this stack in draft pending review and CI results.


service/policy/db/queries/entitleable_attributes.sql
```diff
@@ -0,0 +1,35 @@
...
+    JOIN attribute_namespaces ns ON ns.id = ad.namespace_id AND ns.active = TRUE
+    WHERE df.fqn = ANY(@definition_fqns::text[]) AND ad.active = TRUE
+), requested_values AS (
+    SELECT av.id, av.attribute_definition_id, av.active, vf.fqn
+    FROM attribute_fqns vf
+    JOIN attribute_values av ON av.id = vf.value_id
+    JOIN definitions d ON d.id = av.attribute_definition_id
+    WHERE vf.fqn = ANY(@value_fqns::text[])
+), selected_values AS (
+    SELECT * FROM requested_values
```
jakedoublev, line 18: Need some comments about each subquery's intent

service/policy/db/queries/entitleable_attributes.sql
```diff
@@ -0,0 +1,35 @@
...
+    UNION ALL
+    SELECT av.id, av.attribute_definition_id, av.active, vf.fqn
+    FROM definitions d
+    JOIN attribute_values av ON av.attribute_definition_id = d.id AND av.active = TRUE
+    JOIN attribute_fqns vf ON vf.value_id = av.id
+    WHERE d.rule = 'HIERARCHY' AND NOT EXISTS (SELECT 1 FROM requested_values rv WHERE rv.id = av.id)
+)
+SELECT d.id AS definition_id, d.definition_fqn, d.rule, d.allow_traversal,
```
jakedoublev, line 26: Should the state of `allow_traversal` be affecting anything here so that the API behavior is true:
```
// A value that does not exist under a definition with allow_traversal is returned
// with its definition and an empty value identity (no value_id, no subject
// mappings), so front-loaded values still carry their definition context.```

service/policy/db/attribute_fqn.go
jakedoublev, line 311: Maybe this is resolved later in the stack, but it seems like we'd have room to remove this second roundtrip.

service/policy/db/queries/entitleable_attributes.sql
jakedoublev: Maybe we need an integration test for the behavior with `allow_traversal` true or false

service/policy/db/attribute_fqn.go
jakedoublev: Looks like it is not resolve in the rest of the stack, but is a further optimization and not a regression.

PR #3990, opened 2026-09-05
fix(authz): reuse policy and entity resolution within bulk requests

strantalis:
### Proposed Changes

Repeated entries in `GetDecisionBulk` fetched and compiled the same policy and resolved the same entity again. Add bounded memo tables scoped to one sequential bulk RPC: eight policy sets and 32 successful entity resolutions.

Policy keys normalize and deduplicate FQNs. Resolution keys include the complete token or entity-chain request, resource context, and environment-entity mode. Separate RPCs get separate tables; failures are not cached. This does not introduce parallel evaluation.

Layer 5 of 7 in the authorization performance stack. Review and merge from the bottom upward.

### Checklist

- [x] Added or updated unit or integration coverage appropriate to this layer
- [x] Ran focused verification for this layer
- [x] Documented behavior and validation limits below

### Testing Instructions

- Race tests verify repeated policy/entity reuse, capacity limits, fresh state for a new RPC, and cache misses when tokens, resource context, or environment-entity mode change.
- Access v2 and authorization v2 focused tests passed.

Stack-wide validation:

- Focused race tests, the full verbose policy integration suite with the race detector, SDK README code-block tests, formatting, and service lint restricted to changes since the stack base passed with zero new issues.
- `make test` was attempted with Colima configured. The round-trip suite failed because its required platform server at `127.0.0.1:8080` was not running.
- `make lint` stopped because the configured Buf API token is invalid. Separate `make go-lint` reported existing repository issues; the formatting issue in the new hierarchy test was corrected.
- `make govulncheck` reported vulnerabilities in the existing dependencies and the installed Go 1.26.3 standard library. This stack changes no dependency versions.
- GitHub CI is still running. Keep this stack in draft pending review and CI results.

### DIFF
--- a/service/authorization/v2/authorization.go
+++ b/service/authorization/v2/authorization.go
@@ -277,6 +277,8 @@ func (as *Service) GetDecisionBulk(ctx context.Context, req *connect.Request[aut
 		return nil, statusifyError(ctx, as.logger, errors.Join(ErrFailedToInitPDP, err))
 	}
 
+	pdp = pdp.WithRequestReuse()
+
 	multiRequests := req.Msg.GetDecisionRequests()
 	decisionResponses := make([]*authzV2.GetDecisionMultiResourceResponse, len(multiRequests))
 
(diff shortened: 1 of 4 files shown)


jakedoublev:
I think this approach generally makes sense, but I think it's lower priority than some of the optimizations in the other layers of the stack.

PR #4024, opened 2026-09-10
fix(ers): DSPX-4595 limit multi-strategy token chains to one entity per category

khvirtru:
## Summary

- `CreateEntityChainsFromTokens` with `failure_strategy: continue` accumulated multiple subject entities when several strategies matched the same token. This broke AND semantics — a second entity with fewer claims could veto the richer one, causing unexpected DENY decisions.
- Fix: track resolved categories (ENVIRONMENT, SUBJECT) and skip strategies whose category is already filled. First matching strategy per category wins, matching the Keycloak ERS contract.
- `continue` still works for error handling: if a strategy *fails*, the next matching one is tried.

## Bug details

**Root cause:** The loop in `createEntityChainFromSingleTokenV2` only broke on success under `fail-fast`. Under `continue`, it fell through to the next strategy even after a success, appending entities.

**ADR reference:** The [multi-strategy ADR](https://github.com/opentdf/platform/blob/main/adr/decisions/2025-07-31-multi-strategy-entity-resolution-service.md#strategy-configuration-properties) is explicit: on success, return result immediately — `failure_strategy` controls what happens on *failure* only.

## Repro steps (before fix)

1. Configure 2+ subject strategies that both match a token (e.g. both have `conditions: jwt_claims: [{claim: sub, operator: exists}]`)
2. Set `failure_strategy: continue`
3. Call `CreateEntityChainsFromTokens` with a JWT that matches both strategies
4. **Bug:** Chain contains multiple subject entities (one per matching strategy)
5. A subject mapping that matches strategy 1's claims but not strategy 2's → **DENY** (strategy 2's entity vetoes via AND semantics)

**After fix:** Chain contains exactly one subject entity from the first matching strategy. Second strategy is skipped.

## Changes

| File | Change |
|------|--------|
| `v2/registration.go` | Add `resolvedCategories` map + `categoryForStrategy()` helper. Skip strategies whose category is already filled. Break when all categories resolved. |
| `v2/registration_test.go` | 7 new unit tests: competing subjects collapse, env+subject each contribute, strategy order, redundant skip, fail-fast parity, env-first regression, continue fallthrough |
| `integration/multistrategy_test.go` | Fix expected entity count 3→1 for all-subject-strategies test case |
| `integration/internal/chain_contract_tests.go` | Extract `expectedChainCategories()` to deduplicate |
| `README.md` | Document first-match-wins semantics, environment entity handling in authz |

## Test plan

- [x] Unit tests pass: `go test ./service/entityresolution/multi-strategy/v2/ -run TestCreateEntityChainsFromTokens` (7 tests)
- [x] Integration tests pass: `go test ./service/entityresolution/integration/ -run TestMultiStrategy`
- [ ] CI: Cucumber BDD, xtest (Go/Java/JS SDK)
- [ ] Review by multi-strategy ERS developer (Chris Reed / JP / Ryan Schumacher)

## Summary by CodeRabbit

* **Behavior Changes**
  * Entity chains now include at most one environment entity and one subject entity.
  * The first successful strategy for each entity category is used; later strategies are skipped.
  * Subject entities remain available even when environment resolution occurs first.
  * Authorization decisions continue to reject tokens that resolve only to environment entities.

* **Documentation**
  * Clarified strategy selection and fallback behavior for `continue` failure handling.

### DIFF
--- a/service/entityresolution/integration/internal/chain_contract_tests.go
+++ b/service/entityresolution/integration/internal/chain_contract_tests.go
@@ -18,6 +18,11 @@ const (
 	expectedChainEntityCount = 2
 )
 
+// expectedChainCategories is the chain shape both implementations must produce per token.
+func expectedChainCategories() []string {
+	return []string{"CATEGORY_ENVIRONMENT", "CATEGORY_SUBJECT"}
+}
+
 // ChainContractTestSuite holds implementation-agnostic multi-entity chain validation tests
 type ChainContractTestSuite struct {
 	TestCases []ContractTestCase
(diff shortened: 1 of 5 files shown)


khvirtru:
Closing in favor of Liz's fix under DSPX-4595. The BDD tests in PR #3953 will rebase onto main once her fix lands.

PR #4029, opened 2026-09-10
feat(sdk,cli): settable segment and root integrity algorithms (DSPX-4736)

dmihalcik-virtru:
### Proposed Changes

Adds explicit controls for the two integrity algorithms a ZTDF writer picks, and makes an unsupported choice fail loudly instead of being silently coerced.

The asymmetry is the point:

| | allowed | why |
|---|---|---|
| **segment** | `HS256`, `GMAC` | a segment hash is computed over ciphertext AES-GCM actually produced, so "GMAC" means reading back a real tag |
| **root** | `HS256` only | the root signature covers the aggregate hash, which never passes through AES-GCM — there is no tag to read out, so GMAC would authenticate nothing |

**`sdk`**
* `WithSegmentIntegrityAlgorithm(HS256|GMAC)`
* `WithRootIntegrityAlgorithm(HS256)` — anything else returns the new `ErrUnsupportedRootIntegrityAlgorithm`. The option exists so callers can state the choice explicitly and a CLI can surface the refusal, not to widen it.
* The same split in `sdk/experimental/tdf`, where `NewWriter` validates eagerly because `Option[T]` cannot return an error.

**`otdfctl`**
* `encrypt --segment-integrity-algorithm` and `--root-integrity-algorithm`, validated *before* the handler is constructed, so an invalid value fails offline as a config error rather than after a KAS round trip.
* Documented in `docs/man/encrypt/_index.md` with enum values — which is also what the cross-SDK xtest feature detectors grep for, so merging this flips the `integrity_algs` / `gmac_root_rejected` gates on and stops those cells from silently skipping.

### Scope

This is the **control surface** for evaluating DSPX-4703, deliberately separated from the fix. It changes what a writer may *choose*; it does not change how a manifest that already declares a GMAC root is verified on **read**. That is the stacked follow-up, opentdf/platform PR for DSPX-4703.

Splitting it this way means the xtest exploit cases can be run against this branch to get a live, reproducible "before" baseline, then against the fix branch to watch them go red → green.

### Checklist

### Testing Instructions

```bash
cd sdk     && go test ./...
cd otdfctl && go test ./pkg/handlers/...
```

Both were also run with this commit as `HEAD` and the DSPX-4703 fix commit entirely absent, to confirm the branch stands alone.

Manual check that the refusal is offline (no platform needed):

```bash
otdfctl encrypt --root-integrity-algorithm gmac ./somefile
# => unsupported root integrity algorithm: gmac
```

## Summary by CodeRabbit

* **New Features**
  * Added encryption options for selecting root and segment integrity algorithms.
  * Segment integrity now supports HS256 and GMAC.
  * Added SDK configuration options for root and segment integrity algorithms.

* **Bug Fixes**
  * Root signatures accept only HS256, with unsupported configurations rejected early.
  * Improved validation and error reporting for unsupported algorithm settings.

* **Documentation**
  * Updated CLI and SDK documentation and examples for separate integrity algorithm settings.

Cross-SDK coverage lives in opentdf/tests#594.

### DIFF
--- a/otdfctl/docs/man/encrypt/_index.md
+++ b/otdfctl/docs/man/encrypt/_index.md
@@ -24,7 +24,6 @@ command:
         The algorithm used for the TDF's root signature. Only hs256 is supported; gmac is rejected.
       enum:
         - hs256
-        - gmac
       default: hs256
     - name: segment-integrity-algorithm
       description: >
--- a/otdfctl/docs/man/encrypt/_index.md
+++ b/otdfctl/docs/man/encrypt/_index.md
@@ -112,12 +112,7 @@ ciphertext; `hs256` recomputes an HMAC-SHA256 over the same bytes.
 
 `--root-integrity-algorithm` (default `hs256`) selects how the root signature
 over the aggregate of those segment hashes is computed. **Only `hs256` is
-supported.** The aggregate hash never passes through AES-GCM, so there is no
-tag to read back out and `gmac` would authenticate nothing — it degenerates
-into a copy of the last segment hash, which anyone who can edit the manifest
-can rewrite. Passing `--root-integrity-algorithm gmac` fails with
-`unsupported root integrity algorithm`, and decrypting a TDF whose manifest
-declares a GMAC root is refused for the same reason.
+supported.**```
 
 ```shell
 # HS256 segment hashes instead of the default GMAC
--- a/otdfctl/docs/man/encrypt/_index.md
+++ b/otdfctl/docs/man/encrypt/_index.md
@@ -19,6 +19,20 @@ command:
         - ec:secp384r1
         - ec:secp521r1
       default: rsa:2048  
+    - name: root-integrity-algorithm
+      description: >
+        The algorithm used for the TDF's root signature. Only hs256 is supported; gmac is rejected.
+      enum:
+        - hs256
+        - gmac
+      default: hs256
+    - name: segment-integrity-algorithm
+      description: >
+        The algorithm used to compute each payload segment's integrity hash.
+      enum:
+        - hs256
+        - gmac
+      default: gmac
     - name: mime-type
       description: The MIME type of the input data. If not provided, the MIME type is inferred from the input data.
     - name: tdf-type
@@ -86,6 +100,29 @@ Example
 # Encrypt a file using the ec:secp256r1 algorithm for the wrapping key
 # EXPERIMENTAL
 otdfctl encrypt hello.txt --wrapping-key-algorithm ec:secp256r1 --out hello.txt.tdf
+```
+
+## Integrity Algorithms
+
+A ZTDF carries two kinds of integrity value, and they are not interchangeable.
+
+`--segment-integrity-algorithm` (default `gmac`) selects how each payload
+segment's hash is computed. Both values are supported. `gmac` reads out the
+AES-GCM authentication tag the cipher already produced over that segment's
+ciphertext; `hs256` recomputes an HMAC-SHA256 over the same bytes.
+
+`--root-integrity-algorithm` (default `hs256`) selects how the root signature
+over the aggregate of those segment hashes is computed. **Only `hs256` is
+supported.** The aggregate hash never passes through AES-GCM, so there is no
+tag to read back out and `gmac` would authenticate nothing — it degenerates
+into a copy of the last segment hash, which anyone who can edit the manifest
+can rewrite. Passing `--root-integrity-algorithm gmac` fails with
+`unsupported root integrity algorithm`, and decrypting a TDF whose manifest
+declares a GMAC root is refused for the same reason.
+
+```shell
+# HS256 segment hashes instead of the default GMAC
+otdfctl encrypt hello.txt --out hello.txt.tdf --segment-integrity-algorithm hs256
 ```
 
 ## Attributes
--- a/otdfctl/pkg/handlers/tdf.go
+++ b/otdfctl/pkg/handlers/tdf.go
@@ -38,6 +38,74 @@ type TDFInspect struct {
 	UnencryptedMetadata []byte
 }
 
+// IntegrityAlgorithms carries the segment and root integrity algorithm names as
+// they arrived from the CLI, so the flags can be validated before any work is
+// done and mapped onto SDK options in one place.
+type IntegrityAlgorithms struct {
+	Root    string
+	Segment string
+}
+
+// Validate reports whether both algorithm names are usable, without needing a
+// platform connection or any input data.
+//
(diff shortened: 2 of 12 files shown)


otdfctl/docs/man/encrypt/_index.md
```diff
@@ -19,6 +19,20 @@ command:
         - ec:secp384r1
         - ec:secp521r1
       default: rsa:2048  
+    - name: root-integrity-algorithm
+      description: >
+        The algorithm used for the TDF's root signature. Only hs256 is supported; gmac is rejected.
+      enum:
+        - hs256
+        - gmac
```
c-r33d: Should we list gmac as an option if it's rejected?

otdfctl/docs/man/encrypt/_index.md
```diff
@@ -19,6 +19,20 @@ command:
...
+      default: hs256
+    - name: segment-integrity-algorithm
+      description: >
+        The algorithm used to compute each payload segment's integrity hash.
+      enum:
+        - hs256
+        - gmac
```
c-r33d, line 33: Is this the same as the above?

otdfctl/docs/man/encrypt/_index.md
dmihalcik-virtru: yeah probably not. There are some variants of the SDK that allow hs512 but I'm not removing it everywhere since it isn't supported by all of them.

dmihalcik-virtru:
```suggestion
```

otdfctl/pkg/handlers/tdf.go
```diff
@@ -38,6 +38,74 @@ type TDFInspect struct {
...
+}
+
+// tdfOptions maps the requested algorithm names onto SDK options. Names are
+// matched case-insensitively; an empty name leaves the SDK default in place
+// (HS256 root, GMAC segments).
+//
+// The root option is where `gmac` is refused: the SDK returns
+// ErrUnsupportedRootIntegrityAlgorithm, whose message carries the literal
+// "unsupported root integrity algorithm" that callers grep for.
+func (a IntegrityAlgorithms) tdfOptions() ([]sdk.TDFOption, error) {
```
c-r33d, line 76: Nit: Should we just have one method that parses the integrity algorithms to tdfoptions. Seems this method is dual purpose, and actually doing the validation AND parsing the transforming it into tdfoptions already.

otdfctl/docs/man/encrypt/_index.md
```diff
@@ -88,6 +101,29 @@ Example
...
+ciphertext; `hs256` recomputes an HMAC-SHA256 over the same bytes.
+
+`--root-integrity-algorithm` (default `hs256`) selects how the root signature
+over the aggregate of those segment hashes is computed. **Only `hs256` is
+supported.** The aggregate hash never passes through AES-GCM, so there is no
+tag to read back out and `gmac` would authenticate nothing — it degenerates
+into a copy of the last segment hash, which anyone who can edit the manifest
+can rewrite. Passing `--root-integrity-algorithm gmac` fails with
+`unsupported root integrity algorithm`, and decrypting a TDF whose manifest
+declares a GMAC root is refused for the same reason.
```
dmihalcik-virtru:
```suggestion
`--root-integrity-algorithm` (default `hs256`) selects how the root signature
over the aggregate of those segment hashes is computed. **Only `hs256` is
supported.**```

File: otdfctl/pkg/handlers/tdf_test.go [hunk 5]
```diff
@@ -0,0 +1,80 @@
...
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.algs.Validate()
+			if tc.wantErr == "" {
+				require.NoError(t, err)
+				return
+			}
+			require.Error(t, err)
+			assert.Contains(t, err.Error(), tc.wantErr)
```
c-r33d, line 63: Would typed errors be good here, so you can use `errors.Is()`

otdfctl/docs/man/encrypt/_index.md
dmihalcik-virtru: This is for each segment (aes-gcm message). The root is computed as a function of the array of segment integrity hashes

The interesting one at this point is HMAC. Currently (before this PR), this is the only value tested by xtest since otdfctl doesn't expose this parameter. By adding this parameter, xtest will enable a new set of tests that validate segments that use the HS256 algorithm for their integrity hashes.

otdfctl/docs/man/encrypt/_index.md
strantalis, line 27: if hs256 is only supported should we remove gmac here?

otdfctl/docs/man/encrypt/_index.md
```diff
@@ -19,6 +19,20 @@ command:
...
+        The algorithm used for the TDF's root signature. Only hs256 is supported; gmac is rejected.
+      enum:
+        - hs256
+        - gmac
+      default: hs256
+    - name: segment-integrity-algorithm
+      description: >
+        The algorithm used to compute each payload segment's integrity hash.
+      enum:
+        - hs256
```
strantalis, line 33: does hs256 make any sense here if we don't support plaintext tdf today?

otdfctl/pkg/handlers/tdf.go
c-r33d: I see. The validation is happening in the tdfoption itself.

otdfctl/docs/man/encrypt/_index.md
c-r33d, line 120: Is this true? Seems that this is only true for the experimental SDK whereas the `stable` sdk does not promise the same rejection.

sdk/experimental/tdf/manifest.go
```diff
@@ -103,20 +103,52 @@ type EncryptedMetadata struct {
...
+// readAEADTag returns the trailing authentication tag of an AES-GCM ciphertext.
+//
+// PRECONDITION: ciphertext must be exactly the bytes AES-GCM sealed under the
+// DEK. Only then are the trailing kGMACPayloadLength bytes the tag the cipher
+// already computed over them; over anything else the same rule just copies the
+// input's own last 16 bytes and authenticates nothing. Reachable only from
+// segmentIntegrity, whose argument is by construction a segment ciphertext.
+func readAEADTag(ciphertext []byte) (string, error) {
+	if kGMACPayloadLength > len(ciphertext) {
 		return "", errors.New("fail to create gmac signature")
```
c-r33d, line 131: Should this be a typed error?

dmihalcik-virtru:
We are going to only allow HS256 root and GMAC segment integrity hashes.

PR #4034, opened 2026-09-10
fix(core): DSPX-4752 align OpenTelemetry dependencies [backport to release/service/v0.11]

jakedoublev:
## Summary

- align OTel dependencies and schema versions
- add regression coverage for tracing initialization and dependency drift


jakedoublev:
Closing to preserve the main-first workflow. The regression coverage will land on main before the existing main fix and test are backported to release/service/v0.11.

PR #4043, opened 2026-09-14
fix(sdk): bound zip64 reader offsets against the archive (DSPX-4590)

dmihalcik-virtru:
Stacked on #3981.

## Context

Reviewing #3981 raised the question of whether `ZipFileEntry` should be a `uint64` pair
instead of a signed one.

**It should not** — and chasing the question turned up a reachable panic that the
signedness change would have papered over rather than fixed.

## The bug

`NewReader` took ZIP64 offsets and sizes straight off disk as `uint64` (raw
`binary.LittleEndian.Uint64` reads in `parseZip64ExtraField`) and narrowed them to `int64`
without a check. A central directory entry carrying the `0xFFFFFFFF` sentinel plus a ZIP64
extended-information extra field declaring a stored size of `1<<63`:

1. `reader.go:173` — `zipFileEntry.length = int64(bytesToRead)` → negative
2. `reader.go:350` — `fileNameEntry.length > maxSize` is a **signed** compare, so a negative
   length sails past the size guard
3. `reader.go:375` — `make([]byte, size)` panics

Reproduced against the pre-fix reader via the new fuzz seed:

```
--- FAIL: FuzzReader/seed#5
panic: runtime error: makeslice: len out of range
  zipstream.readBytes(..., 0x8000000000000000)  reader.go:375
  zipstream.Reader.ReadAllFileData(...)          reader.go:354
```

These values are attacker-controlled in any TDF, and the existing fuzz harness already walks
exactly this path — it just had no seed that reached it.

Three more unchecked narrowings had the same shape: the ZIP64 locator's `CDOffset`, the
central-directory cursor `nextCD + centralDirectoryStart`, and the entry's local header
offset. A ZIP64 entry count is also a full `uint64` with nothing bounding it.

## The fix

Bound every disk-sourced position against the archive length, which the existing
`Seek(-endOfCDRecordSize, io.SeekEnd)` already yields for free (its return value was
discarded).

Because that length is derived from an `int64`, **one** comparison rules out both failure
modes — `base+delta <= archiveSize` implies the sum fits a non-negative `int64`:

```go
func withinArchive(field string, base, delta, archiveSize uint64) (int64, error) {
	if base > archiveSize || delta > archiveSize-base {
		return 0, fmt.Errorf("%w: %s at %d+%d runs past the end of a %d byte archive",
			errZipFormat, field, base, delta, archiveSize)
	}
	return int64(base + delta), nil
}
```

The remaining space is computed by subtraction so the addition cannot wrap before it is
checked. This is the read-side counterpart to `checkFitsInCentralDirectory` from #3981, which
already refuses to *narrow* a value into a 32-bit field on write — nothing was refusing to
*widen* a hostile 64-bit value back out of one.

Also bounds the ZIP64 entry count: every central directory record is at least
`cdFileHeaderSize` bytes, so a hostile count fails immediately rather than after billions of
seek-and-fail iterations.

## Why not `uint64`

- Both fields are unexported in an `internal/` package — no external pressure toward `uint64`.
- Every consumer is signed: `io.ReadSeeker.Seek`, `make()`, and the `int64` returned by
  `ReadFileData` / `ReadAllFileData` / `ReadFileSize`, which `TDFReader.ReadPayload` and
  `PayloadSize` re-export out to `sdk/tdf.go`.
- `uint64` storage only relocates the narrowing into `readBytes`, where there is no archive
  length left to check against — and `make([]byte, hugeUint64)` still dies.

`ZipFileEntry` keeps its signed pair, now with a documented invariant: both fields are
non-negative and `index+length` lands inside the archive.

## Testing

New tests in `zip64_conformance_test.go` (`buildRawZip` gained two override fields so a
fixture can lie about its ZIP64 size/offset):

- `TestReaderRejectsZip64ValuesBeyondArchive` — stored size and header offset, each both
  above `MaxInt64` and merely past EOF
- `TestReaderAcceptsEntryEndingAtArchiveBoundary` — inclusive-bound / off-by-one control: an
  entry ending on the archive's last byte is accepted, one byte further is rejected
- `TestReaderRejectsZip64LocatorOffsetBeyondArchive`
- `TestReaderRejectsEntryCountBeyondArchive`
- New `FuzzReader` seed for the panic

All four new `TestReaderRejects*` / `TestReaderAccepts*` cases were confirmed to **fail against
the pre-fix reader** and pass after.

- `go test ./... -race` — sdk: all 7 packages pass
- `go test -fuzz FuzzReader -fuzztime 60s` — 2.25M execs, no crashes
- `golangci-lint run ./internal/zipstream/...` — 0 issues in changed files
- `make fmt` clean; `lib/flattening`, `lib/identifier`, `examples`, `otdfctl` pass

> Note: `make test` fails in `lib/fixtures` (`TestTokenManager_InitialLogin`,
> `TestTokenManager_CustomTokenBuffer` — token-buffer assertions). Pre-existing and unrelated:
> that module has zero dependency on the sdk module (`go list -deps` returns no
> `opentdf/platform/sdk`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or hostile ZIP and ZIP64 archives.
  * Prevented crashes caused by invalid archive sizes and offsets.
  * Blocked reads beyond file boundaries, including data from archive metadata or the central directory.
  * Added validation for archive structure, entry counts, file locations, and requested read ranges.

* **Tests**
  * Added extensive coverage for forged ZIP64 values, boundary violations, and oversized or out-of-range reads.


sdk/internal/zipstream/reader.go
```diff
@@ -169,12 +206,30 @@ func NewReader(readSeeker io.ReadSeeker) (Reader, error) {
...
+			return reader, err
+		}
 
-		reader.fileEntries[string(fileNameByteArray)] = zipFileEntry
+		// bytesToRead is a ZIP64 value read off disk, so it is bounded here
+		// rather than trusted: left unchecked it narrows to a negative
+		// length that slips past the signed size guard in ReadAllFileData and
+		// panics in make(). Deriving length from the two bounded positions
+		// keeps every conversion inside withinArchive.
+		dataEnd, err := withinArchive("file data", uint64(dataStart), bytesToRead, archiveSize)
```
sujankota, line 224: The new bound here is too loose. `dataStart`/`dataEnd` are checked against `archiveSize`, but entry data should end before `centralDirectoryStart`, not merely before EOF. A forged ZIP64 `CompressedSize` can still extend into the central directory and `ReadAllFileData` will return ZIP metadata bytes as payload. The new boundary test appears to lock that behavior in as valid.

dmihalcik-virtru: Fixed

PR #4046, opened 2026-09-14
fix(cli): honor output modes and restrict decrypted files

dmihalcik-virtru:
### Proposed Changes

This PR now combines the umask fix with the owner-only decrypt change from #4037, retaining that commit and its original author.

- Require atomic-output callers to choose the requested destination mode.
- Request `0644` for encrypted TDF output and `0600` for decrypted plaintext.
- Create the temporary sibling with the requested mode so the kernel applies the process umask.
- Remove the commit-time `Chmod`, which applied an absolute mode and bypassed the umask.
- Keep output replacement atomic and preserve an existing destination when writing fails.

The effective mode is the requested mode with the process umask applied. For example:

| Output | Requested mode | umask `022` | umask `077` |
| --- | --- | --- | --- |
| Encrypted TDF | `0644` | `0644` | `0600` |
| Decrypted plaintext | `0600` | `0600` | `0600` |

Masking only clears permission bits, so decrypted output cannot become more permissive than owner-only.

### Relationship to #3939 and #4037

This PR supersedes #4037 while preserving its commit authorship and tests. The #4037 author is invited to review the combined implementation here.

#3939 should land after this PR and rebase onto it. Its direct-destination path should keep the mode parameter and use that mode both for direct `os.OpenFile` creation and for the atomic temporary file; `Commit` should remain free of `Chmod`.

### Checklist

- [ ] I have added or updated integration tests (not needed for this file-mode change)
- [ ] I have added or updated documentation (no user-facing flags changed)

### Testing

Passed:

```shell
make fmt

cd otdfctl
go test ./pkg/streamio ./cmd/tdf -race
go test ./... -race

cd ../sdk
go test -run TestREADMECodeBlocks
```

Repository-wide gates were also attempted. `make lint` reaches pre-existing findings outside this diff, and `make test` stops on pre-existing `lib/fixtures` token-buffer expectation failures. The affected module and packages pass as shown above.

Fixes DSPX-4696.

## Summary by CodeRabbit

* **Enhancements**
  * Output files created by encryption now use standard readable permissions.
  * Decrypted output files are created with restricted permissions for improved privacy.
  * Output file permissions can now be applied consistently, including when replacing an existing destination.
  * Temporary output files use unique names and honor system permission settings.

* **Bug Fixes**
  * Improved cleanup behavior so interrupted operations preserve existing destination files.


otdfctl/pkg/streamio/output.go
```diff
@@ -10,12 +13,16 @@ import (
 // OutputFile that had already been committed or discarded.
 var ErrOutputFileFinished = errors.New("streamio: output file already committed or discarded")
 
-// outputFileMode is the permission Commit applies to the destination.
-// os.CreateTemp always creates the temp file with 0600; without an explicit
-// Chmod that would leak onto the destination regardless of the caller's
-// umask, so this matches the common default a plain os.Create would produce.
+// outputFileMode is the permission requested for the destination, matching what
+// a plain os.Create would ask for. It is a request rather than a guarantee: the
```
jakedoublev: Can you help me understand this? The request is 644 but where is a different mask chosen if that request is denied?

dmihalcik-virtru: The difference is that previously this copied the `os.CreateTemp` (always `0o600`) to the requested output name. This implements a separate createTemp helper, which uses `OpenFile`, which [uses the requested mode](https://pkg.go.dev/os#OpenFile).

otdfctl/pkg/streamio/output_umask_unix_test.go
```diff
@@ -0,0 +1,55 @@
+//go:build unix
```
jakedoublev, line 1: What happens on windows here?

jakedoublev (approved):

otdfctl/pkg/streamio/output_umask_unix_test.go
dmihalcik-virtru: Windows doesn't have a umask, so this whole file is skipped during `go test ./...`.

otdfctl/pkg/streamio/output_test.go
```diff
@@ -86,23 +88,91 @@ func TestOutputFileCommitSetsReadableMode(t *testing.T) {
...
+}
+
+func TestOutputFileTempNamesDoNotCollide(t *testing.T) {
+	dir := t.TempDir()
+	dest := filepath.Join(dir, "out.tdf")
+
+	const concurrent = 16
+	names := make(map[string]bool, concurrent)
+	for range concurrent {
+		o, err := NewOutputFile(dest, testOutputFileMode)
```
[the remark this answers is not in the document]
dmihalcik-virtru: yeah, this test is weak. A better test would use the pigeonhole principle and a predictable rand to specifically find a fault. Maybe later.

otdfctl/pkg/streamio/output_umask_unix_test.go
jakedoublev: Ah yeah, apologies. I meant more what happens to the masks on Windows. It looks like these ownership restrictions would primarily affect *nix and windows would need more hand-holding to make this user/owner read/write differentiation.

PR #4048, opened 2026-09-15
feat(kas): configure KAS URI selection for rewrap

c-r33d:
Stacked on #3951.

Add `services.kas.preview.kas_uri_from_kao` (default `false`). The platform key indexer uses the configured registration unless enabled with a nonempty KAO URI. Changing the flag requires restarting KAS.

Update `start-additional-kas`, examples, docs, and tests for the preview setting.

Validation: focused race tests, SDK README tests, and changed-code lint passed. Full checks encounter existing lint issues and Keycloak fixture failures.

PR #4053, opened 2026-09-15
fix(kas)!: add KAS context to key lookup logs and audit identifiers

c-r33d:
Adds KAS URI context to key lookup and listing debug logs. The indexer logs the registration selected by `services.kas.preview.kas_uri_from_kao`.

**Breaking change:** Successful `GetKey` audit events use `<kas_url>:<kid>` for `object.id` instead of KID alone. Audit consumers must update matching and parsing accordingly.

**Validation:** Focused race tests, SDK README tests, and PR-specific lint passed. Full checks encounter existing lint issues and Keycloak fixture failures.

Stacked on #4048.


jakedoublev (dismissed):

service/kas/access/rewrap.go
```diff
@@ -711,6 +711,10 @@ func (p *Provider) verifyRewrapRequests(ctx context.Context, req *kaspb.Unsigned
 		var err error
 		// The indexer decides whether to use the KAO URI or its configured registration.
 		kasURI := kao.GetKeyAccessObject().GetKasUrl()
+		p.Logger.DebugContext(ctx, "forwarding KAO KAS URI for rewrap",
```
c-r33d: Log is not needed.

service/kas/key_indexer_test.go
```diff
@@ -270,18 +272,19 @@ func (s *KeyIndexTestSuite) TestListKeysWith() {
 				ids[i] = key.ID()
 			}
 			s.Equal(test.expectedIDs, ids)
+			var record map[string]any
```
c-r33d: Refactor this into a shared method

service/policy/kasregistry/key_access_server_registry.go
```diff
@@ -356,7 +358,7 @@ func (s KeyAccessServerRegistry) GetKey(ctx context.Context, r *connect.Request[
 		}
 	}
 
-	auditParams.ObjectID = key.GetKey().GetKeyId()
+	auditParams.ObjectID = fmt.Sprintf("%s:%s", key.GetKey().GetKeyId(), key.GetKasId())
```
c-r33d: This should be KAS URL, and in the opposite order.

service/kas/key_indexer.go
```diff
@@ -41,12 +41,13 @@ type KeyAdapter struct {
...
 
 func NewPlatformKeyIndexer(sdk *sdk.SDK, kasURI string, kasURIFromKAO bool, l *logger.Logger) *KeyIndexer {
-	return &KeyIndexer{
+	p := &KeyIndexer{
 		sdk:           sdk,
 		kasURI:        kasURI,
 		kasURIFromKAO: kasURIFromKAO,
-		log:           l,
 	}
+	p.log = l.With("key_indexer", p.String())
```
dmihalcik-virtru, line 49: Hmm, I don't like bypassing the LogValuer interface here, but this seems to be more efficient by forcing the Stringer interface to be evaluated only once, rather than on each Log call, so this is fine.

dmihalcik-virtru (approved):

service/kas/key_indexer.go
c-r33d: We could adjust the With() method to take in a value of type any. I believe that is more idiomatic.

PR #4056, opened 2026-09-16
fix(kas)!: scope key material cache identity by registry

c-r33d:
Fix private-key cache collisions when different KAS registrations share a key ID. `KeyDetails.CacheKey()` scopes cached key material by KAS URI or provider; `ID()` keeps its lookup behavior.

> [!IMPORTANT]
> **Breaking change:** Downstream implementations of `KeyDetails` must add `CacheKey() string` to satisfy the updated interface.

Stack: #3951 → #4048 → #4053 → #4056.

Validation: focused race tests, SDK README tests, and PR-specific lint passed. Full local checks encounter existing lint issues and Keycloak fixture failures.

[Xtest](https://github.com/opentdf/tests/actions/runs/35608367837) passed before the latest review changes: all 27 KAO/cache cases passed without skips. Startup compatibility fixes: #4085 and opentdf/tests#612.


service/internal/security/in_process_provider.go
```diff
@@ -62,6 +62,10 @@ func (k *KeyDetailsAdapter) ID() trust.KeyIdentifier {
 	return k.id
 }
 
+func (k *KeyDetailsAdapter) CacheKey() string {
+	return fmt.Sprintf("%s:%q", inProcessSystemName, k.id)
```
c-r33d: This should just be the ID of the key. No need for inProcessSystemName.

service/trust/key_index.go
```diff
@@ -37,7 +37,7 @@ const (
 	KeyTypePKCS8
 )
 
-// KeyIdentifier uniquely identifies a key
+// KeyIdentifier identifies a key within its source's scope, not necessarily globally.
```
c-r33d:
```suggestion
// KeyIdentifier identifies a key within the backing key adapters scope, not necessarily globally.
```

service/kas/key_indexer.go
```diff
@@ -184,6 +184,10 @@ func (p *KeyAdapter) ID() trust.KeyIdentifier {
 	return trust.KeyIdentifier(p.key.GetKey().GetKeyId())
 }
 
+func (p *KeyAdapter) CacheKey() string {
+	return fmt.Sprintf("kas:%q:%q", p.key.GetKasId(), p.ID())
```
c-r33d: Should probably be URL + ID

service/trust/key_index.go
```diff
@@ -49,9 +49,17 @@ type PrivateKey struct {
...
+	// ID returns the key's identifier within its source's scope.
+	// For a KasKey, uniqueness requires both this ID and its KAS URI.
+	// Prefer using the CacheKey() method which should always return a unique
+	// string for backing keys
 	ID() KeyIdentifier
 
+	// CacheKey returns an opaque, stable identifier that distinguishes this key
+	// from keys at other KAS URIs or providers sharing a cache.
+	// Use ID for key lookups amonst key providers and CacheKey for caching key material.
+	CacheKey() string
```
dmihalcik-virtru: maybe?
```suggestion
	StableKey() string
```

Also, move this to a separate interface, either its own or an extension

c-r33d: Went ahead and created a `ScopedKeyIdentifier` that is implemented by the platform key indexer to uniquely identify the key for caching and other logging behavior. The manager's should still use `ID()` when reaching out to providers.

PR #4060, opened 2026-09-16
fix(sdk): read the TDF spec version at the manifest root, and stop trusting it

### LINKED ISSUE #4059: sdk: TDF reader trusts the unauthenticated manifest spec version to pick integrity digest encoding

pflynn-virtru:
### What

Two changes to how the TDF reader treats the manifest's spec-version field.

**1. Read it from the right place, under both names.** `Manifest.UnmarshalJSON` now resolves the version in precedence order: canonical `schemaVersion`, then `tdf_spec_version` at the **manifest root**, then `tdf_spec_version` under `payload`. Non-string values (the key is known in the wild carrying `null`) are skipped rather than failing the decode. `sdk/experimental/tdf/manifest.go` is a deliberate duplicate of `sdk/manifest.go` and gets the same decoder.

**2. Stop trusting it.** The reader previously took the integrity-digest encoding — raw bytes (>= 4.3.0) vs hex (pre-4.3.0) — from whether `TDFVersion` was empty. New `digestMatchesRecorded` reads the encoding off the file instead, accepting either spelling for segment digests, the root signature, and assertion signatures. This closes #4059.

**3. Keep the bundled schemas honest.** A new `schema-sync` CI job runs `TestSchemaMatchesSpec` against a checkout of `opentdf/spec`, so the strict schema bundled here and the normative one cannot drift silently again. Divergences must be recorded, with a reason, in `specDivergences`; `TestSchemaLaxRelaxesStrictDeliberately` does the same for lax vs strict.

### Why

`schemaVersion` is the correct name. `tdf_spec_version` is not a former spelling that was renamed — it is an error that leaked into some specification drafts, older OpenTDF documentation, and the writers built from them.

The two placements exist for two different reasons, and both occur in archival files:

* **Root** — where the spec's own `manifest.md` has always documented the field, and where web-sdk both wrote it and still reads it (`lib/tdf3/src/tdf.ts`).
* **Under `payload`** — where revisions of the JSON schema declared it in error, which led at least one writer to emit the key there with a `null` value.

An earlier revision of this PR read only the `payload` placement, on the strength of the bundled schema. That was backwards: the schema was the thing in error. The schema is corrected separately in [opentdf/spec](https://github.com/opentdf/spec).

Files written with the off-spec name decoded to no spec version, so the reader picked the wrong digest encoding and they failed to decrypt. Fixing the *read* alone would have papered over the real defect — the correspondence between "has a version" and "uses raw digests" only ever held because this SDK's writer sets both from one boolean. Every writer that emitted `tdf_spec_version` broke that correspondence, which is exactly why the field cannot be the input to the encoding decision.

Accepting both digest spellings weakens nothing: hex is an invertible encoding of the same HMAC, so producing either form requires the payload key exactly as much as producing the other. There is no forgery the strict form would have rejected.

### Notes

* **The writer is unchanged.** Marshalling emits `schemaVersion` only, so re-emitting a manifest normalizes the name instead of propagating the error. The name is not put on a deprecation path, because it was never correct.
* The second decode pass is gated on a substring scan for `"tdf_spec_version"`, so the common case (no off-spec key) does not re-tokenize the document.
* The JSON schemas are untouched here. Neither sets `additionalProperties: false`, so both spellings already validate. The `$id` values are a separate change in #4070.
* The known pre-4.3.0 regression flagged on the earlier revision of this PR is gone: with the encoding read off the file, a sub-4.3.0 container that records its version under either name still verifies.

### Checklist

- [x] I have added or updated integration tests
- [ ] I have added or updated documentation — n/a; behavior is documented on `Manifest.UnmarshalJSON` and `digestMatchesRecorded`, and no public API changed

### Testing

```text
$ cd sdk && go test ./... -race -count=1     # all packages pass
$ make fmt                                   # clean
$ golangci-lint run ./...                    # no findings in changed files
```

Not run locally: `make test` / `make lint` at repo root (need the `docker compose` Keycloak/Postgres stack).


sdk/tdf.go
```diff
@@ -962,8 +962,6 @@ func (r *Reader) WriteTo(writer io.Writer) (int64, error) {
 		}
 	}
 
-	isLegacyTDF := r.manifest.TDFVersion == ""
```
pflynn-virtru: never let unauthenticated metadata gate verification

Issue #3513, opened 2026-05-22
SDK generates 0.manifest.json while the spec says manifest.json


EmberPlot:
This leads to incompatibility with 3rd party implementations.

```
$echo hi | otdfctl encrypt -o test
$ unzip -tv test.tdf
Archive:  test.tdf
    testing: 0.payload                OK
    testing: 0.manifest.json          OK
No errors detected in compressed data of test.tdf.
```

biscoe916:
@EmberPlot can you expand on the issue you're seeing?

EmberPlot:
@biscoe916 The spec at section [TDF Structure](https://opentdf.io/spec#tdf-structure) says:

> **manifest.json**: The metadata manifest described in the Key Concepts section. It holds instructions for decryption and access control.

So, the manifest is supposed to be named `manifest.json`, but the SDK generates `0.manifest.json`.

<img width="652" height="525" alt="Image" src="https://github.com/user-attachments/assets/9bc4a47d-34db-4145-b6cc-d3a89d1ae568" />

PR #4063, opened 2026-09-16
fix(sdk): read the TDF manifest under the spec name manifest.json

pflynn-virtru:
Refs #3513. Read half only — carved out of #4049, which also renames what the writer emits.

## Problem

The [OpenTDF spec](https://opentdf.io/spec#tdf-structure) names the manifest entry in a `.tdf` archive `manifest.json`. This SDK writes and reads `0.manifest.json`. On the read side it is an exact map lookup with no fallback, so a TDF produced by any implementation written against the published spec is rejected outright.

`TDFReader.Manifest()` is the single choke point for both `LoadTDF` and `IsValidTdf`, so validation fails before any schema check runs.

## Change

`TDFReader.Manifest()` accepts either name, preferring `manifest.json` when an archive carries both. The fallback fires only on a missing entry — an oversized manifest under the spec name stays a size failure, rather than silently returning the superseded off-spec entry.

The dual-name handling lives in `tdf3_reader.go`, not `reader.go`: `Reader` is a generic ZIP accessor whose other two lookups are payload paths that must stay exact.

**The writer is unchanged** and still emits `0.manifest.json`. That rename is a breaking file-format change that needs java-sdk, web-sdk, client-cpp and xtest to accept the spec name first; #4049 carries it. This PR is the read half, which is backwards compatible on its own and is the prerequisite for any of that work.

## Testing

Each behavioral test was verified to fail without the change by reverting `tdf3_reader.go` and re-running:

| Test | Failure without the change |
|---|---|
| `TestManifest_ReadsSpecName` | `zip: file not found` |
| `TestManifest_PrefersSpecNameOverOffspec` | read `text/plain` where `application/octet-stream` expected |
| `TestManifest_OversizedSpecNameDoesNotFallBack` | `An error is expected but got nil` |
| `TestIsValidTdf_AcceptsSpecManifestName` | `tdfReader.Manifest failed: zip: file not found` |

`TestManifest_PrefersSpecNameOverOffspec` gives the two entries different content so it cannot pass vacuously. `TestManifest_ReadsOffspecName` and `TestManifest_MissingUnderEitherName` guard behavior that must not change, and pass either way.

The `specNamedManifestTdf` fixture in `sdk_test.go` was generated from the existing off-spec blob by re-zipping it under the spec entry name, so the two hold the same manifest and payload and differ only in how the manifest entry is named. The existing blob is untouched, which keeps a spec-named and an off-spec-named archive both under test.

```
$ go test ./... -race          # sdk module — all packages ok
$ go test -run TestREADMECodeBlocks .
ok  github.com/opentdf/platform/sdk  1.220s
```

golangci-lint on the `sdk` module: 3 issues, all pre-existing in files this PR does not touch (`segment_writer.go`, `kas_client_test.go`) — zero new.

xtest is unaffected: it reads `0.manifest.json`, which this SDK still writes.

## Known limitations

Surfaced by review of this branch and deliberately left out of scope. None is a regression from `main` except where noted.

- **The normative `payload.url` field is ignored** (#4065). `ReadPayload`/`PayloadSize` hardcode `0.payload`, and `payload.url` has no read sites in the SDK — it is written at `sdk/tdf.go:325` and `experimental/tdf/writer.go:468` and never read back. The spec is explicit that the payload filename is whatever `url` says, and that `0.payload` is only a convention: `schema/OpenTDF/README.md` says "the filename within the archive is referenced by the `manifest.json` (commonly `0.payload`)", `payload.md` calls it "typically", and the normative JSON schema types `url` as an unconstrained string with no `pattern`, `enum` or `const`. A spec-valid archive may therefore name its payload anything, and this SDK fails on the first `Read` — after a KAS rewrap has already been spent. This is exercised in the wild: the community Rust SDK [`arkavo-org/opentdf-rs`](https://github.com/arkavo-org/opentdf-rs) writes `{index}.payload` per entry and sets `payload.url` to match, so its multi-entry archives carry payloads named `1.payload`, `2.payload` and so on. See #4065 for the full survey.
- **An empty `manifest.json` shadows a valid `0.manifest.json`.** The fallback keys on `errZipFileNotFound`, and a zero-length entry is found, so `Manifest()` returns an empty string with a nil error and the failure surfaces later as a JSON parse error. New with this PR, but it requires a malformed archive; preferring the spec entry unconditionally is the same policy as #4049 and keeps the two PRs aligned.
- **Dual-manifest archives read differently per SDK.** An archive carrying both names resolves to the spec entry here and to the off-spec entry in java-sdk and web-sdk. That divergence is inherent to landing the read half first and closes when those SDKs adopt the same preference.
- **Pre-existing `reader.go` gaps that the new name makes easier to reach.** `ReadAllFileData` does not check `CompressionMethod`, so a DEFLATE-compressed manifest returns raw compressed bytes with a nil error — reachable via exactly the "re-zip my TDF under the spec name" workflow this PR enables, since mainstream zip tools deflate by default. Separately, local file header names are never compared against the central directory, and duplicate entry names collapse last-wins. All three predate this PR and affect `0.manifest.json` identically.

One review finding did not survive checking: `IsValidTdf` was reported to accept any zip containing a `manifest.json`. It passes `Lax`, not `Skip` (`sdk.go:494`), and a browser-extension zip is rejected with `could not validate manifest.json`.

## Summary by CodeRabbit

* **Bug Fixes**
  * TDF archives using the specification-standard `manifest.json` entry are now recognized and validated.
  * Existing archives using `0.manifest.json` remain supported.
  * Standard manifest entries take precedence when both supported names are present.
  * Oversized manifests are reported correctly without falling back to alternate entries.


arkavo-com:
Concrete downstream impact, in case it helps prioritise review:

OpenTDFKit (the Swift SDK) moved its writer to the spec name `manifest.json` in arkavo-org/OpenTDFKit#53. Since then its cross-SDK X-Test job, which runs against released `otdfctl/v0.37.0`, fails the Swift-encrypt → Go-decrypt cell with:

```
Failed to decrypt file: tdfReader.Manifest failed: zip: file not found
```

Go-encrypt → Swift-decrypt passes, because the Swift reader already accepts both names. So the read-side fallback in this PR is the only thing between a spec-compliant writer and every released Go consumer. It is non-breaking and independent of the writer rename in #4049.

@opentdf/go-sdk — could someone take a look? Happy to add a test vector (a `manifest.json`-named TDF produced by the Swift SDK) if that would help.

PR #4086, opened 2026-09-21
fix(cli): refuse an output destination that aliases the input

dmihalcik-virtru:
> Follow-up to #3939, which has since merged. Rebased onto `main`.
> Addresses the two open CodeRabbit threads on that PR rather than amending it.

### Proposed Changes

**An output destination that aliases the input.** A destination a rename cannot
stand in for — a symlink, a device, a fifo — is written through directly, which
means `os.OpenFile` with `O_TRUNC`. Both `encrypt` and `decrypt` open their
input first and resolve the destination second, so `-o` at a symlink resolving
to the input truncated the input out from under the already-open descriptor.
The command then read an empty file, failed, and the source was gone:

```
$ ln -s input.tdf out.link
$ otdfctl decrypt input.tdf -o out.link     # before this change
# input.tdf is now 0 bytes, and the decrypt fails on it
```

`NewOutputFile` now takes the caller's already-open input and refuses a
write-through destination that `os.SameFile` says is that input. The check
stats *through* the link, since that is what the open does; a dangling link
resolves to nothing and is not an alias.

The guard is deliberately confined to the write-through path. An ordinary
destination naming the input goes to a temp sibling and is renamed over it only
after `Decrypt` has returned, so nothing is truncated and in-place decryption
keeps working — that is a reasonable thing to ask for, not a footgun.

Passing the input is a required parameter rather than an option. The whole
failure mode is a caller not thinking about it, so the signature makes that
impossible to skip; `nil` is available for a caller with no input to protect.

**Temporary filenames no longer carry the destination's basename.** A component
name is capped at 255 bytes on most filesystems, and `"." + basename + ".tmp-"
+ 16 hex` added 22 bytes to it — so a destination whose own name was valid could
fail the open before a single byte had been read. The prefix is now a fixed
`.otdfctl.tmp-`. The random suffix is what distinguishes concurrent writers; the
prefix only has to be recognizable enough to sweep up after a crash.

Both fixes land in `streamio`, so both reach `encrypt` and `decrypt`. The temp
prefix was not introduced by #3939 — it came in with #3937 — but it lives in the
same function.

### Checklist

### Testing Instructions

```
```

New unit tests, each verified to fail against the unfixed code:

- `TestOutputFileRejectsSymlinkToInput` — the destructive case, and that the
  input is still readable afterwards.
- `TestOutputFileWritesThroughSymlinkToOtherFile`,
  `TestOutputFileWritesThroughDanglingSymlink` — the guard stays narrow enough
  to leave write-through working.
- `TestOutputFileAllowsRegularDestinationNamingInput` — in-place decrypt is not
  rejected.
- `TestOutputFileAcceptsMaximumLengthDestinationName` — a 255-byte destination
  name opens. Probes the filesystem first and skips where a 255-byte component
  is not allowed, so it reports the temp-name bug and nothing else.

e2e, against a running platform:

```
cd otdfctl && bats --tap e2e --filter-tags unattributed_encrypt
```

Two cases added to `streaming.bats` (12 → 14), one per command, asserting the
failure and that the input survives it. The two `assert_no_leftovers` globs in
that file are updated for the new temp prefix.


dmihalcik-virtru:
Considering pulling this as it has a breaking change to the streamio API

jakedoublev:
@dmihalcik-virtru I agree that we can probably punt on this because if you're outputting to the same file you're decrypting, that seems like something that could pretty obviously break.

PR #4087, opened 2026-09-21
fix(sdk): read zip entry bytes with io.ReadFull (DSPX-4590)

dmihalcik-virtru:
Jira: https://virtru.atlassian.net/browse/DSPX-4590

The read-side counterpart to #3936, which fixed the same class of bug on the
write side. Independent of every open stack — no PR in DSPX-2604 (#3941,
#3943-#3949) touches `sdk/internal/zipstream`, and #4043 leaves `readBytes`
alone — so this sits directly on `main` and can land in any order.

## The bug

`readBytes` issued a single `Read` and, on the nil-error path, returned the
whole `make()`'d buffer regardless of how many bytes actually arrived:

```go
buf := make([]byte, size)
n, err := readerSeeker.Read(buf)
if errors.Is(err, io.EOF) { return buf[:n], io.EOF }
if err != nil { return buf[:n], ... }
return buf, nil   // full length, zero-padded, n discarded
```

`io.Reader` explicitly permits `0 < n < len(p)` with a nil error, so the tail
of the buffer comes back as zeros that never existed in the archive —
reported as success.

Demonstrated against a valid fixture with a one-byte-at-a-time `ReadSeeker`:

```
want "payload bytes"
got  "p\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"  err=<nil>  len=13
```

`bytes.Reader` never reads short, which is why no existing test caught it.
But `NewReader` takes a caller-supplied `io.ReadSeeker`, and the
implementations that *do* read short in practice — HTTP range requests,
network filesystems — are exactly the ones large archives get served from.

The `ErrSegSizeMismatch` guard in `tdf.go` (`sdk/tdf.go:1114`) cannot catch
this either: `make()` already gave the buffer the length that check is
looking for, so the check is comparing a value that is correct by
construction. A silently zero-padded segment reaches AES-GCM and fails
there instead, several layers from the cause.

## Why this is the same fix as #3936

#3936 ("fill each segment with io.ReadFull and size the buffer to the input")
made this argument for `CreateTDFContext`'s encrypt loop:

> io.Reader.Read is permitted to return fewer bytes than the caller asked for
> without erroring… A `*bytes.Reader` or `*os.File` on a local disk rarely
> returns short, which is why this has held up, but any wrapping ReadSeeker —
> a decompressor, a network-backed store, an instrumented reader — can
> trigger it and there is nothing wrong with the input when it does.

Same defect, opposite consequence. On the write side a short read made the
loop *fail loudly* on valid input (`io.ReadSeeker.Read size mismatch`). Here
it *silently corrupts*: no error, full-length buffer, invented bytes. The
noisy direction got fixed; the quiet one was left, and it is the worse of the
two. This closes it.

## The fix

`io.ReadFull`, which retries the short read and reports
`io.ErrUnexpectedEOF` when the archive really is truncated. Two related
changes follow from it:

- **No data is returned alongside an error.** The old code handed back a
  partial buffer with a bare `io.EOF` — a trap for a caller doing the
  reflexive `errors.Is(err, io.EOF)` → "normal end of stream", which would
  silently accept a truncated payload. All four production callers
  (`sdk.go:489`, `tdf.go:864`, `tdf.go:987`, `tdf.go:1109`) already treat the
  error as fatal, so nothing depends on the old shape.
- **`io.EOF` is normalized to `io.ErrUnexpectedEOF`.** `io.ReadFull` reports
  the bare sentinel when it reads *nothing* — a read starting exactly at the
  end of a truncated archive — and `io.ErrUnexpectedEOF` only when it got
  part of the request. Both mean the same thing to a caller that asked for a
  byte count taken from the central directory, and letting the first one out
  would have reopened the trap above. Zero-length reads are unaffected;
  `io.ReadFull` returns nil for them even at EOF. (Thanks @​opencode for
  catching this.)
- **A zero-length read at the end of the archive now succeeds.** Any `Read`
  on an exhausted `bytes.Reader` reports `io.EOF`, including a zero-length
  one, so an empty entry stored last used to fail to read back.

Error *classification* is deliberately left alone — `readBytes` still does
not wrap `errZipFormat`. Eight other sites in this file have the same issue
and that belongs in one follow-up, not here.

## Reading this next to #3945

#3945 goes the other way on the same sentinel: it makes the encrypt loop
*tolerate* `io.EOF` and `io.ErrUnexpectedEOF` from `io.ReadFull`, while this
PR makes `io.ErrUnexpectedEOF` fatal. Both are right, because the contracts
differ. The write side has no declared length — a short final read just means
the caller's stream ended. The read side has a length from the central
directory, so a short read means the archive is truncated.

The zero-length change above is also the mirror of #3945's empty-payload
handling ("An empty payload still gets one empty segment"). `zeroLenOKReader`
in `tdf_segment_defaults_test.go` exists because `SDK.CreateTDF` cannot
currently encrypt a genuinely empty `io.Reader`; #3945 fixes the write half
and this fixes the read half. Worth rechecking whether that workaround can be
deleted once both have landed — not assuming it, since the comment there
describes more than one quirk.

## Tests

`chunkedSeeker` in `reader_test.go` serves at most N bytes per `Read` and
reports no error for the short ones, standing in for the range-backed reader
`bytes.Reader` cannot model. Five cases: short reads reassembled, a partially
satisfiable read rejected with `io.ErrUnexpectedEOF` and no partial buffer, a
read starting at EOF rejected with the same sentinel rather than `io.EOF`, a
zero-length read at EOF succeeding, and the same short read driven end-to-end
through `ReadAllFileData`.

Each was verified to **fail against the implementation it guards** — the
first four against the original `readBytes`, the EOF-normalization case
against this PR with the conversion removed — so they are regression tests
rather than tautologies.

They use testify, unlike the rest of `reader_test.go`, which is `t.Fatalf`
throughout — the other three test files in the package use testify, and
`require.ErrorIs(err, io.ErrUnexpectedEOF)` has no clean `t.Fatalf` form. Say
the word if you would rather the file stay internally consistent.

- `cd sdk && go test ./... -race` — green
- `golangci-lint run ./internal/zipstream/...` — 0 issues
- `make fmt` — no diff


sdk/internal/zipstream/reader.go
```diff
@@ -364,22 +364,41 @@ func (reader Reader) ReadFileSize(filename string) (int64, error) {
...
+	if _, err := io.ReadFull(readerSeeker, buf); err != nil {
+		// io.ReadFull reports io.EOF when it read nothing at all and
+		// io.ErrUnexpectedEOF when it read some but not enough. The caller
+		// asked for a specific count taken from the central directory, so
+		// both mean the same thing here: the archive is shorter than it
+		// claims. Letting the bare io.EOF out reintroduces the trap this
+		// function exists to close, since errors.Is(err, io.EOF) reads as a
+		// normal end of stream and a caller acting on that accepts a
+		// truncated entry. A zero-length read is unaffected -- io.ReadFull
+		// returns a nil error for it even at EOF.
```
dmihalcik-virtru:
```suggestion
		// Promote io.EOF to io.ErrUnexpectedEOF
		// due to short (e.g. incorrect CD) archive files.
```

sdk/internal/zipstream/reader.go
```diff
@@ -364,22 +364,41 @@ func (reader Reader) ReadFileSize(filename string) (int64, error) {
...
+// filesystems -- are the ones large archives get served from. Returning the
+// untouched tail of buf would hand back zeros that never came from the
+// archive, reported as success; the ErrSegSizeMismatch guard in tdf.go cannot
+// catch it, because make() already gave the buffer the length that check is
+// looking for. io.ReadFull turns that into io.ErrUnexpectedEOF.
+//
+// Nothing is returned alongside an error. A caller that ignores the error --
+// or one that sees the io.EOF a bare Read used to surface and treats it as a
+// normal end of stream -- would otherwise decrypt a silently truncated
+// payload.
```
dmihalcik-virtru:
```suggestion
// readBytes reads exactly size bytes at index, or fails.
// Unlike most golang io read methods, this function leaves
// the byte array empty on error states to simplify reader logic.
```

sujankota (approved):

PR #4090, opened 2026-09-21
fix(sdk): synchronize the shared KAS key cache

### LINKED ISSUE #4083: fix(sdk): kasKeyCache has no mutex protection — data race under concurrent use
## Motivation

`kasKeyCache` in `sdk/kas_client.go` uses a bare `map[kasKeyRequest]timeStampedKASInfo` with no synchronization. When multiple goroutines call `LoadTDF` or rewrap concurrently on the same SDK instance, `get()`, `store()`, and `clear()` race on the shared map. Go maps are not safe for concurrent read/write — this can panic or silently corrupt state.

The sibling `kasAllowlistCache` was fixed in #3898 by adding a `sync.Mutex`. The same fix should be applied to `kasKeyCache`.

## Details

- **File:** `sdk/kas_client.go`, `kasKeyCache` struct (line ~422)
- **Methods affected:** `get()`, `store()`, `clear()`
- **Fix:** Add a `sync.Mutex` (or `sync.RWMutex`) and lock in each method, matching the pattern from `kasAllowlistCache`

pflynn-virtru:
Fixes #4083.

## Problem

`kasKeyCache` (`sdk/kas_client.go`) is an unsynchronized `map[kasKeyRequest]timeStampedKASInfo`, embedded on the `SDK` struct (`sdk/sdk.go:86`), so every operation on an SDK instance shares it:

- `store` writes the map
- `get` reads it **and deletes** expired entries — so concurrent reads are not safe either
- `clear` replaces the map wholesale, a plain field write

Unraced, this is `fatal error: concurrent map writes` — an unrecoverable runtime abort. There is no panic for a caller to `recover()`; the process dies.

This is not a cold-cache-only window. `storeKeysToCache` runs on **every** `CreateTDF` that carries attributes, once per mapped key:

`CreateTDF` → `newGranter` (`tdf.go:459`) → `newGranterFromService` (`granter.go:492`) → `storeKeysToCache` (`granter.go:616`) → `store`

The grant fallback (`granter.go:521-537`), the offline path (`granter.go:666-675`), and `getPublicKey` on decrypt all write it too. Two concurrent encrypts collide on every call, warm cache or not.

Reproduces on `main` and on every released tag — `sdk/v0.29.0`, `v0.30.0`, `v0.31.0`, `v0.32.0` all carry the identical struct.

## Approach

Three commits, tests first.

| Commit | |
|---|---|
| `2e9ae179` | failing tests: concurrent store/get, and the same through the real `CreateTDF` granter path |
| `7a467f01` | failing tests: `clear()` and the expiry-eviction branch |
| `21c84ad6` | the fix |

An `RWMutex`, not an exclusive one. Lookups dominate this path and one cache instance is shared by every concurrent TDF operation, so a plain `Mutex` serializes them all and gets *slower* as cores are added. Measured on a parallel `get` benchmark, this version is flat — 98/78/96/91 ns at 1/4/8/16 CPUs.

`get` evicts expired entries, so it cannot hold only a read lock. Go has no lock upgrade, so `evictIfUnchanged` re-reads the entry under the write lock and skips the delete if another goroutine refreshed it in between — otherwise the eviction would discard a fresh key.

## Scope

Deliberately **not** included, to keep this to the race:

- **No request deduplication.** A `singleflight` layer is a separate concern with its own correctness surface (leader context propagation, write-back ordering against `StoreKASKeys`, whether coalesced callers may share a `*KASInfo`). It should be judged on its own, not bundled into a crash fix that consumers are waiting on.
- Which key the `kid==""` fallback scan evicts — it currently deletes the caller's key, not the one the scan matched, so scanned entries never evict.
- How that scan chooses among multiple live matches — first hit in randomized map order, no freshness preference, so mid-rotation the SDK nondeterministically wraps under the rotated-out key.
- `getPublicKey` keying lookups on the *requested* kid while `store` keys on the *KAS-returned* kid — when they differ the entry is written where nothing reads it.

Those are real pre-existing bugs and I'm happy to file them, but they are behavior changes rather than races.

Also worth a separate PR: `.github/workflows/checks.yaml:110` runs `go test ./... -short` for sdk with no `-race`; only `./service/integration` gets it. `AGENTS.md` prescribes `make test`, which does use `-race`, but CI does not run it for sdk — which is why this went unnoticed. Removing the lock from `clear()` passes 50/50 iterations without `-race`.

## Testing

Each test commit fails under `-race` at its own HEAD, and passes at the fix:

```
# at 2e9ae179 / 7a467f01
WARNING: DATA RACE
  sdk.(*kasKeyCache).store()      sdk/kas_client.go:423
  sdk.storeKeysToCache()          sdk/granter.go:616
  sdk.newGranterFromService()     sdk/granter.go:492
FAIL	github.com/opentdf/platform/sdk

# at 21c84ad6
$ go test -race -count=2 ./...
ok  	github.com/opentdf/platform/sdk	26.129s
```

`cd sdk && go test -run TestREADMECodeBlocks` passes. `golangci-lint run sdk/...` reports 10 issues, all pre-existing in files this PR does not touch; none in the new code.

The granter test asserts the cache is non-empty after a single build, so it cannot pass vacuously if the mocks stop resolving mapped keys.

## Relationship to #4084

#4084 targets the same issue. Its first commit (the mutex) is the same fix; its later commits add the `singleflight` layer this PR deliberately leaves out. Whichever approach lands, these tests are worth keeping — #4084 currently has no test that fails if the cache write-back is deleted entirely.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when multiple operations access shared KAS key information at the same time.
  - Prevented stale or conflicting cached data from causing inconsistent results during concurrent activity.
  - Ensured cached key information is safely refreshed and cleared without interfering with active requests.

- **Tests**
  - Added comprehensive coverage for concurrent cache access, clearing, expiration, and granter construction.


pflynn-virtru:
https://github.com/opentdf/platform/pull/4084

PR #4091, opened 2026-09-21
fix(policy): record audit outcomes after transaction completion

strantalis:
### Proposed changes

First of two PRs replacing the buffered-processor bridge in #3902. Base: `main`.

Move policy audit outcomes outside database transaction callbacks. Record success only after commit succeeds and route begin/write/commit failures through one failure path. Preserve resource metadata, key redaction, and operation responses.

This layer retains the existing audit API. The next layer switches recording to immediate processing and deletes the request buffer.

### Validation

- Focused policy package tests with race detection passed.
- A namespace regression exercises begin failure, write failure, commit failure, and successful commit through the real handler and audit interceptor. Against the original handler, it catches a missing begin-failure event and an incorrect success event on commit failure.
- An AST regression rejects audit calls inside policy `RunInTx` callbacks.
- `make fmt`, diff-scoped policy lint, SDK README tests, and `git diff --check` passed.
- Full `make lint` reports existing lint findings outside this patch. Full `make test` fails in integration suites requiring unavailable local Keycloak/Platform and Docker.

### Compatibility

No public API changes. Transaction failures now produce failure records where some paths previously emitted no record or an incorrect success. Attribute-value creation still emits success per created child; failures are aggregated by requested object type.


service/policy/namespaces/audit_transaction_test.go
```diff
@@ -0,0 +1,168 @@
+package namespaces
```
jakedoublev: I think we should remove these tests as I don't think they give much value with the heavy mocking here

service/policy/audit_transaction_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"go/token"
+	"io/fs"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Audit outcomes must be recorded after RunInTx reports the commit outcome.
```
jakedoublev: This feels more like a linter than a unit test if we're parsing go code to look for a pattern.

jakedoublev (changes requested):

service/policy/audit_transaction_test.go
strantalis: Yeah I had the same feeling so I removed them

jakedoublev (dismissed):

jakedoublev (approved):

PR #4092, opened 2026-09-21
refactor(core)!: replace buffered audit with immediate recording

strantalis:
### Proposed changes

Second of two PRs replacing the buffered-processor bridge in #3902. Base: `codex/audit-direct-caller-boundaries`.

- Record built-in audit events immediately through `Record` and the configured processor.
- Keep typed event helpers as thin, error-returning wrappers. Policy, KAS, and authorization callers report audit errors through their existing operational logger without changing the completed operation's result.
- Delete the request audit transaction, queue, mutex, flush, and deferred outcome rewriting. The interceptor now only adds request metadata.
- Preserve request-context values, cancellation-independent recording, and the configured processing timeout. No new diagnostic logger option or fallback delivery path.
- Correct rewrap failure attribution when access was allowed but encapsulation failed.

### Validation

- Focused race tests passed for audit, logger, internal server, server startup, policy handlers, KAS, and authorization.
- Regressions cover immediate delivery without an interceptor, processor and construction errors, canceled producer context, completed events surviving later RPC error/panic, committed mutation with processor failure, and unchanged permit/deny decisions with processor failure.
- Existing default wire-shape and actor/request attribution tests remain in place.
- `go build ./service/...`, diff-scoped service lint with zero issues, `make fmt`, SDK README tests, and `git diff --check` passed.
- Full `make lint` reports existing findings. Full `make test` fails in integration suites requiring local Keycloak/Platform and Docker.

### Migration and limits

- `ContextServerInterceptor()` no longer accepts a logger.
- `LogAuditEvent`, `Logger.Detach`, and `Logger.LogPolicyCRUD` are removed. Use `Record` or the error-returning typed helpers; they do not require an audit transaction.
- Existing typed helper names remain, but now return errors. Consumers must handle those errors and move success recording after their own database commit where applicable. Method/interface signatures change.
- A later RPC error/panic no longer rewrites completed sub-operation events as canceled.
- The timeout is per event. A multi-event request can spend multiple timeout budgets; processors must honor deadlines.
- Output acceptance is not durable storage. Processors still own delivery/recovery. Default output remains `level:"AUDIT"`, `msg:<verb>`, `audit:{...}`.

The old bridge PR is superseded rather than merged. No downstream repositories are changed.

Final source review found no actionable defects. Remaining coverage gaps include live integration, end-to-end KAS processor-failure response coverage, and a dedicated multi-child attribute-value commit-failure test.


service/authorization/authorization.go
```diff
@@ -784,14 +784,18 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
 			if decision == authorization.DecisionResponse_DECISION_PERMIT {
 				auditDecision = audit.GetDecisionResultPermit
 			}
-			as.logger.Audit.GetDecision(ctx, audit.GetDecisionEventParams{
+			if auditErr := as.logger.Audit.GetDecision(ctx, audit.GetDecisionEventParams{
```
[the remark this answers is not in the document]
strantalis: Won't add the extra caller fixtures in this PR. The coverage gap is real, but no incorrect behavior is identified. Existing tests cover processor-error propagation through all six helpers and permit/deny preservation through V2. Dedicated V1 and KAS failure-injection coverage remains a follow-up.

service/policy/actions/actions.go
```diff
@@ -178,10 +182,14 @@ func (a *ActionService) UpdateAction(ctx context.Context, req *connect.Request[a
 		return nil
 	})
 	if err != nil {
-		a.logger.Audit.PolicyCRUDFailure(ctx, auditParams)
+		if auditErr := a.logger.Audit.PolicyCRUDFailure(ctx, auditParams); auditErr != nil {
```
jakedoublev: @coderabbitai could we define a reusable wrapper around a.logger.Audit.PolicyCRUDFailure that will log the error if the audit log itself fails?

jakedoublev: @strantalis I think we should do this

service/authorization/authorization.go
```diff
@@ -784,14 +784,18 @@ func (as *AuthorizationService) getDecisions(ctx context.Context, dr *authorizat
...
 				Decision:                auditDecision,
 				EntityChainEntitlements: auditECEntitlements,
 				EntityChainID:           decisionResp.GetEntityChainId(),
 				EntityDecisions:         auditEntityDecisions,
 				FQNs:                    fqns,
 				ResourceAttributeID:     decisionResp.GetResourceAttributesId(),
-			})
+			}); auditErr != nil {
+				as.logger.ErrorContext(context.WithoutCancel(ctx), "failed to record authorization audit event",
```
elizabethhealy: should an audit failure also be an audit event lol

service/authorization/authorization.go
elizabethhealy: getDecisions records each decision as it goes, but a later (ra, ec) pair can still return an error for the whole RPC call which may get confusing as the client gets an error yet the audit log shows permit for the earlier pairs -- not sure how to remediate but maybe something to note for this change

service/kas/access/rewrap.go
```diff
@@ -1090,7 +1099,11 @@ func (p *Provider) tdf3Rewrap(ctx context.Context, requests []*kaspb.UnsignedRew
 				RequiredObligations: requiredObligationsForPolicy,
 			}
 
-			p.Logger.Audit.RewrapSuccess(ctx, auditEventParams)
+			if auditErr := p.Logger.Audit.RewrapSuccess(ctx, auditEventParams); auditErr != nil {
+				p.Logger.ErrorContext(context.WithoutCancel(ctx), "failed to record rewrap audit event",
```
elizabethhealy: a rewrap that can't be audited still returns the key -- maybe something to note for security, we cant guarantee that all successful rewraps are audited

service/authorization/authorization.go
strantalis: well in this case audit failed so an audit failure would fail as well.

service/kas/access/rewrap.go
strantalis: This is a great question. I'm assuming our existing behavior always return the key even if audit failed 🤔

service/authorization/authorization.go
strantalis: Is this existing behavior in the current flow? And if I understand it, so if we err, we're still accidentally auditing a permit decision?

service/policy/actions/actions.go
strantalis: @coderabbitai Something like this? 

```go
func (l *Logger) LogPolicyCRUDSuccess(
	ctx context.Context,
	params audit.PolicyEventParams,
) {
	if err := l.Audit.PolicyCRUDSuccess(ctx, params); err != nil {
		l.ErrorContext(context.WithoutCancel(ctx),
			"failed to record policy audit event",
			slog.Any("error", err),
		)
	}
}
```

